PR #1, opened 2021-09-27
Support wrapping `Real` types

ansley:
Test plan: `test_intarg`

### DIFF
--- a/torchdynamo/symbolic_convert.py
+++ b/torchdynamo/symbolic_convert.py
@@ -4,6 +4,7 @@ import dataclasses
 import functools
 import inspect
 import itertools
+from numbers import Real
 import operator
 import types
 import typing
@@ -178,7 +179,7 @@ class InstructionTranslatorBase(fx.Tracer):
                 state=TracingSupported.YES,
                 guards={Guard(name, GuardSource.LOCAL, GuardBuilder.VALUE_MATCH)},
             )
-        elif value is True or value is False or value is None:
+        elif value is True or value is False or value is None or isinstance(value, Real):
             # For these, just specialize on exact value
             return ConstantVariable(
                 value=value,
--- a/torchdynamo/symbolic_convert.py
+++ b/torchdynamo/symbolic_convert.py
@@ -24,6 +24,7 @@ from .guards import GuardBuilder
 from .guards import GuardSource
 from .variable_tracker import AllowedFunctionOrModuleVariable, PythonModuleVariable
 from .variable_tracker import BaseListVariable
+from .variable_tracker import BasicTypeVariable
 from .variable_tracker import BuiltinVariable
 from .variable_tracker import ConstDictVariable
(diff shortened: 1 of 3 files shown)


torchdynamo/symbolic_convert.py
```diff
@@ -146,7 +147,7 @@ def wrap_local(self, name, value):
                 state=TracingSupported.YES,
                 guards={Guard(name, GuardSource.LOCAL, GuardBuilder.VALUE_MATCH)},
             )
-        elif value is True or value is False or value is None:
+        elif value is True or value is False or value is None or isinstance(value, Real):
```
jansel, line 150: This might not work exactly right.  There are two ways we could support int/floats:

1) Specialize the entire graph based on the exact value.  (The graph gets recompiled whenever the value changes.)  

This is what this PR does, however to make it work you will need to extend how `GuardBuilder.VALUE_MATCH`
works, you will notice it is actually checking `id(val1) == id(val2)`.   Which works for True/False/None because they are singletons, but will have false negatives for integers.  (`id` looks at the address of the object in memory.)  Adding this check should be an easy fix.

2) Treat the value as an input to the graph similar to how `TensorVariable` is implemented.  (See the handling of `isinstance(..., torch.Tensor)` above.)

For this we may want to set `TracingSupported.UNKOWN` (to prevent extraction of a graph with no Tensors) and set the type annotation correctly in the fx Graph.

-----

For None/True/False, going with option 1 makes sense -- as it will only specialize a few times.

For floats, option 2 might be better to prevent over-specialization. 

Integers could go either way.  I might lean towards option 2.  We could also make it configurable, or do some hybrid of specializing small integers but not big integers.

torchdynamo/symbolic_convert.py
```diff
@@ -170,6 +171,14 @@ def wrap_local(self, name, value):
                 state=TracingSupported.YES,
                 guards={Guard(name, GuardSource.LOCAL, GuardBuilder.TYPE_MATCH)},
             )
+        elif isinstance(value, Real):
+            self.graphargs.append(LocalArg(name))
+            self.create_graph_input(name)
+            return ConstantVariable(
```
jansel, line 177: ConstantVariable => TensorVariable

torchdynamo/symbolic_convert.py
```diff
@@ -170,6 +171,14 @@ def wrap_local(self, name, value):
...
+                value=value,
+                state=TracingSupported.UNKNOWN,
+                guards={Guard(name, GuardSource.LOCAL, GuardBuilder.VALUE_MATCH)},
```
jansel, line 180: VALUE_MATCH => TYPE_MATCH

torchdynamo/symbolic_convert.py
jansel: Either way works.  If you want to try adding a new one that could be useful later on.

jansel (approved):

PR #2, opened 2021-09-27
Allow using `nn.Module`s inside a list

ansley:


torchdynamo/symbolic_convert.py
```diff
@@ -645,6 +659,8 @@ def LOAD_ATTR(self, inst):
                     self.push(UserMethodVariable(method, obj, **options))
                 else:
                     unimplemented("nn.Module callable")
+            elif isinstance(subobj, list):
+                self.push(BaseListVariable(subobj))
```
jansel, line 663: 1) BaseListVariable => ListVariable
2) You need to pass **options (to propagate required guards)
3) You should do the conversion of the items in the list to `NNModuleVariable` down here rather than above.  ListVariable expects the items in the list to be VariableTracker instances.

torchdynamo/symbolic_convert.py
```diff
@@ -298,7 +298,21 @@ def call_function(self, fn, args, kwargs):
...
             elif fn.fn is iter and args and isinstance(args[0], BaseListVariable):
                 assert not kwargs and len(args) == 1
-                self.push(ListIteratorVariable(args[0].items, **options))
+                l = args[0].items
+                # If the list contains exclusively nn.Modules, transform
+                # each module into a torchdynamo.NNModuleVariable
+                new_l = []
+                for item in l:
+                    if isinstance(item, torch.nn.Module):
+                        key = unique_id(item.__module__).replace("torch.nn.modules.", "")
```
jansel, line 307: `key` should be based on the attribute name (which you have access to below)

torchdynamo/symbolic_convert.py
```diff
@@ -298,7 +298,21 @@ def call_function(self, fn, args, kwargs):
                 self.push(ListVariable(items, **options))
             elif fn.fn is iter and args and isinstance(args[0], BaseListVariable):
                 assert not kwargs and len(args) == 1
-                self.push(ListIteratorVariable(args[0].items, **options))
+                l = args[0].items
```
jansel, line 301: move this conversion down to the other function.  The items in a BaseListVariable should already be converted to VariableTracker instances.

jansel (changes requested):

torchdynamo/symbolic_convert.py
ansley: Do you mean the line `name = inst.argval` in `LOAD_ATTR`? That would get the name of the list, not the names of its internal elements, as far as I understand right now

jansel: So the current behavior for keys is if you have:
```
self.foo(x)
```
that is mapped to the FX node:
```
call_module   "self_0.foo"   (x,)     {}
```

-----

For this PR  (lists of modules), if you have:
```
self.foo[0](x)
```
that is mapped to the FX node:
```
call_module   "linear_1"   (x,)     {}
```
(or something similar based on the name of the module containing it)

Am suggesting perhaps we name it something like "self.foo_0" for the first item.  "self.foo_1" for the second item. etc. 

The names don't actually matter, it is just for readable graphs.

torchdynamo/symbolic_convert.py
```diff
@@ -645,6 +645,21 @@ def LOAD_ATTR(self, inst):
...
+                l = []
+                for i, item in enumerate(subobj):
+                    if isinstance(item, torch.nn.Module):
+                        key = f"{name}_{i}"
+                        self.nn_modules[key] = item
+                        l.append(NNModuleVariable(
+                                     module_key=key,
+                                     state=TracingSupported.YES,
+                                     ))
+                if len(l) == len(subobj):
```
jansel, line 660: I think if this check fails we should call `unimplemented("....")`  (that means there is some unsupported things in the list)

torchdynamo/symbolic_convert.py
```diff
@@ -645,6 +645,21 @@ def LOAD_ATTR(self, inst):
...
+                # If the list contains exclusively nn.Modules, transform
+                # each module into a torchdynamo.NNModuleVariable
+                l = []
+                for i, item in enumerate(subobj):
+                    if isinstance(item, torch.nn.Module):
+                        key = f"{name}_{i}"
+                        self.nn_modules[key] = item
+                        l.append(NNModuleVariable(
+                                     module_key=key,
+                                     state=TracingSupported.YES,
```
jansel, line 658:
```suggestion
                                     **options
```

jansel (commented):
A few minor things, other than that lgtm

torchdynamo/symbolic_convert.py
```diff
@@ -645,6 +645,24 @@ def LOAD_ATTR(self, inst):
                     self.push(UserMethodVariable(method, obj, **options))
                 else:
                     unimplemented("nn.Module callable")
+            elif isinstance(subobj, list):
+                # If the list contains exclusively nn.Modules, transform
+                # each module into a torchdynamo.NNModuleVariable
+                l = []
+                for i, item in enumerate(subobj):
+                    if isinstance(item, torch.nn.Module):
+                        key = f"{name}_{i}"
```
jansel, line 654:
```suggestion
                        key = f"{obj.module_key}.{name}_{i}"
```
To handle the somewhat unlikely case of:
`a.foo` and `b.foo`

jansel (approved):
other than the minor nit above looks good

torchdynamo/symbolic_convert.py
ansley: Good point, thanks!

ansley: Big oops, I remembered why I didn't have this in the first place: the `get_attr` call with this key breaks the `LayerList` test. (Sending you a paste on Workplace.)

PR #5, opened 2021-10-08
Implement `MAKE_FUNCTION`

ansley:
Trace and inline the inner function if possible


jansel (approved):
lgtm, did you test this with `./torchbench.py`?

PR #7, opened 2021-10-16
Support global loads of bools

ansley:


torchdynamo/symbolic_convert.py
```diff
@@ -504,6 +504,16 @@ def LOAD_GLOBAL(self, inst):
                     },
                 )
             )
+        elif isinstance(value, bool):
+            self.push(
+                ConstantVariable(
+                    value=inst.argval,
+                    state=TracingSupported.YES,
```
jansel, line 511: This should be TracingSupported.UNKNOWN.

TracingSupported  means roughly "contains Tensors" and was intended to prevent generating graphs with only integer operations.  

TracingSupported isn't fully used though, so it actually might make sense to just remove it entirely (though not in this PR).

jansel (changes requested):

PR #8, opened 2021-10-16
Support `len`

ansley:


torchdynamo/symbolic_convert.py
```diff
@@ -376,6 +376,25 @@ def call_function(self, fn, args, kwargs):
             elif fn.fn is iter and args and isinstance(args[0], BaseListVariable):
                 assert not kwargs and len(args) == 1
                 self.push(ListIteratorVariable(args[0].items, **options))
+            elif fn.fn is len:
+                assert not kwargs and len(args) == 1
+                arg = args[0]
+                if isinstance(arg, (TensorVariable, ListVariable, ConstDictVariable)):
```
jansel, line 382: You can treat ListVariable and ConstDictVariable the same as ConstantVariable.  Both of these only support fixed sizes, so their length is a constant.

(This change will also simplify proxy_args below)

jansel (changes requested):

PR #9, opened 2021-10-17
Implement `isinstance`

ansley:


torchdynamo/symbolic_convert.py
```diff
@@ -397,6 +397,24 @@ def call_function(self, fn, args, kwargs):
...
+                arg, *isinstance_types = args
+                get_type = (
+                    lambda x: x.fn if isinstance(x, BuiltinVariable) else x.as_proxy()
+                )
+                isinstance_types = tuple(map(get_type, isinstance_types))
+                if isinstance(arg, TensorVariable):
+                    self.push(
+                        BasicTypeVariable(
+                            self.create_proxy(
+                                "call_function", len, (arg.as_proxy(),), {}
```
jansel, line 411: `len` seems wrong here?

We should have a special case for `isinstance(, Tensor)` that results in a constant.  Since we guard on types, in the common case we don't actually need to run isinstance() as part of the graph.  The guard is checking the type of `x` so checking it again is redundant.

torchdynamo/symbolic_convert.py
```diff
@@ -397,6 +397,24 @@ def call_function(self, fn, args, kwargs):
                     self.push(ConstantVariable(len(item), **options))
                 else:
                     unimplemented(f"`len` with arg type {arg}")
+            elif fn.fn is isinstance:
+                assert not kwargs and len(args) == 2
+                arg, *isinstance_types = args
+                get_type = (
+                    lambda x: x.fn if isinstance(x, BuiltinVariable) else x.as_proxy()
```
jansel, line 404: When would the type be a proxy?  Perhaps we should just require it to be a constant.

jansel (changes requested):

torchdynamo/symbolic_convert.py
```diff
@@ -397,6 +397,19 @@ def call_function(self, fn, args, kwargs):
...
+                arg = (
+                    torch.Tensor
+                    if isinstance(arg, TensorVariable)
+                    else type(arg.as_proxy())
+                )
+                get_type = (
+                    lambda x: x.fn if isinstance(x, BuiltinVariable) else x.as_proxy()
+                )
+                isinstance_types = tuple(map(get_type, isinstance_types))
+                self.push(ConstantVariable(arg in isinstance_types, **options))
```
jansel, line 412: `issubclass` might be better than `in`

torchdynamo/symbolic_convert.py
jansel: Perhaps we could simplify this (and above) by adding a `get_type()` method to `VariableTracker` which errors if we don't know the type.

I worry we will by mistake get a proxy type, then create a incorrect `False` node in the graph

torchdynamo/variable_tracker.py
```diff
@@ -266,6 +300,11 @@ def as_proxy(self):
     def get_key(self):
         return self.__class__, id(self.value)
 
+    def python_type(self):
+        if isinstance(self.value, (torch.Tensor, torch.nn.Module)):
+            return type(self.value)
+        super().python_type()
```
jansel, line 306: missing return (I know it errors, but for future proofing)

torchdynamo/symbolic_convert.py
```diff
@@ -397,6 +397,28 @@ def call_function(self, fn, args, kwargs):
                     self.push(ConstantVariable(len(item), **options))
                 else:
                     unimplemented(f"`len` with arg type {arg}")
+            elif fn.fn is isinstance:
+                assert not kwargs and len(args) == 2
+                arg, *isinstance_types = args
```
jansel, line 402: isinstance_types is always length 1 because of the assert on the line above.

torchdynamo/variable_tracker.py
```diff
@@ -199,16 +215,25 @@ class ListVariable(BaseListVariable):
     def as_proxy(self):
         return list(self._as_proxy())
 
+    def python_type(self):
+        return list
```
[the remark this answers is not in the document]
jansel: You are right, my mistake

torchdynamo/symbolic_convert.py
```diff
@@ -397,6 +397,27 @@ def call_function(self, fn, args, kwargs):
...
+                arg_type = arg.python_type()
+
+                def get_and_assert_python_value(x: VariableTracker):
+                    assert isinstance(
+                        x,
+                        (AllowedFunctionOrModuleVariable, BuiltinVariable),
+                    )
+                    return x.python_value()
+
+                if isinstance(isinstance_types, VariableTracker):
```
jansel, line 412: I believe this will always be true.

jansel: A tuple would turn into TupleVariable (which is a subclass of VariableTracker)

PR #10, opened 2021-10-19
Add support for `float` constructor

ansley:


torchdynamo/symbolic_convert.py
```diff
@@ -397,6 +397,9 @@ def call_function(self, fn, args, kwargs):
                     self.push(ConstantVariable(len(item), **options))
                 else:
                     unimplemented(f"`len` with arg type {arg}")
+            elif fn.fn is float:
```
jansel, line 400: We should have some sort of check on the type of args[0].  This rule won't apply unless it is a constant.

jansel (changes requested):

PR #11, opened 2021-10-20
Support instances of user-defined, non-`nn.Module` classes

ansley:
This PR is a starting-off point for supporting all user-defined classes in torchdynamo. 

Big TODOs for later:
- Allow user-defined subclasses of `nn.Module` to be inlined
- Support user-defined, non-`nn.Module` classes that use an `nn.Module` (not sure if this is worth implementing; can't see anyone doing this in real life)

### DIFF
--- a/torchdynamo/symbolic_convert.py
+++ b/torchdynamo/symbolic_convert.py
@@ -39,6 +39,7 @@ from .variable_tracker import TensorVariable
 from .variable_tracker import TracingSupported
 from .variable_tracker import TupleVariable
 from .variable_tracker import UnsupportedVariable
+from .variable_tracker import UserClassInstanceVariable
 from .variable_tracker import UserFunctionVariable
 from .variable_tracker import UserMethodVariable
 from .variable_tracker import VariableTracker
@@ -123,7 +124,6 @@ def generic_jump(truth_fn: typing.Callable, push: bool):
             if truth_fn(value.value):
                 push and self.push(value)
                 self.jump(inst)
-
         elif isinstance(value, TensorVariable) and self.should_compile_partial_graph():
             # compile a partial subgraph prefix then jump into user code
             assert len(self.stack) == 0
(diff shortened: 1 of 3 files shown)


torchdynamo/symbolic_convert.py
```diff
@@ -359,7 +359,19 @@ def call_function(self, fn, args, kwargs):
...
-            self.inline_user_function(fn.guards, fn.fn, fn.self_args() + args, kwargs)
+            if fn.fn.__class__ is type:
+                options = VariableTracker.propagate(
+                    [
+                        fn,
+                    ]
+                    + list(args)
+                    + list(kwargs.values())
+                )
+                self.push(UserClassInstanceVariable(fn.fn(*args, **kwargs), **options))
```
jansel, line 370: Here you are calling the user code (fn.fn) at compile time.  We should never do this.  This won't work because
1) fn.fn() will only get called once and the result will be hardcoded in the graph (rather than the correct behavior of calling fn.fn() every time the function is called).
2) Calling fn.fn() here could cause arbitrary things to happen, we have no idea it what it does.  So we could lose control of process execution and not capture lots of stuff.

It might be easier to start out by not supporting constructors, and instead just try to support user classes passed as arguments.  These will show up as `UnsupportVariable()`.

When we do try to support constructors, we should leave the constructors in the python code.  I have some pending changes that will make doing something like this easier.

torchdynamo/symbolic_convert.py
```diff
@@ -865,6 +882,20 @@ def LOAD_ATTR(self, inst):
                 self.push(UserFunctionVariable(member, **options))
             else:
                 unimplemented("PythonModuleVariable attribute")
+        elif isinstance(obj, UserClassInstanceVariable):
+            attr = getattr(obj.value, name)
```
jansel, line 886: 1) `getattr()` can execute arbitrary code via properties.  So we need to check for that with inspect.getattr_static().   
2) If the attribute is on the class (and we guard based on the type) then we can know it at compile time.  This is useful for things like method calls.  The other cases above mostly fall into this case.
3) If the attribute is on the instance, then we need to read it at runtime. This code is incorrectly reading the compile-time value.

For case 3, we will need to add guards to handle this correctly.  For `foo.bar = True`, we have a guard checking the type of `foo`, but just because we know the type of `foo` does not mean we know the type of `foo.bar`.   This is solvable, but requires careful accounting of guards.

torchdynamo/symbolic_convert.py
```diff
@@ -865,6 +882,20 @@ def LOAD_ATTR(self, inst):
...
+            if isinstance(attr, types.MethodType):
+                self.push(UserMethodVariable(attr, obj.value, **options))
+            elif isinstance(attr, types.FunctionType):
+                self.push(UserFunctionVariable(attr, **options))
+            elif isinstance(attr, (bool, str, Real, type(None))):
+                self.push(ConstantVariable(attr, **options))
+            elif isinstance(attr, VariableTracker):
```
jansel, line 893: Why is there a compiler type inside a user object?  We shouldn't mix the two.

torchdynamo/symbolic_convert.py
```diff
@@ -1391,7 +1422,7 @@ def inline_call(parent, func, args, kwargs):
         if getattr(func, "__closure__", None) is not None:
             unimplemented("inline with  __closure__")
         if getattr(func, "__self__", None) is not None:
-            unimplemented("inline with  __self__")
+            func = func.__func__
```
jansel, line 1425: Why don't we need to add an extra arg for self here.  Inlining won't work without that.

PR #15, opened 2022-01-26
fix a corner case of list add

shunting314:
Allow adding a BaseListVariable and a ConstantVariable together if the latter is an iterable. It happens in detectron2 which adds a BaseListVariable with an empty tuple.

The PR also fix a few other small things
- fix the usage of json.loads in autotune.py
- in autotune.py, graph_name is hidden by the last backend name iterated
in the for loop
- a typo in bytecode_transformation.py


shunting314:
cc @jansel

jansel (approved):
lgtm, thanks!

PR #16, opened 2022-02-08
Adding functorch to skipfile

anijain2305:
Adding functorch and related packages to the skipfile to avoid dynamo analysis on AOT Autograd code.

@Chillee @jansel

### DIFF
--- a/torchdynamo/skipfiles.py
+++ b/torchdynamo/skipfiles.py
@@ -1,9 +1,11 @@
 import abc
 import collections
+import contextlib
 import copy
 import copyreg
 import dataclasses
 import enum
+import functools
 import importlib
 import inspect
 import linecache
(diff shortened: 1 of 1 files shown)


torchdynamo/skipfiles.py
```diff
@@ -35,10 +38,12 @@
     for m in (
         abc,
         collections,
+        contextlib,
         copy,
         copyreg,
         dataclasses,
         enum,
+        functools,
```
jansel, line 46: This should go in the list below with the try/except (so we don't error if it is not installed).

jansel (changes requested):

torchdynamo/skipfiles.py
anijain2305: Done.

PR #17, opened 2022-02-08
Adding AOT Autograd backend for inference

anijain2305:
@chillee @jansel Added AOT Autograd backend. Let me know if you have any alternate suggestions for the API.

I have tested this for inference for now. It might go through some changes as we work on training.

### DIFF
--- a/torchdynamo/optimizations/backends.py
+++ b/torchdynamo/optimizations/backends.py
@@ -452,6 +452,16 @@ def cudagraphs_inner(model, inputs):
     return run
 
 
+@create_backend
+def aot_autograd(subgraph, **kwargs):
+    from functorch.compile import aot_function, memory_efficient_fusion
+
+    # If kwargs is empty, we fallback to default of memory_efficient_fusion
+    if kwargs:
+        return subgraph.wrap_returns(aot_function(subgraph.model, **kwargs))
+    return subgraph.wrap_returns(memory_efficient_fusion(subgraph.model))
+
+
 def tvm_compile(jit_mod, example_inputs, log_file=None, **kwargs):
     if jit_mod is None:
         return None


torchdynamo/optimizations/backends.py
```diff
@@ -452,6 +452,16 @@ def run(*new_inputs):
     return run
 
 
+@create_backend
+def aot_autograd(subgraph, **kwargs):
+    from functorch.compile import aot_function, memory_efficient_fusion
```
anijain2305, line 457: I kept the imports inside for now to avoid making functorch a dependency.

jansel (approved):
LGTM.  Need to land the other PR first though.

torchdynamo/optimizations/backends.py
jansel: That's perfect

torchdynamo/optimizations/backends.py
```diff
@@ -452,6 +452,16 @@ def run(*new_inputs):
     return run
 
 
+@create_backend
+def aot_autograd(subgraph, **kwargs):
```
jansel, line 456: Note the `@create_backend` decorator might prevent `**kwargs` from working.

anijain2305: It seems to be working for my testcases. But, I will double check.

jansel: Ah cool, I wasn't sure.

Chillee:
Out of curiosity, what makes it not immediately work for training?

anijain2305:
> Out of curiosity, what makes it not immediately work for training?

Actually, it works out of the box here. I am meeting Jason to figure out if there are any gaping holes, or any alternate design.

jansel:
> Actually, it works out of the box here. I am meeting Jason to figure out if there are any gaping holes, or any alternate design.

Oh cool.  I'm actually kind of expecting it to be this easy (these ~5 lines to add the backend).  I feel like 90% of the work will be in measurment and fixing the odd bug or two that surfaces.

PR #18, opened 2022-02-16
Support builtins module

anijain2305:
Looking for a small testcast to add.

### DIFF
--- a/torchdynamo/symbolic_convert.py
+++ b/torchdynamo/symbolic_convert.py
@@ -1095,10 +1095,13 @@ class InliningInstructionTranslator(InstructionTranslatorBase):
         symbolic_locals,
         f_globals,
     ):
+        f_builtins = f_globals["__builtins__"]
+        if not isinstance(f_builtins, dict):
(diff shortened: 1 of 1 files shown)


jansel (approved):
lgtm

anijain2305:
The issue arises because I have the following function in my testing script. TorchDynamo does bytecode analysis and decides to inline it. For this function, the builtins is just a `module` instead of a `dict`. 

~~~
def reduce_out(out):
    """Reduce the output of a model to get scalar loss"""
    if isinstance(out, torch.Tensor):
        return out.sum()
    elif isinstance(out, tuple):
        return sum([reduce_out(x) for x in out])
    raise NotImplementedError("Don't know how to reduce")
~~~

I tried to write a test by simply adding the above function in `tests/misc.py` but it fails to trigger the Inlining. Not sure when inlining is triggered, and how to write a test to force it.


PS  - I have also moved the above code to `torchdynamo/testing.py` to avoid Dynamo analysis.

jansel:
Seems reasonable.  Thanks!

PR #20, opened 2022-02-24
Helper functions for AOT Autograd testing

anijain2305:
More helper functions for AOT Autograd testing

### DIFF
--- a/torchdynamo/testing.py
+++ b/torchdynamo/testing.py
@@ -40,22 +40,6 @@ def reduce_out(out):
     raise NotImplementedError("Don't know how to reduce")
 
 
-def checkpoint_module_params(mod):
-    """Checkpointing state for nn module"""
-    with torch.no_grad():
-        rng_state = torch.clone(torch.random.get_rng_state())
-        module_state = {}
-        for k, v in mod.state_dict().items():
-            module_state[k] = torch.clone(v)
-
-    def restore():
-        with torch.no_grad():
-            torch.random.set_rng_state(rng_state)
-            mod.load_state_dict(module_state)
-
-    return restore
-
-
 def exc_bytecode_offset():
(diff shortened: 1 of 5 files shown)


torchdynamo/testing.py
```diff
@@ -21,6 +21,41 @@
...
+    ):
+        return reduce_out(out.logits)
+    elif type(out).__name__ == "SquashedNormal":
+        return out.mean
+    elif isinstance(out, dict):
+        return sum([reduce_out(value) for value in out.values()])
+    raise NotImplementedError("Don't know how to reduce")
+
+
+def checkpoint_module_params(mod):
```
jansel, line 43: Could this be combined with https://github.com/jansel/torchdynamo/blob/main/torchdynamo/utils.py#L292 ?

torchdynamo/testing.py
```diff
@@ -21,6 +21,41 @@
...
+    with torch.no_grad():
+        rng_state = torch.clone(torch.random.get_rng_state())
+        module_state = {}
+        for k, v in mod.state_dict().items():
+            module_state[k] = torch.clone(v)
+
+    def restore():
+        with torch.no_grad():
+            torch.random.set_rng_state(rng_state)
+            mod.load_state_dict(module_state)
```
jansel, line 54: Is this correct if there is aliasing?

jansel (changes requested):

torchdynamo/testing.py
anijain2305: Hmm, I did not consider aliasing in the parameters itself.
I think I will just use your checkpoint_params from utils.py.

anijain2305:
Changes done.

jansel (approved):

PR #23, opened 2022-03-01
add ltc backend

shunting314:
Add torchdynamo LTC backend. There are 2 version being added
- ltc_trivial : this one does not reuse LTC compiled graph and does tracing everytime. Adding this just for comparison purpose
- ltc_reuse_graph: this one will reuse the compiled LTC graph without retracing.

ltc_reuse_graph is implemented in the LTC PR: https://github.com/pytorch/pytorch/pull/72936 

Command:
LTC_TS_CUDA=1 gpui python torchbench.py --speedup-ltc-trivial -dcuda --only dlrm

cc @jansel

### DIFF
--- a/torchbench.py
+++ b/torchbench.py
@@ -450,6 +450,16 @@ def main():
         help="speedup using experimental fixed_strategy backend",
     )
     group.add_argument(
+        "--speedup-ltc",
+        action="store_true",
+        help="speedup using the ltc backend",
+    )
+    group.add_argument(
+        "--speedup-ltc-trivial",
+        action="store_true",
+        help="speedup using the ltc backend without reusing compiled graph",
+    )
+    group.add_argument(
         "--overhead", action="store_true", help=help(overhead_experiment)
     )
     group.add_argument(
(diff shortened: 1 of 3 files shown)


torchdynamo/optimizations/backends.py
```diff
@@ -582,3 +583,24 @@ def exec_tvm(*args):
         return [to_torch_tensor(m.get_output(i)) for i in range(m.get_num_outputs())]
 
     return exec_tvm
+
+try:
+    import lazy_tensor_core
+    from lazy_tensor_core import _LAZYC
+    # hopefully changing this line to sth like _ltc_init_xla_backend in future
+    # will enable XLA
+    _LAZYC._ltc_init_ts_backend()
```
jansel, line 592: Can we do the imports/setup inside the backend function scope?  This will mean when using other backends we won't need to load extra stuff into memory.

You could follow the same pattern as [_init_tensorflow](https://github.com/jansel/torchdynamo/blob/main/torchdynamo/optimizations/backends.py#L217) to make it happen exactly once.

torchbench.py
```diff
@@ -566,13 +586,15 @@ def main():
         for device in args.devices:
             try:
                 device, name, model, example_inputs = load_model(device, args.only)
-            except NotImplementedError:
+            except NotImplementedError as e:
+                print(f"Fail to load benchmark model {args.only} because it's not fully implemented yet")
```
jansel, line 590: Let's remove debugging prints (or put behind a flag).

torchbench.py
```diff
@@ -566,13 +586,15 @@ def main():
...
                 continue  # bad benchmark implementation
             run_one_model(name, model, example_inputs, optimize_ctx, experiment)
     elif args.isolate:
         os.path.exists(output_filename) and os.unlink(output_filename)
         os.chdir(torchdynamo.config.base_dir)
         for name in iter_model_names(args):
+            print(f"==== running for model {name} ====")
```
jansel, line 597: Let's remove debugging prints (or put behind a flag).

jansel (changes requested):

PR #30, opened 2022-03-11
[DO NOT MERGE] Normalize IR for profiler for debugging

anijain2305:
While debugging `yolov3` with AOT, I found that something weird is going on it `normalize_ir`. 

This PR called normalization for the simple fx_profiler. I am able to repro the `yolov3` error with this 

`python torchbench.py --only=yolov3`

### DIFF
--- a/torchdynamo/profiler.py
+++ b/torchdynamo/profiler.py
@@ -6,6 +6,8 @@ from typing import List
 import torch
 
 from . import config
+from .optimizations.normalize import normalize_ir
+from .utils import checkpoint_params
 from .utils import print_once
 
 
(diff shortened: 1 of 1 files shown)


jansel:
@anijain2305 https://github.com/facebookresearch/torchdynamo/commit/a9a894b8c514b1a0f239d4441fcb01e26f340870 should fix the issue you have here.

Though it introduces a new type of "indirect_mutation" that is not unhandled.  Here is an example:
```
def indirection_mutation(x):
    tmp = x * 10
    torch.sigmoid_(tmp[3])
    return tmp   
```

Before it would change the `sigmoid_` to `sigmoid`.  Now it won't.

anijain2305:
Thanks! `yolov3` is not passing with AOT Autograd!

anijain2305:
Hi Jason, can you explain what `indirect_mutation` means here? Currently, it is preventing the functionalization of simple examples like there

~~~
import torch
import torch.fx
import torchdynamo
from torchdynamo.optimizations.normalize import normalize_ir
from torchdynamo.optimizations.analysis import has_mutation

inputs = [torch.randn(4, 5) for _ in range(4)]


def f(p, q, r, s):
    mul = torch.mul(p, q)
    mul.add_(r)
    sub = mul - s
    return sub


fx_mod = torch.fx.symbolic_trace(f)
print(fx_mod)

res = normalize_ir(fx_mod, inputs)
print(res)

assert not has_mutation(res, inputs)
~~~

jansel:
indirect_mutation means the thing you mutated is reused elsewhere under a different name.  It was mainly trying to catch cases like:
```
b=a.view(1,2,3)
b[5:] = 10
<use a -- which was indirectly mutated via b> 
```

I can see why your example is triggering that.  I think you want to add a pass to rewrite it to something like:
```
def f(p, q, r, s):
    mul1 = torch.mul(p, q)
    mul2 = mul1.add_(r)
    sub = mul2 - s
    return sub
```

Or modify the indirect mutation detection to not fire on `_` or `inplace` methods.

PR #50, opened 2022-03-15
hf_T5 dataclass fields handling

anijain2305:
hf_T5 fails at this place w/o this PR - https://github.com/huggingface/transformers/blob/master/src/transformers/file_utils.py#L2638


anijain2305:
I am not sure if this is the correct implementation as another model `hf_GPT1` fails with an error, and I think it might be related.

jansel (approved):

PR #51, opened 2022-03-16
Using mean instead of sum to have reasonable loss value for backprop

anijain2305:
Helps with `hf_Bart` and `hf_BigBird` models.

### DIFF
--- a/torchdynamo/testing.py
+++ b/torchdynamo/testing.py
@@ -47,7 +47,7 @@ def collect_results(model, prediction, loss, example_inputs):
 def reduce_to_scalar_loss(out):
     """Reduce the output of a model to get scalar loss"""
     if isinstance(out, torch.Tensor):
-        # Mean does not work on float tensors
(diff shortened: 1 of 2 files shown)


torchdynamo/testing.py
```diff
@@ -34,30 +34,35 @@ def collect_results(model, prediction, loss, example_inputs):
...
     return results
 
 
 def reduce_to_scalar_loss(out):
     """Reduce the output of a model to get scalar loss"""
     if isinstance(out, torch.Tensor):
-        return out.sum()
-    elif isinstance(out, tuple):
-        return sum([reduce_to_scalar_loss(x) for x in out])
+        # Mean does not work on float tensors
```
jansel, line 50: Do you mean integer tensors here?

anijain2305: Yeah, let me send a fix

jansel (approved):

PR #63, opened 2022-03-16
Fiddle with C++ and setup.py to fix build on M1 Max Mac

jroesch:
Made a few tweaks to make compilation work on a relatively fresh Mac with default version of clang. Let me know if there is a better way to do this, the ifdef is to handle name not being used when debugging is off, and clang defaults to an older C++ standard.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,6 +22,7 @@ setup(
         CppExtension(
             name="torchdynamo._guards",
             sources=["torchdynamo/_guards.cpp"],
+            extra_compile_args=["-std=c++14"]
         ),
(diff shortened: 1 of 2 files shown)


jansel:
Hey Jared!  Funny seeing you here.  :)  It has been a while since we last talked.

This is great, thanks for fixing the Mac build!  I have only tested on Linux and these fixes look good to me.

Can you sign the CLA and fix the [minor lint](https://github.com/facebookresearch/torchdynamo/runs/5578500346?check_suite_focus=true#step:5:17)?

PR #71, opened 2022-03-20
Fx2trt integration improvement

frank-wei:
This PR will be an improvement on the existing integration. It serves a few purposes:
1. Add the split module by checking each fx graph and split it according to the supported/unsupported op in fx2trt op list. If supported, the subgraph runs on TRT. If unsupported, the subgraph runs on pytorch gpu.
2. Since fx2trt works default on fp16, this PR casts the model and inputs to fp16 if fx2trt is set as backend
3. Due to the nature of TRT, the precision and output loss is controlled by TRT depending on many factors. I set the cosine similarity as the metric to measure loss between reference model and TRT model.


torchbench.py
```diff
@@ -238,9 +238,9 @@ def speedup_experiment(args, model_iter_fn, model, example_inputs):
     timings = np.zeros((args.repeat, 2), np.float64)
     for rep in range(args.repeat):
         # interleave the runs to handle frequency scaling and load changes
-        timings[rep, 0] = timed(model, model_iter_fn, example_inputs)
+        timings[rep, 0] = timed(model, model_iter_fn, example_inputs, times=30)
```
jansel, line 241: Can we make this an option rather than the default?

This script is also used for correctness testing, so this will slow that down 30x.

torchbench.py
```diff
@@ -238,9 +238,9 @@ def speedup_experiment(args, model_iter_fn, model, example_inputs):
...
         with torchdynamo.run():
-            timings[rep, 1] = timed(model, model_iter_fn, example_inputs)
+            timings[rep, 1] = timed(model, model_iter_fn, example_inputs, times=30)
```
jansel, line 243: Same as above.

torchbench.py
```diff
@@ -552,6 +551,9 @@ def main():
         "--speedup-trt", action="store_true", help=help(speedup_experiment_trt)
     )
     group.add_argument("--python-key", action="store_true")
+    group.add_argument(
+        "--speedup-fx2trt", action="store_true", help=help(speedup_experiment)
```
jansel, line 555: I think this `help=` string will be for the wrong thing.

torchbench.py
```diff
@@ -730,18 +739,29 @@ def main():
...
-
     if args.only:
         for device in args.devices:
             try:
                 device, name, model, example_inputs = load_model(
                     device, args.only, args.training, args.check_accuracy
                 )
+
+                if args.speedup_fx2trt:
+                    # cast model and inputs to fp16
```
jansel, line 760: Have you tried using fx2trt with `explicit_precision=True`?  When I tested things this it caused the types to stay float32 and pass the accuracy tests.  I'd be curious to see how much of the speedup is from reduced precision.

Is that mode still supported?  Perhaps we could have a `--speedup-fx2trt-fp16` and an `--speedup-fx2trt` allowing us to compare them both.

Nit: refactor the "convert to fp16" logic into its own function.

jansel (changes requested):
Looks like there are some lint/test issues.

You can fix lint issues automatically with `make lint-deps format`

torchbench.py
frank-wei: I removed the times=30. Functionally, line 239 does a loop with enough iterations already.

torchbench.py
frank-wei: Fixed

torchbench.py
frank-wei: Thanks for pointing out. I created another similar function speedup_experiment_fx2trt

torchbench.py
frank-wei: I am not sure how is that works but according to TRT(EXPLICIT_PRECISION : Specify that the network contains explicit quantization and dequantization scale layers.) It's something related with int8. 
I created a `--speedup-fx2trt-fp16` and an `--speedup-fx2trt` and compare the results on some vision models. Fp32 can not pass the current 1e-4 threshold. (In backend.py:342, I still set explicit_precision=True to see if it works ). Here are the results I got:
Alexnet:
Fp16
tensor([ 0.0450, -1.0811, -1.4688,  ..., -1.4307, -0.8989,  1.1025],
       device='cuda:0')
tensor([ 0.0420, -1.0791, -1.4688,  ..., -1.4307, -0.8994,  1.1055],
       device='cuda:0')

Fp32
tensor([ 0.0459, -1.0811, -1.4692,  ..., -1.4308, -0.8991,  1.1039],
       device='cuda:0')
tensor([ 0.0457, -1.0810, -1.4689,  ..., -1.4307, -0.8994,  1.1040],
       device='cuda:0')

Resnet18:
Fp16:
tensor([1.0850, 2.3184, 2.3320,  ..., 0.3833, 0.4138, 1.2676], device='cuda:0')
tensor([1.0918, 2.3145, 2.3281,  ..., 0.3867, 0.4097, 1.2656], device='cuda:0')

Fp32:
tensor([1.0875, 2.3192, 2.3329,  ..., 0.3841, 0.4151, 1.2665], device='cuda:0')
tensor([1.0907, 2.3191, 2.3355,  ..., 0.3846, 0.4158, 1.2661], device='cuda:0')

Resnet50:
Fp16:
tensor([-0.1942, -0.5420,  1.2812,  ..., -1.0527, -1.2080,  2.0488],
       device='cuda:0')
tensor([-0.1941, -0.5420,  1.2793,  ..., -1.0518, -1.2080,  2.0547],
       device='cuda:0')

Fp32:
tensor([-0.1956, -0.5414,  1.2803,  ..., -1.0534, -1.2087,  2.0509],
       device='cuda:0')
tensor([-0.1929, -0.5416,  1.2817,  ..., -1.0531, -1.2101,  2.0512],
       device='cuda:0')
 
mobilenet_v2
Fp16:
tensor([-1.8945,  0.1934,  0.8633,  ..., -1.1885,  0.0638,  1.7451],
       device='cuda:0')
tensor([-1.8770,  0.2068,  0.8896,  ..., -1.1963,  0.0406,  1.7578],
       device='cuda:0')

Fp32:
tensor([-1.8944,  0.1888,  0.8605,  ..., -1.1921,  0.0652,  1.7422],
       device='cuda:0')
tensor([-1.8936,  0.1904,  0.8655,  ..., -1.2234,  0.0559,  1.7499],
       device='cuda:0')

mnasnet1_0
Fp16:
tensor([-3.9883, -1.2520, -3.3340,  ..., -0.8306,  0.1199,  1.8096],
       device='cuda:0')
tensor([-4.0195, -1.2705, -3.3828,  ..., -0.8257,  0.1372,  1.8242],
       device='cuda:0')
Fp32:
tensor([-3.9908, -1.2514, -3.3324,  ..., -0.8314,  0.1192,  1.8083],
       device='cuda:0')
tensor([-3.9767, -1.2598, -3.3286,  ..., -0.8319,  0.1158,  1.7996],
       device='cuda:0')

jansel (changes requested):
Looks like there are some lint/test errors: https://github.com/facebookresearch/torchdynamo/runs/5630662035?check_suite_focus=true#step:5:13

torchbench.py
```diff
@@ -226,6 +226,27 @@ def coverage_experiment(args, model_iter_fn, model, example_inputs):
     return coverage_result
 
 
+def speedup_experiment_fx2trt(args, model_iter_fn, model, example_inputs):
+    """
+    Measure speedups over eager using the trt inference backend. TRT backend is based fx graph
+    generated by torchdynamo.
+    Writes to ./speedups_fx2trt.csv
+    """
+    timings = np.zeros((args.repeat, 2), np.float64)
```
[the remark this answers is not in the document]
frank-wei: Thanks for the suggestion. Will fix it.

torchbench.py
```diff
@@ -730,18 +805,38 @@ def main():
...
     if args.minimum_call_count:
         torchdynamo.config.minimum_call_count = args.minimum_call_count
-
     if args.only:
         for device in args.devices:
             try:
                 device, name, model, example_inputs = load_model(
                     device, args.only, args.training, args.check_accuracy
                 )
+                # torchbench changed the default precison=fp16 on torchvision net
```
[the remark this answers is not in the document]
frank-wei: I agree. We can leave the vision models work with fp16 only. They will not work with fp32 which I will set an assertion to give a alarm. 
Yes, TRT fp16 is where we can squeeze the most performance out.

torchdynamo/optimizations/backends.py
```diff
@@ -296,29 +296,129 @@ def fx2trt(subgraph):
...
-        input_specs = InputTensorSpec.from_tensors(inputs)
-        interp = TRTInterpreter(model, input_specs, explicit_precision=True)
-        result = interp.run(fp16_mode=False, max_batch_size=len(inputs[0]))
-        trt_mod = TRTModule(result.engine, result.input_names, result.output_names)
-        return subgraph.wrap_returns(trt_mod)
-    finally:
-        signal.alarm(0)
+
+@create_backend
+def fx2trt_fp16(subgraph):
```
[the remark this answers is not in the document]
frank-wei: Nice. The only difference is the fp16_mode=False or fp16_mode=True. 
Actually,  explicit_precision does not need to be set(we already convert the weight to fp16). It is depreciated in new version of TRT. https://docs.nvidia.com/deeplearning/tensorrt/api/c_api/namespacenvinfer1.html#aa8f406be96c14b7dbea548cf19f09a08ac1d5bcbbbe1176cede9cacdf35375e7e

torchbench.py
```diff
@@ -730,18 +776,43 @@ def main():
...
+                    "mobilenet_v2",
+                    "mnasnet1_0",
+                    "squeezenet1_1",
+                    "shufflenetv2_x1_0",
+                    "vgg16",
+                    "resnext50_32x4d",
+                ):
+                    try:
+                        assert (
+                            args.speedup_fx2trt_fp16 == True
```
jansel, line 809: Can we move this to `speedup_experiment_fx2trt()`, you can use the global variable `current_name`.  I believe this disables these models for everything, not just fx2trt.

torchbench.py
```diff
@@ -730,18 +776,43 @@ def main():
...
+                        ), "Do not test vision models in fp32 mode"
+                    except NotImplementedError:
+                        continue  # not supported benchmark implementation
```
jansel, line 812: Unreachable?

jansel (changes requested):
There is still a lint here: https://github.com/facebookresearch/torchdynamo/runs/5633321345?check_suite_focus=true#step:5:12

Note you can run linter locally with `make lint`

torchbench.py
frank-wei: Thanks for pointing that out. 
Hm, we can not put it into speedup_experiment_fx2trt. Since we need to stop it earlier than that. If not, the error will throw in line 923: 
elif not same(correct_result, new_result, cos_similarity):

torchbench.py
```diff
@@ -839,7 +920,7 @@ def run_one_model(
             # check correctness.
             # TODO(jansel): submit upstream fix for this
             pass
-        elif not same(correct_result, new_result):
+        elif not same(correct_result, new_result, cos_similarity):
```
frank-wei, line 923: If we let the fp32 vision models to run, their output is fp32 and reference model output is fp16. They can not be compared.

torchbench.py
frank-wei: Not really, just want to stop fp32 run here. Otherwise, line 923 will throw error. The reason is that TRT output is fp32 and reference model output is fp16. They can not be compared.

jansel: The only thing inside the `try:` is an `assert`   How will we get a NotImplementedError?

torchbench.py
jansel: Could it go inside the `if` on L813?  Just want to make sure this doesn't change things when not using fx2trt.

torchbench.py
jansel: I'm suggesting we still exclude them.  Just only exclude them when running fx2trt.

torchbench.py
frank-wei: Got it. Maybe it is better to have another `if args.speedup_fx2trt: ` and put them under this if

torchbench.py
frank-wei: I made it complicated. Will fix it.

torchbench.py
frank-wei: yea, I changed the previous comment that should exclude them.

jansel (approved):
lgtm, thanks!

PR #84, opened 2022-03-22
Torchbench changes for AOT Autograd

anijain2305:
More command line options in torchbench for AOTAutograd

### DIFF
--- a/torchbench.py
+++ b/torchbench.py
@@ -33,6 +33,7 @@ from torchdynamo.optimizations.inference import offline_autotuner
 from torchdynamo.optimizations.inference import online_autotuner
 from torchdynamo.optimizations.python_key import python_key
 from torchdynamo.optimizations.training import aot_autograd_debug_strategy1
+from torchdynamo.optimizations.training import aot_autograd_nnc_strategy
 from torchdynamo.optimizations.training import aot_autograd_speedup_strategy
 from torchdynamo.profiler import Profiler
 from torchdynamo.profiler import fx_insert_profiling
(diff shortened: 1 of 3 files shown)


torchdynamo/optimizations/training.py
```diff
@@ -4,13 +4,18 @@
...
 from torchdynamo.utils import count_calls
+from torchdynamo.utils import counters
 
 from .analysis import has_mutation
 from .backends import BACKENDS
 from .normalize import normalize_ir
 
 log = logging.getLogger(__name__)
 
+counters["aot_autograd"]["total"] = 0
```
jansel, line 15: Counters are a defaultdict, no need for this -- everything defaults to 0

torchbench.py
```diff
@@ -177,13 +183,17 @@ def print_summary(cls):
...
 
+    @classmethod
+    def aot_summary(cls):
+        return [cls.totals["aot_autograd"]["total"], cls.totals["aot_autograd"]["ok"]]
+
 
-def output_csv(headers, row):
-    assert output_filename
-    existed = os.path.exists(output_filename)
+def output_csv(filename, headers, row):
```
jansel, line 191: Is this change actually used?  What is motivation?

jansel (changes requested):

torchbench.py
anijain2305: I am using this later to write *_stats.csv file.

torchbench.py
```diff
@@ -823,6 +866,12 @@ def main():
                 experiment,
                 cos_similarity,
             )
+        stats_file = output_filename.split(".csv")[0] + "_stats.csv"
```
anijain2305, line 869: @jansel This is the place, I am passing on a new filename. But, yeah, it is not as contained as it should be.

torchbench.py
jansel: Sounds good, I didn't see that

jansel (approved):

PR #97, opened 2022-03-29
Fx2trt pr2

frank-wei:
1. the same function calls miss the cosine_similarity param
2. downgrade the precision of int64 to int32 for some inputs which got called in embedding


frank-wei:
Looks like CI has some problems on "make lint". Local test is good.

jansel:
Thanks, I'll fix the linter problem.  Looks like click 8.1.0 breaks black.

PR #110, opened 2022-04-05
Fx2trt pr3

frank-wei:
Align with new fx2trt api
Some small changes


torchbench.py
```diff
@@ -971,7 +972,11 @@ def run_one_model(
 
         correct_result = model_iter_fn(copy.deepcopy(model), example_inputs)
         torch.manual_seed(1337)
-        if current_name != "pyhpc_turbulent_kinetic_energy":
+        if (
+            current_name != "pyhpc_turbulent_kinetic_energy"
+            and current_name != "demucs"
+            and current_name != "pyhpc_isoneutral_mixing"
+        ):  # diff for demucs is 0.0002
```
jansel, line 979: For these should we have an option like `--correctness-threshold=...` (and change the default for fx2trt).

I believe this will disable correctness checking for all backends, rather than only fx2trt.

jansel (changes requested):

torchbench.py
frank-wei: Hmmm, I was not aware of that fp32 is also affected. line 980 is for 2 times of runs of model on pytorch(not between pytorch and fx2trt). It was weird that demucs has some randomness in its fp16 runs. 
Since we do not have a fp16 flag yet. I will remove these this change first.

jansel: I think the correctness is fine for fp32 on these models.  At least for non-fx2trt backends.

frank-wei: yea, I removed those changes.

jansel (approved):

PR #130, opened 2022-04-08
add an option to randomize inputs

shunting314:
torchbench only provides a single set of input for each model so far. We can not discover issues like caching something we should not cache. This PR add an option to torchbench.py in dynamo repo to generate randomized input.

Note that, we only randomize floating point items. For integral items, randomizing it may generate invalid inputs especially when those integral items representing indices.

The PR also adapt the ltc/ltc_trivial backend to the refactoring made inside LTC recently.

Test plan:
```
LTC_TS_CUDA=1 gpui time python torchbench.py --speedup-ltc -dcuda --nvfuser --randomize-input
```

cc @jansel

### DIFF
--- a/torchbench.py
+++ b/torchbench.py
@@ -161,17 +161,17 @@ def load_model(device, model_name, is_training, use_eval_mode):
     return device, current_name, model, example_inputs
 
 
-def timed(model, model_iter_fn, example_inputs, times=1):
+def timed(model, model_iter_fn, example_inputs, times=1, return_result=False):
     synchronize()
     gc.collect()
     torch.manual_seed(1337)
     t0 = time.perf_counter()
     # Dont collect outputs to correctly measure timing
     for _ in range(times):
-        model_iter_fn(model, example_inputs, collect_outputs=False)
+        result = model_iter_fn(model, example_inputs, collect_outputs=False)
         synchronize()
     t1 = time.perf_counter()
-    return t1 - t0
+    return (t1 - t0, result) if return_result else t1 - t0
 
 
 class Stats:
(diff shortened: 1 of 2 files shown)


xuzhao9:
> torchbench only provides a single set of input for each model so far. We can not discover issues like caching something we should not cache. This PR add an option to torchbench to generate randomized input.
> 
> Note that, we only randomize floating point items. For integral items, randomizing it may generate invalid inputs especially when those integral items representing indices.
> 
> The PR also adapt the ltc/ltc_trivial backend to the refactoring made inside LTC recently.
> 
> Test plan:
> 
> ```
> LTC_TS_CUDA=1 gpui time python torchbench.py --speedup-ltc -dcuda --nvfuser --randomize-input
> ```
> 
> cc @jansel

TorchBench has added an interface, `gen_inputs()`, which will return another set of inputs for the model. It currently only supports timm and vision models, but we are working towards 100% coverage.

anijain2305:
Does it make sense to change the example inputs here - https://github.com/facebookresearch/torchdynamo/blob/main/torchbench.py#L145-L149

This way, you can use randomize inputs for any of the speedup experiments.

jansel (approved):
lgtm

xuzhao9:
> 

Yes, it makes sense to change the example inputs there.

shunting314:
> Does it make sense to change the example inputs here - https://github.com/facebookresearch/torchdynamo/blob/main/torchbench.py#L145-L149
> 
> This way, you can use randomize inputs for any of the speedup experiments.

The problem is we need multiple random inputs while the load_model interface only return a single input. We wound need to change the interface/behavior of this method which may break callers.

xuzhao9:
>  The problem is we need multiple random inputs while the load_model interface only return a single input. We wound need to change the interface/behavior of this method which may break callers.

First, I am okay to merge this without using `gen_inputs()`.
Second, I think it makes sense to change the interface of this method from returning a single input to returning an input generator, but this can be an optional improvement later.

anijain2305 (approved):

shunting314:
> Second, I think it makes sense to change the interface of this method from returning a single input to returning an input generator, but this can be an optional improvement later.

For the second point, should we worry about the caller side. Assume there are other callers to this method. Originally they expect a single batch but now they get an generator/iterator of batches. We would need a way to handle this backward compatibility.

shunting314:
@jansel  @anijain2305  looks like I don't have permission to merge to the facebookresearch/torchdynamo repo. Do you mind do the merge for me?

xuzhao9:
Merged

PR #141, opened 2022-04-13
Handle torch.seed

anijain2305:
@jansel This is probably the incorrect way, looking for feedback.

The generated Fx graph for 

~~~
        def fn(x):
            attention_seed = int(torch.seed() % sys.maxsize)
            torch.manual_seed(attention_seed)
            return ()
~~~

is 
~~~

def forward(self):
    seed = torch.random.seed()
    mod = seed % 9223372036854775807;  seed = None
    int_1 = int(mod);  mod = None
    manual_seed = torch.random.manual_seed(int_1);  int_1 = None
    return ()

~~~

### DIFF
--- a/torchdynamo/variables/builtin.py
+++ b/torchdynamo/variables/builtin.py
@@ -8,6 +8,8 @@ from typing import List
 
 import torch
 
+from torchdynamo.variables.tensor import SeedVariable
+
 from .. import config
 from .. import variables
 from ..allowed_functions import is_disallowed
@@ -94,6 +96,7 @@ class BuiltinVariable(VariableTracker):
     @functools.lru_cache(None)
     def _fx_graph_functions():
         fns = {
+            int,
             operator.pos,
             operator.neg,
             operator.not_,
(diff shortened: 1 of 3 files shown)


torchdynamo/variables/builtin.py
```diff
@@ -94,6 +96,7 @@ def can_constant_fold_through(self):
     @functools.lru_cache(None)
     def _fx_graph_functions():
         fns = {
+            int,
```
jansel, line 99: Perhaps we should be more selective about admitting `int()` into the FX graph.  If someone calls `int(Tensor)` we should likely break the graph on it.  So maybe allow int on a SeedVariable/DynamicShapeVariable but not a general tensor?

jansel (changes requested):
I think this will cause

torchdynamo/variables/builtin.py
anijain2305: I agree. I don't like `int()` in the graph. Yes, I think allowing `int` on SeedVariable/DynamicShapeVariable is acceptable.  Will make the changes.

anijain2305:
Ready for review.

jansel (approved):

PR #143, opened 2022-04-14
Fix Python 3.7 compatibility issues & Add testing action workflow

yanboliang:
Changes:
* Replace ```BEGIN_FINALLY``` with ```LOAD_CONST``` when Python version <= 3.7.
* Add ```SETUP_EXCEPT```.
* Update ```WITH_CLEANUP_START```.
* Add github Python 3.7 testing action workflow.

Tests:
* All unit tests passed.
* torchbench passed.

Issue: https://github.com/facebookresearch/torchdynamo/issues/52

### DIFF
--- a/torchdynamo/resume_execution.py
+++ b/torchdynamo/resume_execution.py
@@ -11,7 +11,6 @@ from .bytecode_transformation import create_instruction
 from .bytecode_transformation import transform_code_object
 from .codegen import PyCodegen
 from .utils import ExactWeakKeyDictionary
-from .utils import istype
 
 # taken from code.h in cpython
 CO_OPTIMIZED = 0x0001
--- a/torchdynamo/resume_execution.py
+++ b/torchdynamo/resume_execution.py
@@ -1,5 +1,6 @@
 import copy
 import dataclasses
+import sys
 import types
 from typing import Any
 from typing import Dict
@@ -29,9 +30,14 @@ class ReenterWith:
 
     def __call__(self, code_options, cleanup):
         with_cleanup_start = create_instruction("WITH_CLEANUP_START")
+        if sys.version_info < (3, 8):
+            begin_finally = create_instruction("LOAD_CONST", 0)
+        else:
+            begin_finally = create_instruction("BEGIN_FINALLY")
+
         cleanup[:] = [
             create_instruction("POP_BLOCK"),
-            create_instruction("BEGIN_FINALLY"),
+            begin_finally,
             with_cleanup_start,
             create_instruction("WITH_CLEANUP_FINISH"),
             create_instruction("END_FINALLY"),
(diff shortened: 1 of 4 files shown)


torchdynamo/resume_execution.py
```diff
@@ -29,9 +30,14 @@ class ReenterWith:
 
     def __call__(self, code_options, cleanup):
         with_cleanup_start = create_instruction("WITH_CLEANUP_START")
+        if sys.version_info < (3, 8):
+            begin_finally = create_instruction("LOAD_CONST", 0)
```
jansel, line 34: This is close, but might not be exactly right.  

[LOAD_CONST](https://docs.python.org/3/library/dis.html#opcode-LOAD_CONST) pushes `code_options["co_consts"][inst.arg]` onto the stack.

I think you are assuming `code_options["co_consts"][0] is None`, which might not be true.

You may need to update co_consts similar to this: https://github.com/facebookresearch/torchdynamo/blob/16f6836e2f9793b99f40ee0f047ebc3a46e9e162/torchdynamo/codegen.py#L214
Ideally you could just call that code.

jansel (changes requested):

torchdynamo/resume_execution.py
```diff
@@ -49,6 +45,24 @@ def __call__(self, code_options, cleanup):
...
         ]
 
+    # For Python 3.7 compatibility
+    def _create_compatible_begin_finally(self, code_options):
+        if sys.version_info < (3, 8):
+            co_consts = code_options["co_consts"]
+            assert istype(co_consts, tuple)
+            index = None
+            for i, v in enumerate(co_consts):
+                if v == None:
```
jansel, line 55: This should be `is None`

torchdynamo/resume_execution.py
yanboliang: Thanks for catching this! I find it's not easy to reuse the ```create_load_const``` as we can't get the ```PyGen``` object in this context, so I write a similar function follow that one. Lmk if it makes sense.

torchdynamo/resume_execution.py
```diff
@@ -49,6 +45,24 @@ def __call__(self, code_options, cleanup):
             create_instruction("POP_TOP"),
         ]
 
+    # For Python 3.7 compatibility
+    def _create_compatible_begin_finally(self, code_options):
+        if sys.version_info < (3, 8):
+            co_consts = code_options["co_consts"]
```
jansel, line 51: Could we refactor this to share code with _create_load_const?  

You could modify `_create_load_const` to have a `@staticmethod` helper function.

yanboliang: Good point! Let me do it!

jansel (approved):

PR #144, opened 2022-04-15
Handle torch.override.is_tensor_like

anijain2305:
To handle new changes in pytorch.distributions

### DIFF
--- a/torchdynamo/variables/tensor.py
+++ b/torchdynamo/variables/tensor.py
@@ -112,11 +112,6 @@ class TensorVariable(VariableTracker):
                 )
         elif example_value is None or proxy.node.target is torch.manual_seed:
             return variables.ConstantVariable(None, **options)
-        elif (
-            isinstance(example_value, bool)
-            and proxy.node.target is torch.overrides.is_tensor_like
-        ):
-            return variables.ConstantVariable(example_value, **options)
         else:
             assert (
                 False
--- a/torchdynamo/variables/tensor.py
+++ b/torchdynamo/variables/tensor.py
@@ -112,6 +112,12 @@ class TensorVariable(VariableTracker):
                 )
         elif example_value is None or proxy.node.target is torch.manual_seed:
             return variables.ConstantVariable(None, **options)
+        elif (
+            isinstance(example_value, bool)
+            and proxy.node.target is torch.overrides.is_tensor_like
+        ):
+            proxy.node.meta["example_value"] = example_value
+            return DynamicShapeVariable(proxy, type(example_value), **options)
         else:
             assert (
                 False
--- a/torchdynamo/variables/tensor.py
+++ b/torchdynamo/variables/tensor.py
@@ -116,8 +116,7 @@ class TensorVariable(VariableTracker):
             isinstance(example_value, bool)
             and proxy.node.target is torch.overrides.is_tensor_like
         ):
-            proxy.node.meta["example_value"] = example_value
-            return DynamicShapeVariable(proxy, type(example_value), **options)
+            return variables.ConstantVariable(example_value, **options)
         else:
             assert (
                 False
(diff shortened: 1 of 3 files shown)


torchdynamo/variables/tensor.py
```diff
@@ -106,6 +106,12 @@ def create(cls, tx, proxy, example_value=None, nnmodule=None, **options):
                 )
         elif example_value is None or proxy.node.target is torch.manual_seed:
             return variables.ConstantVariable(None, **options)
+        elif (
+            isinstance(example_value, bool)
+            and proxy.node.target is torch.overrides.is_tensor_like
+        ):
+            proxy.node.meta["example_value"] = example_value
+            return DynamicShapeVariable(proxy, type(example_value), **options)
```
jansel, line 114: What is the motivation for putting this in the graph.  Could we do any useful computation in it?

This could just be a ConstantVariable with a guard?  (And not go into the graph.)

I haven't seen the use case, so just curious.

anijain2305: One example is Normal distribution - https://github.com/pytorch/pytorch/blob/master/torch/distributions/normal.py#L45

The inputs to this init (loc and scale) are tensors. That's the main reason, I put it in the graph. 

`is_tensor_like` is used here - https://github.com/pytorch/pytorch/blob/master/torch/distributions/utils.py#L11

But, maybe since this is called in __init__, it should be ok to be ConstantVariable?

More generally, how do you decide what to put as Constant vs DynamicShape? For example, would you put `torch.seed()` as ConstantVariable and put a guard on it?

jansel: In those examples the result seems to feed directly into control flow.

I'd suggest either a constant or a graph break (by adding this to disallowed functions).

The basic heuristic I use is to think about if the compiler could do anything useful with it.

`torch.seed()` should *not* be a constant, because we don't want to specialize on it.  It is possibly some random value.

`is_tensor_like(Tensor)` seems like it should just always be true.  We already guard on the fact the Tensor is a Tensor.

anijain2305: Makes sense. Making it a `ConstantVariable`.

anijain2305: IIUC, specializing on `torch.seed()` would be bad, because we would be causing so many recompilations because `torch.rand()` will most possibly be some random value.

But `is_tensor_like` is mostly going to be true, so it is ok to specialize on that. Even if it changes, we will most likely do a couple of recompilations and that will be good enough to cover all the scenarios. And we already have guard on Tensor, so we are not adding any more guards as well. 

Is that the right way to think about it?

jansel: Yes.  Also is_tensor_like() usually flows into control flow, so making it a constant allows bigger graphs.

torchdynamo/variables/tensor.py
```diff
@@ -106,6 +106,11 @@ def create(cls, tx, proxy, example_value=None, nnmodule=None, **options):
                 )
         elif example_value is None or proxy.node.target is torch.manual_seed:
             return variables.ConstantVariable(None, **options)
+        elif (
+            isinstance(example_value, bool)
+            and proxy.node.target is torch.overrides.is_tensor_like
+        ):
+            return variables.ConstantVariable(example_value, **options)
```
jansel, line 113: This is the wrong place to make it a constant.  At this point we already added a node to the FX graph, so it is too late.

You could handle this similar to `torch.is_tensor` which is here:
https://github.com/facebookresearch/torchdynamo/blob/1bc7be4de849aeb3aacee1259bfb593fce79a137/torchdynamo/variables/torch.py#L104

jansel (changes requested):

torchdynamo/variables/tensor.py
jansel: L104 handles the True case.  For the False case you can add it here:
https://github.com/facebookresearch/torchdynamo/blob/1bc7be4de849aeb3aacee1259bfb593fce79a137/torchdynamo/variables/torch.py#L72

anijain2305:
Thanks for the comments. It makes sense now.

Incorporated your changes, and also added a test.

jansel (approved):
lgtm, feel free to merge once merge conflict is fixed

PR #151, opened 2022-04-18
AOT Autograd Training - few more models passing

anijain2305:
As title

### DIFF
--- a/torchbench.py
+++ b/torchbench.py
@@ -59,6 +59,8 @@ SKIP = {
     "tacotron2",
     # https://github.com/facebookresearch/torchdynamo/issues/101
     "detectron2_maskrcnn",
+    # https://github.com/facebookresearch/torchdynamo/issues/145
+    "fambench_xlmr",
 }
 
 # Additional models that are skipped in training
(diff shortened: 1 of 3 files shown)


torchbench.py
```diff
@@ -518,6 +533,10 @@ def forward_pass(mod, inputs, collect_outputs=True):
     return mod(*inputs)
 
 
+# Skip the wrapper function to allow tracing to print the traced filenames
+torchdynamo.eval_frame.skip_code(forward_pass.__code__)
```
jansel, line 537: Let's add a specific API for this one, rather than call this C function directly.  Perhaps name it `torchdynamo.skip` and allow it to be used as a decorator. 

```
def skip(fn=None):
  """
  Skip frames associated with a given function's code, but still process recursively invoked frames.
  """
  if fn is None:
    return skip  # @torchdynamo.skip() decorator syntax
  assert callable(fn), "expected a callable"
  eval_frame.skip_code(fn.__code__)
  return fn  # allow use as a decorator

**Review by jansel — CHANGES_REQUESTED** [review]

File: torchbench.py (hunk 1 above)
Comment (anijain2305) [reply torchbench.py:537]: Good idea!

Comment (jansel) [reply torchbench.py:537]: One more thing, the wrapper should also set:
```
fn._torchdynamo_disable = True
```
Similar to:
https://github.com/facebookresearch/torchdynamo/blob/417a60b5e8208c972e384a7557af4905770969aa/torchdynamo/eval_frame.py#L64

This will prevent the function from getting inlined.

jansel (approved):

PR #153, opened 2022-04-18
Fix import overhead by using `importlib.util.find_spec`

xuzhao9:
Fixes https://github.com/facebookresearch/torchdynamo/issues/133.
This also fixes the performance overhead caused by importing third party libraries in https://github.com/facebookresearch/torchdynamo/issues/107

### DIFF
--- a/torchdynamo/skipfiles.py
+++ b/torchdynamo/skipfiles.py
@@ -122,6 +122,7 @@ for _name in (
 ):
     add(importlib.util.find_spec(_name))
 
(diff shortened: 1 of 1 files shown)


jansel (approved):

jansel:
@anijain2305 is reporting this changes the behavior of TorchDynamo and results in fewer graphs.  So there is an issue with the PR.  We may need to revert it if there isn't an easy fix.

PR #154, opened 2022-04-18
Add support for Python 3.9

yanboliang:
Changes:

* ```_eval_frame.c```
  * Use ```_PyInterpreterState_SetEvalFrameFunc``` to enable/disable eval frame for Python >= 3.9.
  * Refactor ```custom_eval_frame``` as ```_PyFrameEvalFunction``` has a new arg ```PyThreadState *tstate``` if Python >= 3.9
  * Use ```PyFrame_New``` to replace ```_PyFrame_New_NoTrack``` which is not public API after Python 3.9. Meanwhile, the ```PyFrame_New``` includes ```PyObject_GC_UnTrack``` so we have to untrack it after the function call.

* Implement the new version of ```ReenterWith.__call__``` cleanup logics.

* Add several new bytecodes:
  * IS_OP
  * CONTAINS_OP
  * LIST_EXTEND
  * LIST_TO_TUPLE
  * DICT_MERGE
* Bump Python version in setup.py.
* Add Python 3.9 github test workflow.
 
Test:
* All unit tests passed.
* All models in torchbenchmark passed.
  
Issue: https://github.com/facebookresearch/torchdynamo/issues/33

### DIFF
--- a/torchdynamo/_eval_frame.c
+++ b/torchdynamo/_eval_frame.c
@@ -92,20 +92,65 @@ inline static void eval_frame_callback_set(PyObject *obj) {
 #endif
 
 static void ignored(void *obj) {}
-static PyObject *custom_eval_frame(PyFrameObject *frame, int throw_flag);
-static PyObject *custom_eval_frame_run_only(PyFrameObject *frame,
+static PyObject *_custom_eval_frame(PyThreadState *tstate,
+                                    PyFrameObject *frame,
+                                    int throw_flag);
+static PyObject *_custom_eval_frame_run_only(PyFrameObject *frame,
                                             int throw_flag);
+#if PY_VERSION_HEX >= 0x03090000
+static PyObject *custom_eval_frame(PyThreadState *tstate,
+                                   PyFrameObject *frame,
+                                   int throw_flag) {
+  return _custom_eval_frame(tstate, frame, throw_flag);
+}
+static PyObject *custom_eval_frame_run_only(PyThreadState *tstate,
+                                            PyFrameObject *frame,
+                                            int throw_flag) {
+  return _custom_eval_frame_run_only(frame, throw_flag);
+}
+#else
+static PyObject *custom_eval_frame(PyFrameObject *frame, int throw_flag) {
+  PyThreadState *tstate = PyThreadState_GET();
+  return _custom_eval_frame(tstate, frame, throw_flag);
+}
+static PyObject *custom_eval_frame_run_only(PyFrameObject *frame,
+                                            int throw_flag) {
+  return _custom_eval_frame_run_only(frame, throw_flag);
+}
+#endif
+
+inline static PyObject *eval_frame_default(PyFrameObject *frame,
+                                           int throw_flag) {
+  #if PY_VERSION_HEX >= 0x03090000
+  PyThreadState *tstate = PyThreadState_GET();
+  return _PyEval_EvalFrameDefault(tstate, frame, throw_flag);
+  #else
+  return _PyEval_EvalFrameDefault(frame, throw_flag);
+  #endif
+}
 
 inline static void enable_eval_frame(PyThreadState *tstate) {
+  #if PY_VERSION_HEX >= 0x03090000
+  _PyInterpreterState_SetEvalFrameFunc(tstate->interp, &custom_eval_frame);
+  #else
   tstate->interp->eval_frame = &custom_eval_frame;
+  #endif
 }
 
 inline static void disable_eval_frame(PyThreadState *tstate) {
+  #if PY_VERSION_HEX >= 0x03090000
+  _PyInterpreterState_SetEvalFrameFunc(tstate->interp, &_PyEval_EvalFrameDefault);
+  #else
   tstate->interp->eval_frame = &_PyEval_EvalFrameDefault;
+  #endif
 }
 
 inline static void enable_run_only_eval_frame(PyThreadState *tstate) {
+  #if PY_VERSION_HEX >= 0x03090000
+  _PyInterpreterState_SetEvalFrameFunc(tstate->interp, &custom_eval_frame_run_only);
+  #else
   tstate->interp->eval_frame = &custom_eval_frame_run_only;
+  #endif
 }
 
 static inline PyObject *call_callback(PyObject *callable, PyObject *frame,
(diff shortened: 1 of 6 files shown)


torchdynamo/bytecode_analysis.py
```diff
@@ -159,5 +159,5 @@ def stacksize_analysis(instructions):
         return low + 32
 
     assert fixed_point.value, "failed to reach fixed point"
-    assert low >= 0
+    # assert low >= 0
```
yanboliang, line 162: Line 162 doesn't pass 5 unit tests and line 161 breaks ```maml``` model in torchbench. I'm working on debug this, would be happy if you have any suggestion. @jansel

jansel: Usually this error means we generated invalid bytecode.  There is a commented out print earlier in the same function I've used to debug that in the past.

yanboliang: The print out helper finally helped me find & fix the bug, which is because I got messed up between comment & comment out of adding ```RERAISE``` to ```TERMINAL_OPCODES``` as I have to test with Python 3.7/3.8/3.9 at the same time. Now the code is in good shape and passed all unit tests & benchmark models.

jansel: Great!   Glad we caught that one.

torchdynamo/symbolic_convert.py
```diff
@@ -898,6 +898,79 @@ def BUILD_STRING(self, inst):
             result = str_var.value + result
         self.push(ConstantVariable(value=result))
 
+    def IS_OP(self, inst):
+        left, right = self.popn(2)
+        options = VariableTracker.propagate([left, right])
+        op = inst.argval
+        if (
```
jansel, line 905: Is this code copy and pasted from `COMPARE_OP`?  Could we just call `COMPARE_OP` to reuse code?

torchdynamo/symbolic_convert.py
```diff
@@ -898,6 +898,79 @@ def BUILD_STRING(self, inst):
...
+        if op == 1:
+            self.UNARY_NOT(inst)
+
+    def LIST_EXTEND(self, inst):
+        v = self.pop()
+        assert inst.argval > 0
+        obj = self.stack[-inst.arg]
+        assert isinstance(obj, ListVariable)
+        assert obj.mutable_local
+        list.extend(obj.items, list(v.unpack_var_sequence(self)))
```
jansel, line 947: Could we reuse existing code here?  Perhaps something like:
```
obj.call_method(tx, "extend", [v], {})
```

jansel, line 947: Could we reuse existing code here?  Perhaps something like:
```
obj.call_method(tx, "extend", [v], {})
```

torchdynamo/symbolic_convert.py
```diff
@@ -898,6 +898,79 @@ def BUILD_STRING(self, inst):
...
+    def LIST_TO_TUPLE(self, inst):
+        self.push(BuiltinVariable(tuple).call_function(self, [self.pop()], {}))
+
+    def DICT_MERGE(self, inst):
+        v = self.pop()
+        assert inst.argval > 0
+        obj = self.stack[-inst.arg]
+        assert isinstance(obj, ConstDictVariable)
+        assert obj.mutable_local
+        collections.OrderedDict.update(obj.items, v.items)
```
jansel, line 965: Could we reuse existing code here by calling `obj.call_method`?

jansel, line 965: Could we reuse existing code here by calling `obj.call_method`?

torchdynamo/_eval_frame.c
```diff
@@ -238,10 +283,11 @@ inline static PyObject *eval_custom_code(PyThreadState *tstate,
   DEBUG_CHECK(nlocals_new >= nlocals_old);
 
   PyFrameObject *shadow =
-      _PyFrame_New_NoTrack(tstate, code, frame->f_globals, NULL);
+      PyFrame_New(tstate, code, frame->f_globals, NULL);
   if (shadow == NULL) {
     return NULL;
   }
+  PyObject_GC_UnTrack(shadow);
```
jansel, line 290: I think at the end of this function we call `PyObject_GC_Track(shadow)`, so I think this was an optimization I copied from `cpython/.../call.c` to avoid ever starting to track the frame.  Given this patch removes that optimization, perhaps we can simplify things.

I'd suggest removing this line, then replacing L308 to L315 with just `Py_DECREF(shadow)`.

jansel, line 290: I think at the end of this function we call `PyObject_GC_Track(shadow)`, so I think this was an optimization I copied from `cpython/.../call.c` to avoid ever starting to track the frame.  Given this patch removes that optimization, perhaps we can simplify things.

I'd suggest removing this line, then replacing L308 to L315 with just `Py_DECREF(shadow)`.

torchdynamo/_eval_frame.c
```diff
@@ -92,20 +92,65 @@ inline static void eval_frame_callback_set(PyObject *obj) {
...
+static PyObject *custom_eval_frame_run_only(PyFrameObject *frame,
+                                            int throw_flag) {
+  return _custom_eval_frame_run_only(frame, throw_flag);
+}
+#endif
+
+inline static PyObject *eval_frame_default(PyFrameObject *frame,
+                                           int throw_flag) {
+  #if PY_VERSION_HEX >= 0x03090000
+  PyThreadState *tstate = PyThreadState_GET();
```
jansel, line 125: `PyThreadState_GET()` is expensive, which is why it is passed in in 3.9+.

We should avoid adding extra calls to it.  I'd suggest refactoring this to take an optional `PyThreadState *tstate`.  Then only call PyThreadState_GET() if `tstate==NULL`.

Most of the callers will already have a tstate pointer.

jansel, line 125: `PyThreadState_GET()` is expensive, which is why it is passed in in 3.9+.

We should avoid adding extra calls to it.  I'd suggest refactoring this to take an optional `PyThreadState *tstate`.  Then only call PyThreadState_GET() if `tstate==NULL`.

Most of the callers will already have a tstate pointer.

jansel (changes requested):

jansel (approved):

PR #155, opened 2022-04-19
Variable builder - handle slice

anijain2305:
As Title

### DIFF
--- a/torchdynamo/variables/builder.py
+++ b/torchdynamo/variables/builder.py
@@ -40,6 +40,7 @@ from .lists import ListIteratorVariable
 from .lists import ListVariable
 from .lists import NamedTupleVariable
 from .lists import RangeVariable
+from .lists import SliceVariable
 from .lists import TupleVariable
 from .misc import AutogradFunctionVariable
 from .misc import InspectSignatureVariable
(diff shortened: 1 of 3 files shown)


torchdynamo/variables/builder.py
```diff
@@ -265,6 +266,13 @@ def _wrap(self, value):
             return DataClassVariable.wrap(self, value).add_guards(
                 make_guards(GuardBuilder.TYPE_MATCH)
             )
+        elif isinstance(value, slice):
+            start = ConstantVariable(value.start)
+            stop = ConstantVariable(value.stop)
+            step = ConstantVariable(value.step)
+            return SliceVariable(
+                [start, stop, step], guards=make_guards(GuardBuilder.CONSTANT_MATCH)
```
anijain2305, line 274: Is `CONSTANT_MATCH` the right type of guard? The bytecode does not change if we have `slice(0, 2, 1)` or `slice(0, 6, 2)`. So, I think `CONSTANT_MATCH` might be too strict.

I tried `FUNCTION_MATCH`, but that too was causing too many recompilations. Any suggestions?

jansel: Yes I think CONSTANT_MATCH is correct with this handling.  It will recompile with different slices.

jansel (approved):

Issue #133, opened 2022-04-12
Make torchdynamo not import third party package in `skipfiles.py`


jansel:
@xuzhao9 in https://github.com/facebookresearch/torchdynamo/issues/107#issuecomment-1095681515 found that the following line makes alexnet 18% slower: 

https://github.com/jansel/torchdynamo/blob/bf90b8cdbacf35944fa8c12185b1823dc5cb90bb/torchdynamo/skipfiles.py#L123

It seems importing:  "networkx", "omegaconf", "onnx", "pandas",  and "sklearn" cause performance issues.

TorchDynamo is only importing these modules to find the filename, which is also a bit wasteful.  We should rewrite `skipfiles.py` to use [find_spec](https://docs.python.org/3/library/importlib.html#importlib.abc.PathEntryFinder.find_spec) instead, so we don't need to import unused packages.

Also, I think we can cut down the list of modules in skipfiles dramatically.  Most of those were added when TorchDynamo didn't automatically skip backends and supported much less of python, so likely many (most?) can be removed.

xuzhao9:
Can we skip tracing all non-pytorch source code, instead of enumerating all common third party libraries?

jansel:
We could have some sort of heuristic to decide if we should analyze a given frame.  E.g. if neither args nor globals contain `torch.*` objects, just skip the frame.

jansel:
For more on skipping see https://github.com/facebookresearch/torchdynamo/issues/69

Chillee:
Another annoying thing about this is that it makes import times significantly longer.

For example, we have a `skip_dynamo` decorator in AOTAutograd. These imports mean that whenever we import AOTAutograd, we now also need to import {tensorflow, onnx, etc.}

xuzhao9:
Reopening this issue since we are reverting #153

PR #160, opened 2022-04-20
Revert "Fix import overhead by using `importlib.util.find_spec` (#153)"

jansel:
This reverts commit dedd9fa8ef8e7ef9bdf286a58001476491fc30a6.

This PR seems to cause a lot more files to get skipped than before.

cc @xuzhao9 @anijain2305

### DIFF
--- a/torchdynamo/skipfiles.py
+++ b/torchdynamo/skipfiles.py
@@ -85,14 +85,13 @@ FILENAME_ALLOWLIST = {
 }
 
 
-def add(module_spec: typing.Optional[importlib.machinery.ModuleSpec]):
-    if not module_spec:
-        return
+def add(module: types.ModuleType):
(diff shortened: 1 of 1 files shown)


xuzhao9 (commented):
Is there a better way to avoid the trouble caused by importing modules?

xuzhao9 (approved):
Approved to unblock the development. However IMO we should avoid importing modules here and figure out why removing import caused more files to be skipped

anijain2305 (approved):

anijain2305:
Thanks for looking. I could not get a chance today.

jansel:
> Approved to unblock the development. However IMO we should avoid importing modules here and figure out why removing import caused more files to be skipped. Reopening #133

Agreed, we should figure out the bug and resubmit the PR.

PR #162, opened 2022-04-20
enable support for classmethod on superclass

vkuzo:
Summary:

Adds support for tracing through this syntax:

```
class Parent(torch.nn.Module):
    @classmethod
    def foo(cls, x):
        x = x + x
        return x

class Child(Parent):
    @classmethod
    def helper(cls, x):
        // resolving super().foo failed before this PR
        x = super().foo(x)
        return x

    def forward(self, x):
        x = self.helper(x)
        return x
```

This is useful for eventually enabling __torch_function__ support.

Note: I don't know much about the codebase, please feel free to point out
n00b shortcomings. I left some questions in the implementation which
would be good to resolve before landing.

Test plan:

```
pytest -vsk test_super_class_method
```

### DIFF
--- a/tests/test_modules.py
+++ b/tests/test_modules.py
@@ -427,6 +427,24 @@ class SuperModule(BasicModule):
         return x + 10.0
 
 
+class ComplicatedSuperParent(torch.nn.Module):
+    @classmethod
+    def custom_add(cls, x):
+        x = x + x
+        return x
+
+
+class SuperChildCallsClassMethod(ComplicatedSuperParent):
+    @classmethod
+    def child_func(self, x):
+        x = super().custom_add(x)
+        return x
+
+    def forward(self, x):
+        x = self.child_func(x)
+        return x
+
+
 class HasAttrModule(torch.nn.Module):
     def __init__(self):
         super().__init__()
@@ -531,6 +549,7 @@ class NNModuleTests(torchdynamo.testing.TestCase):
     test_stringmember = make_test(StringMember())
     test_modulelist = make_test(ModuleList())
     test_super1 = make_test(SuperModule())
+    test_super_class_method = make_test(SuperChildCallsClassMethod())
     test_children = make_test(Children())
     test_densenet = make_test(DenseNetBlocks())
     test_parameters1 = make_test(ParametersModule1())
--- a/torchdynamo/variables/misc.py
+++ b/torchdynamo/variables/misc.py
@@ -34,8 +34,15 @@ class SuperVariable(VariableTracker):
     def const_getattr(self, tx, name):
         assert self.objvar, "1-arg super not implemented"
         search_type = self.typevar.as_python_constant()
+
+        # We default to the python type of the object. However, if this is a
+        # `type`, then the original object represents the user defined type.
+        type_to_use = self.objvar.python_type()
+        if type_to_use is type:
+            type_to_use = self.objvar.value
+
         # TODO(jansel): there is a small chance this could trigger user code, prevent that
-        return getattr(super(search_type, self.objvar.python_type()), name)
+        return getattr(super(search_type, type_to_use), name)
 
     def call_method(
         self,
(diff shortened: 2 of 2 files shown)


tests/test_modules.py
```diff
@@ -427,6 +427,24 @@ def forward(self, x):
...
 
 
+class ComplicatedSuperParent(torch.nn.Module):
+    @classmethod
+    def custom_add(cls, x):
+        x = x + x
+        return x
+
+
+class SuperChildCallsStaticMethod(ComplicatedSuperParent):
```
vkuzo, line 437: this should be named `SuperChildCallsClassMethod`, will fix before landing

torchdynamo/symbolic_convert.py
```diff
@@ -1125,6 +1123,12 @@ def inline_call_(parent, func, args, kwargs):
             )
 
         try:
+            # TODO(before land): I didn't research this in detail, need to check
+            # what else we need to handle here
+            if func.has_self():
+                # remove self
+                assert args[0] is args[1]
+                args = args[1:]
```
jansel, line 1131: There are two values of `self`:
1) `self` at compile time
2) `self` at call time

I think this will incorrectly use the first at call time, which may be incorrect.

torchdynamo/variables/functions.py
```diff
@@ -77,9 +77,10 @@ class UserFunctionVariable(BaseUserFunctionVariable):
 
     def __init__(self, fn, **kwargs):
         super(UserFunctionVariable, self).__init__(**kwargs)
+        # TODO(before land): verify if this is okay
         assert isinstance(
-            fn, types.FunctionType
-        ), f"expected FunctionType {typestr(fn)} {fn}"
+            fn, (types.FunctionType, types.MethodType)
```
jansel, line 82: I don't think we can inline a method like this, because method has an object embedded in it `method.__self__`. To handle methods properly, we need to "unpack" it.  See `UserMethodVariable` below.

To handle this:
1) Pull out `method.__self__`.  You can construct a `VariableTracker` for `method.__self__` with something like:
```
self_var = VariableBuilder(tx, AttrSource(method_var.source, "__self__"))(method_var.value.__self__)
```
This will take an example value of `__self__` and build a Variable around it.  TorchDynamo will read the value dynamically from the `__self__` attribute.


2) Construct a `UserMethodVariable(method_var.value.__func__, self_var, ...)` (need to guard on this properly, but can ignore that for early testing)

3) Given a `UserMethodVariable`, torchdynamo should inline it automatically.

jansel (changes requested):

torchdynamo/variables/functions.py
vkuzo: to clarify, would the right place for the logic above be as an additional case in `UserMethodVariable.call_function`?

torchdynamo/symbolic_convert.py
vkuzo: just to make sure I understand your comment correctly,
1. current code hackily removes self to ensure the number of arguments is right
2. you are pointing out that there are other things we need to handle, such as value of self during call time not matching value of self during compile time (which my current code doesn't do anything about)

is that a correct understanding?

if yes, would implementing your suggestion below result in a the method call having `is_self` equal to False, because it would be unpacked?

torchdynamo/variables/functions.py
jansel: You want to do this logic when you construct the Variable.  You want `method()` instances to map to UserMethodVariable (not UserFunctionVariable as they do in this PR)

vkuzo: hmm, I might be missing something.  The reason I added the `types.MethodType` here is because `UserMethodVariable` inherits from `UserFunctionVariable`, and this code gets hit as a part of the constructor.  The current code (in misc.py) does create a `UserMethodVariable` already (the changes below, in this PR).

vkuzo: current code in misc.py:

```
        elif isinstance(inner_fn, types.FunctionType):
            return variables.UserFunctionVariable(inner_fn, **options).call_function(
                tx, [self.objvar] + args, kwargs
            )
        elif isinstance(inner_fn, types.MethodType):
            return variables.UserMethodVariable(inner_fn, self.objvar, **options).call_function(
                tx, [self.objvar] + args, kwargs
            )
        else:
            unimplemented(f"non-function or method super: {inner_fn}")
```

This does create `UserMethodVariable`, it just doesn't do any of the special handling of `__self__`.  Should I add the extra self handling there?

jansel: I think you just need to do `inner_fn.__func__` to get the underlying types.FunctionType.
```py
>>> class Foo:
...   def bar(self):
...     pass
... 
>>> type(Foo().bar)
<class 'method'>
>>> type(Foo().bar.__func__)
<class 'function'>
>>> 
```

vkuzo: perfect, thx, that worked

jansel (approved):

PR #163, opened 2022-04-21
Break graph on Tensor grad

anijain2305:


torchdynamo/variables/tensor.py
```diff
@@ -70,6 +70,8 @@ def create(cls, tx, proxy, example_value=None, nnmodule=None, **options):
 
         if isinstance(example_value, torch.Tensor):
             proxy.node.meta["example_value"] = example_value.clone()
+            if example_value.grad is not None:
```
anijain2305, line 73: There is a hiccup here. PyTorch gives this warning message for the added testcase

~~~
/data/home/anijain/miniconda/envs/pytorch_dev/lib/python3.8/site-packages/torch/_tensor.py:1082: UserWarning: The .grad attribute of a Tensor that is not a leaf Tensor is being accessed. Its .grad attribute won't be populated during autograd.backward(). If you indeed want the .grad field to be populated for a non-leaf Tensor, use .retain_grad() on the non-leaf Tensor. If you access the non-leaf Tensor by mistake, make sure you access the leaf Tensor instead. See github.com/pytorch/pytorch/pull/30531 for more informations. (Triggered internally at  aten/src/ATen/core/TensorBody.h:481.)
~~~

jansel: A hrm.  Can we use the `warnings` module to suppress this?

torchdynamo/variables/tensor.py
```diff
@@ -70,6 +70,8 @@ def create(cls, tx, proxy, example_value=None, nnmodule=None, **options):
...
+                proxy.node.meta["example_grad_value"] = example_value.grad.clone()
```
jansel, line 74: Thoughts on `"example_grad_value"` versus `meta["example_value"].grad` ?

The latter might propagate through some PyTorch ops to other example values, so I'd lean towards that -- but perhaps there are pros of this one I havent thought of?

torchdynamo/variables/tensor.py
anijain2305: Done

torchdynamo/variables/tensor.py
anijain2305: I like the `.grad` (latter) version better too. Made the changes.

jansel (approved):

PR #164, opened 2022-04-21
Options to print fx/aot traces

bertmaher:
I've heard from a few different people who want to capture traces of popular workloads for analysis, and dynamo/fx/aot_autograd make this possible, and this script makes running torchbench fairly tractable.


jansel (approved):
Thanks!

anijain2305 (approved):
lgtm

We actually have AOTAutograd backend in TorchDyanmo - https://github.com/facebookresearch/torchdynamo/blob/main/torchdynamo/optimizations/backends.py#L496

PR #167, opened 2022-04-22
enable basic __torch_function__ support

vkuzo:
Summary:

This adds a skeleton for `__torch_function__` support in torchdynamo.

What this is currently doing:
1. in variable builder, check for `__torch_function__` and wrap tensors in `TensorTFOverrideVariable` if found
1. in `TorchVariable.call_function`, inline the `__torch_function__` function of `TensorTFOverrideVariable` arguments
2. in `GetAttrVariable.call_function`, check for `super().__torch_function__` which resolves to the original. If it's found, stop inlining and insert the function call into the graph.

The current test just creates a `__torch_function__` override which doesn't do
anything but call `super().__torch_function__`.

Things left for future PRs:
1. supporting `call_method`
2. supporing actual logic inside the overrides
3. implementing the full `__torch_function__` spec (currently things are hardcoded to first argument only)

Test plan:

```
pytest -vsk test_simple_torch_function
// used to fail with https://www.internalfb.com/phabricator/paste/view/P496842415
// currently passes
```

### DIFF
--- a/tests/test_modules.py
+++ b/tests/test_modules.py
@@ -594,3 +594,28 @@ class NNModuleTests(torchdynamo.testing.TestCase):
         self.assertTrue(torchdynamo.testing.same(out2, out3))
         self.assertTrue(torchdynamo.testing.same(out2, out4))
         self.assertEqual(cnt.frame_count, 3)
+
+    def test_simple_torch_function(self):
+        def foo(x):
+            # function call, twice to test wrapping
+            x = F.sigmoid(x)
+            x = F.sigmoid(x)
+            # TODO(future PR): support method calls
+            # x = x.sigmoid()
+            return x
+
+        class TensorProxy(torch.Tensor):
+            @classmethod
+            def __torch_function__(cls, func, types, args=(), kwargs=None):
+                return super().__torch_function__(func, types, args, kwargs)
+
+        torchdynamo.config.traceable_tensor_subclasses.add(TensorProxy)
+
+        x = torch.randn(1).as_subclass(TensorProxy)
+        cnt = torchdynamo.testing.CompileCounter()
+        out1 = foo(x)
+        with torchdynamo.optimize(cnt, nopython=True):
+            out2 = foo(x)
+
+        self.assertEqual(cnt.op_count, 2)
+        self.assertTrue(torchdynamo.testing.same(out1, out2))
--- a/torchdynamo/config.py
+++ b/torchdynamo/config.py
@@ -40,3 +40,7 @@ dynamic_propagation = True
 
 # run FX normalization passes in optimizer
 normalize_ir = True
+
+# If a tensor subclass type is in this set, torchdynamo will inline the
+# __torch_function__ logic of the subclass.
+traceable_tensor_subclasses = set()
--- a/torchdynamo/variables/misc.py
+++ b/torchdynamo/variables/misc.py
@@ -13,6 +13,7 @@ from ..guards import GuardBuilder
 from ..guards import GuardSource
 from ..source import AttrSource
 from ..utils import identity
+from ..utils import proxy_args_kwargs
 from .base import VariableTracker
 
 
@@ -35,10 +36,14 @@ class SuperVariable(VariableTracker):
         assert self.objvar, "1-arg super not implemented"
         search_type = self.typevar.as_python_constant()
 
-        # We default to the python type of the object. However, if this is a
-        # `type`, then the original object represents the user defined type.
+        # We default to the python type of the object. However,
+        # 1. If this is a `type`, then the original object represents the user
+        # defined type.
+        # 2. If this is `torch._C._TensorMeta`, the original object is the user
+        # defined type of a custom tensor subclass.
+        # TODO(future PR): figure out how to do this in a less hacky way
         type_to_use = self.objvar.python_type()
-        if type_to_use is type:
+        if type_to_use is type or type_to_use is torch._C._TensorMeta:
             type_to_use = self.objvar.value
 
         # TODO(jansel): there is a small chance this could trigger user code, prevent that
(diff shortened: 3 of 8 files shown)


torchdynamo/variables/torch.py
```diff
@@ -148,6 +149,56 @@ def call_function(
...
+                    # does not have `self`
+                    func = tensor_val.__torch_function__.__func__
+
+                    # For now, we inline the object instance's __torch_function__ directly
+                    # In the future, it might be worth considering inlining
+                    # the `torch.overrides.handle_torch_function` method instead.
+
+                    # first, get the source
+                    # TODO: is this the right type of source to use?
+                    source = LocalSource('__torch_function__')
```
jansel, line 175: This line is the root cause of your `NameError: name '__torch_function__' is not defined` error.

`LocalSource('__torch_function__')` tells the guard that it can get this value from the local `'__torch_function__'`, which doesn't exist and triggers the error.

The "minimal" fix would be:
```suggestion
                    assert args[0].source
                    source = AttrSource(AttrSource(args[0].source, '__torch_function__'), '__func__')
```

Which, will cause the guard to read `tensor.__torch_function__.__func__`.


----

You might want to try the above just to convince yourself it works.

However, I think a better approach would be:

1) Check for the existence of `__torch_function__` in `variable_builder.py`
2) Map tensors with `__torch_function__` to a new `TensorSubclassVariable` instead of `TensorVariable`. `TensorSubclassVariable` could contain a `TensorVariable` as a property.  We do this because we want to eliminate the tensor subclass from the generated code, so don't want the default tensor handling.
3) When we do operations on a `TensorSubclassVariable`, just inline calls to `__torch_function__` similar to what you do here.

torchdynamo/variables/torch.py
```diff
@@ -148,6 +149,56 @@ def call_function(
...
+                    # next, create the variable tracker
+                    func_var = VariableBuilder(tx, source)(func)
```
jansel, line 177: The guard (reading the invalid value) is implicitly being generated by this call.  

I don't think we actually need a guard here.  [We already guard on the Python class of the tensor](https://github.com/facebookresearch/torchdynamo/blob/688448565e36f1e295fd213d6ab74b23e1faa4d9/torchdynamo/_guards.cpp#L143), so unless someone monkey patches `__torch_function__` it should be safe to instead do:
```suggestion
                    func_var = UserMethodVariable(func, <cls/self VariableTracker>, **options)
```
Which won't add any new guards (but propagate existing guards through `options`).

torchdynamo/variables/torch.py
```diff
@@ -148,6 +149,56 @@ def call_function(
...
+                    # TODO: source here is reused, probably incorrect
+                    # print('TYPE', type(tensor_val))
+                    type_var = VariableBuilder(tx, source)(type(tensor_val))
```
jansel, line 180: This will add another invalid guard, because `source` doesn't correctly say how to read the type at runtime.  

Since we already guard on the class of the Tensor, I think we can just construct this directly and not create a guard:
https://github.com/facebookresearch/torchdynamo/blob/688448565e36f1e295fd213d6ab74b23e1faa4d9/torchdynamo/variables/user_defined.py#L27

tests/test_modules.py
```diff
@@ -594,3 +596,28 @@ def test_self_mutating1(self):
         self.assertTrue(torchdynamo.testing.same(out2, out3))
         self.assertTrue(torchdynamo.testing.same(out2, out4))
         self.assertEqual(cnt.frame_count, 3)
+
+    def test_simple_torch_function(self):
+        def my_compiler(gm: torch.fx.GraphModule, example_inputs: List[torch.Tensor]):
```
jansel, line 601: For unittests I'd suggest using `CompilerCounter` and asserting something about the number of captured ops.  See above for examples.

torchdynamo/config.py
```diff
@@ -6,9 +6,13 @@
 
 # print out lots of stuff
 debug = False
+# TODO(before land): revert
+debug = True
```
jansel, line 10: remove this

torchdynamo/config.py
```diff
@@ -6,9 +6,13 @@
...
 
 # an unreasonable amount of debug printouts
 trace = False
+# TODO(before land): revert
+trace = True
```
jansel, line 15: remove this

torchdynamo/variables/misc.py
```diff
@@ -35,10 +36,13 @@ def const_getattr(self, tx, name):
...
-        # We default to the python type of the object. However, if this is a
-        # `type`, then the original object represents the user defined type.
+        # We default to the python type of the object. However,
+        # 1. If this is a `type`, then the original object represents the user
+        # defined type.
+        # 2. If this is `torch._C._TensorMeta`, the original object is the user
+        # defined type of a custom tensor subclass.
         type_to_use = self.objvar.python_type()
-        if type_to_use is type:
+        if type_to_use is type or type_to_use is torch._C._TensorMeta:
```
jansel, line 45: I wonder if there is a better way to handle this, as it feels a bit like a hack to do different things based on the value.  Perhaps we could detect we are in a `@classmethod` and do something different.  Or look at the args to `super()` and do this logic based on if `super()` got a type or an object.  Ideally this would mirror what cpython does.

torchdynamo/variables/misc.py
```diff
@@ -269,6 +273,39 @@ def reconstruct(self, codegen):
...
+        if is_original_tensor_torch_function:
+            # Instead of tracing inside torch.Tensor.__torch_function__,
+            # record the `call_function` call into the graph.
+            from . import TensorVariable
+
+            original_torch_variable = args[0]
+            new_args = args[2].items
+            new_kwargs = args[3].items
+            options = VariableTracker.propagate(self, new_args, new_kwargs.values())
+            return TensorVariable.create(
```
jansel, line 299: Inside of `TensorVariable.create` we are constructing an `example_value`.  I believe constructing this example value will trigger a call to user code by calling torch_function again.

I believe there is a way to disable torch_function handling.  We should do that for the example_value construction.

torchdynamo/variables/torch.py
```diff
@@ -147,6 +151,39 @@ def call_function(
...
+                type_var,  # cls
+                self,  # func
+                (type_var,),  # types
+                TupleVariable(args),  # args
+                kwargs,  # kwargs
+            )
+            unwrapped = tx.inline_user_function_return(func_var, tf_args, {})
+            # Wrap the result, to ensure the override survives for the next
+            # function.
+            return TensorWithTFOverrideVariable(unwrapped)
```
jansel, line 186: Does a torch_function always return a wrapped tensor?  Or do we need to check for that?

torchdynamo/variables/torch.py
```diff
@@ -147,6 +151,39 @@ def call_function(
...
+            tensor_val = args[0].proxy.node.meta["example_value"]
+
+            # Note: the __func__ attribute call is to ensure the function
+            # does not have `self`
+            # Note: currently this does not guard for the source code of the
+            # override.
+            func_var = UserFunctionVariable(
+                tensor_val.__torch_function__.__func__, **options
+            )
+            type_var = UserDefinedObjectVariable(type(tensor_val), **options)
```
jansel, line 172: Should this be `UserDefinedClassVariable` (rather than object?)

jansel (changes requested):

jansel:
This is getting pretty close!  Thanks for working on it.

torchdynamo/variables/torch.py
vkuzo: thanks, added the relevant logic

torchdynamo/variables/torch.py
vkuzo: In the first version of this PR variable builder created it as `UserDefinedObjectVariable`.  It works either way, I defer to you which one is more correct, changed it.

jansel (approved):

Issue #69, opened 2022-03-17
Skip non-Tensor/Module frames


jansel:
People may have non-PyTorch code run under TorchDynamo.  We should make sure TorchDynamo does nothing in this case.

If TorchDynamo reaches the end of the frame without finding PyTorch ops, it will [just run the frame normally](https://github.com/facebookresearch/torchdynamo/blob/44971ffd9a7e6798b7868a592c337acb75bd1d2d/torchdynamo/symbolic_convert.py#L1056).  However, if there are unsupported things that prevent a whole-graph, TorchDyanmo could generate specialized frames for non-PyTorch code.   This should be correct, but could add extra overhead.

To improve this we should expand the logic in this function:
https://github.com/facebookresearch/torchdynamo/blob/44971ffd9a7e6798b7868a592c337acb75bd1d2d/torchdynamo/symbolic_convert.py#L1008
That function controls if TorchDynamo should break the graph and generate a `resume_at_xx` function to pick up after an unsupported thing.

The logic I would propose is: examine the stack, locals, and globals referenced by co_names; if there is a tensor/nn.Module/torch.* anywhere then, then keep doing what we do now; if there is not, just bail out and switch to normal execution.

jansel:
> The logic I would propose is: examine the stack, locals, and globals referenced by co_names; if there is a tensor/nn.Module/torch.* anywhere then, then keep doing what we do now; if there is not, just bail out and switch to normal execution.

On further discussion with folks such as @ezyang.  I think we should run this logic at the start of each frame, before we start doing any work to try to reduce overheads of symbolic analysis.

ezyang:
A very minor problem with the heuristics we know how to do is that this won't work:

```
def factory_wrapper(args):
    return torch.empty(args)

def foo(args):
    x = factory_wrapper(args)
    return do_something(x)
```

where `do_something` contains Torch API calls; we won't start analysis on foo. But this seems acceptable (there is no way to determine that we should trace this locally), and maybe if the user adds some type annotations we can pick up on this case too.

jansel:
We should make the decorator syntax:
```
@torchdynamo.optimize(...)
def foo(args):
   ...
```
force `foo.__code__` to pass the checks.  

That way users have a way around a false-fast-exit.

anijain2305:
Happy to look into this. Assigning to myself. Let me know if someone is already looking into this.

ezyang:
[...]

PR #171, opened 2022-04-25
Skip non tensor

anijain2305:
Starting to work on #69 

This remove pytest module related errors while running in PyTorch.

~~~
pytest -v -s test/test_torch.py::TestTorch::test_RNG_after_pickle
~~~

The PR is draft because
* Looking for initial reviews to ensure that direction is good. (cc @jansel)
* Do thorough performance testing with TorchBench on inference and training to ensure that important frames are not missed.
* Do thorough testing with PyTorch tests to see if we need to be more aggressive in skipping more frames.

### DIFF
--- a/torchdynamo/convert_frame.py
+++ b/torchdynamo/convert_frame.py
@@ -1,5 +1,6 @@
 import dis
 import functools
+import inspect
 import itertools
 import os
 import sys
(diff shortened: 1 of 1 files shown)


torchdynamo/convert_frame.py
```diff
@@ -96,6 +97,28 @@ def _fn(*args, **kwargs):
     return _fn
 
 
+def has_tensor_in_frame(frame):
+    # Check if the passed arguments are of type Tensor
+    for _, value in frame.f_locals.items():
```
jansel, line 102:
```suggestion
    for value in frame.f_locals.values():
```

torchdynamo/convert_frame.py
```diff
@@ -96,6 +97,28 @@ def _fn(*args, **kwargs):
...
+        if value is not None and isinstance(value, torch.Tensor):
+            return True
+
+    # Check if there is global import of torch.*
+    allowed_modules = ["torch"]
+    for _, value in frame.f_globals.items():
```
jansel, line 108: `frame.f_code.co_names` contains a list of all the global variable names this function could read.  

So we don't actually need to iterate over all globals, just the ones that the code could reference.

jansel (changes requested):
A few minor comments, but overall approach looks good.

torchdynamo/convert_frame.py
```diff
@@ -96,6 +97,28 @@ def _fn(*args, **kwargs):
...
+        if inspect.ismodule(value) and value.__name__ in allowed_modules:
```
jansel, line 109: For this test, we might want to reuse some code from `allowed_functions.py`

anijain2305:
Duplicate at https://github.com/pytorch/torchdynamo/pull/248

Issue #34, opened 2022-03-11
Add support for Python 3.10


jansel:
Python 3.9 (#33) should be added before 3.10.

Known issues for Python 3.10 support:

Python 3.10 and later have a new method for mapping the bytecode index to line number, deprecating co_lnotab.  See [PEP 626](https://peps.python.org/pep-0626/).  [lnotab_writer](https://github.com/facebookresearch/torchdynamo/blob/052b8dab6df5f027a34edb00dee73870433883d5/torchdynamo/bytecode_transformation.py#L58) in TorchDynamo needs to be rewritten to support the new format.   See #36 for more on line numbers.

There are [9 new/changed bytecodes in 3.10](https://docs.python.org/3.10/library/dis.html)

COPY_DICT_WITHOUT_KEYS
GET_LEN
MATCH_MAPPING
MATCH_SEQUENCE
MATCH_KEYS
MATCH_CLASS
Look like easy to add aliases for already supported things.

MAKE_FUNCTION
Handling of annotations changed.  We should add a test case to make sure annotations work for nested functions.

ROT_N
Seems easy to add.  We should update usage of [rot_n_helper](https://github.com/facebookresearch/torchdynamo/blob/052b8dab6df5f027a34edb00dee73870433883d5/torchdynamo/utils.py#L336) to use this new bytecode instead.

GEN_START
This could break handling of inline generators.  Need to look into this in more detail.

yanboliang:
Fixed by https://github.com/facebookresearch/torchdynamo/pull/172, now TorchDynamo supports Python 3.10!

PR #172, opened 2022-04-25
Add supports for Python 3.10

yanboliang:
Changes:
* Implement the new internal format of the line number table ```co_linetable``` and update the code object constructor keys inside of ```transform_code_object```.
* arg is offset of the instruction line rather than the bytecode for all jabs/jrel since python 3.10, so we have to correct the arg update logic inside of ```devirtualize_jumps```.
* Regarding to ```GEN_START```, the definition is to pop top of stack actually there is no other meaning. There are two ways to implement it: 
  * Just pass
  * Push ```None``` into the stack ahead of ```GEN_START``` and pop it when calling ```GEN_START```.

I tentatively choose option 2 but open to the other one. BTW, the ```GEN_START``` instruction will be replaced by ```POP_TOP``` in Python 3.11. See [bpo-46009](https://bugs.python.org/issue43683).
* Add several new bytecodes:
  * ```GET_LEN```
  * ```MATCH_MAPPING```
  * ```MATCH_SEQUENCE```
  * ```MATCH_KEYS```

* Use ```ROT_N``` rather than [rot_n_helper](https://github.com/facebookresearch/torchdynamo/blob/052b8dab6df5f027a34edb00dee73870433883d5/torchdynamo/utils.py#L336) for Python 3.10+
* The annotations of ```MAKE_FUNCTION``` is a tuple of strings since Python 3.10.
* Add github testing workflow for Python 3.10.

Tests:
* All unit tests passed.
* Torchbench doesn't support Python 3.10, there are 4-5 models need to update. This PR passed all other torchbench models. 

Issue: https://github.com/facebookresearch/torchdynamo/issues/34


yanboliang:
Interesting! both pytest and lint passed in my local env. It seems there are some dependent libraries version changed or not available for Python 3.10 on github env. Let me check it!

jansel:
Option 2 seems right to me.

For the dependency I think it is fine to just remove onnxruntime-gpu from requirements.txt.  I don't think it is actually used.

What about the other new bytecodes in 3.10 (mentioned in the #34)?  Perhaps we need to add some testcases to cover them.

yanboliang:
The the other new bytecodes and test cases will be added soon.

tests/test_functions.py
```diff
@@ -527,3 +527,28 @@ def test_fstrings3(x):
         tmp = f"{x.__class__.__name__} foo"
         if tmp.startswith("Tensor"):
             return x + 1
+
+    # # This is to test the new syntax for pattern matching
+    # # ("match ... case ...") added on python 3.10.
+    # # Uncomment these test cases if you run on 3.10+
+    # @make_test
+    # def test_match_sequence(a):
```
yanboliang, line 535: As Python(<3.10) doesn't support the ```match...case...``` syntax, so there are compile errors if we uncomment it when running py3.7-3.9 tests. Let me know if you have better solution. @jansel

torchdynamo/bytecode_transformation.py
```diff
@@ -121,9 +163,19 @@ def devirtualize_jumps(instructions):
...
+                    # for all jabs/jrel since python 3.10
+                    inst.arg = int(target.offset / 2)
             else:  # relative jump
-                inst.arg = target.offset - inst.offset - instruction_size(inst)
+                if sys.version_info < (3, 10):
+                    inst.arg = target.offset - inst.offset - instruction_size(inst)
+                else:
+                    inst.arg = int(
+                        (target.offset - inst.offset - instruction_size(inst)) / 2
+                    )
```
yanboliang, line 178: Actually this took the most time of the overall work, I don't find any doc describing this change. Is this because it's Python internal changes which is not expected to expose users?

torchdynamo/variables/functions.py
```diff
@@ -284,12 +284,20 @@ def reconstruct(self, codegen):
         if self.kwdefaults:
             flags |= 0x02
             codegen(self.kwdefaults)
-        if isinstance(self.annotations, variables.ConstDictVariable):
+        if isinstance(self.annotations, variables.ConstDictVariable) or isinstance(
```
yanboliang, line 287: There is a test case covering this, but here we don't require ```annotations``` should be ```ConstDictVariable``` so it doesn't break when I test on Python 3.10. Do we need to add ```assert``` before this line? @jansel

yanboliang (commented):
@jansel This is ready to review now. I document all changes in this PR's description. As torchbench doesn't support Python 3.10 yet so there are 4-5 models need to update. Is there anyone else who is working on that?

requirements.txt
```diff
@@ -2,10 +2,9 @@ black==22.1.0
...
 click==8.0.4
+numpy==1.21.2
 ninja
-onnxruntime-gpu
-onnx-tf
 pytest
 tabulate
-tensorflow-gpu
 torch
+Cython
```
jansel, line 10: Why is this needed?

requirements.txt
```diff
@@ -2,10 +2,9 @@ black==22.1.0
 flake8==4.0.1
 isort==5.10.1
 click==8.0.4
+numpy==1.21.2
```
jansel, line 5: Did another version cause issues? What happened here?

tests/test_functions.py
jansel: Let's just leave this commented out for now. I haven't seen a lot of users of this new syntax yet.

torchdynamo/bytecode_transformation.py
jansel: Glad you figure this out.  Did you look in the Python 3.10 release notes?  I'd be surprised if it weren't mentioned there.

torchdynamo/variables/functions.py
jansel: This seems like a reasonable fix.  I'll leave it up to you.

jansel (commented):
Overall looks good!

We should also update README.md.

requirements.txt
yanboliang: If we don't have specific version, it will pull the latest one, which is ```1.22.*```, which requires python 3.8+, so the py37 testing failed.

requirements.txt
yanboliang: Added by mistake, removed.

torchdynamo/bytecode_transformation.py
yanboliang: I didn't find it in Python 3.10 release notes, that what surprised me. I'll open an issue at cython to ask why they make this change and if they document them anywhere.

jansel (approved):

PR #174, opened 2022-04-26
Fixes for pytorch tests - 1/n

anijain2305:
Multiple changes in one PR (happy to break it down if needed)
* Filter torch.testing
* Adding is_complex, and is_contiguous
* Do not raise Dynamo assertion when Torch operation itself raises assertion. This is for cases like follows. Created a new type of assertion - TorchRuntimeError.
~~~
import torch
import torchdynamo

def fn(x, y):
    try:
        x.expand_as(y)
    except RuntimeError:
        return False
    return True

x = torch.randn(4)
y = torch.randn(10)
ref = fn(x, y)

    res = fn(x, y)

print("Success")
~~~

### DIFF
--- a/torchdynamo/allowed_functions.py
+++ b/torchdynamo/allowed_functions.py
@@ -53,6 +53,8 @@ def _allowed_function_ids():
     def _find_torch_objects(module):
         if module.__name__.startswith("torch.distributions"):
             return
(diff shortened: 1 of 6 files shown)


jansel (approved):

anijain2305:

PR #182, opened 2022-04-27
Fixes for pytorch tests 2/n - torch.Size and nn.Parameter

anijain2305:
* Save the Tensor Type information like `torch.nn.Parameter` vs `torch.Tensor`
* torch.Size - Handle reconstruction logic to return torch.Size instead of a tuple
* List check - Use EQUALS_MATCH because tensor can be indexed with a list. This might cause the case of too-much-specialization, but is needed for correctness.


anijain2305:
Looking into the failures. Seems like I broke something fundamental.

jansel (approved):

anijain2305:

PR #184, opened 2022-04-27
Fix Pytorch tests 3/n - Skip exec frame

anijain2305:
If we run the frame created by the `exec` builtin execution, TorchDynamo messes up the f_globals for the next frames.

This PR is a band-aid fix, where I am just skipping the frame created by the `exec`. I find such frames by checking the `frame.f_code.co_filename` equals `<string>`.


jansel (approved):

anijain2305:

PR #195, opened 2022-05-04
Detect x.new(torch.Size) and rewrite to torch.empty(tuple)

yanboliang:
Fix https://github.com/facebookresearch/torchdynamo/issues/191

### DIFF
--- a/torchdynamo/symbolic_convert.py
+++ b/torchdynamo/symbolic_convert.py
@@ -56,6 +56,7 @@ from .variables.functions import UserFunctionVariable
 from .variables.lists import BaseListVariable
 from .variables.lists import ListIteratorVariable
 from .variables.lists import ListVariable
+from .variables.lists import SizeVariable
 from .variables.lists import SliceVariable
 from .variables.lists import TupleVariable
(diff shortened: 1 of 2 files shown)


torchdynamo/symbolic_convert.py
```diff
@@ -553,6 +554,39 @@ def GET_ITER(self, inst):
...
+        # Hack to handle special case in some legacy code.
+        # Convert x.new(torch.Size) into torch.empty(tuple),
+        # as torch.Tensor.new acts differently with a Size input versus a tuple input.
+        if (
+            isinstance(fn, GetAttrVariable)
+            and isinstance(fn.obj, TensorVariable)
+            and fn.name == "new"
+            and len(args) == 1
+            and isinstance(args[0], SizeVariable)
+            and not config.dynamic_shapes
```
jansel, line 567: I'd suggest also checking that `fn.obj.dtype` / `fn.obj.device` are both non-None.  This will currently only happen if you disable dynamic_propagation.

jansel (changes requested):

jansel (approved):

PR #197, opened 2022-05-04
support inlining __torch_function__ with reading from closure

vkuzo:
Summary:

The previous PRs to add `__torch_function__` support inlined through
`__torch_function__` without adding any guards for the function.

This worked for simple cases, but did not work if `__torch_function__`
needs to read a nonlocal variable, for example: https://gist.github.com/vkuzo/a3388fcaa532318d049368e96652b366
The reason it was broken is because the code which bound arguments
during inlining had to have a reference to a source in order to bind
things properly.

One way to fix this is to get the source of the `__torch_function__` attribute
of the original tensor, guard on it, and persist it through
all the rewrapping logic.

I'm flexible if there is a better alternative, lmk.

Test plan:

```
pytest -vsk test_torch_function_with_closure
```

### DIFF
--- a/tests/test_modules.py
+++ b/tests/test_modules.py
@@ -620,3 +620,40 @@ class NNModuleTests(torchdynamo.testing.TestCase):
 
         self.assertEqual(cnt.op_count, 4)
         self.assertTrue(torchdynamo.testing.same(out1, out2))
+
+    def test_torch_function_with_closure(self):
+        def run():
+
+            counter = 0
+
+            def foo(x):
+                # function call, twice to test wrapping
+                x = F.sigmoid(x)
+                x = F.sigmoid(x)
+                # method call, twice to test wrapping
+                x = x.sigmoid()
+                x = x.sigmoid()
+                return x
+
+            class TensorProxy(torch.Tensor):
+                @classmethod
+                def __torch_function__(cls, func, types, args=(), kwargs=None):
+                    nonlocal counter
+                    # for now, only support reads from closure cells
+                    # TODO(future PR): support writes as well
+                    counter + 1
+                    return super().__torch_function__(func, types, args, kwargs)
+
+            torchdynamo.config.traceable_tensor_subclasses.add(TensorProxy)
+
+            x = torch.randn(1).as_subclass(TensorProxy)
+            x = torch.randn(1)
+            cnt = torchdynamo.testing.CompileCounter()
+            out1 = foo(x)
+            with torchdynamo.optimize(cnt, nopython=True):
+                out2 = foo(x)
+
+            self.assertEqual(cnt.op_count, 4)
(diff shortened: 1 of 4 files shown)


tests/test_modules.py
```diff
@@ -620,3 +620,40 @@ def __torch_function__(cls, func, types, args=(), kwargs=None):
 
         self.assertEqual(cnt.op_count, 4)
         self.assertTrue(torchdynamo.testing.same(out1, out2))
+
+    def test_torch_function_with_closure(self):
```
vkuzo, line 624: this has some overlap with `test_simple_torch_function`, but making parts of the code reusable seemed overkill, I'm flexible though

tests/test_modules.py
```diff
@@ -620,3 +620,40 @@ def __torch_function__(cls, func, types, args=(), kwargs=None):
...
+                x = x.sigmoid()
+                return x
+
+            class TensorProxy(torch.Tensor):
+                @classmethod
+                def __torch_function__(cls, func, types, args=(), kwargs=None):
+                    nonlocal counter
+                    # for now, only support reads from closure cells
+                    # TODO(future PR): support writes as well
+                    counter + 1
```
vkuzo, line 644: writing here doesn't actually crash the code, but saving the investigation to guarantee correctness of writing like this for a future PR

torchdynamo/variables/tensor.py
```diff
@@ -366,12 +367,14 @@ class TensorWithTFOverrideVariable(VariableTracker):
...
         self,
         tensor_variable,
+        orig_tensor_variable_source,
         subclass_torch_function__func,
         subclass_type,
         **kwargs,
     ):
         super(TensorWithTFOverrideVariable, self).__init__(**kwargs)
         self.tensor_variable = tensor_variable
+        self.orig_tensor_variable_source = orig_tensor_variable_source
```
jansel, line 377: Should this go in the `self.source` field (defined in the VariableTracker base class)?

The advantage of that is it will work with `resume_at_` continuations.

jansel (changes requested):

torchdynamo/variables/tensor.py
vkuzo: I actually tried that first, that led to problems in this line: https://github.com/pytorch/torchdynamo/blob/main/torchdynamo/codegen.py#L88, that statement would now pass for `TensorWithTFOverrideVariable` and it would lead to incorrect graph outputs.

We could unbreak it by doing something like

```
        if value.source is not None and allow_cache and not isinstance(value, TensorWithTFOverrideVariable):
            output.extend(value.source.reconstruct(self))
```

but that also looked pretty hacky.  WDYT?

jansel: Ah got it, I'm fine with this version then.

jansel (approved):

Issue #192, opened 2022-05-03
Torchdynamo with aot_autograd_speedup_strategy has increased memory usage and long overhead on ResNet50 model


xwang233:
python script (please correct me if I'm using the torchdynamo API wrong)
<details> 
<summary> dynamo-test.py </summary>
<p>

```python
import contextlib
import traceback
import time

import torch
import torchvision
import torchdynamo
from torchdynamo.optimizations.training import aot_autograd_speedup_strategy

N_WARMUP = 100
N_BENCH = 100

def bench(batch_size, use_dynamo):
    model = torchvision.models.resnet50().cuda()
    x = torch.randn(batch_size, 3, 224, 224, dtype=torch.float, device='cuda')

    train_context = torchdynamo.optimize(aot_autograd_speedup_strategy) if use_dynamo is True else contextlib.nullcontext()

    torch.cuda.synchronize()
    t0  = time.time()

    with train_context:
        for _ in range(N_WARMUP):
            out = model(x)
            out.sum().backward()

        torch.cuda.synchronize()
        t1 = time.time()

        for _ in range(N_BENCH):
            out = model(x)
            out.sum().backward()

        torch.cuda.synchronize()
        t2 = time.time()

    print('Training img/s (larger better):', batch_size / ((t2 - t1) / N_BENCH))
    print('Total time incl. overhead (smaller better):', t2 - t0)
    print()


if __name__ == '__main__':
    import argparse
    parser = argparse.ArgumentParser()
    parser.add_argument('-b', '--batch_size', type=int, default=16)
    parser.add_argument('--use_dynamo', action='store_true', default=False)
    args = parser.parse_args()
    print(args)

    bench(args.batch_size, args.use_dynamo)

```
</p>
</details>

bash script
```bash
#!/bin/bash
python dynamo-test.py -b 16
python dynamo-test.py -b 16 --use_dynamo
python dynamo-test.py -b 64
python dynamo-test.py -b 64 --use_dynamo
python dynamo-test.py -b 128
python dynamo-test.py -b 128 --use_dynamo
```

Tested with V100 16GB on
pytorch/pytorch@f6bbecf8b503eb6562d52b66124a49b45a57f14a ,
pytorch/vision@104073ccd5a5f06b1998e5e1a82fe813dced4cce,
torchdynamo @ 0d59ce9bfae18337a88de32743777e3a75c4ad5c,
pytorch/functorch@ac0fdf179748072cbacf0740650669ed763548ff,
[...]

PR #201, opened 2022-05-04
Pass backend-related ctx to TorchDynamo Optimize Context

anijain2305:
Background - For `aot_autograd_speedup_strategy`, the user can assume that `nvfuser` is turned on by default. But, for now, we have to wrap the model call in `with torch.jit.fuser("fuser2")` separately.

This PR extends Dynamo OptimizationCtx to accept extra ctx from the backend, and directly call `__enter__` and `__exit__`.

@Chillee @jansel

### DIFF
--- a/torchdynamo/eval_frame.py
+++ b/torchdynamo/eval_frame.py
@@ -25,7 +25,7 @@ def nothing():
     pass
 
 
-null_context = contextlib.nullcontext()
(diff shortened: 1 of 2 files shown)


jansel (approved):

anijain2305:
Fixes a part of https://github.com/pytorch/torchdynamo/issues/192

Issue #156, opened 2022-04-19
Run TorchDynamo on nn.Modules crawled from github


jansel:
Back in 2020 I did this project: 
https://github.com/jansel/pytorch-jit-paritybench

Which contains 8k `nn.Modules` crawled from 1k github projects.

At the time I used it to find that TorchScript worked on ~30% of projects crawled from github.  

We should run TorchDynamo on those testcases to 
1) get more datapoints on model coverage levels
2) find and fix more bugs in TorchDynamo

yanboliang:
Woohoo, it seems we have a good coverage:
```
TorchDynamo ParityBench:
          total  passing  score
projects   1111     1035  93.2%
tests      7549     7399  98.0%
```
I'll dig into the errors and check the missing features.

anijain2305:
Pretty awesome numbers for the first run :)

jansel:
Wow, awesome!  For context of those reading, TorchScript was ~30% projects and ~60% tests. So more than 3x the number of projects passing.

I assume that is allowing Python fallbacks (`nopython=False`, the default)?  

Seems like 100% might be in reach!

yanboliang:
> I assume that is allowing Python fallbacks (`nopython=False`, the default)?

That's right!

Chillee:
I'm curious about other stats, like "how big are the graphs", "how many graphs does it get split into", etc.

soumith:
this is pretty cool, and yea 100% with nopython=False seems like a no-brainer to aim for.

xuzhao9:
interesting! Curious what are the errors and we should include some delegates in torchbench to further improve coverage

yanboliang:
We hit our 100% goal! To whom might be interested in details of the errors, graph and graph break characteristics, model complexity, etc, please check our post [TorchDynamo Update 8: TorchDynamo passed correctness check on 7k+ github models](https://dev-discuss.pytorch.org/t/torchdynamo-update-8-torchdynamo-passed-correctness-check-on-7k-github-models/663).

voznesenskym:
> We hit our 100% goal! To whom might be interested in details of the errors, graph and graph break characteristics, model complexity, etc, please check our post [TorchDynamo Update 8: TorchDynamo passed correctness check on 7k+ github models](https://fb.workplace.com/notes/755492628976297).

Fantastic! There's also a lot of good low hanging fruit in nopython=True, fwiw, do we have plans to run that and group the most common errors?

[...]

PR #215, opened 2022-05-07
Refine ConstDictVariable python type and add wrapper to class argument

yanboliang:
*This is the 2/N issue found from https://github.com/pytorch/torchdynamo/issues/156.*

Problem:
* Failed test example: [```test_NVIDIA_retinanet_examples.py:MobileNet```](https://github.com/yanboliang/pytorch-jit-paritybench/blob/master/generated/test_NVIDIA_retinanet_examples.py)
* Error: ```TypeError: OrderedDict is not a Module subclass```

Minimal code to reproduce:
```
from typing import List
import torch
import torchdynamo
from torchvision.models import mobilenet

class MyMobileNet(mobilenet.MobileNetV2):

    def __init__(self, outputs=[18], url=None):
        self.stride = 128
        self.url = url
        super().__init__()
        self.outputs = outputs

    def forward(self, x):
        outputs = []
        for indx, feat in enumerate(self.features[:-1]):
            x = feat(x)
            if indx in self.outputs:
                outputs.append(x)
        return outputs

    return gm.forward

    model = MyMobileNet()
    model.eval()
    model(torch.rand([4, 3, 64, 64]))
```

Root causes and solutions
* Error at ```self.features[:-1]``` where calling the [```Sequential.__getitem__```](https://github.com/pytorch/pytorch/blob/c031643e392d2d5dda14c5876761265ac6bcf32a/torch/nn/modules/container.py#L101) and return ```self.__class__(OrderedDict(list(self._modules.items())[idx]))``` to construct ```Sequential```. Then jump to [```Sequential.__init__```](https://github.com/pytorch/pytorch/blob/c031643e392d2d5dda14c5876761265ac6bcf32a/torch/nn/modules/container.py#L82), where it checks ```isinstance(args[0], OrderedDict)```. We should pass this check as ```args[0]``` is ```OrderedDict```. However, torchdynamo wrap ```OrderedDict``` as ```ConstDictVariable``` and return python type ```dict```. Hence we can't pass this check, fall into another branch and throw error.   
* There are lots of functions contain class type arguments, e.g, ```MobileNetV2.__init__```, ```Conv2dNormActivation.__init__``` and ```ConvNormActivation.__init__```. Torchdynamo calls ```wrap_bound_arg``` to wrap arguments into variables, where misses the handling for class type arguments and throw error.

### DIFF
--- a/torchdynamo/symbolic_convert.py
+++ b/torchdynamo/symbolic_convert.py
@@ -585,7 +585,7 @@ class InstructionTranslatorBase(object):
     @break_graph_if_unsupported(push=1)
     def CALL_FUNCTION_EX(self, inst):
         if inst.argval == 0:
-            kwargsvars = ConstDictVariable({})
+            kwargsvars = ConstDictVariable({}, dict)
             argsvars = self.pop()
         elif inst.argval == 1:
             kwargsvars = self.pop()
@@ -721,7 +721,11 @@ class InstructionTranslatorBase(object):
             assert isinstance(k, ConstantVariable)
             result[k.value] = v
         assert len(result) == len(items) / 2
-        self.push(ConstDictVariable(result, mutable_local=MutableLocal(), **options))
+        self.push(
+            ConstDictVariable(
+                result, collections.OrderedDict, mutable_local=MutableLocal(), **options
+            )
+        )
 
     def BUILD_CONST_KEY_MAP(self, inst):
         keys = self.pop()
@@ -734,6 +738,7 @@ class InstructionTranslatorBase(object):
         self.push(
             ConstDictVariable(
                 collections.OrderedDict(zip(keys, values)),
+                collections.OrderedDict,
                 mutable_local=MutableLocal(),
                 **options,
             )
@@ -755,6 +760,7 @@ class InstructionTranslatorBase(object):
             obj,
             ConstDictVariable(
                 items,
+                collections.OrderedDict,
                 **VariableTracker.propagate([obj, k, v]),
             ),
         )
--- a/torchdynamo/variables/dicts.py
+++ b/torchdynamo/variables/dicts.py
@@ -14,11 +14,12 @@ from .base import VariableTracker
 
 
 class ConstDictVariable(VariableTracker):
-    def __init__(self, items, **kwargs):
+    def __init__(self, items, dict_cls=collections.OrderedDict, **kwargs):
         super(ConstDictVariable, self).__init__(**kwargs)
-        self._python_type = type(items)
+        self.dict_cls = dict_cls
         if not isinstance(items, collections.OrderedDict):
             assert isinstance(items, dict)
+            self.dict_cls = dict
             items = collections.OrderedDict((k, items[k]) for k in sorted(items.keys()))
         self.items = items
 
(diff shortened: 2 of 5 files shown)


jansel (changes requested):
1) can we add a test case for this?

2) There are a lot of places where we make copies of ConstDictVariable, I think we need to pass dict_cls through in these cases.  Search for calls to ConstDictVariable constructor.

torchdynamo/symbolic_convert.py
```diff
@@ -721,7 +721,11 @@ def BUILD_MAP(self, inst):
             assert isinstance(k, ConstantVariable)
             result[k.value] = v
         assert len(result) == len(items) / 2
-        self.push(ConstDictVariable(result, mutable_local=MutableLocal(), **options))
+        self.push(
+            ConstDictVariable(
+                result, collections.OrderedDict, mutable_local=MutableLocal(), **options
```
jansel, line 726: This should just be a normal dict

torchdynamo/symbolic_convert.py
```diff
@@ -734,6 +738,7 @@ def BUILD_CONST_KEY_MAP(self, inst):
         self.push(
             ConstDictVariable(
                 collections.OrderedDict(zip(keys, values)),
+                collections.OrderedDict,
```
jansel, line 741: This should just be a normal dict

torchdynamo/symbolic_convert.py
```diff
@@ -755,6 +760,7 @@ def MAP_ADD(self, inst):
             obj,
             ConstDictVariable(
                 items,
+                collections.OrderedDict,
```
jansel, line 763: This should copy `obj.dict_type`

torchdynamo/symbolic_convert.py
yanboliang: I'm not fully understand why this should be normal dict. As we wrap ```result``` as a ```ConstDictVariable``` whose ```dict_type``` should be the type of ```result``` which is ```OrderedDict```?

torchdynamo/variables/dicts.py
```diff
@@ -14,18 +14,22 @@
 
 
 class ConstDictVariable(VariableTracker):
-    def __init__(self, items, **kwargs):
+    def __init__(self, items, dict_cls=collections.OrderedDict, **kwargs):
         super(ConstDictVariable, self).__init__(**kwargs)
+        self.dict_cls = dict_cls
         if not isinstance(items, collections.OrderedDict):
             assert isinstance(items, dict)
+            self.dict_cls = dict
```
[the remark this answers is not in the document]
yanboliang: Agree to just wrap the object and keep the original dict type, either ```dict``` or ```OrderedDict```, depends on user code's type. Actually right now the sorting in L23 may break the original dict's order. Python ```dict``` keep the insert order rather than sorted order.
Another thing I want to discuss is actually the subclass ```DataClassVariable``` has a ```user_cls``` variable which is used as similar purpose. Do you think we can refactor to move this variable to the parent class? Not sure if the name is confusing, but looks more elegant.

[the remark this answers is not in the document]
jansel: I think we should remove OrderedDict entirely from ConstDictVariable, you are correct it gets the order wrong.

Yeah we could move the `user_cls` handling up to this class

yanboliang:
Since we need to do some refactor, so move to a more clean PR https://github.com/pytorch/torchdynamo/pull/226.

PR #216, opened 2022-05-07
[inductor] add heuristic to decide layouts and loop orders

jansel:


bertmaher (approved):
Going to go ahead and Accept to unblock you, and I’ll apply llvm’s tradition of post commit review 🙂

PR #217, opened 2022-05-07
[inductor] support torch.linspace and torch.tensor

jansel:


torchinductor/codegen/triton.py
```diff
@@ -56,7 +56,7 @@ def constant(value, dtype):
             return 'float("inf")'
         elif value == float("-inf"):
             return 'float("-inf")'
-        return super().constant(value, dtype)
+        return OpOverrides.constant(value, dtype)
```
bertmaher, line 60: Okay, showing my python ignorance here- why is OpOverrides preferable to super()?

bertmaher (approved):
Lgtm!

torchinductor/codegen/triton.py
jansel: `super()` doesn't work inside a `staticmethod`.  This was triggering an error for me.

PR #223, opened 2022-05-10
Remove reference cycle

anijain2305:
For the following test, Dynamo was holding on to more memory than necessary. And this memory was released after calling `gc.collect()`. This indicated that their was a reference cycle somewhere.

~~~

def print_mem(name):


def fn1():
    print_mem("Inside func")
    return x.sum()


fn1()
print_mem("Outside func")
print("\n\n\n")

    fn1()
    print_mem("Outside func")
print("\n\n\n")


graph = refcycle.garbage()
sccs = graph.strongly_connected_components()

refcycles = graph.source_components()
for idx, r in enumerate(refcycles):
    r.export_image(f"cycle{idx}.svg")

print(refcycles)
~~~

I used `refcycle` package to find the reference cycles, and following is the culprit.

image

This PR does the following
* Breaks the cycle by setting output_graph.root_tx = None after we are done with tracing
* Also, cleanup the graphargs, the data structure that was holding on to the tensor.

### DIFF
--- a/torchdynamo/convert_frame.py
+++ b/torchdynamo/convert_frame.py
@@ -149,6 +149,7 @@ def convert_frame_assert(compiler_fn: Callable, one_graph=True):
             )
             tracer.run()
             output = tracer.output
+            output.cleanup()
             assert output.output_instructions
             instructions[:] = output.output_instructions
             code_options.update(output.code_options)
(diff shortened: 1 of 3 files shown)


anijain2305:
Cross referencing #69 and #192

jansel (approved):
Awesome, glad we found this one!

PR #227, opened 2022-05-11
Bail out for __setattr__ and fix ClassVariable handling

wconstab:
- avoid compiling __setattr__ functions as they may be difficult
  to correctly handle for arbitrary custom classes, but also aren't
  likely to be useful for torch module optimization
- expand the condition for constructing UserDefinedClassVariable
  to include ABCMeta classes via `inspect.isclass` check


torchdynamo/variables/builder.py
```diff
@@ -238,7 +238,7 @@ def _wrap(self, value):
             return SkipFilesVariable(
                 value, guards=make_guards(GuardBuilder.FUNCTION_MATCH)
             )
-        elif istype(value, type):
+        elif inspect.isclass(value):
```
wconstab, line 241: any ideas on how to test the ABCMeta fix?  I wasn't able to think of a simple unit test that would exercise this bug

torchdynamo/convert_frame.py
```diff
@@ -120,7 +120,10 @@ def _convert_frame_assert(frame: types.FrameType, cache_size: int):
         ):
             # not needed, but cleans up torchbench error stats
             return None
-
+        if code.co_name == "__setattr__":
+            # setattr could be tricky to handle generally,
+            # but also not likely useful to compile- break the graph
```
jansel, line 125: This is actually skipping the frame (the graph is already broken)

torchdynamo/variables/builder.py
jansel: Hrm, I think you test should exercise this... does it not?

jansel (changes requested):
Looks like unit tests are failing.  Perhaps we need to go back to `istype(value, (type, ABCMeta)`.

torchdynamo/variables/builder.py
wconstab: i couldn't get the test to fail with `istype(value, type)` despite trying a few different things with super.  I'll have another go at it.

jansel (approved):
LGTM

Fix the lint error before landing.  You can ignore the inductor failure, not related.

jansel:
CI failure is fixed by #229.  Feel free to merge and ignore the error.

PR #228, opened 2022-05-11
Remove reference cycle - with exceptions

anijain2305:
This is a followup of #223 

#223 was cleaning up the `OutputGraph`. However, it was missing the cleanup when an exception was raised. This led to some memory held alive for the following case. Therefore, moving the cleanup to `finalize` of `try ... except` block.

~~~

@torchdynamo.disable
def print_mem(name):


def fn1(x, y, z):
    out = torch.addcdiv(x, y, z)
    print_mem("Inside func")
    return out.sum()


def fn():
    y = torch.randn(1024, 1024, 1024, device="cuda")
    z = torch.randn(1024, 1024, 1024, device="cuda")
    return fn1(x, y, z)


fn()
print_mem("Outside func")
print("\n\n\n")

    fn()
    print_mem("Outside func")
print("\n\n\n")


gc.collect()
torch.cuda.empty_cache()
print_mem("End")
~~~

### DIFF
--- a/torchdynamo/convert_frame.py
+++ b/torchdynamo/convert_frame.py
@@ -149,7 +149,6 @@ def convert_frame_assert(compiler_fn: Callable, one_graph=True):
             )
             tracer.run()
             output = tracer.output
-            output.cleanup()
             assert output.output_instructions
             instructions[:] = output.output_instructions
             code_options.update(output.code_options)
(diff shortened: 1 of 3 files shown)


jansel (approved):

anijain2305:
The first commit did not account for `InliningInstructionTranslator`

PR #229, opened 2022-05-12
Pin inductor CI to specific pytorch version

jansel:
PyTorch master has been pretty flaky lately.  Hopefully this will give better test signal.


Chillee (approved):
Looks reasonable.

Issue #55, opened 2022-03-16
Support writing to closures while inlining


jansel:
TorchDynamo support most cases of closures, however this one is not supported
```
def make_counter():
    x = torch.randn(10)

    def counter():
        nonlocal x
        x = x + 1
        return x

    return counter

@torchdynamo.optimize(torchdynamo.testing.CompileCounter(), nopython=True)
def fn(counter):
    return counter() + counter()


fn(make_counter())
```

The error (when in `nopython=True` mode) is:
```
...
torchdynamo.exc.Unsupported: write to __closure__ while inlining
Processing original code:
  File "/home/jansel/torchdynamo/tests/test_misc.py", line 959, in fn
    return counter() + counter()
  File "/home/jansel/torchdynamo/tests/test_misc.py", line 953, in counter
    x = x + 1
```

This will work if the closure is in the top-level frame.  It will also work if the closure is defined within the captured scope.  But in this case we can't actually emit a `STORE_DEREF` bytecode because "x" is not in our function's freevars.

To support this case we need to rewrite the [STORE_DEREF](https://docs.python.org/3.8/library/dis.html#opcode-STORE_DEREF) to do something like:
```
def fn(counter):
   v0 = counter.__closure__[0].cell_contents
   v1 = v0 + 1
   v2 = v1 + 1
   counter.__closure__[0].cell_contents = v2
   return v1 + v2
```

We already do the first part here:
https://github.com/facebookresearch/torchdynamo/blob/1d1ef71d111d3df894f535ebf8ff6088c0c2e1e1/torchdynamo/variables/functions.py#L142
When we read the value of the closure.

We need to make it writable though.  For that we need to register the cell using [AttributeMutationExisting](https://github.com/facebookresearch/torchdynamo/blob/1d1ef71d111d3df894f535ebf8ff6088c0c2e1e1/torchdynamo/side_effects.py#L154) and then we can use [side_effects.store_cell()](https://github.com/facebookresearch/torchdynamo/blob/1d1ef71d111d3df894f535ebf8ff6088c0c2e1e1/torchdynamo/side_effects.py#L115) on it.

Other types of cells are handled in
https://github.com/facebookresearch/torchdynamo/blob/1d1ef71d111d3df894f535ebf8ff6088c0c2e1e1/torchdynamo/symbolic_convert.py#L1152
and
[...]

PR #233, opened 2022-05-12
Add fix for writing to closures

eellison:
Fix for https://github.com/pytorch/torchdynamo/issues/55

Rewrites 
```
def make_counter():
    x = torch.randn(10)

    def counter():
        nonlocal x
        x = x + 1
        return x

    return counter

@torchdynamo.optimize(torchdynamo.testing.CompileCounter(), nopython=True)
def fn(counter):
    return counter() + counter()


fn(make_counter())
```

As 
```
def fn(counter):
   v0 = counter.__closure__[0].cell_contents
   v1 = v0 + 1
   v2 = v1 + 1
   counter.__closure__[0].cell_contents = v2
   return v1 + v2
```

By rewriting the `STORE_DEREF` to use side-effects tracking

### DIFF
--- a/torchdynamo/variables/functions.py
+++ b/torchdynamo/variables/functions.py
@@ -140,11 +140,18 @@ class UserFunctionVariable(BaseUserFunctionVariable):
                 elif self.source:
                     from .builder import VariableBuilder
 
-                    source = AttrSource(
-                        GetItemSource(AttrSource(self.source, "__closure__"), idx),
-                        "cell_contents",
-                    )
-                    result[name] = VariableBuilder(parent, source)(cell.cell_contents)
+                    side_effects = parent.output.side_effects
+                    if cell in side_effects:
+                        out = side_effects[cell]
+                    else:
+                        closure_cell = GetItemSource(AttrSource(self.source, "__closure__"), idx)
+                        closure_cell_contents = AttrSource(closure_cell, "cell_contents")
+
+                        # cells are written to with "cell_contents", so the source should just be the closure_cell, not its contents
+                        out = side_effects.track_cell_existing(closure_cell, cell)
+                        side_effects.store_cell(out, VariableBuilder(parent, closure_cell_contents)(cell.cell_contents))
+
+                    result[name] = out
                 else:
                     unimplemented("inline with __closure__")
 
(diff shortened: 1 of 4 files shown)


jansel (approved):
`make lint-deps && make format lint` for the lint issues.

torchdynamo/variables/functions.py
```diff
@@ -140,11 +140,29 @@ def bind_args(self, parent, args, kwargs):
...
+
+                        # cells are written to with "cell_contents",
+                        # so the source should just be the closure_cell, not its contents
+                        out = side_effects.track_cell_existing(closure_cell, cell)
+                        side_effects.store_cell(
+                            out,
+                            VariableBuilder(parent, closure_cell_contents)(
+                                cell.cell_contents
+                            ),
+                        )
```
ezyang, line 162: Sorry, I know this is ancient history, but the new code here looks worse than the old code in one sense: we're unconditionally writing to the cell on function creation, triggering side effects, but in fact we may not actually need to write the closure cell if there aren't any variables that will get mutated in the function. I wonder if there's a way to avoid unconditionally store_cell here (context: I'm rewriting cond() operator to be bytecode based and for simplicity I want to assert that there are no side effects, but benign use of closures in our test cases are triggering side effects.)

ezyang: I think what I am going to try is analyzing the bytecode to see if there is a `STORE_DEREF` on the relevant freevar. If there isn't I can assume I don't need to allocate a cell.

StrongerXi: Sorry, back again from [#137456](https://github.com/pytorch/pytorch/issues/137456).

> I wonder if there's a way to avoid unconditionally store_cell here

Is the purpose of this `side_effects.store_cell` call solely to allow us look up its value via `side_effects.load_cell` later, since that's how Dynamo models cell content at the moment?

It does feel a bit strange that we need to generate a write side-effect for an _existing_ cell, I don't see `track_cell_existing` used anywhere else either. Can we avoid this by putting some special treatment to `load_cell` for these "existing cells" from the outside world?

I think that might be less complexity than the restarting mechanism? The latter has generated some subtle bugs in the Dynamo stack, see [#137510](https://github.com/pytorch/pytorch/pull/137510) and the referenced PRs in its commit message.

PR #234, opened 2022-05-12
Delete example value for unused args

anijain2305:
Fixing memory leak for the following test case

~~~


@torchdynamo.disable
def print_mem(name):


def fn1(x, y, z):
    # Unused arg z
    out = torch.addcdiv(x, y, y)
    print_mem("Inside func")
    return out.sum()


def fn():
    y = torch.randn(1024, 1024, 1024, device="cuda")
    z = torch.randn(1024, 1024, 1024, device="cuda")
    return fn1(x, y, z)


fn()
print_mem("Outside func")
print("\n\n\n")

    fn()
    print_mem("Outside func")
print("\n\n\n")


torch.cuda.empty_cache()
gc.collect()

print_mem("End")
~~~

### DIFF
--- a/torchdynamo/output_graph.py
+++ b/torchdynamo/output_graph.py
@@ -355,6 +355,8 @@ class OutputGraph(fx.Tracer):
 
         for node, arg in list(zip(self.graph.nodes, expanded_graphargs)):
             if arg.uses == 0:
+                if "example_value" in node.meta:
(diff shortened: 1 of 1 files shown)


jansel (approved):

anijain2305:
Cross referencing #69 
A large number of memory leaks are resolved now.

PR #231, opened 2022-05-12
Revert "Fix slicing list returning wrong result (#222)"

jansel:
This reverts commit 243222ea119212fa17ca520020bacac9f2f11a21.

@yanboliang I noticed an error in `./torchbench.py` and `git bisect` says it is this commit.

The error is:
```
 ./torchbench.py -k yolov3 
cpu  eval  yolov3                             ERROR FROM offset=262 filename /home/jansel/torchbenchmark/torchbenchmark/models/yolov3/yolo_models.py 290 AttributeError
========== TorchDynamo Stack Trace ==========
Traceback (most recent call last):
  File "/home/jansel/torchdynamo/torchdynamo/convert_frame.py", line 175, in _convert_frame_assert
    code = transform_code_object(frame.f_code, transform)
  File "/home/jansel/torchdynamo/torchdynamo/bytecode_transformation.py", line 338, in transform_code_object
    transformations(instructions, code_options)
  File "/home/jansel/torchdynamo/torchdynamo/convert_frame.py", line 150, in transform
    tracer.run()
  File "/home/jansel/torchdynamo/torchdynamo/symbolic_convert.py", line 300, in run
    and self.step()
  File "/home/jansel/torchdynamo/torchdynamo/symbolic_convert.py", line 278, in step
    getattr(self, inst.opname)(inst)
  File "/home/jansel/torchdynamo/torchdynamo/symbolic_convert.py", line 149, in wrapper
    self.output.compile_subgraph(self)
  File "/home/jansel/torchdynamo/torchdynamo/output_graph.py", line 247, in compile_subgraph
    pass1.foreach(stack_values)
  File "/home/jansel/torchdynamo/torchdynamo/codegen.py", line 142, in foreach
    self(i)
  File "/home/jansel/torchdynamo/torchdynamo/codegen.py", line 89, in __call__
    output.extend(value.source.reconstruct(self))
  File "/home/jansel/torchdynamo/torchdynamo/source.py", line 111, in reconstruct
    return self.base.reconstruct(codegen) + [
AttributeError: 'NoneType' object has no attribute 'reconstruct'
========== Exception (above) while processing ==========
  File "./torchbench.py", line 1217, in <module>
    main()
  File "./torchbench.py", line 1097, in main
    run_one_model(
  File "./torchbench.py", line 1176, in run_one_model
    new_result = model_iter_fn(model, example_inputs)
  File "./torchbench.py", line 577, in forward_pass
    return mod(*inputs)
[...]

PR #235, opened 2022-05-13
Fix list slice & ConstantVariable to TupleVariable conversion missing source info

yanboliang:
This is to resubmit https://github.com/pytorch/torchdynamo/pull/222.
Fix https://github.com/pytorch/torchdynamo/pull/231
* Root cause: ```ConstantVariable.call_method``` wraps value as ```TupleVariable``` and call its ```call_method```, but doesn't propagate source info in this conversion, so the source chain is missed when codegen.
* Why ```torchbench``` doesn't fail before https://github.com/pytorch/torchdynamo/pull/222?
  * Because it fall into another branch [here](https://github.com/pytorch/torchdynamo/blob/main/torchdynamo/codegen.py#L90) previously. Actually I think we can switch the order to first check python constant:
 ```
if value.is_python_constant() and is_safe_constant(
    value.as_python_constant()
):
    output.append(self.create_load_const(value.as_python_constant()))
elif value.source is not None and allow_cache:
    output.extend(value.source.reconstruct(self))
```  
As this can generate more concise code if the stack value is python constant compatible(only LOAD CONST in this case, and doesn't need to trace the source). I'd like to know if this is aligned with the design philosophy and update this PR if it makes sense.

### DIFF
--- a/torchdynamo/variables/lists.py
+++ b/torchdynamo/variables/lists.py
@@ -6,6 +6,7 @@ import torch
 from .. import variables
 from ..bytecode_transformation import create_instruction
 from ..exc import unimplemented
+from ..source import GetItemSource
 from ..utils import namedtuple_fields
 from .base import MutableLocal
 from .base import VariableTracker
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: torchdynamo/variables/lists.py [hunk 1]
```diff
@@ -40,9 +41,11 @@ def as_proxy(self):
     def getitem_const(self, arg: VariableTracker):
         index = arg.as_python_constant()
         if isinstance(index, slice):
-            return self.clone(items=self.items[index], mutable_local=None).add_options(
-                arg, self
-            )
+            return self.clone(
+                items=self.items[index],
+                source=GetItemSource(self.source, index),
```
jansel, line 46: We should only do this if `self.source is not None`

jansel (changes requested):

PR #236, opened 2022-05-13
Introduce Jiterator Compiler

SherlockNoMad:


torchinductor/codegen/jiterator.py
```diff
@@ -0,0 +1,288 @@
...
+        super(JiteratorKernel, self).__init__(args)
+        self.call_ranges = None
+        self.ranges = None
+        self.itervars = None
+        self.reduction_depth = None
+        self.reduction_vars = {}
+
+    def load(self, name: str, index: sympy.Expr):
+        var = self.args.input(name)
+        index = self.rename_indexing(index)
```
jansel, line 131: Here we aren't actually using the indexing formula.  So if this isn't a basic pointwise op things will fail.

As a short term fix you could add an assert that this is basic indexing by calling the indexer on the source node to see what a naive load looks like.

Longer term, I think the right fix is to rewrite the `index` into a call to `torch.as_strided()`

Something similar to:
https://github.com/jansel/torchdynamo/blob/345b44f3b6fee4e4b5e16a4dff1fad8e1c936b2d/torchinductor/codegen/triton.py#L281

First you can use some sympy `subs` calls to separate out the stride in each iteration axis:
```
offset = index.subs({v: 0 for v in iter_vars})
stride0 = index.subs({iter_vars[i]: 0 for i in range(len(iter_vars)) if i != 0}) - offset
stride1 = index.subs({iter_vars[i]: 0 for i in range(len(iter_vars)) if i != 1}) - offset
stride2 = index.subs({iter_vars[i]: 0 for i in range(len(iter_vars)) if i != 2}) - offset
...
```

Next, in the wrapper code (that calls into jiterator) you would do:
```
torch.as_strided(x, iteration_ranges, (stride0, stride1, stride2), offset)
```

This will work for everything except for indirect loads (from aten.index, embeddings, etc), and masked loads (from padding).

Note you will need to include the index in the hash key, since we might read the same buffer with multiple indexing formulas.  In this case we should pass them in as if they are different tensors.

torchinductor/codegen/jiterator.py
```diff
@@ -0,0 +1,288 @@
...
+        args = self.args
+        call_args = []
+        for name in args.input_buffers.keys():
+            call_args.append(f"{name}")
+
+        output_args = []
+        for name in args.output_buffers.keys():
+            output_args.append(f"{name}")
+
+        wrapper.body.writeline(
```
jansel, line 282: we may want to write this to the header section so it gets compiled head of time.

torchinductor/config.py
```diff
@@ -16,6 +16,8 @@
 # enable certain approximation algorithms
 approximations = False
 
+# select compiler backend for cuda
+cuda_backend = 'Jiteraor'
```
jansel, line 20: In the final version let's leave triton as the default

jansel (changes requested):
Perhaps add some unit tests?

torchinductor/decomposition.py
```diff
@@ -236,3 +238,73 @@ def cudnn_batch_norm(
...
+
+
+@register_decomposition(aten.celu.default, decompositions)
+def celu(input: Tensor, alpha: float = 1.0) -> Tensor:
+    inv_alpha = 1.0 / alpha
+    return aten.elu(input, alpha, 1.0, inv_alpha)
+
+
+@register_decomposition(aten.mish.default, decompositions)
+# @pw_cast_for_opmath
```
Chillee, line 255: Can we just load these in from torch? Not a big deal, but since you're commenting this stuff out maybe might as well?

Chillee: oh, nvm, didn't realize those hadn't landed yet.

torchinductor/codegen/jiterator.py
```diff
@@ -0,0 +1,288 @@
...
+from ..virtualized import V
+from ..virtualized import ops
+from .common import BracesBuffer
+from .common import ExprPrinter
+from .common import IndentedBuffer
+from .common import Kernel
+# from .common import KernelArgs
+from .common import OpOverrides
+
+DTYPE_TO_CPP = {
```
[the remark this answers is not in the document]
SherlockNoMad: yep.

torchinductor/codegen/jiterator.py
```diff
@@ -0,0 +1,288 @@
...
+    torch.int32: "int",
+    torch.int16: "short",
+    torch.int8: "signed char",
+    torch.uint8: "unsigned char",
+    torch.bool: "bool",
+}
+INDEX_TYPE = "long"
+
+
+class CppPrinter(ExprPrinter):
```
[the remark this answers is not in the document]
SherlockNoMad: yep.

torchinductor/codegen/jiterator.py
```diff
@@ -0,0 +1,288 @@
...
+
+    def _print_IndexingDiv(self, expr):
+        x, div = expr.args
+        x = self.paren(self.doprint(x))
+        div = self.paren(self.doprint(div))
+        return f"({x} / {div})"
+
+
+cexpr = CppPrinter().doprint
+class JiteratorOverrides(OpOverrides):
```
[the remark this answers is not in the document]
SherlockNoMad: It's a bit tricky here. 
CUDA code uses cmath library, which has a different namespace for math functions. 
For example, "std::min" doesn't work, and I have to codegen "min" or "::min". 

So there is still a lot of shared code, but need subtle adjustment on the namespace prefix.

torchinductor/codegen/jiterator.py
```diff
@@ -0,0 +1,288 @@
...
+                # first any pointwise sharing same loops
+                for node in scheduler.pop_group((group + reduction_group, ())):
+                    node.run(vars, reduction_vars)
+                    node.mark_fusable()
+
+            kernel_group.finalize_kernel(kernel, scheduler)
+
+        kernel_group.codegen_define_and_call(wrapper)
+
+class JiteratorKernelArgs:
```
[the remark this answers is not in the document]
SherlockNoMad: Yes, Made JiteratorKernelArgs subclass KernelArgs.

torchinductor/codegen/jiterator.py
SherlockNoMad: Yes. I moved the _create_jit_fn declaration to the header. 

Under the hook, the compilation actually happens during the first call. Future calls will read the compiled kernel from a Jiterator internal cache.

torchinductor/config.py
SherlockNoMad: Yes, will do.

torchinductor/codegen/jiterator.py
SherlockNoMad: Jiterator handles indexing internally. 

Codegen just need to define the kernel for a single element, e.g 
```
code_string = "template <typename T> T inverted_where(bool cond, T a, T b){ return !cond ? a : b; }"
```

For more examples: see https://github.com/pytorch/pytorch/pull/77121

torchinductor/codegen/jiterator.py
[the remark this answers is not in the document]
jansel: You could override the stuff you need to change.

torchinductor/codegen/jiterator.py
jansel: I am aware of how Jiterator handles indexing.  `as_strided()` is the way to map from TorchInductor indexing to Jiterator indexing.

For an example try compiling:
```
x[::2] + x[1::2]
```

With your current code it will generate the incorrect result.

If you trace through the codegen you will see:
```
tmp0 = load("x", 2*sympy.Symbol("i0")) 
tmp1 = load("x", 2*sympy.Symbol("i0") + 1) 
tmp2 = ops.add(tmp0, tmp1)
```

Note a "normal" index to "x" would just be `sympy.Symbol("i0")` (without the `2*` or `+1`).

Since the current version just ignores the `index=...`, Jiterator won't do the correct thing.  You need to do the `as_strided()` thing to convey the correct indexing formula to Jiterator.

torchinductor/codegen/jiterator.py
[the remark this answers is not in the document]
SherlockNoMad: yes, this is what I am doing...

jansel:
Closing due to no activity in last month, feel free to reopen.

PR #237, opened 2022-05-13
[inductor] early memory reuse and new operators

jansel:
There is still a bug in the inplacing logic I need to track down.


torchinductor/lowering.py
```diff
@@ -22,6 +22,37 @@
 lowerings = {}
 aten = torch.ops.aten
 
+# TODO(jansel): figure out a way generate this automatically
+#               these come as dtype args to kernels
+# based on https://github.com/pytorch/pytorch/blob/9e3eb329df8f701/c10/core/ScalarType.h#L28
+DTYPE_ID_LOOKUP = {
```
Chillee, line 28: lol, cc: @ezyang

ezyang: This is fixed on master pytorch

bertmaher (approved):
lgtm

torchinductor/lowering.py
jansel: @ezyang I definitely still need this on the latest pytorch master.  Example of an op getting `dtype=6` is `aten.new_ones`, but there are many others.

ezyang: oh I bet this is because optionals are not being handled correctly

PR #238, opened 2022-05-13
Add type checking in Constant match - Fix Pytorch tests 4/n

anijain2305:
A guard like `y == 9.0` where y is a `torch.Tensor` results in boolean Tensor, and fails guard execution. This PR adds a type check in `CONSTANT_MATCH` to prevent such checking in the first place.


anijain2305:
Fixes few segfaults seen in #69

jansel (approved):

PR #239, opened 2022-05-13
Make eval_frame thread safe

voznesenskym:
This is a draft diff that should make eval_frame thread safe. Currently, the eval_frame is a global object, and different threads may step on each other setting a different frame.

This changes the behavior to instead always have a "shim" eval_frame which then routes to the correct behavior by looking at the thread-local associated object. This is thread safe because now the callback object is always thread safe, and we only use it to drive logic at frame eval time, as opposed to at callback registration time. This means that we do not change the global frame on every thread, but rather track instances of threads currently in `optimize` to determine the right time to install or remove our shim.

Currently, the logic for None/False/Callback is kept, but the False case could be easily collapsed behind the shim in a subsequent diff.


voznesenskym:
Just a draft - the "test.py" is an example and will not land.

torchdynamo/_eval_frame.c
```diff
@@ -310,6 +313,26 @@ inline static PyObject *eval_custom_code(PyThreadState *tstate,
   return result;
 }
 
+static PyObject *_custom_eval_frame_shim(PyThreadState *tstate, PyFrameObject *frame,
+                                    int throw_flag) {
+// Shims logic into one of three states. Can probably be refactored into a single func, later:
+  //  - None: disables TorchDynamo
+  //  - False: run-only mode (reuse existing compiles)
+  //  - Python callable(): enables TorchDynamo
+    if (eval_frame_is_paused_for_guard() == true) {
```
jansel, line 322: `== true` is a bit redundant

torchdynamo/_eval_frame.c
```diff
@@ -394,48 +417,22 @@ static PyObject *set_eval_frame(PyObject *new_callback, PyThreadState *tstate) {
...
-    enable_eval_frame(tstate);
-#ifdef TORCHDYNAMO_MULTI_THREAD
-  } // callback is private, so always clear it
-#endif
+  
+  // Always enable eval frame shim.
+  enable_eval_frame_shim(tstate);
 
   Py_INCREF(new_callback);
   Py_DECREF(eval_frame_callback_get());
```
jansel, line 430: Perhaps reuse pointer from above (TLS is expensive)

torchdynamo/_eval_frame.c
```diff
@@ -77,45 +73,49 @@ inline static void eval_frame_callback_set(PyObject *obj) {
   PyThread_tss_set(&eval_frame_callback_key, obj);
 }
 
-#else
-
-static PyObject *eval_frame_callback_ = NULL;
+static Py_tss_t eval_frame_pause_for_guard_key = Py_tss_NEEDS_INIT;
```
jansel, line 76: Do we actually need a new TLS variable for this?  Could we just set the callback to None (and back again) instead?

jansel (commented):
Overall seems like the right track to me.  Thanks!

I'd like to see if we could simplify the code a bit (mainly around merging the run-only logic with the normal flow). 

We should also be able to fully restore original python behavior when we exit.

torchdynamo/_eval_frame.c
voznesenskym: Yea, that works!

voznesenskym:
> Overall seems like the right track to me. Thanks!

Thank you :)

> I'd like to see if we could simplify the code a bit (mainly around merging the run-only logic with the normal flow).

Agreed, but I may save that for a subsequent PR, if you don't mind, if only cause this one already does a bunch. I can follow fast on this one with that change.

> We should also be able to fully restore original python behavior when we exit.

Great point, I implemented the counter mechanism as you suggested, and it seems to work correctly. Let me know if that's what you had in mind.

torchdynamo/_eval_frame.c
```diff
@@ -390,52 +388,55 @@ static PyObject *_custom_eval_frame_run_only(PyThreadState *tstate,
   }
 }
 
+static int active_dynamo_threads = 0;
+
+static pthread_mutex_t dynamo_frame_swap_mutex = PTHREAD_MUTEX_INITIALIZER;
```
jansel, line 393: Will the python GIL make it so we don't need this?  (At least for currently released versions of Python.)

torchdynamo/_eval_frame.c
```diff
@@ -132,29 +108,29 @@ inline static PyObject *eval_frame_default(PyThreadState *tstate,
...
-inline static void disable_eval_frame(PyThreadState *tstate) {
-#if PY_VERSION_HEX >= 0x03090000
-  _PyInterpreterState_SetEvalFrameFunc(tstate->interp,
-                                       &_PyEval_EvalFrameDefault);
+  if (_PyInterpreterState_GetEvalFrameFunc(tstate->interp) != &custom_eval_frame_shim) {
+    _PyInterpreterState_SetEvalFrameFunc(tstate->interp, &custom_eval_frame_shim);
+  }
 #else
-  tstate->interp->eval_frame = &_PyEval_EvalFrameDefault;
+  if (tstate->interp->eval_frame != &custom_eval_frame_shim) {
```
jansel, line 117: Is this check just an optimization?  Curious about the motivation.

torchdynamo/_eval_frame.c
```diff
@@ -310,6 +286,23 @@ inline static PyObject *eval_custom_code(PyThreadState *tstate,
   return result;
 }
 
+static PyObject *_custom_eval_frame_shim(PyThreadState *tstate, PyFrameObject *frame,
+                                    int throw_flag) {
+// Shims logic into one of three states. Can probably be refactored into a single func, later:
+  //  - None: disables TorchDynamo
+  //  - False: run-only mode (reuse existing compiles)
+  //  - Python callable(): enables TorchDynamo
+    PyObject *callback = eval_frame_callback_get();
```
jansel, line 295: Now that we have this shim, perhaps we should change the signature of _custom_eval_frame to accept `callback` as an arg, this would let us avoid a thread-local-storage lookup.

jansel (changes requested):

torchdynamo/_eval_frame.c
voznesenskym: Yes, just an optimization. We can always drop it if you think it doesn't add value.

torchdynamo/_eval_frame.c
voznesenskym: Sure, let me try that.

voznesenskym: Actually, I think I don't understand, but I probably am missing knowledge about a critical mechanism.

This is called from the eval_frame, so this shim gets set on via `tstate->interp->eval_frame = &custom_eval_frame_shim;`. How can we pass in a callback to this as an arg?

torchdynamo/_eval_frame.c
```diff
@@ -310,6 +286,23 @@ inline static PyObject *eval_custom_code(PyThreadState *tstate,
...
+    
+    if (callback == Py_None) {
+      return eval_frame_default(tstate, frame, throw_flag);
+    } else if (callback == Py_False) {
+      return _custom_eval_frame_run_only(tstate, frame, throw_flag);
+    } else {
+      return _custom_eval_frame(tstate, frame, throw_flag);
```
jansel, line 302: You would add the arg here.  (Which lets you skip a TLS call inside the function.)

torchdynamo/_eval_frame.c
jansel: See https://github.com/pytorch/torchdynamo/pull/239/files#r873054114

torchdynamo/_eval_frame.c
voznesenskym: Oh, I see, yes, I misunderstood you, good catch, we can skip the `eval_frame_callback_get()` downstream of the shim.

torchdynamo/_eval_frame.c
```diff
@@ -2,6 +2,7 @@
 #include <Python.h>
 #include <frameobject.h>
 #include <pystate.h>
+#include <pthread.h>
```
jansel, line 5: Is this header still needed

torchdynamo/_eval_frame.c
```diff
@@ -390,52 +386,49 @@ static PyObject *_custom_eval_frame_run_only(PyThreadState *tstate,
...
-    disable_eval_frame(tstate);
-  } else if (new_callback == Py_False) {
-    enable_run_only_eval_frame(tstate);
+    decrement_working_threads(tstate);
   } else {
-    enable_eval_frame(tstate);
-#ifdef TORCHDYNAMO_MULTI_THREAD
-  } // callback is private, so always clear it
-#endif
+    increment_working_threads(tstate);
```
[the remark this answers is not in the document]
voznesenskym: I'm no expert in cpython reference counting, but I think this is correct. We increment the new_callback reference since we take it and store it, and we lower the old one. The shims don't need ref counting, and nothing else really changed in terms of crossing the c/py boundary. LMK if I did something clowny and overlooked it though!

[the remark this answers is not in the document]
jansel: Thanks yeah.  I didn't see any issue, but it is subtle enough I want to double check.

jansel (approved):

PR #240, opened 2022-05-14
enumerate supports start argument

yanboliang:
As title, to support ```enumerate ([1, 2, 3, 4], 2)```

### DIFF
--- a/tests/test_subgraphs.py
+++ b/tests/test_subgraphs.py
@@ -504,3 +504,13 @@ class SubGraphTests(torchdynamo.testing.TestCase):
         with torchdynamo.optimize(cnt):
             self.assertEqual(fn(v1, it1).tolist(), (v1 + 1 + 2 + 3).tolist())
         self.assertEqual(list(it1), [4, 5, 6, 7, 8, 9])
+
+    def test_enumerate_not_break_graph(self):
+        def fn(a, b):
+            for i, x in enumerate(a.shape):
+                b = b + x
+            for i, x in enumerate(b.shape, 0):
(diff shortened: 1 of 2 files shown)


torchdynamo/variables/builtin.py
```diff
@@ -294,15 +294,20 @@ def call_zip(self, tx, *args):
...
 
-    def call_enumerate(self, tx, arg):
-        options = VariableTracker.propagate(self, arg)
-        if arg.has_unpack_var_sequence(tx):
+    def call_enumerate(self, tx, *args):
+        options = VariableTracker.propagate(self, args)
+        if len(args) == 1:
+            start = 0
+        else:
+            assert isinstance(args[1], variables.ConstantVariable)
```
jansel, line 302: lets also assert `len(args)==2`

tests/test_subgraphs.py
```diff
@@ -504,3 +504,13 @@ def fn(x, it):
         with torchdynamo.optimize(cnt):
             self.assertEqual(fn(v1, it1).tolist(), (v1 + 1 + 2 + 3).tolist())
         self.assertEqual(list(it1), [4, 5, 6, 7, 8, 9])
+
+    def test_enumerate_not_break_graph(self):
+        def fn(a, b):
+            for i, x in enumerate(a.shape):
+                b = b + x
+            for i, x in enumerate(b.shape, 0):
```
jansel, line 512: we should have a test with a start not equal to 0

jansel (changes requested):

PR #243, opened 2022-05-16
[inductor] Multi-devices, primtorch decomps, and many new ops

jansel:


bertmaher:
Sorry I'm just now getting to this review -- something that would help me do these faster is if you could split up the diffs into logical components; it's going to take me a while to go through an 800 line change, and I have to mentally figure out which pieces of it are related to which logical change.  Don't worry about breaking up this PR, just, for future reference ;-)

bertmaher (approved):

torchinductor/decomposition.py
```diff
@@ -4,17 +4,27 @@
...
-# decompositions = get_decompositions([...])
-
-# note AOT Autograd decomps are included by default in torchdynamo
-decompositions = {}
+decompositions = get_decompositions(
+    [
+        aten.l1_loss,
+        aten.mse_loss,
+        aten.stack,
+        aten.native_layer_norm,
```
Chillee, line 18: nice :)

torchinductor/ir.py
```diff
@@ -1237,15 +1240,44 @@ def __init__(
...
+
+            def __repr__(self):
+                return self.ref
+
+        tensor_args = [Shim(x.codegen_reference()) for x in self.inputs]
+        constant_args = [Shim(repr(x)) for x in self.constant_args]
+        return list(map(repr, self.unflatten_args(tensor_args, constant_args)))
 
     @classmethod
     def create(cls, kernel, *args):
```
Chillee, line 1258: What is this logic for handling? Is this assuming that we might take in FX graphs where the inputs is not a flat list of tensors?

torchinductor/lowering.py
```diff
@@ -460,19 +472,32 @@ def _embedding_bag(
...
-                aten._cudnn_rnn,
-                *args
-            ),
+            ir.FallbackKernel.create(aten._cudnn_rnn, *args),
         )
     )
-"""
+
+
+@register_lowering(aten.grid_sampler_2d, type_promote=False)
```
Chillee, line 485: The decompositions in core are aiming to accurately capture type promotion logic (although certainly possible that they're not completely correct in some cases).

torchinductor/lowering.py
```diff
@@ -716,22 +789,66 @@ def fn(idx):
...
 
+    return Pointwise.create(
+        device=x.get_device(),
+        dtype=x.get_dtype(),
+        inner_fn=fn,
+        ranges=output_size,
+    )
+
+
+@register_lowering(aten.index_select, type_promote=False)
```
Chillee, line 837: Would a decomposition of `index_select` into gather or the like be preferred?

torchinductor/lowering.py
```diff
@@ -1076,6 +1312,29 @@ def fn(idx):
         # else: torch.exp(torch.log(a) * b)
 
 
+def mutate_to(changed, val):
```
Chillee, line 1315: Is TorchInductor aiming to support in-place ops in full generality?

Chillee (commented):
Don't have enough context for a full review, but read over the code, understood some more components, and had some questions about the lowerings.

torchinductor/lowering.py
jansel: Does this already exist in core?  I didn't see it when I wrote this (last week).

Happy to switch when it is available.

torchinductor/ir.py
jansel: _cudnn_rnn takes a list of tensors as args.  This is just to support that.

torchinductor/lowering.py
Chillee: ah, I was specifically asking about the `type_promote` flag. `grid_sampler_2d` is not in core.

torchinductor/lowering.py
Chillee: So if we functionalized things you just wouldn't have these lowerings/

jansel: There are some things like input mutation which I don't think functionalization will handle.  Some decomposition are introducing mutation also.  So even if you functionalize before AOT autograd, you get graphs with mutation in them.

Maybe we can delete these at some point.  Though I think they will be needed for a while.

Chillee: > Some decomposition are introducing mutation also. So even if you functionalize before AOT autograd, you get graphs with mutation in them.

Well, we'd presumably functionalize during AOTAutograd tracing, right? So it'd go through decompositions.

I do agree with input mutation... not totally sure how to handle that.

PR #244, opened 2022-05-16
cast model in no-isolate mode

zzpmiracle:
I think we need cast model when assign model precision in no-isolate mode.

### DIFF
--- a/torchbench.py
+++ b/torchbench.py
@@ -1094,6 +1094,12 @@ def main():
         for device, name, model, example_inputs in iter_models(args):
             torchdynamo.reset()
             gc.collect()
+            
+            if args.float32:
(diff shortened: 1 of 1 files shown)


jansel:
Thanks for the PR @zzpmiracle!  

LGTM, though could you fix the minor lint: https://github.com/pytorch/torchdynamo/runs/6455971109?check_suite_focus=true#step:5:18

zzpmiracle:
> Thanks for the PR @zzpmiracle!
> 
> LGTM, though could you fix the minor lint: https://github.com/pytorch/torchdynamo/runs/6455971109?check_suite_focus=true#step:5:18

ok, fixed.

PR #248, opened 2022-05-16
Skip non tensor frame

anijain2305:
Reimplementation of https://github.com/pytorch/torchdynamo/pull/171

More details at #69

### DIFF
--- a/torchdynamo/convert_frame.py
+++ b/torchdynamo/convert_frame.py
@@ -117,14 +117,36 @@ def wrap_convert_context(fn):
 
 
 def has_tensor_in_frame(frame):
-    # Check if the passed arguments are of type Tensor
-    for value in frame.f_locals.values():
-        if isinstance(value, torch.Tensor):
-            return True
+    """Check if the frame has torch.* related bits"""
 
     # Check if there is global import of torch.*
-    for value in frame.f_globals.values():
-        if is_allowed(value):
+    for co_name in frame.f_code.co_names:
+        if co_name in frame.f_globals:
+            if is_allowed(frame.f_globals[co_name]):
+                return True
+
+    def has_tensor(obj):
+        """Recursively check if the obj has a tensor"""
+        if isinstance(obj, (torch.Tensor, torch.nn.Module)):
+            return True
+        elif isinstance(obj, (list, tuple)):
+            return any([has_tensor(v) for v in obj])
+        elif isinstance(obj, dict):
+            return any([has_tensor(v) for v in obj.values()])
+        elif isinstance(obj, (str, int, float, type(None), bool)):
+            return False
+        elif hasattr(obj, "__dict__"):
+            return any([has_tensor(v) for v in obj.__dict__.values()])
+        else:
+            if config.debug:
+                print(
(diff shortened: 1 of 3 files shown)


torchdynamo/convert_frame.py
```diff
@@ -101,6 +103,27 @@ def _fn(*args, **kwargs):
...
 
 
+def has_tensor_in_frame(frame):
+    # Check if the passed arguments are of type Tensor
+    for value in frame.f_locals.values():
+        if isinstance(value, torch.Tensor):
+            return True
+
+    # Check if there is global import of torch.*
+    for value in frame.f_globals.values():
```
jansel, line 112: It might be faster to check only `for name in frame.f_code.co_names`.  This will give only the globals used by the function.

torchdynamo/convert_frame.py
```diff
@@ -101,6 +103,27 @@ def _fn(*args, **kwargs):
     return _fn
 
 
+def has_tensor_in_frame(frame):
+    # Check if the passed arguments are of type Tensor
+    for value in frame.f_locals.values():
+        if isinstance(value, torch.Tensor):
```
jansel, line 108: We should also check for torch.nn.Module. 

We should also check for `List[Tensor]`, `Dict[str, Tensor]`, etc.  Perhaps have small helper that walks data structures.

jansel (changes requested):

anijain2305:
@jansel Ready for another round of review. 

Also added the functionality where if somebody decorates the function with `torchdynamo.optimize`, we do not skip the frame.

torchdynamo/convert_frame.py
```diff
@@ -101,6 +103,52 @@ def _fn(*args, **kwargs):
...
+    if frame.f_code in always_optimize_code_objects:
+        return True
+
+    # Check if there is global import of torch.*
+    for co_name in frame.f_code.co_names:
+        if co_name in frame.f_globals:
+            if is_allowed(frame.f_globals[co_name]):
+                return True
+
+    def has_tensor(obj):
```
jansel, line 118: I think this will infinite loop if given:
```
x = []
x.append(x)
```

Memoize and keep track of all `id(obj)` seen to prevent that

torchdynamo/convert_frame.py
```diff
@@ -101,6 +103,52 @@ def _fn(*args, **kwargs):
...
+        """Recursively check if the obj has a tensor"""
+        if isinstance(obj, (torch.Tensor, torch.nn.Module)):
+            return True
+        elif isinstance(obj, (list, tuple)):
```
jansel, line 122: I think it might be possible to call into user code here.
```
class Foo(list):
   def __iter__(self):
       raise Exception()
```

We could use `utils.istype` for a more strict type check.  We could call `list.values(obj)` directly.  

Similar issues below.  We should be careful about user defined objects.

anijain2305:
@jansel I put stricter type checks. I also looked at the torchbench stats, and they look good overall.

Thanks for going through the multiple iterations here with me. This PR is ready for another round of review.

jansel (approved):

PR #251, opened 2022-05-16
enable tracing through frozenset contains of PyTorch ops

vkuzo:
Summary:

Enables tracing through this syntax:

```
funcs = frozenset([torch.add])

def fn(x, func):
    if func in funcs:
        x = torch.add(x, 1.0)
    x = torch.mul(x, 1.0)
    return x
```

This is useful for DBR quantization.

Test plan:

```
pytest -vsk test_frozenset_torch_func_contains
```

### DIFF
--- a/torchdynamo/variables/builder.py
+++ b/torchdynamo/variables/builder.py
@@ -202,6 +202,15 @@ class VariableBuilder:
                 value=value,
                 guards=make_guards(GuardBuilder.CONSTANT_MATCH),
             )
+        elif isinstance(value, frozenset) and (
+            all(is_allowed(x) or ConstantVariable.is_literal(x) for x in value)
+        ):
+            # For frozenset, we can guard by object ID instead of value
+            # equality, this allows us to handle non-literal values
+            return ConstantVariable(
+                value=value,
+                guards=make_guards(GuardBuilder.ID_MATCH),
+            )
         elif isinstance(value, enum.Enum):
             return EnumVariable(
                 value=value,
--- a/tests/test_misc.py
+++ b/tests/test_misc.py
@@ -1339,3 +1339,23 @@ class MiscTests(torchdynamo.testing.TestCase):
 
         self.assertEqual(res1, 3)
         self.assertEqual(res2, 1)
+
+    def test_frozenset_torch_func_contains(self):
+        funcs = frozenset([torch.add])
+
+        def fn(x, func):
+            if func in funcs:
+                x = torch.add(x, 1.0)
+            x = torch.mul(x, 1.0)
+            return x
+
+        x = torch.randn(1)
+        cnts = torchdynamo.testing.CompileCounter()
+        with torchdynamo.optimize(cnts, nopython=True):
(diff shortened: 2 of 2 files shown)


torchdynamo/variables/user_defined.py
```diff
@@ -190,6 +191,17 @@ def call_method(
...
+            if (
+                name == "__contains__"
+                and isinstance(self.value, frozenset)
+                and len(args) > 0
+                and isinstance(args[0], TorchVariable)
+            ):
+                # This handles contains frozenset contains checks for PyTorch
+                # ops.
+                # TODO(future PR): more general support for frozenset
+                return ConstantVariable(args[0].value in self.value, **options)
```
vkuzo, line 203: is it fine to not guard since frozenset is immutable?

jansel: For UserDefindObjectVariable we only guard on TYPE_MATCH:
https://github.com/pytorch/torchdynamo/blob/f21fd574dd7b07e3959103106c88bdfe27ca9e8c/torchdynamo/variables/builder.py#L295

So if you have:
```
funcs1 = frozenset([torch.add])
funcs2 = frozenset([torch.sub])
```

Both of those will pass the guards for one another, because they have the same type.  

I think the best thing to do here would be to not map set/frozenset to UserDefindObjectVariable.  For frozenset, we could just map it to a ConstantVariable -- but we will need to check all the contents of the set are also constants.

A more general solution would be to map them to new type SetVariable, which would be similar to ConstDictVariable.

jansel (changes requested):

torchdynamo/variables/builder.py
```diff
@@ -202,6 +202,13 @@ def _wrap(self, value):
                 value=value,
                 guards=make_guards(GuardBuilder.CONSTANT_MATCH),
             )
+        elif isinstance(value, frozenset):
```
jansel, line 205: I think we also need something like:
```
and all(is_allowed(x) or ConstantVariable.is_literal(x) for x in value)
```

You could have something like:
`frozenset([tensor, dict])`

Then someone could mutate the contents of the frozenset (without actually mutating the containing set).

jansel (changes requested):

torchdynamo/variables/builder.py
vkuzo: sure. Do we need any extra guards, or is this additional check enough?

I'm asking because before currently guarding on something like a `torch.add` is not doable with the current code, so if this needs additional guards maybe we would need to extend the dictionary keys guards to support guarding by ID, currently they only support literals and guard by equality checks.

vkuzo: I'm guessing if ` all(is_allowed(x) or ConstantVariable.is_literal(x) for x in value)` no guards are needed other than ID of the set, since we're enforcing the set members to be immutable, assuming the is_allowed check only passes for immutable things?

jansel (approved):

Issue #179, opened 2022-04-26
Add a backend verification mode


jansel:
We should have a mode where we perform more thorough testing of backend correctness on a graph-by-graph level as a debugging tool.

Some [specific backends include correctness checking](https://github.com/jansel/torchdynamo/blob/8db8efc7681e4ae5851c3a298250993660d16e55/torchdynamo/optimizations/inference.py#L156), though currently it is up to the backends to check this themselves. 

I'd suggest adding a wrapper class that takes a backend and returns a new backend that checks correctness and raises exceptions on correctness failures.

This should be a mode that users can enable to help debug the issue where they get the wrong answer.

anijain2305:
This one is already completed by @pyjhzwh

PR #252, opened 2022-05-16
Implement verify_correctness #179

pyjhzwh:
#179 
Summary
Add a WrapperBackend class. It takes a backend and returns a new backend to check the correctness by example_inputs and raises warning on correctness failure and return the original graph. The correctness verification option is turned on by setting config.verify_correctness as True.

Test plan: ` pytest tests/tests_verify_correctness `
  * For backends that are supported in torchdynamo (like nnc, fixed_strategy), it should passed the correctness verification; 
  * test_incorrect_verify_true() checks if we intentionally build a bad backend that change the graph that are not functionally equal to the original graph, when  config.verify_correctness== True, WrapperBackend will log a warning and return original gm instead. We would expect the correct outputs. 
  * In test_incorrect_verify_false(), the config.verify_correctness == False, the outputs are not the same as original graph's outputs since the graph is not corrected.


torchdynamo/convert_frame.py
```diff
@@ -61,16 +61,26 @@ def fx_forward_from_src_skip_result(*args, **kwargs):
...
     """Expand backend strings to functions"""
-    if compiler_fn == "inductor":
+    if isinstance(compiler_fn, WrapperBackend):
+        # currently, compiler_fn's backend is not callable
+        # if it use default BACKENDS
+        # set compiler_fn.backend as the callable function
+        wrapper_compiler_fn = compiler_fn    
+    else:
+        # convert to WrapperBackend class
+        wrapper_compiler_fn = WrapperBackend(compiler_fn)
```
jansel, line 71: I'd suggest only adding `WrapperBackend` in the case of `verify_correctness=True`, and leave the `verify_correctness=False` behavior the same.  That might make things a bit simpler here.  You can add the wrapper after we decode string backends.

torchdynamo/eval_frame.py
```diff
@@ -25,6 +33,12 @@ def nothing():
     pass
 
 
+def same(left, right):
+    return len(left) == len(right) and all(
```
jansel, line 37: We should be able to merge the `len(left) == len(right)` into the other `same()` function.

jansel (changes requested):

torchdynamo/eval_frame.py
pyjhzwh: I copy the same() function from testing.py. I could not directly import it because of cyclic import issue. Is it ok to duplicate the same() function in eval_frame.py or should I move the testing.same() to another file?

pyjhzwh: Move same() from testing.py to utils.py

torchdynamo/convert_frame.py
```diff
@@ -61,16 +62,32 @@ def fx_forward_from_src_skip_result(*args, **kwargs):
...
+            from .optimizations import BACKENDS
+
+            wrapper_compiler_fn.backend = wrap_compiler_fn(BACKENDS[compiler_fn])
+
+        return wrapper_compiler_fn
 
-        return wrap_compiler_fn(BACKENDS[compiler_fn])
     else:
-        return compiler_fn
+        if compiler_fn == "inductor":
```
jansel, line 81: Let's find a way not to duplicate this code here and above.  Perhaps a helper function?

pyjhzwh: Use a _wrap_compiler_fn() as a helper function to make the code cleaner

jansel (approved):

PR #254, opened 2022-05-17
Add support for STORE_GLOBAL

desertfire:
* __->__ #254

Summary:
1. Create a symbolic_global table to store a global variable name to an
unique object mapping, and the unique object is further used as a key to
index into the store_attr_mutations table in SideEffects.
2. The actual STORE_GLOGAL action is buffered by SideEffects and later
LOAD_GLOBAL just reads from SideEffects when appropriate. STORE_GLOBAL
is eventually applied after the generated graph.

### DIFF
--- a/tests/test_global.py
+++ b/tests/test_global.py
@@ -1,0 +1,253 @@
+#!/usr/bin/env pytest
+import collections
+import copy
+import dataclasses
+import dis
+import functools
+import math
+import sys
+import typing
+import unittest
+
+import numpy as np
+import torch
+
+import torchdynamo.testing
+from torchdynamo import bytecode_transformation
+from torchdynamo.testing import CompileCounter
+from torchdynamo.testing import requires_static_shapes
+from torchdynamo.testing import same
+from torchdynamo.testing import unsupported
+
+from .test_global_declaration import g_tensor_export
+
+torchdynamo.config.debug = True
+
+
+class Pair(object):
+    def __init__(self, x, y):
+        self.x = x
+        self.y = y
+
+
+def Foo():
+    return Pair(1, 1)
+
+
+g_counter = 1
+g_list = [0, 1, 2]
+g_dict = {"a": 0, "b": 1}
+g_object = Foo()
+g_tensor = torch.zeros(10)
+
+
+_name: int = 0
+
+
+def fresh_name() -> str:
+    """create a new unique name for a variable: v0, v1, v2"""
+    global _name
+    r = f"v{_name}"
+    _name += 1
+    return r
+
+
+def reset_name():
+    global _name
+    _name = 0
+
+
+class TestGlobals(torchdynamo.testing.TestCase):
+    def test_store_global_1(self):
+        def fn(x):
+            global g_counter
+            val = x + g_counter
+            g_counter += 1
+            return val
+
+        x = torch.randn(10)
+        cnts = torchdynamo.testing.CompileCounter()
+        with torchdynamo.optimize(cnts):
+            res1 = fn(x)
+        res2 = fn(x)
+        self.assertTrue(same(res2 - res1, torch.ones(10)))
+
+    def test_store_global_2(self):
+        def fn(x):
+            global g_counter
+            val = x + g_counter
+            g_counter += 1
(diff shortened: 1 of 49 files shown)


tests/test_misc.py
```diff
@@ -794,6 +796,27 @@ def fn(x):
...
 
+    def test_store_global_1(self):
+        class MyObj:
+            def __init__(self, a):
+                self.a = a
+
+        def fn(obj):
+            global _gtensor
+            val = obj.a + _gtensor
+            _gtensor += torch.ones(10)
```
jansel, line 807: I'd suggest a few more test cases:
1) writing to a new global: `global _new_global; _new_global = obj.a+1;`
2) writing to a global defined in a different file while inlining a function (this will be harder because we cant use `STORE_GLOBAL` in output code.)

torchdynamo/side_effects.py
```diff
@@ -300,7 +310,12 @@ def codegen(self, cg: PyCodegen):
                     cg.tx.output.update_co_names(name)
                     cg(value)
                     cg(var.mutable_local.source)
-                    suffixes.append([create_instruction("STORE_ATTR", name)])
+                    op = (
+                        "STORE_GLOBAL"
```
jansel, line 314: Doesn't `STORE_GLOBAL` pop 1 from the stack (while STORE_ATTR pops 2).  Are we leaking a value onto the stack here?

torchdynamo/side_effects.py
```diff
@@ -300,7 +310,12 @@ def codegen(self, cg: PyCodegen):
...
+                        if isinstance(var.mutable_local.source, GlobalSource)
```
jansel, line 315: Will this work with (this code should work fine in the main branch)
```
some_global = Foo()

@torchdynamo.optimize(..)
def fn(x: Tensor):
  some_global.y = x + 1
```

I think in that case `var.mutable_local.source` will be `GlobalSource("some_global")`, so we will write to `global y` by mistake.

Perhaps rather than representing globals as
```
load_attr(gvar, gvar.source.name())
```
we should represent them as:
```
load_attr(global_scope_singleton, gvar.source.name())
```

Where global_scope_singleton is some sentinel object.  When the global is from some other module (because of inline), we can just map it to an attribute read directly.

jansel (changes requested):

jansel:
> I am still puzzled by the fact that there is a `20 LOAD_GLOBAL` right before `22 STORE_GLOBAL` in the modified bytecode. I think the to-be-stored value should come from `18 BINARY_SUBSCR`, but to my surprise the test case passes.

I think it is passing the test here because `+=` mutates the tensor orignal tensor.

If you change to `_gtensor = _gtensor + 1` it should fail.

tests/test_global.py
```diff
@@ -0,0 +1,225 @@
+#!/usr/bin/env pytest
+import torch
+
+import torchdynamo.testing
+from torchdynamo.testing import same
+
+from . import test_global_declaration
+
+torchdynamo.config.debug = True
```
jansel, line 9: If you subclass `torchdynamo.testing.TestCase` this should get turned on automatically.

torchdynamo/symbolic_convert.py
```diff
@@ -1279,16 +1305,18 @@ def inline_call_(parent, func, args, kwargs):
...
             tracer = InliningInstructionTranslator(
-                parent, code, sub_locals, closure_cells, func
+                parent, code, sub_locals, parent.symbolic_globals, closure_cells, func
             )
 
         tracer.run()
         assert tracer.symbolic_result is not None
         func.export_freevars(parent, tracer)
+        # Merge symbolic_globals to the parent
+        parent.symbolic_globals.update(tracer.symbolic_globals)
```
jansel, line 1319: We should only do this is we are in the same namespace as the parent.  `f_globals is parent.f_global`

If these come from different files they will have different globals.

jansel (changes requested):
Getting close! A couple minor items.

jansel:
I don't think @pytorchbot knows how for this repo, but maybe I will be surprised :)

PR #256, opened 2022-05-17
[not4land] Test fx graph mode quant + torchdynamo

jerryzh168:
In the compiler function we return a object that supports prepare and convert the model based on `prepare_mode` flag
we can first set the prepare mode to True so that it runs prepare, and then set it to False so that we will run the converted/quantized model

qconfig is hardcoded right now to default_qconfig

Test:
python torchbench.py --quant -x pytorch_CycleGAN_and_pix2pix -x pytorch_stargan -x pytorch_unet -x BERT_pytorch -x Super_SloMo -x attention_is_all_you_need_pytorch -x demucs 1>log 2>log

### DIFF
--- a/torchbench.py
+++ b/torchbench.py
@@ -43,11 +43,16 @@ from torchdynamo.testing import format_speedup
 from torchdynamo.testing import reduce_to_scalar_loss
 from torchdynamo.testing import same
 from torchdynamo.utils import clone_inputs
+from torchdynamo.test import QuantizationCompiler
+from torch.ao.quantization import default_qconfig
+
+quant_compiler = QuantizationCompiler({"": default_qconfig})
 
 os.environ["KALDI_ROOT"] = "/tmp"  # avoids some spam
 torchbench_dir = abspath(
     "../torchbench" if exists("../torchbench") else "../torchbenchmark"
 )
+print(torchbench_dir)
 assert os.path.exists(torchbench_dir)
 os.chdir(torchbench_dir)
 sys.path.append(torchbench_dir)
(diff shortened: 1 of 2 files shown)


torchdynamo/test.py
```diff
@@ -0,0 +1,66 @@
...
+                print("quantized model:", self.quantized_model)
+                self.quantized_model = convert_fx(self.prepared_model)
+            return self.quantized_model(*args, **kwargs)                
+            
+def quant_compiler(model, example):    
+    return QuantizationWrapper(model, example)
+    
+def quantize(m, example_inputs):
+    torchdynamo.config.debug = True
+    prepare_mode = True
```
jansel, line 40: I think you need to add:
```suggestion
    global prepare_mode
    prepare_mode = True
```

torchbench.py
```diff
@@ -1068,13 +1126,15 @@ def run_one_model(
...
     skip_accuracy_check=False,
+    quant_exp=False,
 ):
     tolerance = 1e-4
     # Increase the tolerance for torch allclose
     if is_training and current_device == "cuda" and name in REQUIRE_HIGHER_TOLERANCE:
         tolerance = 1e-3
 
     with pick_grad(name, is_training):
+        original_model = copy.deepcopy(model)
```
jerryzh168, line 1137: Hi @jansel, I'm trying to get a speedup of quantized model to floating point model, do you think this is the correct way to do it?

here I did a deepcopy, and in L1158 to L1164 I used the quant_context to quantize the model, then I feed  the two models to `experiment` function

torchbench.py
```diff
@@ -329,6 +332,47 @@ def speedup_experiment(args, model_iter_fn, model, example_inputs):
...
+
+    for rep in range(args.repeat):
+        inputs = (
+            randomize_input(copy.deepcopy(example_inputs))
+            if should_randomize_input
+            else example_inputs
+        )
+
+        # interleave the runs to handle frequency scaling and load changes
+        with torchdynamo.run():
```
jansel, line 357: I think the issue might be this line here.

TorchDynamo cache on the code object level, so it possible some quantized compiled code is leaking into your baseline.

torchbench.py
jansel: See above.  TorchDynamo's cache is on the PyCode object level.  So when you enable `torchdynamo.run()` you might get code objects from the A side of the experiment leaking into the B side.

If you want to compare dynamo to dynamo, I would suggest:
1) measure the first one 30 times (using torchdynamo.optimize())
2) call torchdynamo.reset(), which clears all caches
3) measure the second one 30 times (using torchdynamo.optimize())

The interleaving thing won't work right if multiple sides of the experiment use dynamo.

jansel:

PR #261, opened 2022-05-19
Dump conv args into file

pyjhzwh:
Summary: 
In the FX graph extracted from python key tracing, when seeing calls to aten.convolution, dump args including shapes (for input/bias/weight) and different options such as padding/stride/kernel size/dilation/etc to a json file.

Usage:
```
with torchdynamo.optimize(conv_args_analysis):
   func_to_analysis()
```
Test:
`pytest tests/test_optimizations.py::TestOptimizations::test_log_conv_args`

### DIFF
--- a/torchdynamo/optimizations/log_args.py
+++ b/torchdynamo/optimizations/log_args.py
@@ -1,0 +1,63 @@
+import torch
+import json
+
+from torchdynamo.optimizations.python_key import python_key_normalize
+from torchinductor.lowering import tensor
+
+aten = torch.ops.aten
+
+
+class ConvArgsAnalysis(torch.fx.Interpreter):
+    """
+    Log arguments like input shape (input, bias, weights shape) 
+    and options(padding/stride/kernel size/dilation/etc) for
+    aten.convolution
+    """
+    def __init__(self, gm: torch.fx.GraphModule):
+        super().__init__(gm)
+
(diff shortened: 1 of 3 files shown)


torchdynamo/optimizations/log_args.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def run_node(self, n: torch.fx.Node):
+        result = super().run_node(n)
+
+        if n.op == "call_function":
+            if n.target == aten.convolution:
+                args, kwargs = self.fetch_args_kwargs_from_env(n)
+                assert len(args) == 9, "aten.convolution should have 9 args"
+                conv_args = {}
+                conv_args["input"] = args[0].shape
```
jansel, line 42: Perhaps also record `args[n].stride()`, so we can tell the difference between channels first vs last.

For completeness `args[n].dtype` might also be useful.

jansel (commented):
Overall looks good.  Are you running this with torchbench.py models?

jansel:
For the test failures (missing functorch), you can add a `unittest.skipIf` functorch is missing.

pyjhzwh:
> Overall looks good. Are you running this with torchbench.py models?

Added option `--log-conv-args` in torchbench.py

jansel (approved):

PR #262, opened 2022-05-19
enable tracing through id(nn_module_variable)

vkuzo:
Summary:

Enables tracing through this syntax:

```
class M(torch.nn.Module):
    def forward(self, x, ref_id):
        self_id = id(self)
        if self_id == ref_id:
            x = torch.mul(x, 1.0)
        x = torch.add(x, 1.0)
        return x
```

This is useful for DBR quant because it uses `id(module)` for some
FQN gymnastics.

Test plan:

```
pytest -vsk test_id_of_nn_module
```

### DIFF
--- a/torchdynamo/variables/builtin.py
+++ b/torchdynamo/variables/builtin.py
@@ -553,3 +553,11 @@ class BuiltinVariable(VariableTracker):
             return variables.TupleVariable(
                 items, **VariableTracker.propagate(self, iterable, *args)
             )
+
+    def call_id(self, tx, *args):
+        if len(args) > 0 and isinstance(args[0], variables.NNModuleVariable):
+            nn_mod_variable = args[0]
+            mod = tx.output.get_submodule(nn_mod_variable.module_key)
+            return variables.ConstantVariable(id(mod))
+        else:
+            unimplemented(f"call_id with args {args}")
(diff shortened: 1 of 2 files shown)


torchdynamo/variables/builtin.py
```diff
@@ -553,3 +553,11 @@ def call_islice(self, tx, iterable, *args):
             return variables.TupleVariable(
                 items, **VariableTracker.propagate(self, iterable, *args)
             )
+
+    def call_id(self, tx, *args):
+        if len(args) > 0 and isinstance(args[0], variables.NNModuleVariable):
+            nn_mod_variable = args[0]
+            mod = tx.output.get_submodule(nn_mod_variable.module_key)
+            return variables.ConstantVariable(id(mod))
```
vkuzo, line 561: n00b question, do I need to guard here?  I looked through other functions in this file and not exactly sure why some of them are using VariableTracker, some are guarding, and some are doing nothing.

jansel: The call to `add_options`, on this line:
https://github.com/pytorch/torchdynamo/blob/08ca9e500a652f08d653e90ae384121974edb20b/torchdynamo/variables/builtin.py#L221

Will copy the guards from the inputs to the outputs.  Since NNModuleVariable already has an ID_GUARD, we don't need a guard here.

jansel (approved):

PR #270, opened 2022-05-20
Add recompile ux tests

wconstab:
A first attempt at proposing
 - some of the actual tests (and the UX we are aiming for)
 - the mechanisms for setting/checking the #recompiles and cache limit to facilitate the testing

### DIFF
--- a/tests/test_recompile_ux.py
+++ b/tests/test_recompile_ux.py
@@ -1,15 +1,15 @@
-from torchdynamo.utils import counters
+import unittest
 from abc import ABC
 from abc import abstractmethod
 from typing import Any
 from typing import List
-import unittest
 
 import torch
 
 import torchdynamo
 import torchdynamo.config
 import torchdynamo.testing
+from torchdynamo.utils import counters
 
 torchdynamo.config.debug = False
 
@@ -22,8 +22,10 @@ class RecompileUxTests(torchdynamo.testing.TestCase):
     def setUpClass(cls):
         super().setUpClass()
         cls._exit_stack.enter_context(
-            unittest.mock.patch.object(torchdynamo.config, "cache_size_limit",
-                                       cls.cache_limit))
+            unittest.mock.patch.object(
+                torchdynamo.config, "cache_size_limit", cls.cache_limit
+            )
+        )
 
     def test_loop_torture(self):
         def loop_torture(input, iters):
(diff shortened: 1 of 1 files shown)


tests/test_recompile_ux.py
```diff
@@ -0,0 +1,92 @@
...
+from torchdynamo.utils import counters
+
+class RecompileUxTests(torchdynamo.testing.TestCase):
+    # TODO(whc) dynamo actualy recompiles one more time than the cache limit
+    max_recompiles = 2
+
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        torchdynamo.config.cache_size_limit = max(1, RecompileUxTests.max_recompiles - 1)
```
jansel, line 22: This will leak state into other tests.  Perhaps:
```suggestion
        cls._exit_stack.enter_context(
           unittest.mock.patch.object(torchdynamo.config, "cache_size_limit", max(1, RecompileUxTests.max_recompiles - 1))
```

Which is using this:
https://github.com/jansel/torchdynamo/blob/5e575216aa7467cda972cfa4cd3443175ca1d6b6/torchdynamo/testing.py#L159

tests/test_recompile_ux.py
```diff
@@ -0,0 +1,92 @@
...
+
+import torchdynamo
+import torchdynamo.config
+import torchdynamo.testing
+
+torchdynamo.config.debug = False
+from torchdynamo.utils import counters
+
+class RecompileUxTests(torchdynamo.testing.TestCase):
+    # TODO(whc) dynamo actualy recompiles one more time than the cache limit
```
jansel, line 16: I'd expect at least one extra compile to come from a graph break.

You should get one graph containing only `torch.randint(low=0, high=1000, size=())`, which gets compiled once.

The limit is on a per-code-object bases.  And graph breaks create new code objects after the break.

tests/test_recompile_ux.py
```diff
@@ -0,0 +1,92 @@
...
+                out = model(x)
+        
+        print(counters)
+        self.assertEqual(counters["frames"]["total"], TortureTests.max_recompiles)
+        self.assertEqual(
+            torchdynamo.utils.counters["frames"]["total"],
+            torchdynamo.utils.counters["frames"]["ok"],
+        )
+
+    def test_nvfuser_guards(self):
```
jansel, line 61: Likely need a requires cuda `unittest.skipIf`.   We have a that in some other files that you could refector into a common place.

jansel:
>     * the actual tests (and the UX we are aiming for)
 
These tests seem to capture graphs that would recompile too much.  I'm not sure about the UX, perhaps we need some sort of warning.

One idea is we could add something similar to `nopython=True` that turns guard failures into errors and explains why the guard failed.  We could extend `guard_fail_hook` to do that:
https://github.com/jansel/torchdynamo/blob/5e575216aa7467cda972cfa4cd3443175ca1d6b6/torchdynamo/guards.py#L388

>     * the mechanisms for setting/checking the #recompiles and cache limit to facilitate the testing

I'd consider using CompileCounter:
https://github.com/jansel/torchdynamo/blob/5e575216aa7467cda972cfa4cd3443175ca1d6b6/torchdynamo/testing.py#L103

Which is counting the number of calls to the backend compiler.   (So if dynamo decides to bail out, it won't get counted.)

tests/test_recompile_ux.py
wconstab: cool, yea that's a handy trick - i'll use it for now.

i'm not sure if I even want to set the cache limit in the test, or if i just want to change the cache limit on master and have the tests verify the intended behavior- so that's another thing I wanted to discuss.

probably makes sense to decide after going a little further with what UX the tests are converging towards

jansel (approved):

PR #273, opened 2022-05-20
Fix inline list/dict mutation

yanboliang:
*This is the 5/N issue found from https://github.com/pytorch/torchdynamo/issues/156.*

Minimal code to reproduce:
```
from typing import List
import torch
import torchdynamo

def f1(my_list):
    my_list.append(torch.tensor([3.0]))
    return my_list

def f2():
    my_list = []
    f1(my_list)
    return my_list

    return gm.forward

    print(f2())
```
The above code's output is:
```
[]
```
which is not correct, actually the output should be:
```
[tensor([3.])]
```

The root cause is the list variable was mutated inside of the inline translator, which was not reflected in the parent translator. The solution is to modify inline translator's ```replace_all``` method and update both parent and inline translator's local variables and stack values.
BTW, this also apply to dict variables and added tests for both.

### DIFF
--- a/torchdynamo/symbolic_convert.py
+++ b/torchdynamo/symbolic_convert.py
@@ -218,6 +218,20 @@ class InstructionTranslatorBase(object):
         )
         self.push(fn.call_function(self, args, kwargs))
 
+    def update_locals_and_stack(
+        self, oldvar: VariableTracker, newvar: VariableTracker, apply_side_effects=False
+    ):
+        def repl(v: VariableTracker):
+            if v.mutable_local is oldvar.mutable_local:
+                return newvar
+            return v
+
+        if apply_side_effects:
+            self.output.side_effects.apply(repl)
+        self.stack = [VariableTracker.apply(repl, x) for x in self.stack]
+        for k, x in self.symbolic_locals.items():
+            self.symbolic_locals[k] = VariableTracker.apply(repl, x)
+
     def replace_all(self, oldvar: VariableTracker, newvar: VariableTracker):
         if isinstance(
             oldvar.mutable_local, torchdynamo.side_effects.MutableSideEffects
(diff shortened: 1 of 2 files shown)


torchdynamo/symbolic_convert.py
```diff
@@ -1362,6 +1368,23 @@ def LOAD_CLOSURE(self, inst):
...
+    def is_from_parent(self, var: VariableTracker):
+        for k, v in self.parent.symbolic_locals.items():
+            if var.mutable_local is v.mutable_local:
+                return True
+        return False
+
+    def replace_all(self, oldvar: VariableTracker, newvar: VariableTracker):
+        if isinstance(
+            oldvar, (ListVariable, ConstDictVariable)
+        ) and self.is_from_parent(oldvar):
```
jansel, line 1380: I think we may just want to do this all the time.  It could be the case that we are mutating something nested inside a data structure.

yanboliang: Oh, yes! if this variable is not from parent, then parent translator's ```update_locals_and_stack``` do nothing.

torchdynamo/symbolic_convert.py
```diff
@@ -1362,6 +1365,12 @@ def LOAD_CLOSURE(self, inst):
         assert inst.argval in self.cell_and_freevars()
         self.push(self.closure_cells[inst.argval])
 
+    def replace_all(self, oldvar: VariableTracker, newvar: VariableTracker):
+        newvar = super().replace_all(oldvar, newvar)
+        # check and update parent's locals and stack in case oldvar is from parent
+        self.parent.update_locals_and_stack(oldvar, newvar)
```
jansel, line 1371: Will this work for multiple layers of recursive nesting?

jansel (approved):

PR #274, opened 2022-05-22
[Draft] Light Refactor + Add support for torch.autograd.profiler.record_profile function

voznesenskym:
This diff takes GradModeVariable's logic and pulls it partially into a more generic ContextWrappingVariable base class intended for making it easier to write context managed code.

This works, and passes all tests, but I am somewhat conflicted on the refactor. On one hand, it pulls a lot of boilerplate out, on the other hand, so far we dont have a ton of cases and it does make it more complicated to read. I also do not know what the n+1 case will look like, in terms of shared vs overriden logic.

### DIFF
--- a/torchdynamo/variables/misc.py
+++ b/torchdynamo/variables/misc.py
@@ -122,7 +122,6 @@ class ContextWrappingVariable(ContextManagerVariable):
         self._call_func(tx, self.initial_value)
         return variables.ConstantVariable(None, **VariableTracker.propagate(self))
 
-
     def reconstruct(self, codegen, target_inst=None):
         """
         Generate following Python Bytecode, with a `torch._C._set_grad_enable` call
@@ -239,10 +238,12 @@ class ContextWrappingVariable(ContextManagerVariable):
     def _initial_value(self):
         raise NotImplementedError("_initial_value called on base")
 
+
 class GradModeVariable(ContextWrappingVariable):
     def __init__(self, target_value, initial_value=None, **kwargs):
-        super(GradModeVariable, self).__init__(target_value=target_value, initial_value=
-        initial_value, **kwargs)
+        super(GradModeVariable, self).__init__(
+            target_value=target_value, initial_value=initial_value, **kwargs
+        )
 
     def enter(self, tx):
         assert self.initial_value == torch.is_grad_enabled()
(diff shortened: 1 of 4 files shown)


torchdynamo/variables/misc.py
```diff
@@ -243,6 +227,84 @@ def set_grad_insts(mode):
...
+        return torch.is_grad_enabled()
+
+    def fn_name(self):
+        if self.target_value:
+            return "enable_grad"
+        else:
+            return "no_grad"
+
+
+class ProfileRunModeVariable(ContextWrappingVariable):
```
jansel, line 271: I'd suggest `ProfileRecordFunctionVariable`

_record_function_enter doesn't actually change a mode, it is a way of inserting labels into a profile.

torchdynamo/variables/torch.py
```diff
@@ -180,6 +182,12 @@ def call_function(
                 tensor_with_tf_override.subclass_torch_function__func,
                 tensor_with_tf_override.subclass_type,
             )
+        elif self.value is torch.autograd.profiler.record_function:
+            assert len(args) == 1
+            return ProfileRunModeVariable(str(args[0].as_proxy()), **options)
+        elif self.value is torch.autograd.profiler.profile:
+            # Passthrough
+            return ConstantVariable(torch.autograd.profiler.profile, **options)
```
jansel, line 190: This isn't really a ConstantVariable.  I curious why this is needed.

jansel (changes requested):
Overall seems like a good approach.  There are a few more ones like this, so I like the refactor.

torchdynamo/variables/torch.py
voznesenskym: Good question. I don't know the correct way to make this pass-through, I want it to no-op. If it goes down to `return TensorVariable.create`, it will reach:

`assert (
                False
            ), f"torch.* op returned non-Tensor {typestr(example_value)} {proxy.node.op} {proxy.node.target}"
`
Is UserDefinedVariable more correct? or perhaps we need a new type here?

voznesenskym: Returning None actually works here, and is probably more correct. Thoughts?

torchdynamo/variables/misc.py
voznesenskym: agreed.

torchdynamo/variables/torch.py
jansel: ConstantVariable() will turn it into a constant (evaluated at compile time), which I don't think is what we want.

If we want to put this into the graph, then we should modify `TensorVariable.create` to detect this function wrap the return value properly.  

I think we may just want to break the graph on this by adding it to the disallowed list in `allowed_functions.py`.

jansel (approved):

Issue #276, opened 2022-05-23
Attribute and method priority inconsistent between Python and TorchDynamo NNModuleVariable


yanboliang:
*This is the 6/N issue found from https://github.com/pytorch/torchdynamo/issues/156.*

Minimal code to reproduce:
```
class HeadModule(nn.Module):

    def __init__(self):
        super(HeadModule, self).__init__()

    def forward(self, x):
        return x + 1

class MyModule(nn.Module):

    def __init__(self, **kwargs):
        super(MyModule, self).__init__(**kwargs)
        self.features = None
        self.head = HeadModule()

    def head(self, x):
        return x

    def forward(self, x):
        return self.head(x)

x = torch.ones([2, 2])
model = MyModule()
model.eval()

def my_compiler(gm: torch.fx.GraphModule, example_inputs: List[torch.Tensor]):
    return gm.forward

with torchdynamo.optimize(my_compiler):
    print(model(x))
```
w/o torchdynamo, the output is:
```
tensor([[1., 1.],
        [1., 1.]])
```
w torchdynamo, the output is:
```
tensor([[2., 2.],
        [2., 2.]])
```
The problem is: the attribute ```head``` and the method ```head``` are the same name, Python favors the method, however, torchdynamo prioritizes the attribute. I checked the code in [```NNModuleVariable.var_getattr```](https://github.com/pytorch/torchdynamo/blob/main/torchdynamo/variables/nn_module.py#L108) and found it seems intended behavior. I'd like to ask if this is intended and should we print warning here if we found method of the same name? @jansel

jansel:
This is a bug.  We should change TorchDynamo to ignore `self.__class__.head` if `self.__dict__["head"]` is defined.

yanboliang:
> We should change TorchDynamo to ignore self.__class__.head if self.__dict__["head"] is defined.
I think you mean ignore ```self.__dict__["head"]``` if ```self.__class__.head``` is defined, right?

jansel:
We should match whatever normal python does.

I thought object members took precedence: 
```
>>> class Foo:
...   bar = "on the class"
...   def __init__(self):
...     self.bar = "on the object"
... 
>>> Foo().bar
'on the object'
>>> 
```

Though perhaps I was wrong there.  You should base it on correct python behavior.

jansel:
Trying a bit more with `nn.Modules`:
```
>>> class Bar(torch.nn.Module):
...   def __init__(self):
...     super().__init__()
[...]

PR #278, opened 2022-05-24
Prioritize class method if there is duplicated attribute name

yanboliang:
Fix https://github.com/pytorch/torchdynamo/issues/276

### DIFF
--- a/torchdynamo/variables/nn_module.py
+++ b/torchdynamo/variables/nn_module.py
@@ -100,25 +100,19 @@ class NNModuleVariable(VariableTracker):
 
         base = tx.output.get_submodule(self.module_key)
         base_dict = object.__getattribute__(base, "__dict__")
+        base_class_dict = object.__getattribute__(base.__class__, "__dict__")
         class_member = True
 
         if not self.source:
             unimplemented("GETATTR with no source")
 
-        # class attribute has the highest priority
-        try:
-            subobj = inspect.getattr_static(base.__class__, name)
-            class_member = False
-        except AttributeError:
-            pass
-
-        if class_member and name in base_dict:
+        if name in base_dict:
             subobj = base_dict[name]
-        elif class_member and name in base_dict["_modules"]:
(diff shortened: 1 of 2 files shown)


torchdynamo/variables/nn_module.py
```diff
@@ -105,13 +105,20 @@ def var_getattr(self, tx, name):
         if not self.source:
             unimplemented("GETATTR with no source")
 
-        if name in base_dict:
+        # class attribute has the highest priority
+        try:
+            subobj = inspect.getattr_static(base.__class__, name)
+            class_member = False
```
yanboliang, line 111: I still use ```class_member``` but I think it's a confusing name, ```object_member``` should be better? Correct me if I'm wrong. @jansel

jansel:
So normal python objects have the opposite behavior
```
>>> class Foo:
...   def __init__(self):
...     self.hook = lambda: "on the object"
...   def hook(self):
...     return "on the class"
... 
>>> Foo().hook()
'on the object'
>>> 
```


Trying a bit more with `nn.Modules`:
```
>>> class Bar(torch.nn.Module):
...   def __init__(self):
...     super().__init__()
...     self.hook1 = torch.nn.ReLU()
...     self.hook2 = lambda: "on the object"
...     self.hook3 = "on the object"
...   def hook1(self):
...     return "on the class"
...   def hook2(self):
...     return "on the class"
...   def hook3(self):
...     return "on the class"
... 
>>> Bar().hook1
<bound method Bar.hook1 of Bar(
  (hook1): ReLU()
)>
>>> Bar().hook2()
'on the object'
>>> Bar().hook3
'on the object'
>>> 
```

As far as I can tell the object one wins.  What is different in this case?

jansel (changes requested):
We should update the logic to match what is in the linked comment above.

yanboliang:
See more details at https://github.com/pytorch/torchdynamo/issues/276#issuecomment-1136550186

jansel (approved):

PR #281, opened 2022-05-25
Collect guard failures into one warning at cache limit hit

wconstab:
- avoid warning on each guard failure separately (in cases cache limit >  1)
- instead, bundle a summary of gaurd failure warnings together at the  time of cache limit hit

Example Output:

> torchdynamo hit recompilation cache limit (2) for function 'model' (warn.py:39), due to the following guard failures: [['___check_tensors(input)'], ['___check_tensors(input)']]

Repro:
```
torchdynamo.config.cache_size_limit = 2

def model(input):
    return input + input

for i in range(10):
    bsz = torch.randint(low=0, high=1000, size=())
    x = torch.randn((bsz, 3, 4))
        out = model(x)

```

### DIFF
--- a/torchdynamo/guards.py
+++ b/torchdynamo/guards.py
@@ -1,6 +1,7 @@
 import collections
 import dataclasses
 import enum
+import logging
 import os
 import re
 import textwrap
@@ -27,6 +28,14 @@ from .utils import istype
 from .utils import rename_implicit
 from .utils import tuple_iterator_getitem
 from .utils import tuple_iterator_len
+
+log = logging.getLogger(__name__)
+
+# map from transformed code back to original user code
+orig_code_map = {}
+
+# keep a record of code_obj -> list of guard failure reasons for logging
+guard_failures = collections.defaultdict(list)
 
 CLOSURE_VARS = collections.OrderedDict(
     [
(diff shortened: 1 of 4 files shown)


torchdynamo/guards.py
```diff
@@ -28,6 +29,14 @@
 from .utils import tuple_iterator_getitem
 from .utils import tuple_iterator_len
 
+log = logging.getLogger(__name__)
+
+# map from transformed code back to original user code
+orig_code_map = {}
```
jansel, line 35: I'd suggest:
https://github.com/pytorch/torchdynamo/blob/c8e2f7aacd307b76fb294533eddc6b3093603bda/torchdynamo/utils.py#L42

So handle garbage collection a bit better.  This also lets you avoid the calls to `id()`.

torchdynamo/guards.py
```diff
@@ -28,6 +29,14 @@
...
+
+# keep a record of code_obj -> list of guard failure reasons for logging
+guard_failures = collections.defaultdict(list)
```
jansel, line 38: we should clear this and above in `torchdynamo.reset()`

jansel (changes requested):

jansel (approved):

PR #287, opened 2022-05-25
Implement verbose tensor guards check 

wconstab:
Verbose guard checks are guards used outside of the hot path
for providing specific failure information to the user on compile
cache miss.

This PR adds support for verbose guards and implements one for the
tensor guard, leaving other guards alone.

Note: depends on unmerged https://github.com/pytorch/torchdynamo/pull/281 - 
when reviewing this PR, only look at the latest commit "Implement verbose tensor guards check"

### DIFF
--- a/torchdynamo/_guards.cpp
+++ b/torchdynamo/_guards.cpp
@@ -1,5 +1,6 @@
 #define PY_SSIZE_T_CLEAN
 #include <Python.h>
+#include <sstream>
 #include <torch/extension.h>
 
 namespace {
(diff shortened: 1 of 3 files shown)


torchdynamo/_guards.cpp
```diff
@@ -59,6 +60,45 @@ class TensorCheck {
     return true;
   }
 
+  std::string check_verbose(const LocalState &state, const at::Tensor &v) {
```
wconstab, line 63: yes.. this is ugly.

i wasn't sure if it is worth expanding all the ||'d conditions into individual ones to get precise messages (yes?)

or if it is worth keeping this function totally separate from the main check function to avoid introducing more branches into it (also yes?)

torchdynamo/_guards.cpp
```diff
@@ -151,8 +191,38 @@ PyObject *TensorGuards_check(TensorGuards *self, PyObject *args) {
...
+  ssize_t len = PyTuple_GET_SIZE(args);
+
+  if (static_cast<ssize_t>(checks.size()) != len) {
+    PyErr_SetString(PyExc_TypeError, "wrong length");
+    return NULL;
+  }
+
+  LocalState state;
+
+  for (ssize_t i = 0; i < len; ++i) {
```
wconstab, line 209: it seems more conceivable that we would introduce some branches in `TensorGuards_check` to avoid code duplication, and keep the 'check_verbose' part separate?

torchdynamo/guards.py
```diff
@@ -334,34 +334,45 @@ def compile_check_fn(self, local_builder, global_builder):
         code_parts = (
             ["___guarded_code.valid"] + local_builder.code + global_builder.code
         )
+        # TODO(whc) maybe only the 'check_tensors' one is ambiguous? if so we can be less general..
```
wconstab, line 337: i'm thinking that for a bunch of the guards, just printing out their 'part' from the code (e.g. printing `guard failed: ___guarded_code.valid`) is specific enough to be useful, but, for some like odict_keys, it would be nice to actually print the mismatched keys

torchdynamo/_guards.cpp
jansel: I'd be fine with this as long we measure the overheads to make sure it doesn't hurt things

torchdynamo/guards.py
jansel: Yeah.  The list might be small though.  There are keys, lengths, constants.

wconstab:
> Overall approach lgtm. Just wondering if there is any major performance overhead.

I didn't benchmark, but, I implemented it in a way that trades off being more verbose/duplicating code in order to avoid changing the fast path.  In other words, i think all of my changes only affect the path that runs when the cache misses, so i think it should be OK.

One thing to consider (pending benchmarking) is to try to make the code less duplicated but that requires adding more branches in the 'real' guard functions, which might not be worth the trouble.

jansel (approved):

PR #289, opened 2022-05-26
[inductor] Improved indexing simplification and loop body representation

jansel:


torchinductor/sizevars.py
```diff
@@ -207,3 +280,67 @@ def codegen_shape_tuple(self, shape):
         if len(parts) == 1:
             return f"({parts[0]}, )"
         return f"({', '.join(parts)})"
+
+
+def join_dimensions(expr: sympy.Expr) -> sympy.Expr:
+    """
+    ModularIndexing(i0, 1, 32) + 32 * ModularIndexing(i0, 32, 4)
```
ngimel, line 287: wow, that's pretty cool!

tests/test_torchinductor.py
```diff
@@ -165,6 +169,84 @@ class SweepInputsCpuTest(SweepInputs2, TestCase):
 SweepInputsCpuTest.populate()
 
 
+class TestIndexingSimplification(unittest.TestCase):
+    def test_indexing_simplification(self):
```
ngimel, line 173: tests are pretty cool and show how simplifications are supposed to work, but (if we are to use them as documentation) would benefit from brief comments on why those results are expected)

torchinductor/codegen/triton.py
```diff
@@ -279,8 +284,12 @@ def indexing(self, index: sympy.Expr):
         offset = index.subs({v: 0 for v in chain(iter_vars, reduction_vars)})
         base_part = index.subs({v: 0 for v in reduction_vars}) - offset
         reduction_part = index.subs({v: 0 for v in iter_vars}) - offset
-
-        assert index == offset + base_part + reduction_part
+        assert index == offset + base_part + reduction_part, (
```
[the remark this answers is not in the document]
jansel: fixed

tests/test_torchinductor.py
jansel: fixed

ngimel (approved):
That's very cool

PR #290, opened 2022-05-26
All base class methods take precedence if it's a nn.Module

yanboliang:
Literally this is a follow up of https://github.com/pytorch/torchdynamo/issues/276.
I found a similar bug in the github 1k models, if the class attribute is defined in the base class, it still takes precedence if it's a ```nn.Module```.

### DIFF
--- a/torchdynamo/variables/nn_module.py
+++ b/torchdynamo/variables/nn_module.py
@@ -100,15 +100,17 @@ class NNModuleVariable(VariableTracker):
 
         base = tx.output.get_submodule(self.module_key)
         base_dict = object.__getattribute__(base, "__dict__")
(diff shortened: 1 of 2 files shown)


torchdynamo/variables/nn_module.py
```diff
@@ -100,28 +100,30 @@ def var_getattr(self, tx, name):
 
         base = tx.output.get_submodule(self.module_key)
         base_dict = object.__getattribute__(base, "__dict__")
-        base_class_dict = object.__getattribute__(base.__class__, "__dict__")
-        class_member = True
+        object_member = True
```
yanboliang, line 103: Changing from ```class_member``` to ```object_member``` is not necessary for this PR, just make it more clear.

jansel (approved):

PR #294, opened 2022-05-27
Fx2trt pr4

frank-wei:
1. add 3 inplace ops in normalization
2. disable int64->int32 input transform as it may affect model accuracy


anijain2305 (approved):
LGTM

PR #296, opened 2022-05-28
Fx2trt pr5

frank-wei:
1. Add a threshold for falling back to non-TRT. Models like hf_Reformer will benefit from it.


anijain2305 (approved):
lgtm!

Just wondering if its the number of pieces (which I am guessing means subgraphs within a FX graph) or the length of those pieces that define the performance?

jansel (approved):

frank-wei:
> lgtm!
> 
> Just wondering if its the number of pieces (which I am guessing means subgraphs within a FX graph) or the length of those pieces that define the performance?

Under this scenario(for most cases, graph has size less than 1000 lines), the number of pieces matters since the overhead of starting a TRT piece/subgraph is relatively big. If users want to squeeze the performance, this threshold should be tuned.

Issue #165, opened 2022-04-21
CUDA+inference correctness issue with torchdynamo eager mode on TorchBench


xuzhao9:
I did a TorchBench sweep run workflow on V100 and the result is available [here](https://github.com/pytorch/benchmark/actions/runs/2201634569). Or view the spreadsheet result [Meta internal only](https://docs.google.com/spreadsheets/d/1eQw4Y5x4ynOZx0Fgpm47ooWAortaTwM4SPxc1y1A-ew/edit#gid=966748872).
In this workflow, I have found the following correctness issues with TorchDynamo eager mode on CUDA inference tests:

- [x] dlrm (similarity: 0.0)
- [ ] fastNLP_Bert (similarity: 0.978)
- [ ] moco (similarity: 0.0)
- [x] pyhpc_turbulent_kinetic_energy (similarity: 0.0)
- [x] timm_efficientnet (similarity: 0.01)

The cosine similarity value is calculated comparing with the fp32 result.

jansel:
@xuzhao9 these results contradict [the correctness checking in torchbench.py](https://github.com/facebookresearch/torchdynamo/blob/26cb2b01f389172c8bbce900a44009e698b818b5/torchbench.py#L1074).

When I run these models with `torchbench.py` they pass:
```
$ ./torchbench.py -k dlrm -k fastNLP_Bert -k moco -k timm_efficientnet -dcuda --backend=eager -n2
cuda eval  dlrm                               1.009x SAME
cuda eval  fastNLP_Bert                       0.989x SAME
cuda eval  moco                               1.001x SAME
cuda eval  timm_efficientnet                  1.019x SAME
speedup gmean=1.03x mean=1.03x
```

Note torchbench.py doesn't check `pyhpc_turbulent_kinetic_energy` because it is non-determinstic.  It fails `torch.allclose()` checks even if you just run it twice in eager mode and is not self-consistent.  But the other 4 pass.  Because of this I have excluded it from my results.

These tests are using [torch.allclose](https://github.com/facebookresearch/torchdynamo/blob/26cb2b01f389172c8bbce900a44009e698b818b5/torchdynamo/testing.py#L100), which I believe is a harder test to pass than cosine similarity.

Can you try to figure out what is different compared to your experimental setup?  Perhaps one of them have a bug.  I know torchbench.py does a lot of stuff to setup random state correctly.

xuzhao9:
[...]

PR #297, opened 2022-05-28
Add guard check within NN_MODULE_PARAM_NAMES to ensure parameter obj ids have not changed

voznesenskym:
Context: https://github.com/pytorch/torchdynamo/issues/165

### DIFF
--- a/tests/test_modules.py
+++ b/tests/test_modules.py
@@ -558,10 +558,10 @@ class ModuleAttributePrecedence(ModuleAttributePrecedenceBase):
         return self.activation(self.linear(self.initializer + x)) * self.scale
 
 
-def make_test(fn, expected_ops=None):
+def make_test(fn, expected_ops=None, expected_frame_count=4):
     def test_fn(self):
         return torchdynamo.testing.standard_test(
-            self, fn=fn, nargs=1, expected_ops=expected_ops
+            self, fn=fn, nargs=1, expected_ops=expected_ops, expected_frame_count=expected_frame_count
         )
 
     fn.eval()
@@ -588,26 +588,26 @@ class NNModuleTests(torchdynamo.testing.TestCase):
     test_viamodulecall = make_test(ViaModuleCall())
     test_isnonelayer = make_test(IsNoneLayer())
     test_layerlist = make_test(LayerList())
-    test_tensorlist = make_test(TensorList())
+    test_tensorlist = make_test(TensorList(), expected_frame_count=1)
     test_intarg = make_test(IntArg())
     test_cfgmod = make_test(CfgModule())
     test_stringmember = make_test(StringMember())
     test_modulelist = make_test(ModuleList())
     test_moduledict = make_test(ModuleDict())
     test_super1 = make_test(SuperModule())
-    test_super_class_method = make_test(SuperChildCallsClassMethod())
+    test_super_class_method = make_test(SuperChildCallsClassMethod(), expected_frame_count=1)
     test_children = make_test(Children())
-    test_densenet = make_test(DenseNetBlocks())
+    test_densenet = make_test(DenseNetBlocks(), expected_frame_count=1)
     test_parameters1 = make_test(ParametersModule1())
     test_parameters2 = make_test(ParametersModule2())
-    test_parameters3 = make_test(ParametersModule3(), expected_ops=5)
+    test_parameters3 = make_test(ParametersModule3(), expected_ops=20 if torchdynamo.config.guard_nn_modules else 5)
     test_hasattr = make_test(HasAttrModule())
-    test_enumvalues = make_test(EnumValues())
+    test_enumvalues = make_test(EnumValues(), expected_frame_count=1)
     test_module_class_method = make_test(ModuleClassMethodCall())
-    test_module_property = make_test(ModuleProperty())
+    test_module_property = make_test(ModuleProperty(), expected_frame_count=1)
     test_forward_directly = make_test(CallForwardDirectly())
     test_module_name_string = make_test(ModuleNameString())
-    test_module_attribute_precedence = make_test(ModuleAttributePrecedence())
+    test_module_attribute_precedence = make_test(ModuleAttributePrecedence(), expected_frame_count=1)
 
     def test_unsupportedmethod(self):
         m = UnsupportedMethodCall()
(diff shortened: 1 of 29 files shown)


jansel:
I think this will just cause a recompile for every new input.

```
@torchdynamo.optimize("eager")
def foo(x):
    return x + 1

for _ in range(100):
   fn(torch.randn(10))
```

Without this PR, the above example would compile once.  With this PR the above example would compile 64 times (every iteration), then hit the [cache_size_limit](https://github.com/pytorch/torchdynamo/blob/26c4c1b593bebf4246e566749dade38254b59ffb/torchdynamo/config.py#L23) and disable itself for that PyCode object. 

We need to compile graphs that are input value agnostic to be useful.  If we don't, then a compiled graph is unlikely to every be reused.  (And since it only every gets called with a single input, we could just make it return a constant.) 

For debugging the linked issue I'd suggest:

First:
Figure out which PyCode object is causing the issue by iteratively enabling and disabling TorchDynamo on a per-frame basis.  If you [return None from _convert_frame_assert](https://github.com/pytorch/torchdynamo/blob/26c4c1b593bebf4246e566749dade38254b59ffb/torchdynamo/convert_frame.py#L201), then TorchDynamo will skip the given PyCode object.

Often there are just a few frames, so you can search with TORCHDYNAMO_DEBUG_FUNCTION.  In the past I've done some rudimentary binary searches with patterns like:
```
global global_counter
global_counter+=1
if global_counter % 2 != 1 or (global_counter>>1) % 2 != 0:
   return None
```

I never checked that type of stuff in, but perhaps we should make that type of thing more general as a debugging tool.  :)

Next:
Hopefully you have narrowed things down to a specific function at this point.  (see `f"{code.co_filename}:{code.co_firstlineno}"` to map code objects to functions)

At this point a usually look at the source function, the generated FX graph, and the guards see if something jumps out at me.  Often the function is doing something odd/unique.

Since, adding a value-guard fixes this issue.  You could also add a print to the function (in the user code) to see how it is being called.  

If I am still stumped, I try to extract a minimal repro.  I will iteratively remove stuff from the source model to get a smaller and smaller piece of code to reproduce the issue.  This is where most of the stuff in test_repros.py came from.

voznesenskym:
Ah understood. That seems reasonable. Thank you for the suggestion, I’ll
try that.

torchdynamo/guards.py
```diff
@@ -268,12 +268,19 @@ def DICT_KEYS(self, guard):
...
         ref = self.arg_ref(guard)
         value = self.get(guard.name)
         keys = {k for k, v in value.named_parameters()}
+        values = {v for k, v in value.named_parameters()}
         self.code.append(f"___check_type_id({ref}, {self.id_ref(type(value))})")
-        self.code.append(f"{{k for k, v in {ref}.named_parameters()}} == {keys!r}")
+        self.code.append(f"{{k for k, v in {ref}.named_parameters()}} == {keys!r}")     
+        # TODO after draft diff, before push - make a config value for this?   
+        values_collapsed = f"{values}".replace(' ', '').replace('\n', '')
+        self.code.append(f"{{v for k, v in {ref}.named_parameters()}} == {values_collapsed!r}")
```
jansel, line 281: Guarding on exact tensor values like this will:
1) Be very expensive, since it is constructing a Tensor inline.  (Which could result in a guard string of possibly megabytes in size.)
2) Will cause this guard to fail every iteration in training, because the parameters are being learned and change.
3) I think this guard will just always fail:
```
>>> value=torch.nn.Linear(2,2)
>>> values = {v for k, v in value.named_parameters()}
>>> values_collapsed = f"{values}".replace(' ', '').replace('\n', '')
>>> ref = "value"
>>> code = f"{{v for k, v in {ref}.named_parameters()}} == {values_collapsed!r}"
>>> eval(code)
False
>>> code
"{v for k, v in value.named_parameters()} == '{Parametercontaining:tensor([[0.6959,0.2447],[-0.5361,0.3697]],requires_grad=True),Parametercontaining:tensor([0.5704,0.1800],requires_grad=True)}'"
>>> {v for k, v in value.named_parameters()} == '{Parametercontaining:tensor([[0.6959,0.2447],[-0.5361,0.3697]],requires_grad=True),Parametercontaining:tensor([0.5704,0.1800],requires_grad=True)}'
False
>>> {v for k, v in value.named_parameters()}
{Parameter containing:
tensor([[ 0.6959,  0.2447],
        [-0.5361,  0.3697]], requires_grad=True), Parameter containing:
tensor([0.5704, 0.1800], requires_grad=True)}
>>> 


```
Note that `eval(code)` returned false even for an identical `value`

For a tensor on an NN_MODULE perhaps we want to use `ID_MATCH` instead.  That will check to see if the user is swapping out parameters dynamically, which will confuse things as NNModuleVariable assumes that doesn't happen.  If that is the case, we may want to try to switch the class in question to [UnspecializedNNModuleVariable](https://github.com/pytorch/torchdynamo/blob/063bc13ee2e9f41e25e4b4ac96a62adcb944697b/torchdynamo/variables/nn_module.py#L327).

jansel (changes requested):

torchdynamo/guards.py
voznesenskym: Great suggestions, let me try.

And yes, this is prohibitively expensive.

torchdynamo/guards.py
```diff
@@ -272,6 +274,14 @@ def NN_MODULE_PARAM_NAMES(self, guard):
         ref = self.arg_ref(guard)
         value = self.get(guard.name)
         keys = {k for k, v in value.named_parameters()}
+        if config.eval_model_named_parameter_checks_enabled and not value.training:
```
jansel, line 277: I'd actually expect this to work in training.  The ID should not change, just the values.

torchdynamo/config.py
```diff
@@ -66,3 +66,12 @@
     "torch.distributions",
     "torch.testing",
 }
+
+# Enable this check if you expect your eval model to change on forward passes - this ensures guards do a deeper check on parameter names and any named parameters you expect to change (see: eval_model_named_parameter_checks).
+# Note: guard_nn_modules must be True for this to run.
+eval_model_named_parameter_checks_enabled = True
```
jansel, line 72: How expensive is this?  (You can check with torchbench.py --overhead.)

I'd be tempted just to turn this on for everything (and all names) if the cost is low.

jansel (changes requested):

tests/test_modules.py
```diff
@@ -558,10 +558,14 @@ def forward(self, x):
         return self.activation(self.linear(self.initializer + x)) * self.scale
 
 
-def make_test(fn, expected_ops=None):
+def make_test(fn, expected_ops=None, expected_frame_count=4):
```
jansel, line 561: Why is the 4 needed?  Are we sure it is right?  I don't think most of these should be swapping out parameters.

jansel (changes requested):

PR #302, opened 2022-05-29
update_locals_and_stack should use shared cache

yanboliang:
*This is the 7/N issue found from https://github.com/pytorch/torchdynamo/issues/156.*

Failed models: [```test_LoSealL_VideoSuperResolution.py```](https://github.com/yanboliang/pytorch-jit-paritybench/blob/master/generated/test_LoSealL_VideoSuperResolution.py#L589)

Minimal code to reproduce:
```
from typing import List
import torch
import torchdynamo

def fn(x):
    perm = [0, 3, 5]
    perm = [i for i in range(min(perm))] + perm
    perm.extend(i for i in range(x.dim()) if i not in perm)
    return perm

    return gm.forward

input = torch.rand([2, 2, 2, 2, 2, 2])

    print(fn(input))
```
w/o torchdynamo, the output is: ```[0, 3, 5, 1, 2, 4]```
w torchdynamo, the output is: ```[0, 3, 5]```

Root cause and proposed solution:
* ```perm``` which is ```ListVariable``` occurs twice in the stack, so inside of [```compile_subgraph```](https://github.com/pytorch/torchdynamo/blob/main/torchdynamo/output_graph.py#L204) it should be listed in the ```pass2.tempvars```. ```pass2.tempvars``` is decided by [```self.uses```](https://github.com/pytorch/torchdynamo/blob/main/torchdynamo/codegen.py#L122) counter, whose key is the variable object(actually is memory address).
* Every time when a list change happens, ```replace_all``` will be called and meanwhile update variable objects in ```side_effect, stack and symbolic_locals```. However, the current implementation use separate cache, which makes the same variable links to different object(as [```VariableTracker.apply```](https://github.com/pytorch/torchdynamo/blob/main/torchdynamo/variables/base.py#L93) returns a new cloned object). We should use shared cache, then the same variable at ```side_effect, stack and symbolic_locals``` are the same object.

### DIFF
--- a/tests/test_misc.py
+++ b/tests/test_misc.py
@@ -1458,3 +1458,17 @@ class MiscTests(torchdynamo.testing.TestCase):
             res = fn(sample)
 
         self.assertTrue(same(ref, res))
+
+    def test_update_locals_and_stack_uses_shared_cache(self):
+        def fn(x):
+            perm = [0, 3, 5]
+            perm = [i for i in range(min(perm))] + perm
+            perm.extend(i for i in range(x.dim()) if i not in perm)
+            return perm
+
+        x = torch.rand([2, 2, 2, 2, 2, 2])
+        res1 = fn(x)
+        cnts = torchdynamo.testing.CompileCounter()
+        with torchdynamo.optimize(cnts):
+            res2 = fn(x)
+        self.assertTrue(same(res1, res2))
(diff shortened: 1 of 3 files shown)


anijain2305 (approved):
LGTM! But let's wait for Jason's review. I have not worked on side_effects portion.

jansel (approved):

PR #303, opened 2022-05-30
PyTorch tests - 6/n - Add type check for list/tuple elems in CONSTANT_MATCH

anijain2305:
For guards like

~~~
isinstance(args, list) # ---> its fast equivalent
and args = (110, 11)
~~~

if `args` actual runtime value is a list of tensor like 

~~~
(tensor(...), )
~~~

The type check passes and it proceeds to comparing Tensor and a scalar value, leading to boolean Tensor and guard executing failure. This PR adds an extra type checks on each element of the list/tuple when it is a CONSTANT_MATCH guard.

This PR also add list length guard recursively for list.

### DIFF
--- a/torchdynamo/guards.py
+++ b/torchdynamo/guards.py
@@ -221,6 +221,12 @@ class GuardBuilder:
 
         # Add a type check to prevent boolean check of a tensor and non-tensor.
         self.code.append(f"___check_type_id({ref}, {self.id_ref(type(val))})")
+        if istype(val, (list, slice, tuple)):
+            for idx, elem in enumerate(val):
+                self.code.append(
+                    f"___check_type_id({ref}[{idx}], {self.id_ref(type(elem))})"
+                )
+
         self.code.append(f"{ref} == {val!r}")
 
     def CONSTANT_MATCH(self, guard: Guard):
--- a/torchdynamo/guards.py
+++ b/torchdynamo/guards.py
@@ -263,6 +263,18 @@ class GuardBuilder:
         self.code.append(f"___check_type_id({ref}, {self.id_ref(type(value))})")
         self.code.append(f"len({ref}) == {len(value)}")
 
+        def add_guard_recursively(prefix_index_str, lst):
+            for idx, elem in enumerate(lst):
+                if istype(elem, (list, tuple)):
+                    r = f"{ref}{prefix_index_str}[{idx}]"
+                    self.code.append(
+                        f"___check_type_id({r}, {self.id_ref(type(elem))})"
+                    )
+                    self.code.append(f"len({r}) == {len(elem)}")
+                    add_guard_recursively(f"{prefix_index_str}[{idx}]", elem)
+
+        add_guard_recursively("", value)
+
     def TUPLE_ITERATOR_LEN(self, guard):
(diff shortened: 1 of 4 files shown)


torchdynamo/guards.py
```diff
@@ -221,6 +221,12 @@ def EQUALS_MATCH(self, guard: Guard):
 
         # Add a type check to prevent boolean check of a tensor and non-tensor.
         self.code.append(f"___check_type_id({ref}, {self.id_ref(type(val))})")
+        if istype(val, (list, slice, tuple)):
+            for idx, elem in enumerate(val):
+                self.code.append(
+                    f"___check_type_id({ref}[{idx}], {self.id_ref(type(elem))})"
```
jansel, line 227: I think we need to add a length check before reading `[{idx}]` or the guard could throw in out of range exception.

torchdynamo/guards.py
```diff
@@ -257,6 +263,18 @@ def LIST_LENGTH(self, guard):
         self.code.append(f"___check_type_id({ref}, {self.id_ref(type(value))})")
         self.code.append(f"len({ref}) == {len(value)}")
 
+        def add_guard_recursively(prefix_index_str, lst):
```
jansel, line 266: If we are going to add this to the `LIST_LENGTH` guard, then we should rename it to something else since it is now checking more than just length.

For length of nested lists, I would expect there to be a different guard that checks the inner length.  For something like `[[Foo()]]`, I'd expect:
```
ListVariable([
      ListVariable([
          UserDefinedObjectVariable(..., guards={TYPE_MATCH of inner object})
        ],
        guards={LIST_LENGTH of inner list}
      ),
   ],
   guards={LIST_LENGTH of outer list}
)
```

So there would be 3 different guards to check the 2 lengths and the type of the inner value.  If this is failing somewhere, perhaps we need to do something like copy guards from the items of the list up to the containing list.   Keeping the guards smaller and sperate will allow them to get automatically de-dupliated.

jansel (changes requested):

torchdynamo/guards.py
anijain2305: Hmm! You are right. I think currently for some reason, we are not generating ListVariable for the inner list. I am making it more complicated than necessary. Let me check.

torchdynamo/guards.py
anijain2305: Added a LIST_LENGTH guard at L222.

torchdynamo/guards.py
anijain2305: Removed this. Now, recursively collecting guards by walking the TupleVariable.

anijain2305:
Let's hold on this. Too many guards on a real model :(

image

anijain2305:
@jansel This is ready for review. The additional guards added above are correct.

jansel (approved):

ezyang:
This PR eliminated the type check for list/tuple EQUALS_MATCH: as in, we will accept a list where a tuple was previously expected. Is this expected?

ezyang:
OK, I can trigger a soundness error without this guard, going to submit a fix

ezyang:
https://github.com/pytorch/pytorch/pull/96421

Issue #107, opened 2022-04-05
Get TorchBench result on Dynamo inference TRT


xuzhao9:
The CI is ready, working on understanding the results - quite different from what we get from `torchbench.py`

xuzhao9:
Update: the result is available in an internal spreadsheet. However, the result is different from what we get from `torchbench.py` because we are compiling contexts everytime using `with torchdynamo.optimize(<backend>)` while `torchbench.py` uses `torchdynamo.run()`, which caches stuff and avoids re-compilation overheads.

jansel:
@xuzhao9 are you are `.run()` is the difference?  I would expect `optimize()` and `run()` to do the exact same thing unless there are recompiles (which there shouldn't be in most benchmarks).

xuzhao9:
@jansel You're right. Looks like the difference is not caused by `optimize()` and `run()`.
I managed to create a script to reproduce the problem on A100, please help verify:
```
import torch
import time
import gc
import numpy as np
from torchbenchmark import load_model_by_name
import argparse

def synchronize():
    torch.cuda.synchronize()

def timed(model, example_inputs, times=1, dynamo=False):
    synchronize()
    gc.collect()
    torch.manual_seed(1337)
    t0 = time.time_ns()
    # Dont collect outputs to correctly measure timing
    if dynamo:
        with torchdynamo.run():
            result = model(*example_inputs)
    else:
        result = model(*example_inputs)
    synchronize()
    t1 = time.time_ns()
    return (t1 - t0) / 1_000_000

def speedup_experiment(model, example_inputs, dynamo=False):
    repeat = 100
    timings = np.zeros((repeat, 2), np.float64)
    for rep in range(repeat):
        # interleave the runs to handle frequency scaling and load changes
        timings[rep, 0] = timed(model, example_inputs)
        if dynamo:
            timings[rep, 1] = timed(model, example_inputs, dynamo=True)
    median = np.median(timings, axis=0)
    print(f"Eager Latency: {median[0]} ms")
    if dynamo:
        print(f"TorchDynamo Eager latency: {median[1]} ms")
        print(f"speedup: {median[0]/median[1]} ")

if __name__ == "__main__":
    parser = argparse.ArgumentParser()
    parser.add_argument("--torchdynamo", action="store_true", help="load torchdynamo library")
    args = parser.parse_args()
    if args.torchdynamo:
        import torchdynamo
        optimize_ctx = torchdynamo.optimize("eager")
[...]

PR #305, opened 2022-05-31
Fix slowdown due to generation_tagging_new

mlazos:
* Fix slowdown due to monkey-patching new() of torch modules

Generation tagging was previously implemented by patching new() to populate a map of object -> generation tag. This changes the implementation to patch init() and setstate() to directly populate a generation attribute on the object. 

The comparison below was made by running alexnet ([test script](https://github.com/pytorch/torchdynamo/issues/107#issuecomment-1102872948)) in eager mode with and without torchdynamo, and computing the ratio of torch dynamo eager latency / eager mode latency * 100.

Ceiling was computed by making every module non-dynamic (not tracking generation tags) which shows the overhead of torchdynamo in isolation. The other values were computed by averaging 10 runs of the [test script](https://github.com/pytorch/torchdynamo/issues/107#issuecomment-1102872948).

| |ceiling|patched init with map | patched new | patched init with attr
-- | -- | -- | -- | --
Result  | 96.74 | 94.74 | 91.22 | 95.11

### DIFF
--- a/torchdynamo/mutation_guard.py
+++ b/torchdynamo/mutation_guard.py
@@ -55,12 +55,11 @@ def ensure_patched(cls):
 
 class GenerationTracker:
     generation = 0
-    db = ExactWeakKeyDictionary()
     dynamic_classes = ExactWeakKeyDictionary()
 
     @classmethod
     def tag(cls, obj):
-        cls.db[obj] = cls.generation
+        obj.generation = cls.generation
 
     @staticmethod
     def mark_class_dynamic(cls):
@@ -69,16 +68,7 @@ class GenerationTracker:
 
     @classmethod
     def check(cls, obj):
-        return cls.db.get(obj, -1) == cls.generation
-
-
-def generation_tagging_new(cls, *args, **kwargs):
-    try:
-        obj = object.__new__(cls)
-    except TypeError:
-        obj = object.__new__(type(cls))
-    GenerationTracker.tag(obj)
-    return obj
+        return getattr(obj, "generation", -1) == cls.generation
 
 
 def is_dynamic_nn_module(obj):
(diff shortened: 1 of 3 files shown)


anijain2305:
For lint, you could do `make lint-deps format lint`

anijain2305 (approved):

torchdynamo/mutation_guard.py
```diff
@@ -88,15 +78,24 @@ def is_dynamic_nn_module(obj):
...
     """
-    assert (
-        Module.__dict__.get("__new__") is generation_tagging_new
-        or Module.__new__ is object.__new__
-    )
-    Module.__new__ = generation_tagging_new
-    GenerationTracker.db.clear()
+    init = Module.__init__
+
+    def patched_init(self, *args, **kwargs):
```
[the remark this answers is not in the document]
mlazos: Jason and I agreed that we think there's a fast-path for calls to an unmodified new() in CPython. It isn't monkey-patching itself that's causing the slowdown (that's only done once when we enter the torchdynamo context).

For two vs one method, we need to patch two methods now because when you unpickle, it calls new() to allocate, but doesn't call init(), it calls setstate(). This impl makes more sense anyway, because rather than having some map floating somewhere with id(obj) as the key mapping to the tag, we just put an attribute for the tag directly on the object.

torchdynamo/mutation_guard.py
```diff
@@ -88,15 +78,24 @@ def is_dynamic_nn_module(obj):
...
+        init(self, *args, **kwargs)
+        GenerationTracker.tag(self)
+
+    Module.__init__ = patched_init
```
jansel, line 92: We should make sure this patch only happens once.  I think currently it will happen again and again (once for each call to torchdynamo.optimize)

jansel (changes requested):

torchdynamo/mutation_guard.py
mlazos: Yeah you're right, that's a good catch. I'll push a fix.

jansel (approved):

PR #309, opened 2022-06-01
[inductor] Minor fixes for latest PyTorch and benchmark harness codegen

jansel:


torchinductor/microbench.py
```diff
@@ -22,7 +21,6 @@
 
 def timed(model, example_inputs, times=1):
     synchronize()
-    gc.collect()
```
voznesenskym, line 25: jw/ why did we GC here, and why did we remove it?

jansel: gc defeats the caching allocator, we want to allow cached allocations.

PR #312, opened 2022-06-02
Add a basic compilation profiler

wconstab:


jansel (approved):

wconstab:
@jansel it would be nice if we could wrap real backends in the profiler and get profiling information that respects the true set of guards a backend would apply.  In the current form I guess there are a lot of generic guards but any guards that come in downstream due to some backend optimizations would be missed.  Otoh, I don't think we want the profiler to actually run compilation in case it's expensive, but may be that's a small price to pay.

I guess I could make it optional: If you pass a backend to your profiler, it actually compiles using the backend which is slower but more accurate; if not, it just does what it does today which is faster but has less guards.

jansel:
> @jansel it would be nice if we could wrap real backends in the profiler and get profiling information that respects the true set of guards a backend would apply. In the current form I guess there are a lot of generic guards but any guards that come in downstream due to some backend optimizations would be missed. Otoh, I don't think we want the profiler to actually run compilation in case it's expensive, but may be that's a small price to pay.
> 
> I guess I could make it optional: If you pass a backend to your profiler, it actually compiles using the backend which is slower but more accurate; if not, it just does what it does today which is faster but has less guards.

I agree.  Currently, no guards come from the backend up to the TorchDynamo level.  So we would need a way to detect recompiles coming from the lower level caches.  Currently, that is the profiling executor cache and the nvfuser cache.

wconstab:
> I agree. Currently, no guards come from the backend up to the TorchDynamo level. So we would need a way to detect recompiles coming from the lower level caches. Currently, that is the profiling executor cache and the nvfuser cache.

oh, i thought that inductor was introducing more guards, and i also wondered if AOTAutograd would introduce more guards.  I wasn't considering the Profiling Executor or nvfuser guards.

jansel:
> oh, i thought that inductor was introducing more guards, and i also wondered if AOTAutograd would introduce more guards. I wasn't considering the Profiling Executor or nvfuser guards.

The plan is to bubble up guards to have a single cache, but that is not yet implemented.

PR #313, opened 2022-06-02
Add constant checks for list of numpy integers

anijain2305:
cc @voznesenskym


torchdynamo/variables/builder.py
```diff
@@ -273,19 +273,7 @@ def _wrap(self, value):
...
-                np.int16,
-                np.int32,
-                np.int64,
-                np.uint8,
-                np.uint16,
-                np.uint32,
-                np.uint64,
-            ),
-        ):
+        elif is_numpy_int_type(value):
```
anijain2305, line 276: On the `main` branch, there is no `elif`, but I think that was a typo.

Pointing it out to ensure review does not miss this.

jansel (approved):

PR #314, opened 2022-06-02
[WIP] [inductor] Topologically sort nodes based on the dependency

desertfire:
Summary: The goal is to reduce the number of generated kernels

### DIFF
--- a/torchbench.py
+++ b/torchbench.py
@@ -893,6 +893,7 @@ def main():
     elif args.inductor or args.inductor_dynamic:
         import torchinductor.config
 
+        torchinductor.config.debug = args.verbose
         if args.threads:
             torchinductor.config.cpp.threads = args.threads
 
--- a/torchinductor/scheduler.py
+++ b/torchinductor/scheduler.py
@@ -379,6 +379,7 @@ class Scheduler:
 
         self.compute_users()
         self.dead_node_elimination()
+        self.topo_sort_nodes()
         self.enqueue(self.nodes)
 
     def compute_users(self):
@@ -473,6 +474,39 @@ class Scheduler:
             return
         if node.can_remove_buffer(broadcast_after_reduce=broadcast_after_reduce):
             V.graph.removed_buffers.add(name)
+
+    def topo_sort_nodes(self):
+        free_nodes = set()
+        sorted_nodes = []
+        for node in self.nodes:
+            if not node.unmet_dependencies:
+                free_nodes.add(node.get_name())
+                sorted_nodes.append(node)
+
+        i = 0
+        while i < len(sorted_nodes):
+            for user in sorted_nodes[i].users:
+                node = user.node
+                if not isinstance(node, BaseSchedulerNode):
+                    continue
+                if node.unmet_dependencies:
+                    node.unmet_dependencies = {
+                        dep
+                        for dep in node.unmet_dependencies
+                        if dep.name not in free_nodes
+                    }
+                    if not node.unmet_dependencies:
+                        # All the dependency has been satisfied
+                        free_nodes.add(node.get_name())
+                        sorted_nodes.append(node)
+            i += 1
+
+        # Append any remaining nodes
+        for node in self.nodes:
+            if node.get_name() not in free_nodes:
+                sorted_nodes.append(node)
+
+        self.nodes = sorted_nodes
 
     def enqueue(self, node):
         if isinstance(node, (tuple, list)):


jansel (approved):

torchinductor/dependencies.py
```diff
@@ -70,11 +69,11 @@ def __init__(self, size):
         self._index_exprs = set()
         self._size = tuple([x for x in size if x != 1])
 
-    def load(self, name: str, index: sympy.Expr, upcast: bool = False):
+    def load(self, name: str, index: sympy.Expr, upcast: bool = False, store_cache_key: sympy.Expr = None):
         self._reads.add(MemoryDep(name, index, self._size))
```
jansel, line 73: I'd expect us to need to normalize this, both the index and the size.

jansel (commented):
Another option here would be to make the store_cache not use the index (this should be fine because it is only used for single elements), and rely on dependencies.py to make sure addresses match.

jansel:
@desertfire this is getting super close.  I needed some stuff from here so I cherry picked it into [my branch](https://github.com/jansel/torchdynamo/tree/inductor).  I made a few fixes/improvements:
```patch
diff --git a/torchinductor/dependencies.py b/torchinductor/dependencies.py
index 51d6333..ecfdb57 100644
--- a/torchinductor/dependencies.py
+++ b/torchinductor/dependencies.py
@@ -64,31 +64,32 @@ class ReadWrites:
 
 
 class RecordLoadStore(V.MockHandler):
-    def __init__(self, size, var_ranges):
+    def __init__(self, var_ranges, normalize):
         super(RecordLoadStore, self).__init__()
         self._reads = set()
         self._writes = set()
         self._index_exprs = set()
-        self._size = tuple([x for x in size if x != 1])
         self._var_ranges = var_ranges
+        self._normalize = normalize
 
     def canonicalize(self, index):
         sizes = list(self._var_ranges.values())
-        if len(self._var_ranges) > 1:
-            # Try to further simplify the indexes even if simplify_loops didn't
-            # convert it to the simpliest form because of the interference from
-            # different indexing formulas.
-            index_vars = list(self._var_ranges.keys())
-            new_sizes, reindex, prune = _simplify_loops(index_vars, sizes, [index])
-            if len(new_sizes) < len(sizes):
-                replacement = dict(zip(index_vars, reindex(prune(index_vars))))
-                index = (
-                    sympy.expand(index)
-                    .subs(replacement)
-                    .subs(dict(zip(index_vars, reversed(index_vars))))
-                )
-                sizes = new_sizes
-        return index, tuple([x for x in sizes if x != 1])
+        if not self._normalize:
+            return index, tuple([x for x in sizes if x != 1])
+
+        # Try to further simplify the indexes even if simplify_loops didn't
+        # convert it to the simpliest form because of the interference from
+        # different indexing formulas.
+        index_vars = list(self._var_ranges.keys())
+        new_sizes, reindex, prune = _simplify_loops(index_vars, sizes, [index])
+
+        # assign new variables each dimension to deal with numbering mismatches
+        # d0, d1, d2 could become d0, d2 -- which won't match d0, d1
+        _, add_var = var_builder("c")
+        replacement = dict(zip(index_vars, reindex([add_var(x) for x in new_sizes])))
+
+        index = sympy.expand(index).subs(replacement)
+        return index, tuple(new_sizes)
 
     def load(self, name: str, index: sympy.Expr, upcast: bool = False):
         canonicalized_index, canonicalized_size = self.canonicalize(index)
@@ -104,7 +105,8 @@ class RecordLoadStore(V.MockHandler):
         return self.store(name, index, f"reduce_{reduction_type})({value})")
 
     def index_expr(self, index, dtype):
-        self._index_exprs.add(IndexExprDep(index, self._size))
+        canonicalized_index, canonicalized_size = self.canonicalize(index)
+        self._index_exprs.add(IndexExprDep(canonicalized_index, canonicalized_size))
         return f"index_expr({index}, {dtype})"
 
 
@@ -141,9 +143,9 @@ def index_vars_squeeze(*argsizes, prefix="d"):
     return new_sizes, args, var_ranges
 
 
-def extract_read_writes(fn, *argsizes):
+def extract_read_writes(fn, *argsizes, normalize=False):
     new_sizes, args, var_ranges = index_vars_squeeze(*argsizes)
-    rw = RecordLoadStore(new_sizes[0], var_ranges)
+    rw = RecordLoadStore(var_ranges, normalize=normalize)
     with V.set_ops_handler(rw):
         fn(*args)
     return ReadWrites(rw._reads, rw._writes, rw._index_exprs)
diff --git a/torchinductor/scheduler.py b/torchinductor/scheduler.py
index bb47ed0..5878e87 100644
--- a/torchinductor/scheduler.py
+++ b/torchinductor/scheduler.py
@@ -188,7 +188,9 @@ class SchedulerNode(BaseSchedulerNode):
         ) = node.simplify_reorder_and_tile()
 
         self.group = (node.get_device(), group_fn(self._sizes))
-        self.set_read_writes(dependencies.extract_read_writes(self._body, *self._sizes))
+        self.set_read_writes(
+            dependencies.extract_read_writes(self._body, *self._sizes, normalize=True)
+        )
 
     def can_remove_buffer(self, check_group):
         if (

```

This should fix the test failure.  Use `git apply` to apply the patch.

PR #315, opened 2022-06-02
Fix repro test to unblock internal sync

anijain2305:
Currently `is_allowed` reads the PyTorch imports and also treats packages like `inspect` as torch objects.

For example parsing of this line - https://github.com/pytorch/pytorch/blob/master/torch/ao/quantization/utils.py#L11 results in an allowed object - `torch.ao.quantization.utils.signature`.

Ideally, we should filter out such stuff from `_allowed_ids`. For now, just sending a quick PR to unblock internal sync.

### DIFF
--- a/torchdynamo/allowed_functions.py
+++ b/torchdynamo/allowed_functions.py
@@ -2,6 +2,7 @@ import builtins
 import collections
 import copy
 import functools
+import inspect
 import itertools
 import math
 import operator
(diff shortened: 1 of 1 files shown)


jansel:
Would it work to just add `inspect.signature` to this disallowed function list: 
https://github.com/pytorch/torchdynamo/blob/6154c314b26d2f7f11584f84fc6244375bc39bd0/torchdynamo/allowed_functions.py#L20

jansel (approved):

Issue #292, opened 2022-05-26
dtype info lost during tensor -> numpy -> tensor conversion


yanboliang:
*This is the 7/N issue found from https://github.com/pytorch/torchdynamo/issues/156.*

Minimal code to reproduce:
```
from typing import List
import torch
import torchdynamo

def fn(x):
    x_np = x.numpy()
    weight = []
    z = x_np[0] / 20
    weight.append(z)
    return torch.tensor(weight)

def my_compiler(gm: torch.fx.GraphModule, example_inputs: List[torch.Tensor]):
    return gm.forward

x = torch.ones([1], dtype=torch.int64)
with torchdynamo.optimize(my_compiler):
    print(fn(x).dtype)
```
The output is ```torch.float32```, however, it's ```torch.float64``` on pytorch.

I'm working this issue and found ```x_np[0]``` is python ```float``` rather than ```numpy.float64```, so sth wrong happens before that. Want to check if anyone has encountered similar issue or have any thought? @jansel

jansel:
I tink this issue is coming from here:
https://github.com/pytorch/torchdynamo/blob/77b561823b4a0330cc7db2f68292c16735b6d694/torchdynamo/variables/builder.py#L286

Where we convert it to an int and specialize on it.

A related issue is: https://github.com/pytorch/torchdynamo/issues/70 
Which this example will also suffer from by over-specializing on the value of `x_np[0]`.

PR #321, opened 2022-06-03
Initial implementation of UnspecializedPrimitiveVariable

yanboliang:
Changes:
* Add ```UnspecializedPrimitiveVariable``` which is a 1-element tensor representing unspecialized primitive type.
* Add heuristic to decide when to specialize and when not to:
  * If the variable is numpy int or float, we wrap it as unspecialized.
* More heuristic will be added later if this is the right direction. 

Fix https://github.com/pytorch/torchdynamo/issues/292
Partially contribute to https://github.com/pytorch/torchdynamo/issues/70

### DIFF
--- a/tests/test_misc.py
+++ b/tests/test_misc.py
@@ -651,8 +651,8 @@ class MiscTests(torchdynamo.testing.TestCase):
         with torchdynamo.optimize(cnts):
             self.assertTrue(same(fn(*args), correct))
             self.assertTrue(same(fn(*args), correct))
-        self.assertEqual(cnts.frame_count, 1)
-        self.assertEqual(cnts.op_count, 1)
+        self.assertEqual(cnts.frame_count, 0)
+        self.assertEqual(cnts.op_count, 0)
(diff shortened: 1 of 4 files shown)


anijain2305:
LGTM. would be good to add a couple of tests that check the number of compilations per function when you pass on different integer values on different invocation.

tests/test_misc.py
```diff
@@ -652,7 +652,7 @@ def fn(x, a, b):
             self.assertTrue(same(fn(*args), correct))
             self.assertTrue(same(fn(*args), correct))
         self.assertEqual(cnts.frame_count, 1)
-        self.assertEqual(cnts.op_count, 1)
+        self.assertEqual(cnts.op_count, 2)
```
anijain2305, line 655: Is the extra op because now we have a % b as part of graph?

yanboliang: Correct!

anijain2305 (approved):

Issue #319, opened 2022-06-03
Dynamic module inside dynamo context is not evaluated correctly?


voznesenskym:
Minimum repro:

```
# -*- coding: utf-8 -*-
import torch
import math
import torchdynamo
from typing import List
import torchdynamo.testing
from torch import nn

class Polynomial3(nn.Module):
    def __init__(self):
        super().__init__()

    def forward(self, x):
        return 1 + x

with torchdynamo.optimize("eager"):
    model = Polynomial3()
    x = 10
    for t in range(2000):
        y_pred = model(x)
    print(y_pred)
```

Produces:

```
AttributeError: 'Polynomial3' object has no attribute 'training'
```

From:
```
    def NN_MODULE(self, guard: Guard):
        self.ID_MATCH(guard)
        ref = self.arg_ref(guard)
        val = self.get(guard.name)
        assert istype(val.training, bool)
        if val.training:
            self.code.append(f"{ref}.training")
        else:
            self.code.append(f"not {ref}.training"
```

I believe this is due to incorrect determination as to what is, and is not a dynamic module.

Hardcoding `def is_dynamic_nn_module(obj):` to return `True` gets me past the bug. This is not a real fix, but is indicative of us not tracking this module as dynamic correctly.

voznesenskym:
cc: @mlazos (who reprod, huge ty) @anijain2305 @jansel

anijain2305:
Ok, I can repro. I think this might be related to recent PR - https://github.com/pytorch/torchdynamo/pull/305. Removing the `tagging_init` passes the test.

All `torch.nn.Module` instances have a `training` member. I think monkey patching `__init__` might be upsetting it.

voznesenskym:
It's actually not monkey patching init. It's more subtle/ annoying, but also easier to hack fix :P

So, the order of critical events is:

1) install_generation_tagging_init
2) VariableBuilder's __call__
3) _convert_frame_assert
4) NN_MODULE guard creation (which calls assert istype(val.training, bool))

Now... what's missing? We have not yet actually called init on the object! We cannot read properties off something that is not instantiated.

Adding:

```
    def __new__(cls):
        super(Polynomial3, cls).__new__(cls)
```

To the model above fixes it. This is not a real fix, but it helps us understand the root cause and how to get it around it. I am going to look into make this work correctly for inits.

voznesenskym:
[...]

PR #322, opened 2022-06-03
Prevent guard creation from accessing objects between __new__ and __init__

voznesenskym:
Discussion on #319

### DIFF
--- a/torchdynamo/guards.py
+++ b/torchdynamo/guards.py
@@ -242,11 +242,12 @@ class GuardBuilder:
         self.ID_MATCH(guard)
         ref = self.arg_ref(guard)
         val = self.get(guard.name)
-        assert istype(val.training, bool)
-        if val.training:
-            self.code.append(f"{ref}.training")
-        else:
-            self.code.append(f"not {ref}.training")
+        if not val._awaiting_init:
+            assert istype(val.training, bool)
+            if val.training:
+                self.code.append(f"{ref}.training")
+            else:
+                self.code.append(f"not {ref}.training")
 
     def FUNCTION_MATCH(self, guard: Guard):
         """things like torch.add and user defined functions"""
@@ -340,6 +341,7 @@ class GuardedCode:
         args += [a for a in local_builder.argnames if a != "___implicit0"]
         args += ["**___kwargs_ignored"]
         args = ",".join(args)
+
         code_parts = (
             ["___guarded_code.valid"] + local_builder.code + global_builder.code
         )
--- a/torchdynamo/guards.py
+++ b/torchdynamo/guards.py
@@ -250,16 +250,11 @@ class GuardBuilder:
             assert istype(val.training, bool)
             self.code.append(f"___training_state({ref}, {val.training})")
 
-        if hasattr(val, "_awaiting_init"):
+        if hasattr(val, "training"):
             # There are cases where a monkeypatched object has a guard made between __new__ and __init__
-            if not val._awaiting_init:
-                setup_guard()
-            else:
-                unimplemented(f"Guard setup for uninitialized class {type(val)}")
-
-        else:
-            # Object was not monkeypatched on __new__ and __init__ by us
             setup_guard()
+        else:
+            unimplemented(f"Guard setup for uninitialized class {type(val)}")
 
     def FUNCTION_MATCH(self, guard: Guard):
         """things like torch.add and user defined functions"""
--- a/torchdynamo/guards.py
+++ b/torchdynamo/guards.py
@@ -25,8 +25,9 @@ from ._guards import check_type_id
 from .eval_frame import set_guard_error_hook
 from .eval_frame import set_guard_fail_hook
 from .exc import unimplemented
-from .utils import ExactWeakKeyDictionary
+from .utils import guard_failures
 from .utils import istype
+from .utils import orig_code_map
 from .utils import rename_implicit
 from .utils import training_state
 from .utils import tuple_iterator_getitem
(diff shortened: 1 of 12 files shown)


torchdynamo/guards.py
```diff
@@ -242,11 +242,21 @@ def NN_MODULE(self, guard: Guard):
...
+        # There are cases where a patched object has a guard made between __new__ and __init__
+        def setup_guard():
+            assert istype(val.training, bool)
+            if val.training:
+                self.code.append(f"{ref}.training")
+            else:
+                self.code.append(f"not {ref}.training")
+
+        if hasattr(val, "_awaiting_init"):
+            if not val._awaiting_init:
```
voznesenskym, line 255: Another thing we can do is setup a guard that always fails here, if we want to be safe.

voznesenskym: Did a little more debugging - we do not need to, __init__ makes a new object and so ID_MATCH already takes care of that for us.

jansel (changes requested):
This might be cleaner if we introduce a helper function that is called from the generated guards.  You can add helper function to the guard namespace here: https://github.com/jansel/torchdynamo/blob/fc71a467ff7aa59c0c2a7651a08a005002246930/torchdynamo/guards.py#L41

If a `nn.Module` is in a partially initialized state, I don't think we should be trying to specialize on it since we expect it to change and the guard to fail.  Perhaps we just want to introduce a graph break in these cases.

torchdynamo/variables/misc.py
```diff
@@ -61,7 +61,7 @@ def call_method(
             self, args, kwargs.values(), self.objvar, self.typevar
         )
         inner_fn = self.const_getattr(self, name)
-        if inner_fn is object.__init__:
+        if name == "__init__":
```
voznesenskym, line 64: this was breaking on patched init. The name is __init__ but the fn is not __init__. The name check here is safer, as using the const_getattr and then checking for object.__init__ (correctly) does not respect monkeypatching. See the added test in test_modules.py wherein we create a dynamic module from within an optimize ctxt.

torchdynamo/mutation_guard.py
```diff
@@ -83,14 +83,23 @@ def install_generation_tagging_init():
     Monkey patch torch.nn.Module.__init__ and torch.nn.Module.__setstate__
     so we can detect nn.Module instances created dynamically inside forward methods.
     """
-
     if getattr(Module, "___needs_generation_tag_patch", True):
         init = Module.__init__
 
+        def patched_new(cls, *args, **kwargs):
```
anijain2305, line 89: Patching new takes us back to the original problem of #305. Patching `__new__`  makes it slower.

IIUC, If we patch `__new__`, we don't need to patch `__init__`.

voznesenskym: Good question. 

I ran this on the same bench as https://github.com/pytorch/torchdynamo/pull/305 (same script, 10 runs, averaged)

Without patched new: 98.88617604838024
With patched new: 98.93933895271553
And with neither: 98.00966040366337

So I think this is within an acceptable margin of slowdown, thoughts?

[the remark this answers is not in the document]
voznesenskym: Yea, we could do that. Let me give that a whirl after I bench.

voznesenskym: Thanks, with eval batch down to 16, I get:

Without patched new: 97.34619600294747 
With patched new: 97.60542088329098
And with neither: 96.98565560222735 

Thoughts?

mlazos: There was a PR which increased the batch size of AlexNet which hides the issue. I reverted this PR when doing my benchmarking. https://github.com/pytorch/benchmark/pull/919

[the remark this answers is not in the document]
voznesenskym: Well, before I do that, how do we distinguish this from modules that did not go through our new? those modules would not have `__init_called` set on them. LMK if I am being silly or missing something.

mlazos: for my sanity can you try benchmarking and moving everything into `__new__` to verify that the slowdown still occurs?

torchdynamo/guards.py
```diff
@@ -242,11 +245,21 @@ def NN_MODULE(self, guard: Guard):
...
         val = self.get(guard.name)
-        assert istype(val.training, bool)
-        if val.training:
-            self.code.append(f"{ref}.training")
+
+        def setup_guard():
+            assert istype(val.training, bool)
+            self.code.append(f"___training_state({ref}, {val.training})")
+
+        if hasattr(val, "_awaiting_init"):
```
jansel, line 253: Would changing this to `hasattr(val, "training")` allow us to remove the `_awaiting_init` patching?

Patching `__new__` results in some noticeable slowdowns.

torchdynamo/guards.py
```diff
@@ -238,11 +241,16 @@ def NN_MODULE(self, guard: Guard):
         self.ID_MATCH(guard)
         ref = self.arg_ref(guard)
         val = self.get(guard.name)
-        assert istype(val.training, bool)
-        if val.training:
-            self.code.append(f"{ref}.training")
+
+        def setup_guard():
+            assert istype(val.training, bool)
+            self.code.append(f"___training_state({ref}, {val.training})")
```
anijain2305, line 247: Is it faster to just directly check the equality here, instead of a function jump to `__training_state` at runtime?

anijain2305 (approved):
LGTM.

Just added one comment about creating a new function vs directly doing a equality check performance.

torchdynamo/guards.py
jansel: I suggested the helper before to allow either hasattr or try/except in the body of it.  Since we aren't doing that I think it is fine to just inline the check.

anijain2305 (approved):

anijain2305:
@jansel Can you PTAL again? This is causing many PyTorch tests to fail.

jansel (approved):

PR #324, opened 2022-06-03
Filter out unimportant modules from allowed modules

anijain2305:
With this a large number of PyTorch tests in `test/distributed/rpc/test_tensorpipe_agent.py` start passing. 

One side effect was filtering `typing` module as well, and this caused unnecessary graph breaks because they are used only for annotating the function args. So, I added a TypingVariable to deal with it.

### DIFF
--- a/torchdynamo/allowed_functions.py
+++ b/torchdynamo/allowed_functions.py
@@ -118,6 +118,8 @@ def _allowed_function_ids():
                     or inspect.getmodule(obj).__name__.startswith("math")
                 ):
                     torch_object_ids[id(obj)] = f"{module.__name__}.{name}"
+                elif inspect.getmodule(obj) is None:
+                    torch_object_ids[id(obj)] = f"{module.__name__}.{name}"
 
     _find_torch_objects(torch)
     _find_torch_objects(math)
--- a/torchdynamo/allowed_functions.py
+++ b/torchdynamo/allowed_functions.py
@@ -16,6 +16,7 @@ import numpy
 import torch
 
 from . import config
+from .utils import is_safe_constant
 
 
 def make_function_id_set(lazy_initializer):
@@ -103,10 +104,14 @@ def _allowed_function_ids():
     torch_object_ids = dict()
 
     def _is_allowed_module_prefix(obj):
-        allowed_module_name_prefix_list = ("torch", "math")
-        return inspect.getmodule(obj) and inspect.getmodule(obj).__name__.startswith(
-            allowed_module_name_prefix_list
-        )
+        allowed_modules = ("torch", "math")
+        allowed_modules_dot = tuple([x + "." for x in allowed_modules])
+        module = inspect.getmodule(obj)
+        if module is None:
+            return False
+
+        mod_name = module.__name__
+        return mod_name in allowed_modules or mod_name.startswith(allowed_modules_dot)
 
     def _find_torch_objects(module):
         if any(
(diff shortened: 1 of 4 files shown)


torchdynamo/allowed_functions.py
```diff
@@ -113,7 +113,10 @@ def _find_torch_objects(module):
                     if obj.__name__.startswith("torch."):
                         torch_object_ids[id(obj)] = f"{module.__name__}.{name}"
                         _find_torch_objects(obj)
-                else:
+                elif inspect.getmodule(obj) and (
+                    inspect.getmodule(obj).__name__.startswith("torch")
+                    or inspect.getmodule(obj).__name__.startswith("math")
+                ):
```
voznesenskym, line 119: `.startswith` supports a tuple iirc, so we can do:
```
allowed_module_name_prefix_list = ["torch", "math"]
```
...
```
inspect.getmodule(obj).__name__.startswith(tuple(allowed_module_name_prefix_list))
```

torchdynamo/variables/builder.py
```diff
@@ -226,6 +228,11 @@ def _wrap(self, value):
                 value,
                 guards=make_guards(GuardBuilder.FUNCTION_MATCH),
             )
+        elif value is List:
+            return TypingVariable(
+                value,
+                guards=make_guards(GuardBuilder.FUNCTION_MATCH),
+            )
```
voznesenskym, line 235: 2 questions:

1) Should we follow the isintance pattern used in the other checks?
2) FUNCTION_MATCH is just:

```
if guard.is_local():
            return self.ID_MATCH(guard)
```
If you care about `guard.is_local()`, that's fine, but if not, we should just use `ID_MATCH`.

Aside: If we use `FUNCTION_MATCH` everywhere to actually mean `is_local` + `ID_MATCH`, maybe we should just rename it to `LOCAL_ID_MATCH`?

torchdynamo/variables/misc.py
```diff
@@ -556,6 +556,26 @@ def call_function(
...
+
+    def call_method(
+        self,
+        tx,
+        name,
+        args: "List[VariableTracker]",
+        kwargs: "Dict[str, VariableTracker]",
+    ) -> "VariableTracker":
+        if name == "__getitem__":
+            return variables.ConstantVariable(
```
voznesenskym, line 572: should we assert args len is 1 here?

torchdynamo/allowed_functions.py
anijain2305: Nice!

torchdynamo/variables/builder.py
anijain2305: `ID_MATCH` sounded better for this one. `typing.List` is not really a function. Thanks for pointing it out.

torchdynamo/allowed_functions.py
```diff
@@ -100,6 +102,12 @@ def _allowed_function_ids():
     torch.distributions.Distribution.set_default_validate_args(False)
     torch_object_ids = dict()
 
+    def _is_allowed_module_prefix(obj):
```
jansel, line 105: I think this will match stuff like `torchvision` or `mathfoo` which could lead to false positives.

anijain2305: Thanks.

It was earlier missing objects like `torch.set_grad_enabled`. So, I kept `torch` instead of `torch.`.

I updated the check to be more restrictive but still allow `torch` objects.

jansel (approved):

PR #325, opened 2022-06-04
Make TorchInductor use Triton's MM implementation in codegen

vesuppi:
The motivation is that torchinductor currently generates a call to `aten.mm.out` for `torch.mm` operations, and we'd like to use Triton's MM implementation in torchinductor's code generation, which makes fusion possible and guarantee the performance of MM in the same time. Our first step is to just make torchinductor generate a call to Triton's mm implementation for `torch.mm`, and make sure this can work. 

The modifications are
- Added a directory called `triton_ops/` in `torchinductor`
- Added Triton's MM implementation in `triton_ops/` and made an "out" version
- Added an option `use_triton_mm` in `torchinductor/config.py`, which defaults to `False`
- Some minor changes to `ir.py` and `wrapper.py` to use triton's mm (out version) function

A simple example to use is like the following:

```python
import torch
import torchdynamo
import torchinductor.config as inductor_config
inductor_config.debug = True
inductor_config.use_triton_mm = True

def toy_example(a, b):
    return torch.mm(a, b)

N = 1024
a = torch.rand([N, N], device='cuda', dtype=torch.float16)
b = torch.rand([N, N], device='cuda', dtype=torch.float16)
c1 = torch.mm(a, b)

with torchdynamo.optimize("inductor"):
    for _ in range(100):
        c = toy_example(a, b)

assert(torch.allclose(c1, c))
```

which generates the following code:

```python
from ctypes import c_void_p, c_long
import torch
from torch import empty_strided, as_strided
from torchinductor.codecache import CppCodeCache, TritonCodeCache, grid
try:
    import triton
    import triton.language as tl
except ImportError:
    pass
aten = torch.ops.aten
from torchinductor.triton_ops.matmul import matmul_out as triton_mm_out


def call(arg0, arg1):
    arg0_size = arg0.size()
    s0 = arg0_size[0]
    buf0 = empty_strided((s0, s0), (s0, 1), device='cuda', dtype=torch.float16)
    triton_mm_out(arg0, arg1, out=buf0)
    return (buf0, )

if __name__ == "__main__":
    from torchinductor.microbench import print_performance
    arg0 = rand_strided((1024, 1024), (1024, 1), device='cuda', dtype=torch.float16)
    arg1 = rand_strided((1024, 1024), (1024, 1), device='cuda', dtype=torch.float16)
    print_performance(lambda: call(arg0, arg1))

```

### DIFF
--- a/tests/test_torchinductor.py
+++ b/tests/test_torchinductor.py
@@ -1146,6 +1146,19 @@ class CommonTemplate:
         y_correct = torch.conv2d(x, w, bias, stride, padding, dilation, groups)
         self.assertTrue(same(y, y_correct, cos_similarity=True, tol=0.1))
 
+    @patch.object(config.triton, "use_mm", True)
+    def test_triton_mm2(self):
+        @torchdynamo.optimize("inductor", nopython=True)
+        def fn(x, y):
+            return torch.mm(x, y)
+
+        N = 1024
+        a = torch.randn([N, N], device=self.device, dtype=torch.float32)
+        b = torch.randn([N, N], device=self.device, dtype=torch.float32)
+        c1 = torch.mm(a, b)
+        c = fn(a, b)
+        assert torch.allclose(c1, c, atol=1e-3, rtol=1e-3)
+
     def test_std(self):
         def fn(x):
             return (
--- a/torchinductor/codegen/wrapper.py
+++ b/torchinductor/codegen/wrapper.py
@@ -73,6 +73,16 @@ class WrapperCodeGen(CodeGen):
                 aten = torch.ops.aten
             """
         )
+
+        if config.triton.use_mm:
+            self.header.splice(
+                """
+                try:
+                    from torchinductor.triton_ops.matmul import matmul_out as triton_mm_out
+                except ImportError:
+                    pass
+                """
+            )
         self.prefix.writelines(
             ["", "", f"def call({', '.join(V.graph.graph_inputs.keys())}):"]
         )
--- a/torchinductor/config.py
+++ b/torchinductor/config.py
@@ -61,3 +61,7 @@ class triton:
 
     # use triton.autotune?
     autotune = True
+
+    # enable codegen to use Triton's mm
+    use_mm = False
+
--- a/torchinductor/config.py
+++ b/torchinductor/config.py
@@ -64,4 +64,3 @@ class triton:
 
     # enable codegen to use Triton's mm
     use_mm = False
-
--- a/torchinductor/ir.py
+++ b/torchinductor/ir.py
@@ -1653,6 +1653,15 @@ class ExternKernelAlloc(ExternKernel):
 class MatrixMultiply(ExternKernelOut):
     kernel = "aten.mm.out"
 
+    def __init__(self, layout, inputs, constant_args=(), output_view=None):
+        super().__init__(layout, inputs, constant_args, output_view)
+        if (
+            config.triton.use_mm
+            and len(inputs) > 0
+            and inputs[0].get_device().type == "cuda"
+        ):
+            self.kernel = "triton_mm_out"
+
     @classmethod
     def create(cls, a, b):
         *m, k1 = a.get_size()
--- a/torchinductor/triton_ops/matmul.py
+++ b/torchinductor/triton_ops/matmul.py
@@ -1,0 +1,228 @@
+import torch
+import triton
+import triton.language as tl
+from triton.ops.matmul import get_configs_io_bound
+from triton.ops.matmul_perf_model import early_config_prune
+from triton.ops.matmul_perf_model import estimate_matmul_time
+
+
+@triton.heuristics(
+    {
+        "EVEN_K": lambda args: args["K"] % (args["BLOCK_K"] * args["SPLIT_K"]) == 0,
+    }
+)
+@triton.autotune(
+    configs=[
+        # basic configs for compute-bound matmuls
+        triton.Config(
+            {"BLOCK_M": 128, "BLOCK_N": 256, "BLOCK_K": 32, "SPLIT_K": 1},
+            num_stages=3,
+            num_warps=8,
+        ),
+        triton.Config(
+            {"BLOCK_M": 256, "BLOCK_N": 128, "BLOCK_K": 32, "SPLIT_K": 1},
+            num_stages=3,
+            num_warps=8,
+        ),
+        triton.Config(
+            {"BLOCK_M": 256, "BLOCK_N": 64, "BLOCK_K": 32, "SPLIT_K": 1},
+            num_stages=4,
+            num_warps=4,
+        ),
+        triton.Config(
+            {"BLOCK_M": 64, "BLOCK_N": 256, "BLOCK_K": 32, "SPLIT_K": 1},
+            num_stages=4,
+            num_warps=4,
+        ),
+        triton.Config(
+            {"BLOCK_M": 128, "BLOCK_N": 128, "BLOCK_K": 32, "SPLIT_K": 1},
+            num_stages=4,
+            num_warps=4,
+        ),
+        triton.Config(
+            {"BLOCK_M": 128, "BLOCK_N": 64, "BLOCK_K": 32, "SPLIT_K": 1},
+            num_stages=4,
+            num_warps=4,
+        ),
+        triton.Config(
+            {"BLOCK_M": 64, "BLOCK_N": 128, "BLOCK_K": 32, "SPLIT_K": 1},
+            num_stages=4,
+            num_warps=4,
+        ),
+        triton.Config(
+            {"BLOCK_M": 128, "BLOCK_N": 32, "BLOCK_K": 32, "SPLIT_K": 1},
+            num_stages=4,
+            num_warps=4,
+        ),
+        triton.Config(
+            {"BLOCK_M": 64, "BLOCK_N": 32, "BLOCK_K": 32, "SPLIT_K": 1},
+            num_stages=5,
+            num_warps=2,
+        ),
+        # good for int8
+        triton.Config(
+            {"BLOCK_M": 128, "BLOCK_N": 256, "BLOCK_K": 128, "SPLIT_K": 1},
+            num_stages=3,
+            num_warps=8,
+        ),
+        triton.Config(
+            {"BLOCK_M": 256, "BLOCK_N": 128, "BLOCK_K": 128, "SPLIT_K": 1},
+            num_stages=3,
+            num_warps=8,
+        ),
+        triton.Config(
+            {"BLOCK_M": 256, "BLOCK_N": 64, "BLOCK_K": 128, "SPLIT_K": 1},
+            num_stages=4,
+            num_warps=4,
+        ),
+        triton.Config(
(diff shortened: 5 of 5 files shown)


torchinductor/codegen/wrapper.py
```diff
@@ -56,19 +56,22 @@ def __init__(self):
...
+            from {codecache.__name__} import CppCodeCache, TritonCodeCache, grid
+            try:
+                import triton
+                import triton.language as tl
+            except ImportError:
+                pass
+            aten = torch.ops.aten
             """
+        if config.use_triton_mm:
+            import_header += 'from torchinductor.triton_ops.matmul import matmul_out as triton_mm_out\n'
```
jansel, line 72: `self.header.writeline()` might be cleaner

torchinductor/config.py
```diff
@@ -42,3 +42,7 @@ class cpp:
 class triton:
     cudagraphs = True
     hackery = False
+
+
+# enable codegen to use Triton's mm
+use_triton_mm = False
```
jansel, line 48: put this in the triton namespace

torchinductor/ir.py
```diff
@@ -1515,7 +1515,9 @@ def should_allocate(self):
 
 
 class MatrixMultiply(ExternKernelOut):
-    kernel = "aten.mm.out"
+    kernel = "aten.mm.out"    
+    if config.use_triton_mm:
```
jansel, line 1519: 1) we should check this dynamically (this current version runs at import time)
2) check for the device.type=="cuda" to filter out CPU use cases

torchinductor/triton_ops/matmul.py
```diff
@@ -0,0 +1,175 @@
...
+import triton
+import triton.language as tl
+from .matmul_perf_model import early_config_prune, estimate_matmul_time
+
+
+def init_to_zero(name):
+    return lambda nargs: nargs[name].zero_()
+
+
+def get_configs_io_bound():
```
jansel, line 12: could we just import this from triton.ops?

torchinductor/triton_ops/matmul.py
```diff
@@ -0,0 +1,175 @@
...
+    C = C + (rm[:, None] * stride_cm + rn[None, :] * stride_cn)
+    mask = (rm < M)[:, None] & (rn < N)[None, :]
+    # handles write-back with reduction-splitting
+    if SPLIT_K == 1:
+        tl.store(C, acc, mask=mask)
+    else:
+        tl.atomic_add(C, acc, mask=mask)
+
+
+class _matmul(torch.autograd.Function):
```
jansel, line 115: Is this used?

torchinductor/triton_ops/matmul_perf_model.py
```diff
@@ -0,0 +1,157 @@
+import heapq
```
jansel, line 1: Did we make a change here?  Perhaps just import from triton.

jansel (changes requested):

jansel:
@vesuppi for the Facebook CLA thing, you need to link your github account to your Meta intern profile.  Once you do that ping me and I can give you access.

anijain2305:
@vesuppi Its looking good, we are getting there.

Can you also add a test where we compare `torch.matmul` eager (just plain PyTorch) vs Inductor-Triton (just accuracy)? It is ok to have high tolerance. The goal here is to have a test that guards future PRs from doing something silly and break this feature.

You might need a ContextManager (https://www.geeksforgeeks.org/context-manager-in-python/) to turn on the `use_mm` flag only for your test and does not affect other tests.

torchinductor/codegen/wrapper.py
```diff
@@ -70,6 +70,10 @@ def __init__(self):
                 aten = torch.ops.aten
             """
         )
+        if config.triton.use_mm:
+            self.header.writeline(
+                "from torchinductor.triton_ops.matmul import matmul_out as triton_mm_out"
```
pyjhzwh, line 75: Not sure if my comments or not. 
But is it possible to use 
"import triton
import triton.ops.matmul as triton_mm_out"
to call the triton matmul directly rather than putting another copy of triton.ops.matmul in torchinductor directory. Or does the import module has to be inside torchinductor?

vesuppi: Hi Yunjie, yes directly importing `triton.ops.matmul` would work too. The thing here is that we created a `out` variant of the original triton mm implementation, and we may need to add some additional arguments to do the template approach. Basically we'd like to do some customization of the original implementation.

pyjhzwh: That makes sense. Thanks for the explanation.

torchinductor/codegen/wrapper.py
```diff
@@ -70,6 +70,10 @@ def __init__(self):
                 aten = torch.ops.aten
             """
         )
+        if config.triton.use_mm:
```
pyjhzwh, line 73: Would it be better if override codegen of MatrixMultiplication class
```
if self.kernel == "aten.mm.out":
   wrapper_writeline("...")
elif self.kernel == "triton_mm_out":
   wrapper_writeline("...")
```
Because if we find inputs type is not cuda, we won't import triton ops in this. case.

vesuppi:
> @vesuppi Its looking good, we are getting there.
> 
> Can you also add a test where we compare `torch.matmul` eager (just plain PyTorch) vs Inductor-Triton (just accuracy)? It is ok to have high tolerance. The goal here is to have a test that guards future PRs from doing something silly and break this feature.
> 
> You might need a ContextManager (https://www.geeksforgeeks.org/context-manager-in-python/) to turn on the `use_mm` flag only for your test and does not affect other tests.

Added tests to `test_torchinductor.py`, thanks!

torchinductor/codegen/wrapper.py
vesuppi: That makes sense! I think that can work too.

vesuppi: Actually I just took a look, and this class WrapperCodeGen doesn't have the concept (field) of kernel, I think it's just about headers.

tests/test_torchinductor.py
```diff
@@ -1102,6 +1102,36 @@ def fn(x, y):
...
+    def test_triton_mm1(self):
+        config.triton.use_mm = True
+
+        @torchdynamo.optimize("inductor", nopython=True)
+        def fn(x, y):
+            return torch.mm(x, y)
+
+        N = 1024
+        a = torch.randn([N, N], device="cuda", dtype=torch.float16)
+        b = torch.randn([N, N], device="cuda", dtype=torch.float16)
```
anijain2305, line 1114: Test failures might be because you are setting device to be always "cuda". I think other tests use `self.device`.

tests/test_torchinductor.py
```diff
@@ -1102,6 +1102,36 @@ def fn(x, y):
...
+
+    def test_triton_mm2(self):
+        config.triton.use_mm = True
+
+        @torchdynamo.optimize("inductor", nopython=True)
+        def fn(x, y):
+            return torch.mm(x, y)
+
+        N = 1024
+        a = torch.randn([N, N], device="cuda", dtype=torch.float32)
```
anijain2305, line 1128: Same here.

vesuppi:
Removed the float16 test due to the following error when `device` is set `self.device`:

```
____________________________________________ CpuTests.test_triton_mm1_cpu ____________________________________________
                                                                                                                      self = <tests.test_torchinductor.CpuTests testMethod=test_triton_mm1_cpu>
                                                                     
    @patch.object(config.triton, "use_mm", True)
    def test_triton_mm1(self):                                                                                       
        @torchdynamo.optimize("inductor", nopython=True)
        def fn(x, y):                                                                                                
            return torch.mm(x, y)                                                                                    
                                                                                                                     
        N = 1024                                                                                                     
        a = torch.randn([N, N], device=self.device, dtype=torch.float16)
        b = torch.randn([N, N], device=self.device, dtype=torch.float16)                                              
>       c1 = torch.mm(a, b)                                                       
E       RuntimeError: "addmm_impl_cpu_" not implemented for 'Half'                                                                                                                                                                          tests/test_torchinductor.py:1114: RuntimeError
```

vesuppi:
`pytest tests/test_torchinductor.py` failing for not being able to import triton in the added matmul file while the same test ran fine on the cluster machine.

```python
----------------------------- Captured stderr call -----------------------------
----------------------------------------
TORCHDYNAMO: backend compiler failed
  File "/home/runner/work/torchdynamo/torchdynamo/torchdynamo/output_graph.py", line 321, in call_user_compiler
    compiled_fn = self.compiler_fn(gm, self.example_inputs())
  File "/home/runner/work/torchdynamo/torchdynamo/torchinductor/compile_fx.py", line 123, in compile_fx
    compiled_fn = wrap(graph.compile_to_fn())
  File "/home/runner/work/torchdynamo/torchdynamo/torchinductor/graph.py", line 235, in compile_to_fn
    mod = PyCodeCache.load(code)
  File "/home/runner/work/torchdynamo/torchdynamo/torchinductor/codecache.py", line 122, in load
    exec(code, mod.__dict__, mod.__dict__)
  File "/tmp/torchinductor_runner/yl/cylopd5pi5kb3rhg72l7ueui3urmcjg2tyme66vb6ftpvkjs3pyx.py", line 12, in <module>
    from torchinductor.triton_ops.matmul import matmul_out as triton_mm_out
  File "/home/runner/work/torchdynamo/torchdynamo/torchinductor/triton_ops/matmul.py", line 2, in <module>
    import triton
ModuleNotFoundError: No module named 'triton'
```

pyjhzwh:
> `pytest tests/test_torchinductor.py` failing for not being able to import triton in the added matmul file while the same test ran fine on the cluster machine.
> 
> ```python
> ----------------------------- Captured stderr call -----------------------------
> ----------------------------------------
> TORCHDYNAMO: backend compiler failed
> Traceback (most recent call last):
>   File "/home/runner/work/torchdynamo/torchdynamo/torchdynamo/output_graph.py", line 321, in call_user_compiler
>     compiled_fn = self.compiler_fn(gm, self.example_inputs())
>   File "/home/runner/work/torchdynamo/torchdynamo/torchinductor/compile_fx.py", line 123, in compile_fx
>     compiled_fn = wrap(graph.compile_to_fn())
>   File "/home/runner/work/torchdynamo/torchdynamo/torchinductor/graph.py", line 235, in compile_to_fn
>     mod = PyCodeCache.load(code)
>   File "/home/runner/work/torchdynamo/torchdynamo/torchinductor/codecache.py", line 122, in load
>     exec(code, mod.__dict__, mod.__dict__)
>   File "/tmp/torchinductor_runner/yl/cylopd5pi5kb3rhg72l7ueui3urmcjg2tyme66vb6ftpvkjs3pyx.py", line 12, in <module>
>     from torchinductor.triton_ops.matmul import matmul_out as triton_mm_out
>   File "/home/runner/work/torchdynamo/torchdynamo/torchinductor/triton_ops/matmul.py", line 2, in <module>
>     import triton
> ModuleNotFoundError: No module named 'triton'
> ```

Maybe you need 
```
@unittest.skipIf(not torch.cuda.is_available(), "requires cuda")
```

vesuppi:
Revised to be more robust.

anijain2305 (approved):

anijain2305:
@jansel Can you please do another round of review?

jansel (approved):

PR #326, opened 2022-06-05
[RFC] Enable guard_nn_modules - Observe module changes to drive invalidation instead of id_match + training guards

voznesenskym:
Title says a lot of it, but this is a prototype / request for comments for a proposal wherein we replace nn_module guards with invalidation on module changes. 

This allows us to enable guard_nn_modules, as the cost is low enough now.

Pre PR:
1.15x without GUARD_NN_MODULES
0.89x with GUARD_NN_MODULES

Post PR:
1.15x without GUARD_NN_MODULES
1.076x with GUARD_NN_MODULES


jansel (changes requested):
1) Why are we still seeing so much overhead from GUARD_NN_MODULES?  I'd expect with an approach like this we should be able to drive it down close to zero.  Perhaps we also need to be checking for submodules being modified?  Can you look at the generated guard code and see where the overhead is coming from?

2) When are hooks removed?  I am worried we will just accumulate more and more hooks and slow things down.  If you aren't careful it is possible the same code could get invalidated O(n) times (resulting in O(n^2) overheads).

voznesenskym:
> 1. Why are we still seeing so much overhead from GUARD_NN_MODULES?  I'd expect with an approach like this we should be able to drive it down close to zero.  Perhaps we also need to be checking for submodules being modified?  Can you look at the generated guard code and see where the overhead is coming from?

Good question - I was a little surprised as well. I plan to look into this further. Interestingly, if we replace the NN_MODULE guard with:

```
    def NN_MODULE(self, guard: Guard):
        return 

```

We get: 1.076x - so this specific codepath where we do observation is not the source of the overhead. Looking at this, it appears like most of the guards left on nn_modules come from submodules -> params:

(Truncated for sanity)
```
GUARDS ___guarded_code.valid and ___check_obj_id(self.embedding.position.pe, 140696704147024) and ___check_obj_id(self.embedding.token.sparse, 94583822895360) and ___check_obj_id(self.embedding.token.weight, 140697039664448) and ___check_obj_id(self.embedding.segment.sparse, 94583822895360) and ___check_obj_id(self.embedding.segment.weight, 140685152492176) and ___check_obj_id(self.embedding.token.max_norm, 94583823007168) and ___check_type_id(self.embedding.token.norm_type, 94583823023552) and self.embedding.token.norm_type == 2.0 and ___check_obj_id(self.embedding.segment.max_norm, 94583823007168) and ___check_type_id(self.embedding.segment.norm_type, 94583823023552) and self.embedding.segment.norm_type == 2.0 and ___check_type_id(self.embedding.token.padding_idx, 94583823018848) and self.embedding.token.padding_idx == 0 and ___check_type_id(self.embedding.segment.padding_idx, 94583823018848) and self.embedding.segment.padding_idx == 0 and ___check_obj_id(self.embedding.token.scale_grad_by_freq, 94583822895360) and ___check_type_id(self.transformer_blocks[0].lambda_module, 94583938533168) and ___check_type_id(self.transformer_blocks[1].lambda_module, 94583938533168) and ___check_type_id(self.transformer_blocks[2].lambda_module, 94583938533168) and ___check_type_id(self.transformer_blocks[3].lambda_module, 94583938533168) and ___check_type_id(self.transformer_blocks[4].lambda_module, 94583938533168) and ___check_type_id(self.transformer_blocks[5].lambda_module, 94583938533168) and ___check_type_id(self.transformer_blocks[6].lambda_module, 94583938533168) and ___check_type_id(self.transformer_blocks[7].lambda_module, 94583938533168) and ___check_type_id(self.transformer_blocks[8].lambda_module, 94583938533168) and ___check_type_id(self.transformer_blocks[9].lambda_module, 94583938533168) ...
```

> 2. When are hooks removed?  I am worried we will just accumulate more and more hooks and slow things down.  If you aren't careful it is possible the same code could get invalidated O(n) times (resulting in O(n^2) overheads).

Never :P That's not in yet. I had a few thoughts, let me know which you prefer:

1) Only support a single change hook at a time for now
2) Dump all hooks on model change
3) Accumulate and store until we go out of a dynamo context, and then clear them all (We may want to do this anyway after, even if we do 1/2 above, if only so that the code stays *exactly* the same as before we did an optimization pass)

voznesenskym:
A little more detail: The extra weight is from:

1) self._modules going through dict (keys) match 
2) equals matching on `elif not istype(val, torch.Size):` 

1 - we can try to drop - I think this might be covered by the current invalidation listener added here
2 - Need to look deeper here, not sure if we can drop this safely (I could see cases where resizing some types internally does not get covered by the current invalidation listener)

voznesenskym:
@jansel 1.122x on latest PR ;) getting there.

voznesenskym:
Still not ready for review - prints everywhere, missing logic, etc

voznesenskym:
gentle nag @jansel

torchdynamo/variables/misc.py
```diff
@@ -405,6 +405,20 @@ def call_method(
         )
 
 
+class NoOpVariable(VariableTracker):
```
jansel, line 408: Where is this used?

jansel (changes requested):

torchdynamo/convert_frame.py
```diff
@@ -295,6 +295,13 @@ def debug_print(prefix):
                 print()
             assert output.guards is not None
             CleanupManager.instance[code] = output.cleanups
+
+            if "self" in frame.f_locals:
+                self_name = frame.f_locals["self"].__class__.__name__
+                if self_name == "ModuleList":
```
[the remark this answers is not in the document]
voznesenskym: Good point, let me try that and see.

torchdynamo/guards.py
```diff
@@ -34,6 +39,8 @@
 
 log = logging.getLogger(__name__)
 
+# map from modules to guarded code
+module_code_map = ExactWeakKeyDictionary()  # {nn.Module: GuardedCode}
```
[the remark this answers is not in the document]
voznesenskym: Good catch.

torchdynamo/guards.py
```diff
@@ -48,6 +55,72 @@
...
 
+class NNModuleChangeTrackerUtil:
+    @staticmethod
+    def setup(module, guarded_code):
+        modulecls = module.__class__
+        if getattr(modulecls, "__dynamo_module_patch", True):
+            modulecls.__dynamo_module_patch = False
+
+            # Setup hook dict
+            modulecls._module_change_hooks = OrderedDict()
```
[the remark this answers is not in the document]
voznesenskym: Yes, this is a mistake. The patching should be on the cls, the data structure should be on the instance.

torchdynamo/variables/misc.py
voznesenskym: It's not, forgot to rm.

torchdynamo/guards.py
```diff
@@ -48,6 +55,72 @@
...
+                NNModuleChangeTrackerUtil._dynamo__setattr__(self, key, value)
+                return original_setattr(self, key, value)
+
+            modulecls.__setattr__ = custom_setattr
+
+        if module not in module_code_map:
+            module_code_map[module] = list()
+        module_code_map[module].append(guarded_code)
+
+    def _register_on_module_change_hook(
```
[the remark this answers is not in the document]
voznesenskym: I can try a version of that and see!

torchdynamo/guards.py
```diff
@@ -152,17 +225,26 @@ def arg_ref(self, guard: Guard):
...
+            return
+
         # ___check_type_id is same as `id(type(x)) == y`
         self.code.append(
             f"___check_type_id({self.arg_ref(guard)}, {self.id_ref(type(self.get(guard.name)))})"
         )
 
     def ID_MATCH(self, guard: Guard):
+        if guard.is_nn_module():
+            # Protected by module invalidation, see NN_MODULE
```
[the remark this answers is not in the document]
voznesenskym: Agreed, will add those.

torchdynamo/guards.py
```diff
@@ -179,6 +261,9 @@ def HASATTR(self, guard: Guard):
             self.code.append(f"not hasattr({ref}, {attr!r})")
 
     def EQUALS_MATCH(self, guard: Guard):
+        if guard.is_nn_module():
```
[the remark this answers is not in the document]
voznesenskym: Agreed, I'll add a test for this too.

torchdynamo/guards.py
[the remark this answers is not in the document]
voznesenskym: Might try both on the instance, actually.

torchdynamo/guards.py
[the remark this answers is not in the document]
voznesenskym: Resolved

torchdynamo/guards.py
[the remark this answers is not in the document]
voznesenskym: LMK what you think about this approach.

torchdynamo/convert_frame.py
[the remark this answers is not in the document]
voznesenskym: Tested it, I do not see guards generated for that self.module_list.

torchdynamo/convert_frame.py
```diff
@@ -296,6 +296,11 @@ def debug_print(prefix):
                 print()
             assert output.guards is not None
             CleanupManager.instance[code] = output.cleanups
+
+            if "self" in frame.f_locals and istype(frame.f_locals["self"], torch.nn.ModuleList):
+                # Protected by module invalidation, see NN_MODULE
+                output.guards = []
```
jansel, line 302: Won't this remove the call to `NNModuleChangeTrackerUtil.setup()` so nothing will be protected?

torchdynamo/guards.py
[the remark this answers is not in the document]
jansel: This seems the same as the prior one, perhaps we should chat in more detail on Monday.

torchdynamo/guards.py
```diff
@@ -274,20 +382,20 @@ def TUPLE_ITERATOR_LEN(self, guard):
         self.code.append(f"___tuple_iterator_len({ref}) == {tuple_iterator_len(value)}")
 
     def DICT_KEYS(self, guard):
+        if guard.is_nn_module():
+            # Protected by module invalidation, see NN_MODULE
+            return
```
jansel, line 387: I think this will be incorrect:
```
class Foo(nn.Module):
   def __init__(self):
        super().__init__()
        self.my_dict = collection.OrderedDict()

   def forward(self, x):
        if "x" in self.my_dict:
              return x + 1
        else:
              return x - 1

m = Foo()
   m(torch.randn(10))
   m.my_dict["x"] = 1
   m(torch.randn(10))
```

jansel (changes requested):
I think we need more testing to convince ourselves this is correct.  I'd suggest adding a bunch of tests to confirm we invalidate things when we expect to.

voznesenskym:
Okay, this feels more correct (Added some tests to cover cases @jansel  brought up, others as well) and is now a little faster (maybe?) than the other codepath. 

1.794x w/ guard_nn_modules
1.77x w/o

torchdynamo/convert_frame.py
voznesenskym: Only for the module list, which should already be a property on the module. I can add a specific test for this though.

jansel: Could this remove unrelated guards?  (For example globals or inputs?)

torchdynamo/guards.py
```diff
@@ -50,6 +60,98 @@
...
+            mod_id = id(mod)
+            module_associated_guarded_ids.add(mod_id)
+            module_to_guard_ids[module].add(mod_id)
+
+        for parameter in module.parameters():
+            parameter_id = id(parameter)
+            module_associated_guarded_ids.add(parameter_id)
+            module_to_guard_ids[module].add(parameter_id)
+
+        if getattr(modulecls, "__dynamo_module_patch", True):
```
jansel, line 106: If you have a class hierarchy, `A->B->C->D->nn.Module`:
I think it is possible we patch each class in the hierarchy and end up getting multiple calls to `invalidate_module`.

Perhaps we should just patch this single torch.nn.Module base class.

torchdynamo/guards.py
```diff
@@ -50,6 +60,98 @@
...
+
+        for buffer in module.buffers():
+            buffer_id = id(buffer)
+            module_associated_guarded_ids.add(buffer_id)
+            module_to_guard_ids[module].add(buffer_id)
+
+        for mod in module.modules():
+            mod_id = id(mod)
+            module_associated_guarded_ids.add(mod_id)
+            module_to_guard_ids[module].add(mod_id)
```
jansel, line 99: What if you have two modules, both with the same submodule.  
- A has submodule C (in frame 1)
- B has submodule C (in frame 2)

Both A and B get guarded, then A gets invalidated.  

I think `module_associated_guarded_ids` should not be global, be local to a specific guard creation session.

torchdynamo/utils.py
```diff
@@ -80,6 +80,9 @@ def clear(self):
         self.values.clear()
         self.refs.clear()
 
+    def __delitem__(self, idx):
+        self._remove_id(id(idx))
```
jansel, line 84: Update refs?

jansel (changes requested):

torchdynamo/utils.py
voznesenskym: I think `_remove_id` does:

```
def _remove_id(self, idx):
        if idx in self.refs:
            del self.values[idx]
            del self.refs[idx]
```
Or did you mean something else?

torchdynamo/convert_frame.py
voznesenskym: Yea, maybe, let me try without this, its a bit more heavy handed than I'd like anyway.

torchdynamo/guards.py
voznesenskym: Yea, I saw that, but I wasn't sure if I'd miss any edge cases with this. I'll simplify it down to just nn.Module and see how it goes.

torchdynamo/guards.py
voznesenskym: Yea, I agree. Is there an object whose lifecycle matches the guard creation session neatly? Perhaps GuardBuilder?

torchdynamo/utils.py
jansel: You are correct

voznesenskym:
Benchmarks:

With guard_nn_modules:

cuda eval  BERT_pytorch                       2.383x p=0.00
cuda eval  Background_Matting                 1.274x p=0.00
cuda eval  LearningToPaint                    1.177x p=0.00
cuda eval  alexnet                            1.177x p=0.00
cuda eval  dcgan                              1.014x SAME
cuda eval  demucs                             1.129x p=0.00
cuda eval  densenet121                        1.665x p=0.00
cuda eval  fambench_dlrm                      1.049x SAME
cuda eval  hf_Albert                          4.631x p=0.00


Without guard_nn_modules:

cuda eval  BERT_pytorch                       1.860x p=0.00
cuda eval  Background_Matting                 1.276x p=0.00
cuda eval  LearningToPaint                    1.187x p=0.00
cuda eval  alexnet                            1.178x p=0.00
cuda eval  dcgan                              1.061x p=0.01
cuda eval  demucs                             1.132x p=0.00
cuda eval  densenet121                        1.670x p=0.00
cuda eval  fambench_dlrm                      1.061x p=0.08
cuda eval  hf_Albert                          4.339x p=0.00

tests/test_misc.py
```diff
@@ -29,6 +29,13 @@ def my_custom_function(x):
     return x + 1
 
 
+def ifdyn(count1, count2):
```
jansel, line 32: refactor this into a common place (torchdynamo.testing) rather than copying

torchdynamo/guards.py
```diff
@@ -35,6 +37,8 @@
 
 log = logging.getLogger(__name__)
 
+# map from modules to guarded code
```
jansel, line 40: explain why this is needed and what it is used for in the comment

torchdynamo/guards.py
```diff
@@ -154,12 +244,22 @@ def arg_ref(self, guard: Guard):
...
+            return
+
         # ___check_type_id is same as `id(type(x)) == y`
         self.code.append(
             f"___check_type_id({self.arg_ref(guard)}, {self.id_ref(type(self.get(guard.name)))})"
         )
 
     def ID_MATCH(self, guard: Guard):
+        val = self.get(guard.name)
+        if id(val) in self.module_associated_guarded_ids or (
```
jansel, line 258: Should this be checking "is this a member of a guarded module"?

If you have:
```
a.b.c
```
We can skip the guard if `a.b` is in the guarded set.

`id(a.b.c)` isn't exactly right, because there is possible reuse issues.  You could have multiple references, some going through a nn.Module, and some not.

torchdynamo/variables/tensor.py
```diff
@@ -336,6 +344,8 @@ def call_method(
             constant_result = ConstantVariable(self.is_contiguous, **options)
         elif name == "is_complex" and self.is_complex is not None:
             constant_result = ConstantVariable(self.is_complex, **options)
+        elif name == "item" and self.item is not None:
+            constant_result = ConstantVariable(self.item, **options)
```
jansel, line 348: This is incorrect.  item() does not return a compile-time constant, it is a dynamically changing value.

voznesenskym: I think this is bad merge from another PR - my apologies.

torchdynamo/guards.py
voznesenskym: That's true, good call, let me fix that.

torchdynamo/guards.py
jansel: Not fixed?

torchdynamo/guards.py
```diff
@@ -246,28 +350,28 @@ def CONSTANT_MATCH(self, guard: Guard):
...
-
-        def setup_guard():
-            assert istype(val.training, bool)
-            self.code.append(f"{ref}.training == {val.training}")
 
+        val = self.get(guard.name)
         if hasattr(val, "training"):
-            # There are cases where a monkeypatched object has a guard made between __new__ and __init__
-            setup_guard()
+            self.register_module_for_invalidation(val, self.guarded_code)
```
jansel, line 356: I'd suggest registering this only on the first guard that this would remove.  It could be the case module has no guarded members and this isn't needed.

voznesenskym: I like this, let me try a lazy install here.

voznesenskym: I made some changes to make this possible, and it feels kinda gross, this feels cleaner, and if a module has no guarded members this invalidation is still pretty low overhead. Do you want me to commit up the alternative, lazy, approach to compare?

test6.py
```diff
@@ -0,0 +1,20 @@
+import torch
```
jansel, line 1: Did you mean to add this file?

jansel (changes requested):
I think this is close now, seems like there is an extra file though.

test6.py
voznesenskym: No - sorry, fat fingered a commit I guess. My apologies.

jansel:
Looks like there are also some merge conflicts.

voznesenskym:
> Looks like there are also some merge conflicts.

Yes this had some overlap w/ exports, so whichever landed first was going to conflict with the second. Will fix up.

jansel (approved):

voznesenskym:
Ran benchmarks, no regressions seen, speedups only.

I may RFC a benchmark comparator tool I have hacked together, to make it easier to post side-by-sides...

PR #333, opened 2022-06-07
Fix test skip when sympy is missing

jansel:
Hopefully this will fix https://www.internalfb.com/diff/D36937905


wconstab (approved):
worth a shot. But I wonder if we plan to add sympy in fbcode eventually?  or if it's already decided not to, should we add a step to OSS CI that doesn't install sympy and ensures some set of functionality still works?

jansel:
> worth a shot. But I wonder if we plan to add sympy in fbcode eventually? or if it's already decided not to, should we add a step to OSS CI that doesn't install sympy and ensures some set of functionality still works?

Yeah, we get sympy installed in the internal build.  This is more of a short term workaround.

Issue #330, opened 2022-06-07
tacotron2 hit recompilation cache limit


jansel:
```
$ ./torchbench.py -dcuda -k tacotron2
[...]

PR #335, opened 2022-06-07
Add config to print all failed guards

voznesenskym:
Small followup from debugging https://github.com/pytorch/torchdynamo/issues/330

### DIFF
--- a/torchdynamo/config.py
+++ b/torchdynamo/config.py
@@ -67,4 +67,6 @@ allowed_functions_module_string_ignorelist = {
     "torch.testing",
 }
 
+# Enable this to print every guard that failed once we have exhausted the chain.
+# Note: This does not print guards as they fail to reduce noise, but rather accumulates them until the last has failed.
 print_all_failed_guards = False
--- a/torchdynamo/config.py
+++ b/torchdynamo/config.py
@@ -66,3 +66,5 @@ allowed_functions_module_string_ignorelist = {
     "torch.distributions",
     "torch.testing",
 }
+
+print_all_failed_guards = False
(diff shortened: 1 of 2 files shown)


wconstab (approved):
LGTM!

jansel:
If we have multiple functions, could this mix the guard failure reasons between them?  Especially if not all the guard_fn's fail for the first one.

voznesenskym:
> If we have multiple functions, could this mix the guard failure reasons between them? Especially if not all the guard_fn's fail for the first one.

Yes, it will, but it keeps them ordered, and it keeps the related failures in a single array, and makes a 2d array of failures.

In the case for #330, It would go from:

```
... ['len(alignments) == 0'], ['len(alignments) == 0'], ...
``` 

to:

```
...
['len(alignments) == 0'], 
['len(alignments) == 1', 'len(alignments) == 0'], 
['len(alignments) == 2', 'len(alignments) == 1', 'len(alignments) == 0'],
['len(alignments) == 3', 'len(alignments) == 2', 'len(alignments) == 1', 'len(alignments) == 0'],
...
```

If we had an output like this, imo, it would be easier to tell that the invalidation was happening every single frame for the same value, being monotonically incremented. 

Which, if we had different failures, would ostensibly look something like:

```
...

['__some_foo_guard(x)', '__some_bar_guard(y)', 'len(alignments) == 0'],
...
```

@wconstab  and I chatted a little about better organization, and there's probably another PR here somewhere after this one that makes the printing clearer w/r/t tracking which failure came from which `code`, adding indices, etc.

jansel (approved):
I think splitting this out per-function would be better, but approving as this is primarily for debugging.

PR #337, opened 2022-06-07
[inductor] Benchmark harness for training

jansel:


torchbench.py
```diff
@@ -115,22 +119,47 @@
...
-        "hf_Bart",  # 400s
-        "hf_Bert",  # 334s
-        "hf_DistilBert",  # 187s
-        "hf_GPT2",  # 470s
-        "hf_Reformer",  # 141s
-        "speech_transformer",  # 317s
-        "vision_maskrcnn",  # 99s
-    },
     *VERY_SLOW_BENCHMARKS,
+    "BERT_pytorch",  # 137s
```
Chillee, line 123: What are these times measuring? how long they take to run?

Chillee (approved):
Makes sense to me.

torchbench.py
jansel: Yes correct.

jansel: This implements `--fast` which skips slow CPU benchmarks.  (A minor quality of life feature for correctness testing.)

torchinductor/compile_fx.py
```diff
@@ -106,16 +107,27 @@ def compile_fx(
...
     if os.environ.get("TORCHINDUCTOR_CHECK_OPS") == "1":
         wrap(CheckEachNode(gm).run)(*example_inputs)
 
+    return compile_fx_inner(gm, example_inputs, wrap=wrap, cudagraphs=cudagraphs)
+
+
+def compile_fx_inner(
+    gm: torch.fx.GraphModule,
+    example_inputs: List[torch.Tensor],
+    wrap=identity,
```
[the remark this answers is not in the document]
jansel: It is this:
https://github.com/jansel/torchdynamo/blob/df478179e881c6fc17e91a4fe075659a29407777/torchdynamo/optimizations/python_key.py#L126

It adds all the module parameters via `*params_flat` in the call.

Basically this is a lower overhead version of `autograd.Function`.  Rather than creating new `autograd.Function`, it gives you a python closure.

desertfire (approved):

PR #338, opened 2022-06-07
[inductor] Minor float16 fix

jansel:
This works around https://github.com/openai/triton/issues/537 and a low precision numerics issue with `test_l1_loss`


Chillee (approved):
Does this hard-code all fp16 ops to be run in fp32? Or just specifically `to(torch.float32)` calls?

jansel:
> Does this hard-code all fp16 ops to be run in fp32? Or just specifically `to(torch.float32)` calls?

https://github.com/pytorch/torchdynamo/pull/304 did that, this is just a minor fix on top of that.

PR #345, opened 2022-06-08
Add _refs and _prims to the allowlist

ezyang:
* __->__ #345

This won't get exercised by real models but it's necessary so we
can test that PrimTorch decomps work under dynamo.

### DIFF
--- a/tests/test_recompile_ux.py
+++ b/tests/test_recompile_ux.py
@@ -111,7 +111,10 @@ class RecompileUxTests(torchdynamo.testing.TestCase):
 
     def test_verbose_tensor_check(self):
         def func(a):
-            return torch.sum(a)
+            # Warning: choose a function here whose meta implementation lives
+            # entirely in C++.  If you do a Python one, Dynamo will dive into
+            # torch._refs which is OK but it will muddy up the warnings
+            return torch.add(a, 4)
 
         def cache_fail_test(cached_input, missed_input, expected_failure):
             # TODO(whc) maybe its hacky to have a 'test within a test' but this seemed convenient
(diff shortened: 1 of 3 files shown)


ezyang:
I think this is not enough. Need something like

```
diff --git a/torchdynamo/allowed_functions.py b/torchdynamo/allowed_functions.py
index d5ea31a..6b092fc 100644
--- a/torchdynamo/allowed_functions.py
+++ b/torchdynamo/allowed_functions.py
@@ -110,6 +110,12 @@ def _allowed_function_ids():
         if module is None:
             return False
 
+        if any(
+            module.__name__.startswith(mod_name)
+            for mod_name in config.allowed_functions_module_string_ignorelist
+        ):
+            return False
+
         mod_name = module.__name__
         return mod_name in allowed_modules or mod_name.startswith(allowed_modules_dot)
 
@@ -122,12 +128,10 @@ def _allowed_function_ids():
         torch_object_ids[id(module)] = module.__name__
         for name, obj in list(module.__dict__.items()):
             if id(obj) not in torch_object_ids:
-                if isinstance(obj, types.ModuleType):
-                    if obj.__name__.startswith("torch."):
-                        torch_object_ids[id(obj)] = f"{module.__name__}.{name}"
-                        _find_torch_objects(obj)
-                elif _is_allowed_module_prefix(obj):
+                if _is_allowed_module_prefix(obj):
                     torch_object_ids[id(obj)] = f"{module.__name__}.{name}"
+                    if isinstance(obj, types.ModuleType):
+                        _find_torch_objects(obj)
                 elif inspect.getmodule(obj) is None and not is_safe_constant(obj):
                     torch_object_ids[id(obj)] = f"{module.__name__}.{name}"
 

```

ezyang:
But this is also not quite right as it breaks type information for dtypes like torch.bfloat16

jansel (approved):

ezyang:
oops i forgot this is a ghstack haha

PR #346, opened 2022-06-08
Skip check_fn if no guards were produced upstream

voznesenskym:


voznesenskym:
I’m not 100% sure this is the behavior we want - @jansel thoughts?

voznesenskym:
This is right upstream of that, so we don’t have a chance to make a valid
check here yet. I figured, probably incorrectly, that if we go into
GuardedCode creation with no guards, we can skip creating this guarded
code.

PR #352, opened 2022-06-09
Add coldstart benchmark

wconstab:
This actually adds 2 new benchmark metrics:
* coldstart: measures the worst of t_eager_compile / t_dynamo_compile as a 'speedup', where dynamo compiles twice to exercise profiling executor
* breakeven: predicts the number of iterations dynamo would have to run to 'break even' with eager, considering the amortization of its compile cost

Test with
`python torchbench.py --cold-start -d cuda --training --use-eval-mode --nvfuser --isolate`

Not yet tested with inference or cpu, may have some other issues.

Should probably adjust to repeat the whole cold-start process several times and median, but for now just does this once.

### DIFF
--- a/torchbench.py
+++ b/torchbench.py
@@ -361,6 +361,72 @@ def randomize_input(inputs):
             f"randomize_input can not handle input of type {type(inputs)}"
         )
 
+def cold_start_experiment(args, model_iter_fn, model, example_inputs, optimize_ctx):
+    compile_iters = 2
+    total_iters = compile_iters + 2
+    timings = np.zeros((total_iters, 2), np.float64)
+    # if we randomize the input, we should also check the result is correct
+    should_check_result = should_randomize_input = args.randomize_input
+    is_correct = True
+
+    for rep in range(total_iters):
+        inputs = (
+            randomize_input(copy.deepcopy(example_inputs))
+            if should_randomize_input
+            else example_inputs
+        )
+
+        # interleave the runs to handle frequency scaling and load changes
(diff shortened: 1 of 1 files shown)


jansel (approved):

torchbench.py
```diff
@@ -954,6 +1036,16 @@ def main():
...
         output_filename = "overheads.csv"
+    elif args.cold_start:
+        optimize_ctx = torchdynamo.optimize(
+            aot_autograd_speedup_strategy, nopython=args.nopython
+        )
+        experiment = cold_start_experiment
+        backend_str = "nvfuser" if args.nvfuser else "nnc"
+        output_filename = "cold_start_{backend_str}.csv"
+        args.isolate = True
+        # horace said so
```
wconstab, line 1047: oops.. haha i should remove this. 

but i did want to confirm whether putting it just in this cold-start benchmark makes sense- isn't it something that should be applied across all the (maybe training only) benchmarks more uniformly?

torchbench.py
```diff
@@ -362,6 +362,85 @@ def randomize_input(inputs):
...
+        ("dev", "name", "cold-start speedup", "breakeven iters"),
+        [
+            current_device,
+            current_name,
+            float(speedup),
+            breakeven(dynamo_times, eager_times),
+        ],
+    )
+
+    print([f"{t:.2f}" for t in dynamo_times], [f"{t:.2f}" for t in eager_times])
```
wconstab, line 428: oops meant to remove this.

PR #356, opened 2022-06-09
Only wrap in TorchVariable if is allowed, not if not disallowed

ezyang:
* (to be filled)

If you don't do this, allowed_functions_module_string_ignorelist
doesn't actually affect if we try to trace these functions into
the graph, since the disallowed list doesn't actually respect
this config.

### DIFF
--- a/torchdynamo/variables/builtin.py
+++ b/torchdynamo/variables/builtin.py
@@ -13,7 +13,7 @@ from torchdynamo.variables.tensor import DynamicShapeVariable
 
 from .. import config
 from .. import variables
-from ..allowed_functions import is_disallowed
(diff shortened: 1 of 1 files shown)


ezyang:
This actually causes the primtorch test to fail on master but CI here doesn't see it since no master=

jansel (approved):

tests/test_repros.py
```diff
@@ -1256,9 +1256,18 @@ def fn(x):
         self.assertTrue(same(ref1, res1))
 
     @unittest.skipIf(not HAS_REFS, "requires recent PT version")
+    @unittest.expectedFailure
```
jansel, line 1259: Why is this failing?

Issue #353, opened 2022-06-09
torchinductor ir.py ExternKernel codegen_size_asserts failed for channel-last layout


pyjhzwh:
Steps to repro:
```
import torch
import triton
import torchdynamo

import torchinductor.config

torchinductor.config.debug = True


BATCH, IN_C, IN_H, IN_W, KERNEL_N, KERNEL_H, KERNEL_W, stride, padding, dilation, groups, dtype, layout = \
    32, 128, 32, 32, 32, 1, 1, (1, 1), (0, 0), (1, 1), 1, torch.float32, "nhwc"

torch.manual_seed(0)

# allocate inputs
x = torch.randn((BATCH, IN_C, IN_H, IN_W), dtype=dtype, device='cuda')
w = torch.randn((KERNEL_N, IN_C // groups, KERNEL_H, KERNEL_W),
                dtype=dtype, device='cuda')
bias = torch.randn((KERNEL_N), dtype=dtype, device='cuda')
if layout == "nhwc":
    x = x.to(memory_format=torch.channels_last)
    w = w.to(memory_format=torch.channels_last)


def wrap_conv(
    x, w, bias,
    stride, padding,
    dilation, groups,
):
    y = torch.conv2d(x, w, bias, stride, padding, dilation, groups)
    return y

with torchdynamo.optimize("inductor"):
    y = wrap_conv(
        x, w, bias,
        stride, padding,
        dilation, groups,
    )

# baseline
y_baseline = torch.conv2d(x, w, bias, stride, padding, dilation, groups)

print('y_baseline', y_baseline.stride())
triton.testing.assert_almost_equal(y, y_baseline, decimal=1)
```

The generated code is 
```
from ctypes import c_void_p, c_long
import torch
from torch import empty_strided, as_strided
from torchinductor.codecache import CppCodeCache, TritonCodeCache, grid
try:
    import triton
    import triton.language as tl
except ImportError:
    pass
aten = torch.ops.aten


def call(arg0, arg1, arg2):
    arg0_size = arg0.size()
    s0 = arg0_size[0]
    s1 = arg0_size[1]
    buf0 = aten.convolution(arg0, arg1, arg2, (1, 1), (0, 0), (1, 1), False, (0, 0), 1)
    assert buf0.size() == (32, 32, 32, 32)
    assert buf0.stride() == (32768, 1024, 32, 1)
    return (buf0, )

if __name__ == "__main__":
    from torchdynamo.testing import rand_strided
[...]

PR #358, opened 2022-06-09
order strides for channel-last layout in Convolution

pyjhzwh:
Fix memory layout strides of outputs in Convolution. Issue #353 
Output layout is the same as x layout

### DIFF
--- a/torchinductor/ir.py
+++ b/torchinductor/ir.py
@@ -2027,7 +2027,9 @@ class Convolution(ExternKernelAlloc):
         # memory layout order of output depends on x
         # symbolic stride of x, e.g. [s1**2*s0, 1, s0*s1, s1]
         # order according to .atoms(), but constant "1" < "s1"
-        x_stride_order = [len(expr.atoms()) - int(expr.is_number) for expr in x.get_stride()]
+        x_stride_order = [
+            len(expr.atoms()) - int(expr.is_number) for expr in x.get_stride()
+        ]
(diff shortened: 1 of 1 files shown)


torchinductor/ir.py
```diff
@@ -2024,12 +2024,19 @@ def create(
                 V.graph.sizevars.guard_static_shape(output_size[-1])
             )
 
+        # memory layout order of output depends on x
+        # symbolic stride of x, e.g. [s1**2*s0, 1, s0*s1, s1]
+        # order according to .atoms(), but constant "1" < "s1"
+        x_stride_order = [
+            len(expr.atoms()) - int(expr.is_number) for expr in x.get_stride()
```
jansel, line 2031: It might be better to use `V.graph.sizevars.size_hint`.

There is some similar stuff (for a memory address, not a list of strides) here:
https://github.com/pytorch/torchdynamo/blob/488fad00afc7e2a1e11adc9b13587aa25466f5b4/torchinductor/sizevars.py#L232

jansel (changes requested):

torchinductor/ir.py
pyjhzwh: Use `V.graph.sizevarssize_hint` to get actual strides. Thanks

jansel (approved):

pyjhzwh:
Resolved in #364

PR #359, opened 2022-06-09
Track failures with more details, prevent guard installation and graph break on failure threshold 

voznesenskym:
ref https://github.com/pytorch/torchdynamo/issues/330


anijain2305:
How is it different from setting cache_size_limit to 3? IIUC, for a code object, if the number of recompilations reach a threshold (64), we stop recompiling.

Your PR is working at a guard abstraction, instead of a code object. This brings extra overhead, but I am not sure what benefit it brings.

voznesenskym:
> How is it different from setting cache_size_limit to 3? IIUC, for a code object, if the number of recompilations reach a threshold (64), we stop recompiling.
> 
> This PR is working at a guard abstraction, instead of a code object. Maybe this brings extra control, but I am not sure what benefit it brings. It will definitely have perf overhead as each guard now accesses a global state, increments, make extra comparisons etc.

Hmm, I do not think the guards access global state, increments, or make comparisons. This is only happening at guard creation and guard failure. The guards themselves stay exactly as they were before this PR.

anijain2305:
Ah, that makes sense. Yes, we are checking the count when there is a guard failure. So, no extra perf overhead in common case.

However, I still want to know if this is better than `cache_size_limit` https://github.com/pytorch/torchdynamo/blob/main/torchdynamo/config.py#L31

voznesenskym:
> Ah, that makes sense. Yes, we are checking the count when there is a guard failure. So, no extra perf overhead in common case.
> 
> However, I still want to know if this is better than `cache_size_limit` https://github.com/pytorch/torchdynamo/blob/main/torchdynamo/config.py#L31

Wonderful question! It's not better than, strictly, but it's a little different. Specifically, it allows us to decouple total failures (the cache size limit) from identical failures. If we look at the doc for `cache_size_limit`: 

`# disable (for a function) when cache reaches this size`

That means that if we have a func `foo()` and it fails 64 times on *any* failure (imagine first it fails on one guard, then on another, then on a third, and so forth and 64 different failures of any combination), then we have broken the cache_size_limit. 

what `same_failure_tolernace_threshold` introduces is a count for failures on a *specific* failure. A specific failure means the same func, but also the same guard type on the same guarded object.

anijain2305:
I see. The answer makes sense to me.

I think this is a pretty good reactive action when a particular guard is going to be continuously failing. 

Longer-term, we should also go for a more pro-active action, where if we skip compiling if we expect a guard to be failing constantly. Even in that case, we might still need this PR for cases that we missed.

torchdynamo/guards.py
```diff
@@ -136,6 +141,23 @@ def __init__(
         self.tensor_check_names = []
         self.tensor_check_examples = []
         self.guarded_code = guarded_code
+        self.full_guarded_code_loc = (
+            f"{full_guarded_code.co_filename}:{full_guarded_code.co_firstlineno}"
+        )
+        self.check_code_to_check_type_dict = dict()
+
+    def append_code_for_check_type(self, code, check_type, name):
```
anijain2305, line 149: Is it possible to convert this in a decorator?

voznesenskym: Should be possible - I do find decorators harder to read, but I dont have a strong opinion on style.

voznesenskym:
> I see. The answer makes sense to me.
> 
> I think this is a pretty good reactive action when a particular guard is going to be continuously failing.
> 
> Longer-term, we should also go for a more pro-active action, where if we skip compiling if we expect a guard to be failing constantly. Even in that case, we might still need this PR for cases that we missed.

Thanks :) I plan to actually followup on this with diffs that do analysis on well-known paths that might invalidate (hot loops setting consts, lists changing size, etc etc) - not installing guards in the first place beats the pants off of throwing them out after N failures! However, that's for after this.

jansel:
1) Does this add any overheads?  Have we measured it?

2) We may want to treat some failures differently than others.  For example ID_MATCH is common from different layers using the same code, so we might want a higher tolerance for that.

voznesenskym:
> 1. Does this add any overheads?  Have we measured it?

I think so! 

```
cpu  eval  BERT_pytorch                       1.043x SAME
cpu  eval  Background_Matting                 1.011x SAME
cpu  eval  LearningToPaint                    1.010x SAME
cpu  eval  Super_SloMo                        1.012x p=0.05
cpu  eval  alexnet                            0.997x SAME
```

on `./torchbench.py --overhead`. So I think it doesn't. I would not expect it to, as it's pretty light weight to store, and pretty lightweight to access, and only happens in failures. 


> 2. We may want to treat some failures differently than others.  For example ID_MATCH is common from different layers using the same code, so we might want a higher tolerance for that.

I thought of that, but I also felt like it might make it a little too complicated. I can add this component now, or add it in another PR. Your call as to what you prefer!

jansel:
@wconstab would you make taking a look at this?

I think we definitely need to improve the way we handle cache limits and falling back, but I am worried this way of doing it might add too much complexity and not solve the underlying problem.  It will be hard for users to reason about the multiple cache limits (per-error-type vs per-code-object), which will make our error messages the recompile profiler more confusing.

@wconstab has been proposing just setting the cache limit to 1 (which would remove the need for this PR), though I am curious if he thinks a cache limit of 1 per error type would be better.

I am kind of leaning towards cache limit of 1, but I think we should make the limit-of-1 apply at the nn.Module-instance/closure level rather than at the PyCodeObject level as it does today.

voznesenskym:
> @wconstab would you make taking a look at this?
> 
> I think we definitely need to improve the way we handle cache limits and falling back, but I am worried this way of doing it might add too much complexity and not solve the underlying problem. It will be hard for users to reason about the multiple cache limits (per-error-type vs per-code-object), which will make our error messages and the recompile profiler more confusing.
> 
> @wconstab has been proposing just setting the cache limit to 1 (which would remove the need for this PR), though I am curious if he thinks a cache limit of 1 per error type would be better.
> 
> I am kind of leaning towards cache limit of 1, but I think we should make the limit-of-1 apply at the nn.Module-instance/closure level rather than at the PyCodeObject level as it does today.

I'm fine dropping this PR in favor of analysis. I certainly agree it does not solve the underlying problem. I also agree that a limit of 1 invalidates this PR. 1 per instance seems interesting, why do you prefer that over the PyCodeObject approach?

jansel:
Suppose we have a cache limit of 1, what does that break?

One super common case would be if you have: 
```
layer1 = Foo()
layer2 = Foo()
layer3 = Foo()
y = layer3(layer2(layer1(x)))
```

`Foo.forward` will need 3 cache entries at the PyCodeObject level, but only 1 cache entry per module instance.

If we run real models with cache limit 1, we can find other cases like that.  Then try to refine our approach such that common stuff works, but we also fall back to eager quickly.

torchdynamo/guards.py
```diff
@@ -153,8 +184,10 @@ def arg_ref(self, guard: Guard):
 
     def TYPE_MATCH(self, guard: Guard):
         # ___check_type_id is same as `id(type(x)) == y`
-        self.code.append(
-            f"___check_type_id({self.arg_ref(guard)}, {self.id_ref(type(self.get(guard.name)))})"
+        self.append_code_for_check_type(
```
wconstab, line 187: I like the aspect of your PR that refactors how guards are accumulated such that there is an opportunity to organize them and potentially use this to also help the error messages we print.

wconstab:
I think it'd be nice to split out the part of the PR that refactors/improves existing behavior and land that.

For behavior change, i'm wary of adding complexity until we know more.  I'd favor a simpler approach first.

If we go with the nn.module cache, I think it'd likely be a net win.  Still, there could be a failure mode where a model with 100s of layers recompiles that many times- do we think this requires additional prevention?

jansel:
> If we go with the nn.module cache, I think it'd likely be a net win. Still, there could be a failure mode where a model with 100s of layers recompiles that many times- do we think this requires additional prevention?

Is this fundamentally different from a model having hundreds of functions?  

IMO background compilation might be the fix for both those cases.

wconstab:
> Is this fundamentally different from a model having hundreds of functions?

nope, guess it is not.

i'm also not sure how to model the cost of doing one whole-graph compile over such a model, vs doing 100s of individual, small compiles.  if the two are about the same, then this is probably a moot point.

but you've convinced me that it is OK to compile the same function 100s of times from different invocation points.

PR #361, opened 2022-06-09
Disabling TorchDynamo inside torch.jit and torch.onnx compiler

anijain2305:
* Helps with many PyTorch JIT tests.  Only 7 test/jit/* failures now. 
* All 136 failures with test/onnx/* are resolved.

### DIFF
--- a/torchdynamo/allowed_functions.py
+++ b/torchdynamo/allowed_functions.py
@@ -100,7 +100,6 @@ def _allowed_function_ids():
     Walk torch.* and get the ids of all the stuff in it
     """
     warnings.filterwarnings("ignore", category=UserWarning, module="torch.distributed")
-    torch.distributions.Distribution.set_default_validate_args(False)
     torch_object_ids = dict()
 
     def _is_allowed_module_prefix(obj):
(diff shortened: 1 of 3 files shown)


jansel (changes requested):
Can we do this without monkey patching?  Perhaps something similar to skipfiles?  Skipfunctions?

anijain2305:
Its not just the frame of `torch.jit.trace`, instead it's the inner frames in the execution of `torch.jit.trace` which causes problems. TorchDynamo changes the bytecode of the inner frames, which changes `torch.jit.trace` result, and causes test failures.

So, I need something like `disable`. IIUC, `skip` just skips one frame, and not the inner frames. What do you think?

anijain2305:
> I see, in that case lets create a file like `torchdynamo/patch_torch.py` to make these easier to find. Let's do the patches on the first call to `torchdynamo.optimize` rather than at import time.
> 
> You could move this one to that file as well:
> 
> https://github.com/pytorch/torchdynamo/blob/85561d0f4f656af71e0f3fa9f56af61466ac8f7c/torchdynamo/allowed_functions.py#L103

Nice! This is cleaner. Let me do that.

anijain2305:
Also, should I put I unpatch it using __enter__ and __exit__ if we are in `torchdynamo.optimize`?

jansel:
No, I'd just patch it once with something like `functools.lru_cache()`

jansel (approved):

Issue #366, opened 2022-06-10
pyyaml missing from requirements.txt


shingjan:
The below two tests will fail if `pyyaml` is not installed.
```
FAILED tests/test_dynamic_shapes.py::DynamicShapesMiscTests::test_sample_input_dynamic_shapes - ModuleNotFoundError: No module named 'yaml'
FAILED tests/test_misc.py::MiscTests::test_sample_input - ModuleNotFoundError: No module named 'yaml'
```

jansel:
Fixed by #367

PR #367, opened 2022-06-10
add yaml to requirements.txt

shingjan:
This PR intends to fix issue #366 by adding pyyaml as a python dependency


jansel (approved):
Thanks!

PR #378, opened 2022-06-10
Fix missing ir.Reduction.default_value

### LINKED ISSUE #375: ./torchbench.py -dcuda --inductor -k BERT_pytorch failing
Asked for repro, got logs from Bin: https://gist.github.com/voznesenskym/ee1c4339f54683da71f49faaa875502b (cc: @desertfire)

jansel:
Fixes #375

### DIFF
--- a/torchinductor/ir.py
+++ b/torchinductor/ir.py
@@ -298,6 +298,12 @@ class Reduction(Loops):
             )
         )
 
+    @staticmethod
+    def default_value(reduction_type, dtype):
+        return {"sum": 0, "max": float("-inf"), "min": float("inf"), "any": 0}[
+            reduction_type
+        ]
+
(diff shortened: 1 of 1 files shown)


voznesenskym (approved):
LGTM! 

Only two super minor nit/ feedback. (1) is that since this class has both `ir.Reduction.create` and `ir.Reduction.default_value` it looks like a static ctor (In fact, I assumed that what this should have been when I saw the error!). If it just returns a dict, it may make the API feel a little more predictable to move it to utils. (2) is that I noticed we use "sum"/"max"/"min"/"any" as meaningful strings around reduction workflows across many files (lowering, cpp, etc). Moving those into some consts in a util file somewhere might not be amiss.

PR #381, opened 2022-06-11
[inductor] Refactor fallback kernel handling

jansel:


tests/test_torchinductor.py
```diff
@@ -60,6 +61,21 @@
 torchinductor.config.triton.autotune = False  # too slow
 
 
+def requires_decomp(fn):
+    """Decorator to disable test of a decomp is missing"""
```
voznesenskym, line 65: nit: if a decomp is missing?

torchinductor/decomposition.py
```diff
@@ -250,3 +252,13 @@ def masked_fill_(x, mask, value):
 @register_decomposition([aten.log1p])
 def log1p(x):
     return torch.log(x + 1)
+
+
+@register_decomposition([aten.baddbmm])
+def baddbmm(self, batch1, batch2, beta=1, alpha=1):
+    result = torch.bmm(batch1, batch2)
+    if not isinstance(alpha, (int, float)) or alpha != 1:
```
voznesenskym, line 260: nit: I think this can be `isinstance(alpha, numbers.Number)`

same below.

torchinductor/ir.py
```diff
@@ -1577,6 +1577,12 @@ def copy_input(x):
 
     @classmethod
     def realize_input(cls, x):
+        if x is None:
+            return V.graph.add_tensor_constant(torch.tensor(()))
```
voznesenskym, line 1581: Am pretty new to understanding the IR, so apologies if this is a dumb question - why does None make a default tensor constant?

jansel: This is a workaround in that some other code expects the graph to return only tensors.  The None values are coming from `aten.convolution_backwards` and are unused, so it doesn't matter.

voznesenskym: Understood! Would we eventually ever have a specialized "NoneTensor" or something like that, to signal that this came from a None?

torchinductor/lowering.py
```diff
@@ -47,6 +47,15 @@
 }
 
 
+def has_torchvision_roi_align():
+    try:
+        from torchvision.ops import roi_align  # noqa
+
+        return roi_align is not None
+    except (ImportError, ModuleNotFoundError):
+        return False
```
voznesenskym, line 56: I wonder if importlib can help avoid exceptions for control flow here, no strong opinions though.

voznesenskym (approved):

torchinductor/ir.py
jansel: I'm not convinced None carries any useful signal we want to preserve in the backend.  In Dynamo, None is useful, but in the backend you can't do anything with None other than ignore it.

PR #383, opened 2022-06-12
Adding nan

anijain2305:


jansel (changes requested):
This one might not work because:
```
>>> float("nan")==float("nan")
False
```

So if this is coming from CONSTANT_MATCH we need to use math.isnan()

anijain2305:
@jansel Ready for review.

jansel (approved):

PR #384, opened 2022-06-12
Fix for `any` decomposition

ngimel:
Previously it produced wrong results for tensor consisting of negative and 0 elements


torchinductor/lowering.py
```diff
@@ -1489,7 +1491,6 @@ def copy_(dst, src, non_blocking=False):
 square = register_pointwise(aten.square)
 sub = register_pointwise(aten.sub)
 
-register_pointwise(aten.abs)
```
jansel, line 1492: Did you mean to remove this?

jansel (changes requested):

jansel (approved):

PR #391, opened 2022-06-13
Support for lazy modules

mlazos:
Added fixes to support [lazy modules](https://pytorch.org/docs/stable/generated/torch.nn.modules.lazy.LazyModuleMixin.html) in torchdynamo. The main issue that needed to be addressed is that LazyModules register some hooks which are run when a module is called. Torchdynamo typically calls the `forward` method instead of `__call__`  so these hooks were never run. In the case of LazyModules we now run and trace the `__call__` method, and allow the original module to be mutated. In the future, we could do this for all modules, but there were cases where torchdynamo does not yet support functionality used in all hooks.

### DIFF
--- a/tests/test_modules.py
+++ b/tests/test_modules.py
@@ -808,33 +808,59 @@ class NNModuleTests(torchdynamo.testing.TestCase):
 
     def test_lazy_module(self):
         input_shape = (16, 3, 6, 7, 8)
+
         cnt = torchdynamo.testing.CompileCounter()
+        module = LazyModule()
         with torchdynamo.optimize(cnt):
 
-            def test_dynamic_module():
+            def test_static_module():
+                input = torch.ones(*input_shape)
+                module(input)
+
+            test_static_module()
+
+        self.assertTrue(
+            isinstance(module, MaterializedModule),
+            "Module should be transformed to an instance of MaterializedModule.",
+        )
+        self.assertEqual(module.param.shape, input_shape)
+
+        # test when mapped to UnspecializedNNModule
+        module = LazyModule()
+        with torchdynamo.optimize(cnt):
+
+            def test_unspecialized():
+                nonlocal module
                 module = LazyModule()
                 input = torch.ones(*input_shape)
                 module(input)
-                self.assertTrue(
-                    isinstance(module, MaterializedModule),
-                    "Module should be transformed to an instance of MaterializedModule.",
-                )
-                self.assertEqual(module.param.shape, input.shape)
 
-            test_dynamic_module()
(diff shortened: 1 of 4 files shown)


tests/test_modules.py
```diff
@@ -776,3 +805,36 @@ def forward(self, x):
...
+    def test_lazy_module(self):
+        input_shape = (16, 3, 6, 7, 8)
+        cnt = torchdynamo.testing.CompileCounter()
+        with torchdynamo.optimize(cnt):
+
+            def test_dynamic_module():
+                module = LazyModule()
+                input = torch.ones(*input_shape)
+                module(input)
+                self.assertTrue(
```
jansel, line 818: Let's move the assert outside torchdynamo.optimize()

tests/test_modules.py
```diff
@@ -776,3 +805,36 @@ def forward(self, x):
...
+
+    def test_lazy_module(self):
+        input_shape = (16, 3, 6, 7, 8)
+        cnt = torchdynamo.testing.CompileCounter()
+        with torchdynamo.optimize(cnt):
+
+            def test_dynamic_module():
+                module = LazyModule()
+                input = torch.ones(*input_shape)
+                module(input)
```
jansel, line 817: we should call this multiple times and assert cnt.frame_count is 1 (to check for guards failing)

tests/test_modules.py
```diff
@@ -776,3 +805,36 @@ def forward(self, x):
...
         self.assertTrue(torchdynamo.testing.same(pre, opt_pre))
         self.assertTrue(torchdynamo.testing.same(out1, out_post))
+
+    def test_lazy_module(self):
+        input_shape = (16, 3, 6, 7, 8)
+        cnt = torchdynamo.testing.CompileCounter()
+        with torchdynamo.optimize(cnt):
+
+            def test_dynamic_module():
+                module = LazyModule()
```
jansel, line 815: Construct this outside of torchdynamo.optimize(), otherwise it gets mapped to UnspecializedNNModule()

jansel (changes requested):

tests/test_modules.py
jansel: To make sure it isn't getting optimized away.

tests/test_modules.py
mlazos: Do I want to test that case too? I also have a check in UnspecializedNNModule to handle lazy modules.

jansel: Yes we should check them both.

mlazos: Should be good now

PR #392, opened 2022-06-13
Step 2 of supporting UnspecializedNumpyVariable & UnspecializedPythonVariable

yanboliang:
This is follow up of https://github.com/pytorch/torchdynamo/pull/321, so please aware of changes in that PR when you review this one. (I think we should combine these two PRs together, my bad!)

[TorchDynamo supports unspecialized primitive type design doc](https://docs.google.com/document/d/1UxbO0XmXixXqukOeFkWNV-9-6F6AZ71dpNmdNL1cWQw/edit#)

major changes:
* Add ```UnspecializedNumpyVariable``` and ```UnspecializedPythonVariable```.
* Update the builtin function logic:
  * Tensor op others -> Tensor
  * Unspecialized op (Unspecialized, Constant) -> Unspecialized
 * Initial implementation to support ```random``` functions.

### DIFF
--- a/tests/test_misc.py
+++ b/tests/test_misc.py
@@ -1614,7 +1614,7 @@ class MiscTests(torchdynamo.testing.TestCase):
             r1 = random.random()
             y = x + random.uniform(10, 20)
             r2 = random.randint(2, 18)
-            return y + r1 - r2
+            return y + r1, r2
 
         x = torch.tensor([[1.0, 2.0], [3.0, 4.0]])
         random.seed(1)
--- a/tests/test_misc.py
+++ b/tests/test_misc.py
@@ -1593,6 +1593,20 @@ class MiscTests(torchdynamo.testing.TestCase):
         self.assertEqual(cnts.frame_count, 1)
         self.assertEqual(cnts.op_count, 2)
 
+    def test_unspecialized_primitive_variable3(self):
+        # test unspecialized primitive max/min
+        def fn(x, y, z):
+            return z + 1, max(x, y), min(x - 4, y)
+
+        x = np.int64(12)
+        y = 10
+        z = torch.tensor([[1.0, 2.0], [3.0, 4.0]], dtype=torch.float64)
+        res1 = fn(x, y, z)
+        cnts = torchdynamo.testing.CompileCounter()
+        with torchdynamo.optimize(cnts):
+            res2 = fn(x, y, z)
+        self.assertTrue(same(res1, res2))
+
     def test_side_effects_codegen_update_mutated(self):
         # codegen to update mutated variables with side effect
         # should after stack value's codegen
--- a/tests/test_misc.py
+++ b/tests/test_misc.py
@@ -6,6 +6,7 @@ import dis
 import enum
 import functools
 import math
+import random
 import sys
 import typing
 import unittest
@@ -1606,6 +1607,22 @@ class MiscTests(torchdynamo.testing.TestCase):
         with torchdynamo.optimize(cnts):
             res2 = fn(x, y, z)
         self.assertTrue(same(res1, res2))
+
+    def test_unspecialized_primitive_variable4(self):
+        # no recompilations on random.random
+        def fn(x):
+            y = random.random()
+            x.sum().item()  # break graph
+            return x + y
+
+        x = torch.tensor([[1.0, 2.0], [3.0, 4.0]])
+
+        cnts = torchdynamo.testing.CompileCounter()
+        with torchdynamo.optimize(cnts):
+            for i in range(10):
+                fn(x)
+        self.assertEqual(cnts.frame_count, 2)
+        self.assertEqual(cnts.op_count, 2)
 
     def test_side_effects_codegen_update_mutated(self):
         # codegen to update mutated variables with side effect
--- a/tests/test_misc.py
+++ b/tests/test_misc.py
@@ -1609,11 +1609,12 @@ class MiscTests(torchdynamo.testing.TestCase):
         self.assertTrue(same(res1, res2))
 
     def test_unspecialized_primitive_variable4(self):
-        # no recompilations on random.random
+        # test random functions
         def fn(x):
-            y1 = random.random()
-            y2 = random.random()
-            return x + y1, y1, y2
+            r1 = random.random()
+            y = x + random.uniform(10, 20)
+            r2 = random.randint(2, 18)
+            return y + r1 - r2
 
         x = torch.tensor([[1.0, 2.0], [3.0, 4.0]])
         random.seed(1)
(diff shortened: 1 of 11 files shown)


torchdynamo/codegen.py
```diff
@@ -110,6 +118,18 @@ def __call__(self, value, allow_cache=True):
                 self._create_load_const(graph_outputs[graph_outputs_key].index)
             )
             output.append(create_instruction("BINARY_SUBSCR"))
+            if isinstance(value, UnspecializedPrimitiveVariable):
```
jansel, line 121: Are we baking in the assumption here that all UnspecializedPrimitiveVariable's are numpy?  Perhaps we should have a specific class for numpy.

torchdynamo/variables/builtin.py
```diff
@@ -248,12 +263,22 @@ def _call_min_max(self, tx, a, b):
...
                     tx, [a], kwargs
                 )
             else:
                 fn = {max: torch.maximum, min: torch.minimum}[self.fn]
-                return variables.TorchVariable(fn).call_function(tx, [a, b], {})
+                result = variables.TorchVariable(fn).call_function(tx, [a, b], {})
+
+            if (
+                self.tensor_variable_class(a, b)
+                == variables.UnspecializedPrimitiveVariable
```
jansel, line 275: `is` is usually used for type comparisons (or `issubclass`)

torchdynamo/codegen.py
```diff
@@ -289,8 +320,37 @@ def make_call_generated_code(self, fn_name: str) -> List[Instruction]:
         self.extend_output(self.load_function_name(fn_name))
 
         graphargs = self.tx.output.graphargs
+
+        # to handle random calls
+        if len(self.tx.random_calls) > 0:
+            self.tx.output.install_global(
+                "_generate_random_values",
```
jansel, line 327: If we have multiple functions using random this name could collide.

torchdynamo/codegen.py
```diff
@@ -289,8 +320,37 @@ def make_call_generated_code(self, fn_name: str) -> List[Instruction]:
...
+                ]
+            )
+
         for arg in graphargs:
-            self.extend_output(arg.load(self))
+            if arg.is_unspecialized:
+                self.extend_output(
+                    [
+                        self.create_load_global("torch", add=True),
+                        self.create_load_attr("tensor"),
```
jansel, line 343: Make sure to test that this works on GPU.

jansel (changes requested):

torchdynamo/codegen.py
```diff
@@ -110,6 +118,18 @@ def __call__(self, value, allow_cache=True):
...
+                output.extend(
+                    [
+                        self.create_load_attr("numpy"),
+                        create_instruction("CALL_FUNCTION", 0),
+                        self.create_load_attr("reshape"),
+                        self._create_load_const(1),
+                        create_instruction("CALL_FUNCTION", 1),
+                        self._create_load_const(0),
+                        create_instruction("BINARY_SUBSCR"),
+                    ]
```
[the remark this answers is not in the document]
yanboliang: Agree that converting back and forth is not guaranteed to preserve types, actually [this](https://github.com/pytorch/torchdynamo/pull/392#discussion_r896211756) is one of the discrepancies I hit.
But I still don't think adding a property is the best solution, as we put the unspecialized into ```fx.Graph```, so it may go through many ops from input to output. If we add any property, we have to define all these rules explicitly. E.g, ```np.int32 / int = np.float64```. I think we wrap them as 1-element tensor is to leverage the facility the tensor ops provided. Correct me if I'm wrong.

torchdynamo/variables/tensor.py
```diff
@@ -172,9 +173,14 @@ def create(cls, tx, proxy, example_value=None, nnmodule=None, **options):
...
-                tx=tx, proxy=proxy, example_value=torch.tensor(example_value)
-            )
+            if isinstance(example_value, np.number):
+                return UnspecializedNumpyVariable.create(
+                    tx=tx, proxy=proxy, example_value=torch.tensor(example_value)
+                )
+            else:
+                return UnspecializedPythonVariable.create(
+                    tx=tx, proxy=proxy, example_value=torch.tensor(example_value)
+                )
```
yanboliang, line 183: I think this change comes from https://github.com/pytorch/torchdynamo/pull/417. I can understand we are trying to put ```item``` op into the graph. I revisited and found it's better we add another unspecialized type(which is not reflected in current PR yet), as here we don't need to unwrap the value after graph, but other places need. Adding a property to identify the difference is another way to fix it, but we have to maintain the propagation rules explicitly. Any thoughts? @jansel @voznesenskym

jansel: We could do something during shape/type propagation in `TensorVariable.create` using example values.

torchdynamo/codegen.py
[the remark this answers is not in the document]
yanboliang: Update: it seems there is a way to do value/type propagation. If unspec ops tensor, it will return tensor and don't need the original type to unwrap; if unspec ops unspec/const, it will return unspec and need the original type for unwrap. So I made change in ```BuiltinVariable.call_function``` to calculate the value during compiling and store the result as ```raw_value``` in unspec variable. The downside is max/min is a bit ugly as it needs special handling. Let me know if this looks good to you. Thank you!

tests/test_misc.py
```diff
@@ -1546,9 +1547,9 @@ def fn(x, y, z):
                 "z": z,
                 "a": np_y.sum(),
                 "b": xy,
-                "c": np_y[0][0] / 68,
+                "c": np_x[0][0] / 68,
```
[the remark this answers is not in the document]
yanboliang: This has been resolved with the most recent commit.

torchdynamo/codegen.py
```diff
@@ -110,6 +130,28 @@ def __call__(self, value, allow_cache=True):
...
             output.append(create_instruction("BINARY_SUBSCR"))
+
+            if isinstance(value, UnspecializedNumpyVariable):
+                unspec_var = self.tx.output.new_var("unspec")
+                output.extend(
+                    [
+                        self.create_load_attr("item"),
+                        create_instruction("CALL_FUNCTION", 0),
+                        self.create_store(unspec_var),
+                        self.create_load_global("type", add=True),
```
jansel, line 141: The user could use the name `type` in their program

torchdynamo/codegen.py
```diff
@@ -110,6 +130,28 @@ def __call__(self, value, allow_cache=True):
...
+                        self.create_load_const(value.raw_value),
```
jansel, line 142: since `type(value.raw_value)` is a constant, perhaps just compute it at compile time?

torchdynamo/codegen.py
```diff
@@ -289,8 +328,36 @@ def make_call_generated_code(self, fn_name: str) -> List[Instruction]:
...
+        # to handle random calls
+        if len(self.tx.random_calls) > 0:
+            rand_fn_name = unique_id("__gen_rand_values")
+            rand_fn = torchdynamo.disable(_get_gen_rand_values_fn(self.tx.random_calls))
+            self.tx.output.install_global(rand_fn_name, rand_fn)
+            self.extend_output(self.load_function_name(rand_fn_name))
+            self.extend_output(
+                [
+                    create_instruction("CALL_FUNCTION", 0),
+                    self.create_store(self.tx.output.random_values_var),
```
yanboliang, line 341: Currently it doesn't support graph break between multiple random calls, I'm thinking to store random values as global to support, which can be one of the follow-ups if it's necessary.

jansel (approved):

PR #393, opened 2022-06-14
Added option use_bmm to enable triton codegen for bmm

vesuppi:
Added an initial triton implementation of bmm and an option `use_bmm` to enable that in torchinductor. Modified `ir.py` and `wrapper.py` similarly as before.

### DIFF
--- a/torchinductor/triton_ops/batched_matmul.py
+++ b/torchinductor/triton_ops/batched_matmul.py
@@ -1,0 +1,254 @@
+import torch
+import triton
+import triton.language as tl
+from triton.ops.matmul_perf_model import early_config_prune
+from triton.ops.matmul_perf_model import estimate_matmul_time
+
+
+def init_to_zero(name):
+    return lambda nargs: nargs[name].zero_()
+
+
+def get_configs_io_bound():
+    configs = []
+    for num_stages in [2, 3, 4, 5, 6]:
+        for block_m in [16, 32]:
+            for block_k in [32, 64]:
+                for block_n in [32, 64, 128, 256]:
+                    num_warps = 2 if block_n <= 64 else 4
+                    configs.append(
+                        triton.Config(
+                            {
+                                "BLOCK_M": block_m,
+                                "BLOCK_N": block_n,
+                                "BLOCK_K": block_k,
+                                "SPLIT_K": 1,
+                            },
+                            num_stages=num_stages,
+                            num_warps=num_warps,
+                        )
+                    )
+                    # split_k
+                    for split_k in [2, 4, 8, 16]:
+                        configs.append(
+                            triton.Config(
+                                {
+                                    "BLOCK_M": block_m,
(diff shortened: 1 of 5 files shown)


torchinductor/triton_ops/batched_matmul.py
```diff
@@ -0,0 +1,261 @@
...
+    rm = pid_m * BLOCK_M + tl.arange(0, BLOCK_M)
+    rn = pid_n * BLOCK_N + tl.arange(0, BLOCK_N)
+    C = C + (rm[:, None] * stride_cm + rn[None, :] * stride_cn)
+    C += bid * M * N
+    mask = (rm < M)[:, None] & (rn < N)[None, :]
+    # handles write-back with reduction-splitting
+    if SPLIT_K == 1:
+        tl.store(C, acc, mask=mask)
+    else:
+        tl.atomic_add(C, acc, mask=mask)
```
jansel, line 225: Does this require the output buffer to be zero initialized?  (TorchInductor doesn't zero out new buffers.)

vesuppi: no, it does not require the output buffer to be zero initialized. The accumulator is zero initialized, and the final result is written to the output buffer.

jansel: My understanding is 
```
tl.atomic_add(C, acc, mask=mask)
```

Would implement:
```
C[i] += acc[i]
```
So `acc` gets added to the existing data in `C`

Does atomic_add do something else?

vesuppi: That makes sense, thank you for pointing that out. I think `atomic_add` looks like it's doing that. And in that case, this code seems potentially buggy. I guess for now because `SPLIT_K` is always 1 so the `atomic_add` path is never executed, but when it does, it may require the out buffer to be zero initialized. Emm I'll try some examples to check that (passing a non zero out buffer). Actually the `matmul` implementation is doing the same thing too.

pyjhzwh: https://github.com/openai/triton/blob/master/python/test/unit/operators/test_matmul.py#L76-L83
You may use this to force the SPILT_K to be > 1

vesuppi: Great thanks

> https://github.com/openai/triton/blob/master/python/test/unit/operators/test_matmul.py#L76-L83 You may use this to force the SPILT_K to be > 1

Great thanks for the reference! I'll see how to use that.

anijain2305:
@vesuppi Can you rebase this and ping when its ready for review?

vesuppi:
The updated batched matmul has disabled the perf model, which improved the performance a bit. The SPLIT_K thing is not an issue after a closer look. Request for a review? Thanks! @anijain2305

jansel (approved):

PR #394, opened 2022-06-14
Ensure dtype instances are not mapped to TorchVariable()

### LINKED ISSUE #354: isinstance test on dtype causes graph break
```
diff --git a/tests/test_repros.py b/tests/test_repros.py
index 4d590f7..3ede478 100755
--- a/tests/test_repros.py
+++ b/tests/test_repros.py
@@ -1246,3 +1246,11 @@ class ReproTests(torchdynamo.testing.TestCase):
 
         self.assertTrue(same(ref0, res0))
         self.assertTrue(same(ref1, res1))
+
+    def test_isinstance_dtype(self):
+        @torchdynamo.optimize("eager", nopython=True)
+        def fn(x):
+            isinstance(torch.bfloat16, torch.dtype)
+            return x
+
+        fn(torch.randn(3))
```

you get

```
  File "/raid/ezyang/torchdynamo/torchdynamo/convert_frame.py", line 278, in _convert_frame_assert                               
    code = transform_code_object(frame.f_code, transform)                                                                        
  File "/raid/ezyang/torchdynamo/torchdynamo/bytecode_transformation.py", line 338, in transform_code_object                     
    transformations(instructions, code_options)                                                                                  
  File "/raid/ezyang/torchdynamo/torchdynamo/convert_frame.py", line 254, in transform                                           
    tracer.run()                                                                                                                 
  File "/raid/ezyang/torchdynamo/torchdynamo/symbolic_convert.py", line 306, in run                                              
    and self.step()                                                                                                              
  File "/raid/ezyang/torchdynamo/torchdynamo/symbolic_convert.py", line 284, in step                                             
    getattr(self, inst.opname)(inst)                                                                                             
  File "/raid/ezyang/torchdynamo/torchdynamo/symbolic_convert.py", line 145, in wrapper                                          
    return inner_fn(self, inst)                                                                                                  
  File "/raid/ezyang/torchdynamo/torchdynamo/symbolic_convert.py", line 619, in CALL_FUNCTION                                    
    self.call_function(fn, args, {})                                                                                             
  File "/raid/ezyang/torchdynamo/torchdynamo/symbolic_convert.py", line 220, in call_function                                    
    self.push(fn.call_function(self, args, kwargs))                                                                              
  File "/raid/ezyang/torchdynamo/torchdynamo/variables/builtin.py", line 220, in call_function                                   
    result = handler(tx, *args, **kwargs)                                                                                        
  File "/raid/ezyang/torchdynamo/torchdynamo/variables/builtin.py", line 354, in call_isinstance                                 
    arg_type = arg.python_type()                                                                                                 
  File "/raid/ezyang/torchdynamo/torchdynamo/variables/torch.py", line 67, in python_type                                        
    return super().python_type()                                                                                                 
  File "/raid/ezyang/torchdynamo/torchdynamo/variables/base.py", line 137, in python_type                                        
    raise NotImplementedError(f"{self} has no type")                                                                             
NotImplementedError: TorchVariable() has no type 
```

mlazos:
Any `torch.dtype` instance currently gets mapped to a `TorchVariable()` when encountered by TorchDynamo which can cause dynamo to introduce graph breaks when calling `isinstance` on the dtype instance. This is because `TorchVariable` is really only intended to be either modules or methods in `torch.*`, and attempting to access the python type of any other object (not modules or methods) will throw an exception. This change disallows all `torch.dtype` instances so they will now map to `ConstantVariable()`, which is able to access its python type, and is also more "correct" because the `torch.dtype` instances are in fact constants.

Closes #354


anijain2305:
Are you certain that they are getting converted to `ConstantVariable`? From reading `builder.py`, it seems they should get converted to `UserDefinedObject`.

mlazos:
`dtype` has its own little carve out here: https://github.com/pytorch/torchdynamo/blob/main/torchdynamo/variables/builder.py#:~:text=)-,elif%20ConstantVariable.is_literal(value)%20or%20istype(,),-elif%20isinstance(

jansel (approved):

PR #396, opened 2022-06-14
Suppress warnings during the pre-checks on a frame

anijain2305:
Few PyTorch tests test for number of warnings. This resolves at least 13 failures there.

### DIFF
--- a/torchdynamo/convert_frame.py
+++ b/torchdynamo/convert_frame.py
@@ -7,6 +7,7 @@ import sys
 import traceback
 import types
 import typing
+import warnings
 from typing import Callable
 
 import torch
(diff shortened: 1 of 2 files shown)


torchdynamo/convert_frame.py
```diff
@@ -127,6 +128,7 @@ def _fn(*args, **kwargs):
 
 def has_tensor_in_frame(frame):
     """Check if the frame has torch.* related bits"""
+    warnings.filterwarnings("ignore", category=UserWarning, module="torch.distributed")
```
jansel, line 131: Should we do this in the patch pytorch function you added?

anijain2305:
@jansel The warnings are ignored within the scope. So, we can't switch it like others done in patch_torch.

But, I have created a new decorator and created a new class to capture these PyTorch related patches at one place.

jansel (approved):

PR #397, opened 2022-06-15
Enable tf32 in torchbench.py

anijain2305:
torchbench repo itself might be a better place to do this.


anijain2305:
cc @Chillee @xuzhao9

jansel (approved):

PR #400, opened 2022-06-15
[inductor] Fix convolution output stride order

jansel:


torchinductor/ir.py
```diff
@@ -1994,18 +2027,20 @@ def create(
...
-        # memory layout order of output depends on x
-        size_hint = V.graph.sizevars.size_hint
-        x_stride = [size_hint(i) for i in x.get_stride()]
-        order = sorted(range(len(x_stride)), key=x_stride.__getitem__)
+        if any(k != 1 for k in output_size[-len(stride) :]) and in_channels_stride == 1:
+            # channels last format
+            order = [0] + list(reversed(range(1, len(kernel_size) + 1)))
+            if len(order) < len(output_size):
+                # add batch dim if it exists
+                order = [len(order)] + order
```
pyjhzwh, line 2035: Could we get fill order directly for channel last format
```
order = [1] + list(reversed(range(2, len(kernel_size)+2))
if len(order) < len(output_size):
  # add batch dim if it exists
  order = order + [0]
```
The `FlexbileLayout.fill_ordered(order)`?
Do we need `FlexibleLayout.stride_ordered()`?

pyjhzwh (approved):

torchinductor/ir.py
jansel: Yeah, we could.  IMO this is a bit more readable, though I suppose it is preference.

PR #401, opened 2022-06-15
[inductor] Fix typo preventing some fusions

jansel:


ngimel (approved):
Should there be a test that we are getting a single fusion group where expected?

jansel:
Yes, though it might be tricky to write such a test.

PR #410, opened 2022-06-16
Triton conv reformat

pyjhzwh:
- `#fmt off` to reduce lines in code

- And reformat to serve as a better template to use.

Remove 
```
class _conv: 
   def _call():
       ...
   def forward():
       _conv.call()
_conv = conv.forward
```
to single function
```
def conv():
    ...
```


jansel:
Let's keep using black formatting.  It helps reduce merge conflicts and make code more readable.

If you want to make things more compact.  Perhaps introduce a helper function to construct the `triton.Config` objects.

For the args, if you want to group them I'd suggest adding comments.

pyjhzwh:
That makes sense. Close this PR

PR #414, opened 2022-06-16
Shape guard and isinstance fixes

mlazos:
Fixes two issues:
* isinstance would return false when called on a tensor and tuple containing the tensor type
* shape guards are sometimes be evaluated before tensor guards, causing guards to crash if the function involved was later called on a type without the shape attribute, this adds an additional type guard when adding shape guards

See the attached issues for more details.

### DIFF
--- a/torchdynamo/convert_frame.py
+++ b/torchdynamo/convert_frame.py
@@ -311,7 +311,9 @@ def convert_frame_assert(compiler_fn: Callable, one_graph=True):
                 print()
             assert output.guards is not None
             CleanupManager.instance[code] = output.cleanups
-            return GuardedCode(code, output.guards, frame.f_locals, frame.f_globals)
+            return GuardedCode(
+                code, sorted(output.guards), frame.f_locals, frame.f_globals
+            )
         except (Unsupported, TorchRuntimeError):
             debug_print("WONT CONVERT")
             raise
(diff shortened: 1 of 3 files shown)


torchdynamo/convert_frame.py
```diff
@@ -311,7 +311,9 @@ def debug_print(prefix):
                 print()
             assert output.guards is not None
             CleanupManager.instance[code] = output.cleanups
-            return GuardedCode(code, output.guards, frame.f_locals, frame.f_globals)
+            return GuardedCode(
+                code, sorted(output.guards), frame.f_locals, frame.f_globals
```
mlazos, line 315: Note: remove this

torchdynamo/variables/tensor.py
```diff
@@ -269,6 +277,15 @@ def var_getattr(self, tx, name):
         if name == "__class__":
             return TorchVariable(self.python_type(), **options)
 
+        # Add a guard for type matching, these guards are checked before tensor guards
+        # In some cases, a <tensor>.<attr> guard can be evaluated first, and break if
+        # <tensor> is later changed another type
+        try:
+            if result is not None:
```
jansel, line 284: perhaps we need `and self.source is not None`

mlazos: but isn't control flow with exceptions "pythonic"? jk done.

jansel (approved):

PR #417, opened 2022-06-17
Config driven support for torch.Tensor .item() 

voznesenskym:
Repro that fails before PR, passes after:

```
from typing import List

import torch
import torch.fx

import torchdynamo


    gm.graph.print_tabular()
    return gm.forward


class MyMod(torch.nn.Module):
    def forward(self, x):
        z = torch.max(x)
        return z.int().item()


with torchdynamo.optimize(my_compiler, nopython=True):
    x = torch.tensor([[0.6763, 0.7445, -2.2369]])
    model = MyMod()
    y = model(x)
    print(y)
```


voznesenskym:
Note: This specializes on an item's value, now, which we may not want. Perhaps a "call_function" may be better.

torchdynamo/variables/tensor.py
```diff
@@ -313,6 +321,8 @@ def call_method(
             constant_result = ConstantVariable(self.is_contiguous, **options)
         elif name == "is_complex" and self.is_complex is not None:
             constant_result = ConstantVariable(self.is_complex, **options)
+        elif name == "item" and self.item is not None:
+            constant_result = ConstantVariable(self.item, **options)
```
jansel, line 325: Ths will just burn in the compile-time value?

jansel (changes requested):
I'm not sure if it is possible to handle `Tensor.item()` properly.

We could put it in the graph pretty easily with a `call_method` node, but no backend compilers support python `float()` types afaik.

I'd put this in the same category as `tensor.numpy()`, it is something we should allow but always break the graph on.

voznesenskym:
> I'm not sure if it is possible to handle `Tensor.item()` properly.
> 
> 
> 
> We could put it in the graph pretty easily with a `call_method` node, but no backend compilers support python `float()` types afaik.
> 
> 
> 
> I'd put this in the same category as `tensor.numpy()`, it is something we should allow but always break the graph on.

Understood. There's a use case for item on integer tensors in executorch I was playing around with. Thoughts on supporting item for integer tensors?

jansel:
@suo what do you think here?

My first instinct is if people have a model with `.numpy()` or `.item()`, that model can't be exported out of python, and to export it user need to rewrite it to use PyTorch tensors rather than python types.

Does the mobile backend support python float/int types?

suo:
the thing coming out of `item()` is a scalar type, right? Yeah, the mobile backend supports those. And we need to be able to capture `.item()` calls in some way. 

At some point @Chillee was talking about rewriting `item()` to something that returns a Tensor instead of a scalar so it was easier to thread through aotautograd. @Chillee any thoughts about that.

voznesenskym:
> At some point @Chillee was talking about rewriting `item()` to something that returns a Tensor instead of a scalar so it was easier to thread through aotautograd. @Chillee any thoughts about that. (this could be explicit as @jansel suggests)

I think I might be misunderstanding this. Let me reason through this so you can tell me where I am off. So, as I understand it, `.item()` on a tensor returning a tensor would essentially be a no op, right? It would go from:

```
>>> x = torch.tensor([1.0])
>>> x.item()
1.0
```
to:
```
>>> x = torch.tensor([1.0])
>>> x.item()
tensor([1.])
```

At which point, that's just the same as `x`.

eellison:
Should we consider introducing an `UnspecializedScalar` in the same way that we have relaxed other specializations ?

suo:
> At which point, that's just the same as `x`.

yeah ok lol my proposal doesn't make much sense, + you generally want to use `.item` to slice or something which won't work unless we add overloads for 0d tensors.

voznesenskym:
> Should we consider introducing an `UnspecializedScalar` in the same way that we have relaxed other specializations ?

I like that idea - Looking around at Unspecialized* - I also see we have UnspecializedPrimitiveVariable:

```

class UnspecializedPrimitiveVariable(TensorVariable):
    """
    This is a 1-element tensor represents unspecialized primitive type,
    e.g, int, float, np.int, np.float, etc
    """

    pass

```

This is actually the only valid call for `item` (it throws a runtime error if the tensor > 1 element) 

Today creation is gated via:

`        elif is_numpy_int_type(value) or is_numpy_float_type(value):`

eellison:
> yeah ok lol my proposal doesn't make much sense, + you generally want to use .item to slice or something which won't work unless we add overloads for 0d tensors.

Note that I think for pytorch purposes a 0-dim tensor and a python scalar are equivalent afaik

jansel:
cc @yanboliang re UnspecializedPrimitiveVariable

I think using this is fine.  If we could remove the call to `.item()` that could help other backends too, though capturing it for mobile seems fine too.

voznesenskym:
I played w/ `UnspecializedPrimitiveVariable`, but it didn't feel like it added much. Going with a method_call + constant variable seems fine, but I could very well be missing something w/r/t other backends. Would folks prefer a UnspecializedPrimitiveVariable over a ConstantVariable here?

jansel:
Constant variables are constants.  Whatever value it has at compile time is burned in.  Try running your example with a different input.

voznesenskym:
```
opcode         name    target                                                  args      kwargs
-------------  ------  ------------------------------------------------------  --------  --------
placeholder    x       x                                                       ()        {}
call_function  max_1   <built-in method max of type object at 0x7f0b9612fce0>  (x,)      {}
call_method    int_1   int                                                     (max_1,)  {}
call_method    item    item                                                    (int_1,)  {}
output         output  output                                                  ((),)     {}
```

> Constant variables are constants. Whatever value it has at compile time is burned in. Try running your example with a different input.

I have a test with this in it, and it seems to change correctly, however, if that turns into a value guard + recompile on change, that's not good. I'll make it a UnspecializedPrimitiveVariable

jansel:
Note the graph has no outputs, so we aren't actually even using the result of `.item()`.

Make sure you alternate input has the same shape.  (Or enable `config.dynamic_shapes`.)

voznesenskym:
> Note the graph has no outputs, so we aren't actually even using the result of `.item()`.
> 
> Make sure you alternate input has the same shape. (Or enable `config.dynamic_shapes`.)

Makes sense! Changed PR based on feedback, new graph looks like it has output ;)

image

jansel:
This looks good now, but I think we should put this behind a flag.   AFAIK most backends don't support `.item()`.

tests/test_misc.py
```diff
@@ -407,7 +407,27 @@ def fn(a):
             self, fn=fn, nargs=1, expected_ops=5, expected_ops_dynamic=8
         )
 
-    def test_tensor_item(self):
+    def test_tensor_item_capture(self):
```
jansel, line 410:
```suggestion
    @patch.object(torchdynamo.config, "capture_scalar_outputs", True)
    def test_tensor_item_capture(self):
```

Might make the tests more readable.

jansel (approved):

PR #422, opened 2022-06-20
[Inductor] Triton template for conv

pyjhzwh:
Summary:
In scheduler.py, `TemplateKernels` specify which Kernels will be code gen by templates. During `codegen_extern_call`, if we find that the node type is in `TemplateKernels` list, we will call `template_codegen()` to code gen triton kernels with existing templates. 
Right now we only support code-gen Convolution triton kernel (fusion not supported yet)

Added files:
- torchinductor/codegen/template.py - `TritonTemplateKernel` will load templates and generate the Triton kernel using templates and rendered args; `template_codegen()` will iterate the scheduler runable nodes to find fusable pointwise kernels to be fused with the current `TritonTemplateKernel`
- torchinductor/codegen/triton_conv.j2 - a jinja2 templates for conv

Modified files:
- torchinductor/ir.py - for ir.Convolution, prepared ordered dict for the triton kernel args and other args for precomputation on host (e.g. delta_x_ptr)
- torchinductor/scheduler.py - allocate buffer for ir.Convolution if it is in `TritonTemplateKernel` list even if it is the subclass of `ExternKernelAlloc` because triton kernel could not allocate buffer itself.

TritonTemplateKernel in detail:
`__init__()`: load templates
`codegen_body()`: render templates with args
`codegen_kernel()`: call `codegen_body()`
`map_args()`: get ordered dict from ir.Convolution
`precompute()`: code gen precompute functions, like `_conv._delta_x_ptr()` 
`call_kernel()`: code gen call to the triton kernel


torchinductor/codegen/template.py
```diff
@@ -0,0 +1,162 @@
+from jinja2 import Environment
```
jansel, line 1: I'd suggest naming the file triton_template

torchinductor/codegen/template.py
```diff
@@ -0,0 +1,162 @@
...
+template_dict = {ir.Convolution: "triton_conv"}
+
+
+class TritonTemplateKernel(TritonKernel):
+    def __init__(self, node: ir.ExternKernel, *groups):
+        super(TritonTemplateKernel, self).__init__(*groups)
+        self.node = node
+        template_name = template_dict[type(node)]
+        env = Environment(
+            loader=FileSystemLoader("torchinductor/codegen"),
```
jansel, line 18: make path relative to `__file__` (our working directory could be different)

torchinductor/codegen/triton_conv.j2
```diff
@@ -0,0 +1,247 @@
...
+            & (off_x_w < IN_W)
+        )[:, None] & (off_x_crs < CRS)[None, :]
+        mask_w = (off_x_crs < CRS)[:, None] & (off_w_k < KERNEL_N)[None, :]
+        # ------ prefetch ------
+        # ------ load x ------
+        matrix_x = tl.load(x_ptrs, mask=mask_x)
+        # ------ load w ------
+        matrix_w = tl.load(w_ptrs, mask=mask_w)
+
+        {{ pointwise_computation }}
```
jansel, line 199: Should this be outside the loop?  I'm assuming the pointwise we will run will operate on `(BLOCK_M, BLOCK_N)`?

torchinductor/scheduler.py
```diff
@@ -118,15 +124,24 @@ def get_priority(self):
 
 
 class ExternKernelSchedulerNode(BaseSchedulerNode):
+    def __init__(self, scheduler: "Scheduler", node: ir.ExternKernel, group_fn):
+        super().__init__(scheduler, node)
+        if type(node) in TemplateKernels and node.get_device == "cuda":
```
jansel, line 129: You can use https://github.com/pytorch/torchdynamo/blob/e5e2366821c0e26a2791508d0359598e1d6e030e/torchinductor/ir.py#L112

torchinductor/scheduler.py
```diff
@@ -118,15 +124,24 @@ def get_priority(self):
...
     def can_remove_buffer(self, **kwargs):
         return False
 
     def run(self, codegen_extern_call):
         self.allocate()
         self.scheduler.run_count += 1
         self.scheduler.pending_buffer_names.add(self.get_name())
-        self.scheduler.kernels.append(self.node)
-        codegen_extern_call(self.node)
+        if type(self.node) not in TemplateKernels:
```
jansel, line 141: Perhaps refactor this check into a `should_use_template` helper

torchinductor/scheduler.py
```diff
@@ -596,12 +614,16 @@ def flush(self):
...
         self.flush()
-        node.codegen(V.graph.wrapper_code)
-        self.barrier()
-        self.maybe_free_buffers()
+        if type(node) in TemplateKernels and node.get_device() == "cuda":
+            template_codegen(self, scheduler_node)
+        else:
+            node.codegen(V.graph.wrapper_code)
+            self.barrier()
+            self.maybe_free_buffers()
```
jansel, line 626: I think we still might want the self.barrier()/self.maybe_free_buffers().

jansel (changes requested):

torchinductor/scheduler.py
pyjhzwh: https://github.com/pytorch/torchdynamo/pull/422/files#diff-eee4a893649579118ce96233528ba599634b8e0d8a04c1c516fb9c45008c2cd6R161-R162
At the end of template_codegen(), there will be `self.barrier()/self.maybe_free_buffers()`. I put them in template_codegen() in case we need to iter_fixed_point again later

jansel: Makes sense, I missed that

torchinductor/codegen/triton_conv.j2
pyjhzwh: My bad, it should be outside the loop

pyjhzwh:
Should I use jinja2? Or I could just store the template in a python string. The jinja2 template won't pass test-inductor on github CI because jinja is not installed.

jansel:
> Should I use jinja2? Or I could just store the template in a python string. The jinja2 template won't pass test-inductor on github CI because jinja is not installed.

You can install extra packages on CI by adding them here:
https://github.com/pytorch/torchdynamo/blob/e5e2366821c0e26a2791508d0359598e1d6e030e/Makefile#L45

pyjhzwh:
Should I create a separate PR for epilog fusion of triton conv? Or should I continue submitting on this PR?

jansel:
Smaller PRs might be easier to review

PR #423, opened 2022-06-20
Removing generation field from the patched nn.Module

anijain2305:
Asking for comments here.

Current patching adds a new field `generation` into the Module object. `torch.jit.save` and `torch.jit.load` causes problems because of this.


voznesenskym:
> torch.jit.save and torch.jit.load causes problems because of this.

What problems are we seeing?

jansel (approved):

PR #424, opened 2022-06-20
Extend python_key_normalize with support for PythonTensor class override and a post trace hook

voznesenskym:


torchdynamo/optimizations/python_key.py
```diff
@@ -121,13 +140,63 @@ def unpack(x):
 
         return tuple(unpack(x) for x in out)
 
+    class PythonKeyTracer(Tracer):
```
jansel, line 143: How is this different from `functorch._src.python_key.PythonKeyTracer`?

Can we just extend PythonKeyTracer to take a `python_tensor_cls` arg?

torchdynamo/optimizations/python_key.py
```diff
@@ -44,7 +49,13 @@ def debug_node(n: Node):
     return f"{n.op} {target} {n.args} {n.kwargs}"
 
 
-def python_key_normalize(gm: torch.fx.GraphModule, example_inputs, decompositions={}):
+def python_key_normalize(
+    gm: torch.fx.GraphModule,
+    example_inputs,
+    decompositions={},
+    PythonTensorOverrideClass=None,
```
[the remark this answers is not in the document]
voznesenskym: sure thing

torchdynamo/optimizations/python_key.py
jansel: Let's avoid copy any pasting it.

torchdynamo/optimizations/python_key.py
```diff
@@ -121,13 +139,32 @@ def unpack(x):
 
         return tuple(unpack(x) for x in out)
 
+    class DynamoPythonKeyTracer(PythonKeyTracer):
+        def __init__(self):
+            super().__init__()
+
+        def _module_getattr(self, attr, attr_val, parameter_proxy_cache):
```
voznesenskym, line 146: @jansel I was incorrect earlier - we don't need to copy, but we do need to provide our own `_module_getattr` over the `Tracer` one. We could implement it like this, OR change the main torch.fx `Tracer` to accept a `python_tensor_cls`. Do you have a preference?

voznesenskym: I was going to say we change behavior too much to call super, then I realized that kind of violates Liskov, so let me actually add this to the base Tracer, it feels more correct as such.

jansel (approved):

PR #426, opened 2022-06-20
Add Fake Tensor Propagation

eellison:
Run and propagate the fx graph with FakeTensors, which are a TensorSubclass with `__torch_dispatch__` defined that do device propagation built on top of meta tensors. 

A few notes:
- FakeTensors/MetaTensors do consistent alias/storage tracking. For that, there is a cache for converting non-fake tensors to fake tensors so that the fake tensors will have the same alias/tensor id. Every Fake Tensor contains a corresponding FakeTensorMode which stores the cache from real->fake tensors and an op with fake tensor inputs can only be run if the inputs all have the same mode. 

- Meta Tensors don't have complete op coverage, but we get pretty far in FakeTensors by falling back to CPU and running tensors there. However, there are a few major categories of ops that don't support Meta, which is sparse, complex, and quantized. Complex should be pretty easily doable soon / is in the works but the others are farther away. 
I added a `fake_tensor_propagation` config option for this case although in the end state it would be great rid of this (cc @ezyang ). 

- Dynamic Shape operators will now cause a graph break, because we are not able to infer the output shape with meta inputs. The only case in the code base I observed this was `repeat_interleave`. There are various ways forward of supporting this (specializing on the data-dependent input / generating symbolic output shapes) but we can leave that to future dynamic shape work.
 
Depends on https://github.com/pytorch/pytorch/pull/79741, https://github.com/pytorch/pytorch/pull/79809 for tests to pass but I think those are landing shortly/accepted.


torchdynamo/variables/tensor.py
```diff
@@ -60,40 +65,94 @@ def visit(n: torch.fx.Node):
...
+            return fake_mode.from_tensor(e)
+        else:
+            return e
+
+    @staticmethod
+    def wrap_fake_exception(fn):
+        try:
+            return fn()
+        except UnsupportedFakeTensorException as e:
+            raise Exception(
```
jansel, line 92: Perhaps add an exception type to `torchdynamo.exc.*`

jansel:
If you want to change the pytorch version used in CI you can change it here:
https://github.com/jansel/torchdynamo/blob/3a0ef2eb9b2d22b44997cb79c5df4eebf6dccd46/.github/workflows/test-py38.yml#L18

jansel:
Overall this looks good to me.  One question:

Will this break PyTorch 1.12 support?  Would it be easy to fall back to non-fake tensors for PyTorch 1.12?

I'm wondering if we should do a branch cut for 1.12 support before merging this.

eellison:
> Will this break PyTorch 1.12 support? Would it be easy to fall back to non-fake tensors for PyTorch 1.12?

It would break as of now, but I can add a check on the pytorch version and fallback to non-Fake Tensors if it isn't recent enough. Should I add that?

jansel:
Yes, thanks!

jansel (approved):

torchdynamo/variables/tensor.py
```diff
@@ -65,45 +77,101 @@ def visit(n: torch.fx.Node):
...
+        try:
+            return fn()
+        except UnsupportedFakeTensorException as e:
+            raise FakeTensorError(
+                f"Unsupported: {e.reason} with fake tensor propagation. "
+                "Run with config.fake_tensor_propagation=False"
+            ) from e
+
     @classmethod
     def create(cls, tx, proxy, example_value=None, nnmodule=None, **options):
```
ezyang, line 110: We should factor this out into an FX pass that can live in PyTorch fx passes proper

PR #428, opened 2022-06-20
Break graph on torch.Storage types

anijain2305:
Alternative way might be to support this. But, I am not sure who benefits from it. So, currently putting graph break.

### DIFF
--- a/torchdynamo/allowed_functions.py
+++ b/torchdynamo/allowed_functions.py
@@ -99,7 +99,7 @@ def _disallowed_function_ids():
     storage = [
         obj
         for obj in torch.__dict__.values()
-        if isinstance(obj, type(torch.FloatStorage))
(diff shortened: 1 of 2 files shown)


jansel (approved):

torchdynamo/allowed_functions.py
```diff
@@ -96,6 +96,12 @@ def _disallowed_function_ids():
         obj for obj in torch.__dict__.values() if isinstance(obj, type(torch.float32))
     ]
     remove += dtypes
+    storage = [
+        obj
+        for obj in torch.__dict__.values()
+        if isinstance(obj, type(torch.FloatStorage))
```
jansel, line 102: I'd consider `torch.storage._LegacyStorageMeta`, but I don't feel strongly there.  We aren't guaranteed there will always be a metaclass here.

PR #429, opened 2022-06-20
Disable TorchDynamo on frames created by fx symbolic tracing

anijain2305:
@jansel Does it make sense? Basically, symbolic tracing creates new frames and Dynamo starts to modify them and then given them back to Fx tracer. Therefore, I am just disabling it.


jansel (approved):
Makes sense to me.

PR #430, opened 2022-06-21
Fix for disabling triton templates

jansel:
Some tests started failing after
https://github.com/pytorch/torchdynamo/pull/422

### DIFF
--- a/torchinductor/scheduler.py
+++ b/torchinductor/scheduler.py
@@ -147,7 +147,7 @@ class ExternKernelSchedulerNode(BaseSchedulerNode):
         self.allocate()
         self.scheduler.run_count += 1
         self.scheduler.pending_buffer_names.add(self.get_name())
-        if should_use_template(self.node):
+        if not should_use_template(self.node):
             # TemplateKernelSchedulerNode will be added to scheduler in template_codegen
             self.scheduler.kernels.append(self.node)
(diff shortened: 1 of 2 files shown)


torchinductor/scheduler.py
```diff
@@ -142,7 +147,7 @@ def run(self, codegen_extern_call):
         self.allocate()
         self.scheduler.run_count += 1
         self.scheduler.pending_buffer_names.add(self.get_name())
-        if type(self.node) not in TemplateKernels:
+        if should_use_template(self.node):
```
pyjhzwh, line 150: `if not should_use_template(self.node)`

torchinductor/scheduler.py
```diff
@@ -19,15 +19,20 @@
 from .sizevars import SimplifyIndexing
 from .virtualized import V
 
-TemplateKernels = [ir.Convolution]
+template_kernels = [ir.Convolution]
```
pyjhzwh, line 22: I think we could set template_kernels = [] to disable TritonTemplateKernel

jansel: This is defined before the configuration might get changed, so we will need to move it out of global scope for that to work.

pyjhzwh: oh I missed the config. Makes sense

PR #431, opened 2022-06-21
[inductor] Add some prims

jansel:


Chillee:
haha I think this is because of https://github.com/pytorch/pytorch/pull/79413

Chillee (approved):
LGTM

PR #433, opened 2022-06-21
[inductor] Add heuristic to set num_warps

jansel:


ngimel (approved):

ngimel:
Are there any benchmarks to go with this?

jansel:
My (very limited) benchmarking showed warps don't matter.

You suggested this heuristic in a prior comment, so I mostly just implemented your suggestion.

ngimel:
YEah, I can believe that for pointwise it wouldn't matter, for reduction there should be more effect (especially with bigger RBLOCKs for bigger reductions)

PR #434, opened 2022-06-21
[inductor] Support scatter operations

jansel:
Generated code for index_put:
```
@triton.jit
def kernel1(in_ptr0, in_ptr1, out_ptr0, ks0, ks1, ks2, xnumel, XBLOCK : tl.constexpr):
    xmask = xindex < xnumel
    x0_next = xindex // (ks2*(ks1*ks1))
    x1 = x0_next
    x2 = xindex
    x0 = xindex % (ks2*(ks1*ks1))
    tmp0 = tl.load(in_ptr0 + x1, xmask)
    tmp1 = tl.load(in_ptr1 + x2, xmask).to(tl.float32)
    tl.store(out_ptr0 + x0 + (ks2*tmp0*(ks1*ks1)) + tl.zeros([XBLOCK], tl.int32), tmp1, xmask)
```

And CPU:
```
                       const long* __restrict__ in_ptr1,
                       const float* __restrict__ in_ptr2,
                       float* __restrict__ out_ptr0,
                       const long ks0,
                       const long ks1,
                       const long ks2,
                       const long ks3)
{
    #pragma omp parallel num_threads(16)
    {
        #pragma omp for nowait
        for(long i0=0; i0<ks1*ks2*(ks0*ks0); ++i0)
        {
            auto tmp0 = in_ptr0[i0];
            out_ptr0[i0] = tmp0;
        }
        #pragma omp barrier
        #pragma omp for nowait
        for(long i0=0; i0<ks3; ++i0)
        {
            #pragma GCC ivdep
            for(long i1=0; i1<ks2*(ks0*ks0); ++i1)
            {
                auto tmp0 = in_ptr1[i0];
                auto tmp1 = in_ptr2[i1 + (i0*ks2*(ks0*ks0))];
                out_ptr0[i1 + (ks2*tmp0*(ks0*ks0))] = tmp1;
            }
        }
    }
}
```

### DIFF
--- a/torchinductor/codegen/triton.py
+++ b/torchinductor/codegen/triton.py
@@ -2,6 +2,7 @@ import contextlib
 import dataclasses
 import functools
 import itertools
+import logging
 import math
 from typing import Dict
 from typing import List
@@ -19,6 +20,8 @@ from .common import ExprPrinter
 from .common import IndentedBuffer
 from .common import Kernel
 from .common import OpOverrides
+
+log = logging.getLogger(__name__)
 
 
 class TritonPrinter(ExprPrinter):
@@ -491,10 +494,15 @@ class TritonKernel(Kernel):
 
         return tmp
 
-    def store(self, name, index, value):
+    def store(self, name, index, value, mode=None):
         var = self.args.output(name)
         index, mask = self.indexing(index, value)
-        line = f"tl.store({var} + {index}, {value}, {mask})"
+        if mode is None:
+            line = f"tl.store({var} + {index}, {value}, {mask})"
+        elif mode == "atomic_add":
+            line = f"tl.atomic_add({var} + {index}, {value}, {mask})"
+        else:
+            raise NotImplementedError(f"store mode={mode}")
         self.stores.writeline(line)
         if not self.inside_reduction:
             self.outside_loop_vars.add(value)
--- a/torchinductor/decomposition.py
+++ b/torchinductor/decomposition.py
@@ -1,3 +1,4 @@
+import logging
 import math
 import numbers
 
@@ -9,6 +10,7 @@ from torch._decomp import get_decompositions
 from torchinductor import config
 
 aten = torch.ops.aten
+log = logging.getLogger(__name__)
 
 decompositions = get_decompositions(
     [
@@ -27,19 +29,21 @@ decompositions = get_decompositions(
         aten.glu_backward,
         aten.grid_sampler_2d,
         aten.hardsigmoid,
+        aten.hardsigmoid_backward,
         aten.hardswish,
+        aten.hardswish_backward,
         aten.hardtanh,
+        aten.hardtanh_backward,
         aten.l1_loss,
         aten.leaky_relu,
         aten.leaky_relu_backward,
         aten._log_softmax_backward_data,
         aten.logsumexp.default,
-        aten.masked_fill_,
         aten.max_pool2d_with_indices_backward,
         aten.mse_loss,
+        aten.narrow,
         aten.native_batch_norm,
         aten.native_batch_norm_backward,
-        aten.native_dropout_backward,
         aten.native_group_norm,
         aten.native_layer_norm,
         aten.native_layer_norm_backward,
@@ -49,6 +53,8 @@ decompositions = get_decompositions(
         aten.select_backward,
         aten.select_scatter,
         aten.sigmoid_backward,
+        aten.silu_backward,
+        aten.slice_backward,
         aten._softmax_backward_data,
         aten.stack,
         aten.tanh_backward,
@@ -61,6 +67,9 @@ decompositions.update(aot_autograd_decompositions)
 
 
 def register_decomposition(ops):
+    for op in [ops] if callable(ops) else ops:
+        if op in decompositions:
+            log.warning(f"duplicate decomp: {ops}")
     return functorch._src.decompositions.register_decomposition(ops, decompositions)
 
 
@@ -263,3 +272,13 @@ def baddbmm(self, batch1, batch2, beta=1, alpha=1):
     if not isinstance(beta, numbers.Number) or beta != 1:
         self = self * beta
     return self + result
+
+
+@register_decomposition([aten.index_put])
+def index_put(self, indices, values, accumulate=False):
+    return torch.index_put_(self.clone(), indices, values, accumulate)
+
+
+@register_decomposition([aten.narrow])
+def narrow(self, dim, start, length):
+    return aten.slice(self, dim, start, start + length)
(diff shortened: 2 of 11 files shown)


torchinductor/lowering.py
```diff
@@ -846,6 +964,46 @@ def _new_constant(
...
+    size, stride, *, dtype=None, layout=None, device=None, pin_memory=None
+):
+    assert isinstance(size, (list, type))
+    assert isinstance(stride, (list, type))
+    assert not pin_memory
+    assert not layout or layout == torch.strided
+    dtype = decode_dtype(dtype) or torch.get_default_dtype()
+    device = device or torch.tensor(0.0).device
+    pointwise = _full(fill_value=0, device=device, dtype=dtype, size=size)
+    if tuple(ir.FlexibleLayout.contiguous_strides(size)) == tuple(stride):
```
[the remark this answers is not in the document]
jansel: No it is a symbolic comparison.  Sympy `__eq__` is only true if it is the same formula.

torchinductor/decomposition.py
```diff
@@ -263,3 +272,13 @@ def baddbmm(self, batch1, batch2, beta=1, alpha=1):
     if not isinstance(beta, numbers.Number) or beta != 1:
         self = self * beta
     return self + result
+
+
+@register_decomposition([aten.index_put])
+def index_put(self, indices, values, accumulate=False):
+    return torch.index_put_(self.clone(), indices, values, accumulate)
```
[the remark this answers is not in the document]
jansel: Yes, this turns out-of-place into inplace.  I think in general decomps don't preserve stides.

torchinductor/ir.py
```diff
@@ -1739,6 +1766,65 @@ def should_allocate(self):
...
+            MutationLayout(x),
+            self.unwrap_storage([x, *indices, values]),
+            [accumulate],
+        )
+        self.name = V.graph.register_buffer(self)
+
+
+class InplaceBernoulliFallback(ExternKernel):
+    """
+    This needs to be a custom class to handle mutation properly
```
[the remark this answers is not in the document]
jansel: Need to override `get_mutation_names` below.  The normal fallback kernel doesn't handle that.

torchinductor/lowering.py
```diff
@@ -693,6 +721,96 @@ def inner_fn(index):
...
+    return Pointwise.create(
+        device=x.get_device(),
+        dtype=x.get_dtype(),
+        inner_fn=inner_fn,
+        ranges=list(x.get_size()),
+    )
+
+
+@register_lowering(aten.slice_scatter)
+def slice_scatter(x, src, dim=0, start=None, end=None, step=1):
```
ngimel, line 750: THat's probably not the most efficient decomposition (an efficient one would take a slice of the tensor with the new strides and sizes, and copy the `src` tensor to it, similar to aten implementation https://fburl.com/code/2ne9ady3. Since you already have a slice_ lowering, that should be something like clone + slice_ + copy?

torchinductor/lowering.py
```diff
@@ -846,6 +964,46 @@ def _new_constant(
...
+    assert isinstance(stride, (list, type))
+    assert not pin_memory
+    assert not layout or layout == torch.strided
+    dtype = decode_dtype(dtype) or torch.get_default_dtype()
+    device = device or torch.tensor(0.0).device
+    pointwise = _full(fill_value=0, device=device, dtype=dtype, size=size)
+    if tuple(ir.FlexibleLayout.contiguous_strides(size)) == tuple(stride):
+        # fast path, no need to realize it
+        return pointwise
+    print(size, stride)
```
ngimel, line 981: stray print?

torchinductor/lowering.py
```diff
@@ -846,6 +964,46 @@ def _new_constant(
...
+def empty_strided(
+    size, stride, *, dtype=None, layout=None, device=None, pin_memory=None
+):
+    assert isinstance(size, (list, type))
+    assert isinstance(stride, (list, type))
+    assert not pin_memory
+    assert not layout or layout == torch.strided
+    dtype = decode_dtype(dtype) or torch.get_default_dtype()
+    device = device or torch.tensor(0.0).device
+    pointwise = _full(fill_value=0, device=device, dtype=dtype, size=size)
```
ngimel, line 977: filling with 0's is typically extra overhead that's not needed for `empty`

torchinductor/lowering.py
jansel: Ah interesting.  You think two kernels would be better.  I had assumed it would be memory bound, so this would be faster.

torchinductor/codegen/triton.py
```diff
@@ -492,10 +495,15 @@ def load(self, name: str, index: sympy.Expr, upcast: bool = False):
...
 
-    def store(self, name, index, value):
+    def store(self, name, index, value, mode=None):
         var = self.args.output(name)
         index, mask = self.indexing(index, value)
-        line = f"tl.store({var} + {index}, {value}, {mask})"
+        if mode is None:
+            line = f"tl.store({var} + {index}, {value}, {mask})"
+        elif mode == "atomic_add":
+            line = f"tl.atomic_add({var} + {index}, {value}, {mask})"
```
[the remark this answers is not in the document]
jansel: Is there a way to do that in Triton?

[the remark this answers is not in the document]
ngimel: Probably not yet, given how their fp16 atomic add doesn't work at all

torchinductor/lowering.py
ngimel: I don't know for sure, your way indeed might turn out faster. 
Where is slice_scatter coming from? Are you running functionalization pass? (Because I thought it's used in functionalization only)

torchinductor/lowering.py
jansel: Agreed.  Inductor doesn't support uninitialized memory, though I think this gets optimized away anyway.

torchinductor/lowering.py
jansel: This is coming from AOT Autograd, which I believe does functionalization.  We may want to turn this off, as I worry this will de-optimize things.

ngimel (approved):

torchinductor/lowering.py
jansel: I looked a bit more into this.  In this case `x` is an unrealized1 zero tensor, so I think this decomp might be fine since `x_loader(...)` just gets mapped to `0`.

I think we might need to take a pass over the decomps with performance in mind though.  The decomps are de-inplacing things, then in codegen we are re-inplacing things -- sometimes creating extra copies in output code.

PR #450, opened 2022-06-23
Add support for crossentropy

voznesenskym:


torchdynamo/variables/torch.py
```diff
@@ -244,6 +246,20 @@ def fake_softmax(input):
 
         return variables.LambdaVariable(fake_softmax, **options)
 
+    def _call_cross_entropy_loss(self, tx, args, kwargs, options):
```
jansel, line 249: Perhpas assert args/kwargs are empty?  Are there possible args we need to pass to `torch.nn.functional.cross_entropy`

torchdynamo/variables/torch.py
```diff
@@ -244,6 +246,20 @@ def fake_softmax(input):
...
+        def fake_cel(input, target):
```
jansel, line 250: cel?

voznesenskym: cross entropy loss, but I'll rename it, the characters don't come out of my paycheck ;)

torchdynamo/variables/torch.py
voznesenskym: Yes, great call, thanks for forcing me to take a closer look - the args are:

functional: `input, target, weight=None, size_average=None, ignore_index=- 100, reduce=None, reduction='mean', label_smoothing=0.0`

non functional ctor: `weight=None, size_average=None, ignore_index=- 100, reduce=None, reduction='mean', label_smoothing=0.0`

non functional `loss` call: `input, target, optional_output`

so it really should be shuffled, much the same way that you take `dim` from softmax.

torchdynamo/variables/torch.py
```diff
@@ -244,6 +246,79 @@ def fake_softmax(input):
...
+        reduce_arg = (
+            args[3] if args else kwargs.get("reduce", variables.ConstantVariable(None))
+        )
+        reduction = (
+            args[4]
+            if args
+            else kwargs.get("reduction", variables.ConstantVariable(None))
+        )
+        label_smoothing = (
+            args[5]
```
jansel, line 280: This will throw an exception if you only give one arg.  I'd suggest:

```
def normalize_args(weight=ConstantVariable(None), size_average=ConstantVariable(None), ignore_index=ConstantVariable(-100), reduce=ConstantVariable(None), reduction=ConstantVariable('mean'), label_smoothing=ConstantVariable(0.0)):
   return weight, size_average, ignore_index, reduce, reduction, label_smoothing
   
weight, size_average, ignore_index, reduce, reduction, label_smoothing = normalize_args(*args, **kwargs) 
```

jansel (approved):

PR #454, opened 2022-06-23
[Request for feedback / discussion] [WIP] Do we want to support custom __get_attribute__?

voznesenskym:


jansel (changes requested):
The correct way to support getattribute in the general case is to turn it into a function call an inline it (on evey attribute access).

For things like HF PretrainedConfig, we could have a list of classes where we ignore the getattribute and pretend it doesn't exist.  This works for special cases, but not in general.

Issue #328, opened 2022-06-06
Use TorchDynamo to capture optimizers


jansel:
We should explore using TorchDynamo to capture/optimize optimizer code.

I don't anticipate any major issues here, but we should make sure it plays nice with AOT Autograd.

mlazos:
linking issue related to performance of this: https://github.com/pytorch/torchdynamo/issues/1803

PR #464, opened 2022-06-27
[WIP] Adding AMP support in benchmark infra

anijain2305:
Few things to notice

* Tests are failing accuracy left and right. Need to debug further why thats the case.
* AMP and optimizers are discussed in conjunction due to loss scaling. However, we have not yet tried optimizers with Dynamo. We will pick up the work after https://github.com/pytorch/torchdynamo/issues/328

cc @ngimel @Chillee

### DIFF
--- a/benchmarks/common.py
+++ b/benchmarks/common.py
@@ -572,12 +572,72 @@ def cast_to_fp32(model, inputs):
     return model, inputs
 
 
+class DummyGradScaler:
+    def scale(self, loss):
+        return loss
+
+
 class BenchmarkRunner:
+    def __init__(self):
+        self.use_amp = False
+        self.grad_scaler = DummyGradScaler()
+        self.autocast = NullContext
+        self._args = None
+
+    def setup_amp(self):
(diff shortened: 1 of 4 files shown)


jansel (approved):

model_benchmark/common.py
```diff
@@ -572,12 +572,72 @@ def cast_to_fp32(model, inputs):
...
+        self.use_amp = False
+        self.grad_scaler = DummyGradScaler()
+        self.autocast = NullContext
+        self._args = None
+
+    def setup_amp(self):
+        if self.args.amp and self.args.training:
+            assert self.args.devices == ["cuda"], "AMP is supported only for CUDA"
+            self.grad_scaler = torch.cuda.amp.GradScaler()
+            self.autocast = torch.cuda.amp.autocast
```
ngimel, line 591: Nit: you can use generic autocast manager with `device="cuda"` https://pytorch.org/docs/stable/amp.html#torch.autocast

PR #465, opened 2022-06-27
Add support for named_params and named_modules

voznesenskym:


torchdynamo/variables/nn_module.py
```diff
@@ -226,8 +226,10 @@ def call_method(
         ):
             return ConstantVariable(True, **options)
 
-        if not all(x.is_python_constant() for x in itertools.chain(args, kwargs)):
-            raise unimplemented(f"non-const NNModule method {name}")
+        for x in itertools.chain(args, kwargs):
```
jansel, line 229:
```suggestion
        for x in itertools.chain(args, kwargs.values()):
```
This looks like a bug in the original code to me.

voznesenskym: Yea, I wasn't sure if we did that intentionally, so I took a safer approach. Happy to make this change higher up in the flow.

jansel (approved):

PR #469, opened 2022-06-27
Add (experimental) support for exporting a graph and guards

voznesenskym:
Sample usage:

```
import torch
import torch.fx

import torchdynamo

inp = torch.randn(2, 2)
l = torch.nn.Linear(2, 2)
net = torch.nn.Sequential(l, l)
x = inp


def foo(k):
    global x
    for idx, m in enumerate(net.named_modules()):
        x = x + m[1](inp) * k


out = torchdynamo.export_experimental(foo, 3)
print(x)
print(out)
```

Sample output:

```
(GraphModule(
  (self_0): Linear(in_features=2, out_features=2, bias=True)
), {Guard(name='self[0]', source=<GuardSource.LOCAL_NN_MODULE: 2>, create_fn=<function GuardBuilder.NN_MODULE at 0x7f0e6af60f70>, guard_type=None, code=None, obj_id=None, guarded_class=None), Guard(name='self', source=<GuardSource.LOCAL: 0>, create_fn=<function GuardBuilder.LIST_LENGTH at 0x7f0e6af621f0>, guard_type='LIST_LENGTH', code='___check_type_id(self, 94335645731376) and len(self) == 2', obj_id='___check_type_id(self, 94335645731376) and len(self) == 2', guarded_class=<class 'torch.nn.modules.container.Sequential'>), Guard(name='self[1]', source=<GuardSource.LOCAL_NN_MODULE: 2>, create_fn=<function GuardBuilder.NN_MODULE at 0x7f0e6af60f70>, guard_type=None, code=None, obj_id=None, guarded_class=None), Guard(name='self', source=<GuardSource.LOCAL: 0>, create_fn=<function GuardBuilder.NN_MODULE at 0x7f0e6af60f70>, guard_type='ID_MATCH', code='___check_obj_id(self, 139702325633424)', obj_id='___check_obj_id(self, 139702325633424)', guarded_class=<class 'torch.nn.modules.container.Sequential'>), Guard(name='input', source=<GuardSource.LOCAL: 0>, create_fn=<function GuardBuilder.TENSOR_MATCH at 0x7f0e6af625e0>, guard_type=None, code=None, obj_id=None, guarded_class=None)})
```


torchdynamo/eval_frame.py
```diff
@@ -228,18 +228,47 @@ def toy_example(a, b):
...
+        assert graph is None, "whole graph export entails exactly one graph"
+        graph = gm
+        return gm.forward
+
+    backend_ctx_ctor = null_context
+
+    with optimize_assert(
+        dynamo_normalization_capturing_compiler, backend_ctx_ctor, guard_export_print
+    ):
+        f(*args)
```
voznesenskym, line 259: self review / todo: add better support for args, kwargs

torchdynamo/convert_frame.py
```diff
@@ -193,7 +193,7 @@ def has_tensor(obj):
     return False
 
 
-def convert_frame_assert(compiler_fn: Callable, one_graph=True):
+def convert_frame_assert(compiler_fn: Callable, guard_export_fn=None, one_graph=True):
     """Fully convert a frame into an FX graph"""
     compiler_fn = wrap_compiler_fn(compiler_fn)
```
voznesenskym, line 198: we may need a wrap_guard_export_fn

torchdynamo/guards.py
```diff
@@ -321,6 +427,8 @@ def TENSOR_MATCH(self, guard: Guard):
             self.tensor_check_names.append(self.arg_ref(guard))
             self.tensor_check_examples.append(self.get(guard.name))
 
+        # TODO(voz): Add support for tensor match export
```
wconstab, line 430: haha, yea that is the annoying one :)

torchdynamo/guards.py
```diff
@@ -210,32 +250,45 @@ def EQUALS_MATCH(self, guard: Guard):
...
         if istype(val, float) and math.isnan(val):
-            self.code.append(f"___check_type_id({ref}, {self.id_ref(type(val))})")
-            self.code.append(f"__math_isnan({ref})")
+            code = f"___check_type_id({ref}, {self.id_ref(t)})"
+            export_code = code
+            self.code.append(code)
+            code = f"__math_isnan({ref})"
+            self.code.append(code)
+            export_code = f"{export_code} and {code}"
+            guard.set_export_info("EQUALS_MATCH", t, export_code, obj_id)
```
wconstab, line 269: at least for now this seems fine, but i wonder if it makes sense to accumulate 'guard objects' which have their guard expression built in into self, rather than putting the guard code into self.code separately from the guard object itself.

wconstab:
overall i think this is a cool idea.

from a UX perspective i think there is an alternative approach we should at least consider pros/cons of.

instead of tracing and returning the guards, what if the user constructs proxy objects to feed into the export (as args/kwargs to f)?  The proxy tensors could be assumed to have no specialization when constructed, and would trigger errors inside dynamo whenever guarding on them would occur.  Users could opt-into specific specializations by setting those properties on their proxies.

e.g.

x = Proxy()
x.size(1) = 2
trace

This would trace successfully as long as specialization was limited to x.size(1).

TBH This approach sounds like a lot more work, has a tricky UI to nail down, and might not be better.  I had a feeling that people doing export would generally know up front what they want to specialize on and having a way to express that might be nicer than making them sift through the returned guards and see if they are appropriate.

torchdynamo/guards.py
voznesenskym: Yeah, I think a very reasonable refactor is to stop passing code around as str, and instead make guard objects. This was a sort of path-of-least-resistance for a proposal in today's code.

suo:
@wconstab the two approaches are not mutually exclusive, right? I'm betting we'll need to have both kinds of functionality. For example, using specialized input proxies will not remove all guards (for example, nn module attribute guards e.g. `if self.training`), so we will likely still need guard reporting

wconstab:
> @wconstab the two approaches are not mutually exclusive, right? I'm betting we'll need to have both kinds of functionality. For example, using specialized input proxies will not remove all guards (for example, nn module attribute guards e.g. `if self.training`), so we will likely still need guard reporting

yea, i think the way in this PR is an easier way to start out and also may be useful long term.

torchdynamo/guards.py
```diff
@@ -97,7 +97,18 @@ def __lt__(self, other):
         return self.sort_key() < other.sort_key()
 
     def __str__(self):
-        return f"{self.source.name.lower()} {repr(self.name)} {self.create_fn.__name__} EXPORT_INFO: {{'guard_type': {self.guard_type}, 'code': {self.code}, 'obj_id': {self.obj_id}, 'guarded_class': {self.guarded_class}}}"
+        s = f"{self.source.name.lower()} {repr(self.name)} {self.create_fn.__name__}"
```
wconstab, line 100: i'm not up to speed on this but why is it only possible to print the details such as self.guard_type under export?  It looks like it was previously possible all the time, unless you are now just preventing trying to print when printing is not possible

voznesenskym: No, this has nothing to do with possible/impossible printing, but rather with keeping things as they were.

Before this pr, the print statement was:

```
f"{self.source.name.lower()} {repr(self.name)} {self.create_fn.__name__}"
```

When `export_guards` is disabled, I'd like to stay exactly as it is before this PR. If we were to not have this config check, this info would be null as the other parts of the PR responsible for setting this data will not have been called, so it would be useless anyway.

torchdynamo/guards.py
```diff
@@ -273,34 +336,83 @@ def PYMODULE_MATCH(self, guard: Guard):
...
-        self.code.append(f"len({ref}) == {len(value)}")
+        t = type(value)
+
+        code_type_check = f"___check_type_id({ref}, {self.id_ref(t)})"
+        self.code.append(code_type_check)
+        code_len_check = f"len({ref}) == {len(value)}"
+        self.code.append(code_len_check)
+
+        guard.set_export_info(
+            "LIST_LENGTH", t, f"{code_type_check} and {code_len_check}", id(value)
```
jansel, line 347: This pattern of calling `set_export_info` is a bit ugly as now there is a lot of redundancy in guard setup that could be easy for things to get out of sync.

I wonder if we could have a helper method that both updates `self.code` and the export info.  We might be able to automate things like the guard name.  

It could be more useful to provide pointers to objects rather than id(value).

jansel (changes requested):

torchdynamo/convert_frame.py
```diff
@@ -306,13 +306,32 @@ def debug_print(prefix):
                 print("MODIFIED BYTECODE")
                 # print(dis.Bytecode(code).info())
                 print(dis.Bytecode(code).dis())
+
+                # When export guard is enabled, more detailed guards can be printed after creating GuardedCode
+                if not config.export_guards:
```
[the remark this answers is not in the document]
voznesenskym: GH just misses the debug line in the context, L304: `if config.debug:`

torchdynamo/convert_frame.py
```diff
@@ -306,13 +306,32 @@ def debug_print(prefix):
...
+                    print("\nGUARDS:")
+                    for guard in sorted(output.guards):
+                        print(" -", str(guard))
+                print()
```
voznesenskym, line 315: Self nit: dedup this with the print below for readability

torchdynamo/eval_frame.py
```diff
@@ -228,18 +228,47 @@ def toy_example(a, b):
...
-        return optimize_assert(backend, backend_ctx_ctor)
+        return optimize_assert(backend, backend_ctx_ctor, guard_export_fn=None)
     return _optimize_catch_errors(
-        convert_frame.convert_frame(backend), backend_ctx_ctor
+        convert_frame.convert_frame(backend, guard_export_fn=None), backend_ctx_ctor
     )
 
 
-def optimize_assert(backend, backend_ctx_ctor=null_context):
+def export_experimental(f, *args):
```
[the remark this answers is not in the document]
voznesenskym: Agreed, good call.

torchdynamo/eval_frame.py
```diff
@@ -228,18 +228,73 @@ def toy_example(a, b):
...
-        return optimize_assert(backend, backend_ctx_ctor)
+        return optimize_assert(backend, backend_ctx_ctor, guard_export_fn=None)
     return _optimize_catch_errors(
-        convert_frame.convert_frame(backend), backend_ctx_ctor
+        convert_frame.convert_frame(backend, guard_export_fn=None), backend_ctx_ctor
     )
 
 
-def optimize_assert(backend, backend_ctx_ctor=null_context):
+def export(f, *args, **kwargs):
```
voznesenskym, line 237: From @jansel, copying as it got lost due to renames:

I'd like to see some checks in this function that the inputs/outputs match.

One issue is if the user gives a function that doesn't take a flat list of tensors:

If the user does:

foo(a, [b,c], {"d": d})
The graph will just take foo_graph(a,b,c,d).

You have the same issue for return values. We could use pytrees to flatten things. But at the very lease we should check for this case an throw and exception.

-----

Let me know if this is what you had in mind, I was also not sure how to verify return type consistency here.

torchdynamo/guards.py
```diff
@@ -309,17 +401,60 @@ def GRAD_MODE(self, guard: Guard):
...
+            # We accumulate tensor names, then do a single install of `___check_tensors`.
+            # See _guards.cpp and TensorGuard for more information.
+            # TODO(voz): Add tensor matching code to export
+            # Note: this is a bit of a special case, and so does not use _produce_guard_code
+            guard.set_export_info(
+                "TENSOR_MATCH", type(value), None, weakref.ref(value), id(value)
+            )
+
+    # A util that appends guarded code, or, in the case of export, adds data onto guards
+    def _produce_guard_code(self, guard, code_list, provided_guarded_object=None):
```
voznesenskym, line 429: self nit, reading this name, it feels a little off, as it doesn't produce guard code so much as add it to existing structures... might rename

torchdynamo/config.py
```diff
@@ -93,3 +93,5 @@
 # Not all backends support scalars. Some calls on torch.Tensor (like .item()) return a scalar type.
 # When this flag is set to False, we introduce a graph break instead of capturing.
 capture_scalar_outputs = False
+
+export_guards = True
```
jansel, line 97: This config controls two things:
1) Prints
2) If `guard_export_fn!=None` will trigger an assert error.

These seem like seperate things.  For 1, we should either tie it to debug/trace or give it a print related name.

For 2, could we just look at `guard_export_fn is None` to decide what to do?  I'm not sure why we need both a config and a kwarg.

torchdynamo/guards.py
```diff
@@ -91,7 +100,19 @@ def __lt__(self, other):
         return self.sort_key() < other.sort_key()
 
     def __str__(self):
-        return f"{self.source.name.lower()} {repr(self.name)} {self.create_fn.__name__}"
+        s = f"{self.source.name.lower()} {repr(self.name)} {self.create_fn.__name__}"
+        if config.export_guards is True:
```
jansel, line 104:
```suggestion
        if config.export_guards:
```

This in this case we could just check if the actual values are `None` and may not need a config.

torchdynamo/guards.py
```diff
@@ -76,6 +78,13 @@ class Guard:
     source: GuardSource
     create_fn: Callable
 
+    # Export only. These values are written to at time of guard check_fn creation.
+    guard_types: Optional[Set[str]] = None
+    code_list: Optional[List[str]] = None
+    obj_weakref: Optional[Any] = None
+    obj_id: Optional[int] = None
```
jansel, line 85: seems redundant with obj_weakref

torchdynamo/guards.py
```diff
@@ -76,6 +78,13 @@ class Guard:
...
+    guarded_class: Optional[type] = None
```
jansel, line 86: We may want to weakref the class too to prevent memory leaks.

jansel (changes requested):

jansel (approved):

Issue #447, opened 2022-06-22
[torchinductor] Add a counter to record the number of generated kernel


desertfire:
This should help to check our fusion unit tests do fuse kernels as we expected.

PR #476, opened 2022-06-28
[inductor] Add a metric to count the number of generated kernels

desertfire:
Summary: This can be used to prevent a regression on our fusion result.


desertfire:
https://github.com/pytorch/torchdynamo/issues/447

jansel (approved):

Issue #477, opened 2022-06-29
create skip leaks CacheEntry


ezyang:
```
    DEBUG_TRACE("create skip %s", name(frame));
    Py_DECREF(result);
    set_extra(frame->f_code, SKIP_CODE);
    // Re-enable custom behavior
    eval_frame_callback_set(callback);
    return eval_frame_default(tstate, frame, throw_flag);
```

set extra doesn't deallocate the previous extra entry of f code. If it was a cache entry, that means it will be leaked.

PR #478, opened 2022-06-29
Don't leak cache entry on skip

ezyang:
* (to be filled)

Fixes https://github.com/pytorch/torchdynamo/issues/477

### DIFF
--- a/torchdynamo/_eval_frame.c
+++ b/torchdynamo/_eval_frame.c
@@ -369,6 +369,7 @@ static PyObject *_custom_eval_frame(PyThreadState *tstate, PyFrameObject *frame,
   } else {
     DEBUG_TRACE("create skip %s", name(frame));
     Py_DECREF(result);
+    destroy_cache_entry(extra);
     set_extra(frame->f_code, SKIP_CODE);
     // Re-enable custom behavior
     eval_frame_callback_set(callback);


ezyang:
I wrote a test for this and it seems like dynamo is still holding onto the code too long

voznesenskym:
> I wrote a test for this and it seems like dynamo is still holding onto the code too long

Do you mind sharing the test?

ezyang:
Pushed. I also pushed one of the refcount fixes. According to the refcount there's still one more ref to code and guards after the first fix.

ezyang:
I'm putting this down for the evening, maybe someone else can find the other two smoking refcounts

ezyang:
oh gfd, gc.collect() makes it die now, that means there is a ref cycle

```
diff --git a/tests/test_recompile_ux.py b/tests/test_recompile_ux.py
index 81fc871..d4b9389 100644
--- a/tests/test_recompile_ux.py
+++ b/tests/test_recompile_ux.py
@@ -50,6 +50,8 @@ class RecompileUxTests(torchdynamo.testing.TestCase):
             with torchdynamo.optimize(compiler):
                 model(x, i)
         print("trace")
+        import gc
+        gc.collect()
 
         self.assertTrue(triggered)
        
```

jansel (approved):
LGTM overall.  Fix the tests/lints before landing.

I think @anijain2305 used a tool to track down ref cycles in another case.

jansel:
Hrm, we could move `__compiled_fn0` out of global scope and into a closure.

A simpler fix might be to add a weakref hook to trigger `CleanupManager` based on a different object being deleted, since `CleanupManager` should break this cycle.  Perhaps the GuardedCode object pointed to by the cache?

ezyang:
Hmm my analysis may not be quite right, since the gc is sufficient to cause the code to die, that means code does die!

ezyang:
OK, I nailed it. The problem is that GuardedCode is holding on a reference to code, whose death is supposed to trigger the GC of the global. Once I remove this reference the cycle is disposed of promptly.

Issue #332, opened 2022-06-07
TorchInductor failing TorchBench models tracker


jansel:
This is meta issues to track multiple issues with TorchInductor inference models in TorchBench.

Failing with `RuntimeError: CUDA: Error- invalid value`:
- [ ] dlrm
- [ ] vision_maskrcnn

Failing with `LLVM ERROR: Broken function found, compilation aborted!` and missing `torch.randn`:
- [ ] hf_Reformer

Crash with no error message:
- [ ] Super_SloMo
- [ ] fastNLP_Bert

To reproduce:
1) Use [my branch](https://github.com/jansel/torchdynamo/tree/inductor) (working on getting everything merged, so hopefully this won't be needed soon)
2) `./torchbench.py -dcuda --inductor --float16 --no-skip -k <name>`

ngimel:
dlrm, vision_maskrcnn - save constants as cpu tensors and send them to triton kernel. 
fastNLP_Bert:
codegen error: 
```
def kernel0(in_ptr0, in_ptr1, out_ptr0, out_ptr1, out_ptr2, out_ptr3, out_ptr4, xnumel: tl.constexpr, rnumel: tl.constexpr, XBLOCK : tl.constexpr, RBLOCK : tl.constexpr):
    xoffset = tl.program_id(0) * XBLOCK
    xindex = xoffset + tl.reshape(tl.arange(0, XBLOCK), [XBLOCK, 1])
    xmask = xindex < xnumel
    rbase = tl.reshape(tl.arange(0, RBLOCK), [1, RBLOCK])
    _tmp3 = tl.zeros([XBLOCK, RBLOCK], tl.int1) + 0
    tmp6 = tl.load(in_ptr1 + tmp0, None)
                             ^
 ```
 tmp0 is not defined

jansel:
#405 works around the wrong device issue, though both of those models still fail for other reasons.

PR #446, opened 2022-06-22
[inductor] Improve fusing of tiled + untiled

jansel:
@pyjhzwh this is the improvement I promised.  It should make fusing conv_template+pointwise easier.

Before:
```
@triton.jit
def kernel0(in_ptr0, in_ptr1, out_ptr0, ks0, xnumel, ynumel, XBLOCK : tl.constexpr, YBLOCK : tl.constexpr):
    xoffset = tl.program_id(0) * XBLOCK
    xindex = xoffset + tl.reshape(tl.arange(0, XBLOCK), [XBLOCK, 1])
    xmask = xindex < xnumel
    yoffset = tl.program_id(1) * YBLOCK
    yindex = yoffset + tl.reshape(tl.arange(0, YBLOCK), [1, YBLOCK])
    ymask = yindex < ynumel
    x0 = xindex
    y1 = yindex
    tmp0 = tl.load(in_ptr0 + x0, xmask).to(tl.float32)
    tmp1 = tl.load(in_ptr1 + y1, ymask).to(tl.float32)
    tmp2 = tmp0 + tmp1
    tl.store(out_ptr0 + y1 + (ks0*x0), tmp2, xmask & ymask)

@triton.jit
def kernel1(in_ptr0, out_ptr0, ks0, xnumel, XBLOCK : tl.constexpr):
    xoffset = tl.program_id(0) * XBLOCK
    xindex = xoffset + tl.reshape(tl.arange(0, XBLOCK), [XBLOCK])
    xmask = xindex < xnumel
    x0 = xindex
    tmp0 = tl.load(in_ptr0 + x0, xmask).to(tl.float32)
    tmp1 = 1
    tmp2 = tmp0 + tmp1
    tl.store(out_ptr0 + x0, tmp2, xmask)
```

After
```
@triton.jit
def kernel0(in_ptr0, in_ptr1, in_ptr2, out_ptr0, out_ptr1, ks0, xnumel, ynumel, XBLOCK : tl.constexpr, YBLOCK : tl.constexpr):
    xoffset = tl.program_id(0) * XBLOCK
    xindex = xoffset + tl.reshape(tl.arange(0, XBLOCK), [XBLOCK, 1])
    xmask = xindex < xnumel
    yoffset = tl.program_id(1) * YBLOCK
    yindex = yoffset + tl.reshape(tl.arange(0, YBLOCK), [1, YBLOCK])
    ymask = yindex < ynumel
    x0 = xindex
    y1 = yindex
[...]

PR #484, opened 2022-06-29
Compute multilayer reductions in higher precision

eellison:
triton will automatically compute reductions in fp32 if reducing over fp16/bf16
within the kernel. keep the intermediate in fp32 so as to keep the whole reduction
in fp32 and not reduce precision by breaking up the kernel into multiple layers

Fix Super_SloMo in https://github.com/pytorch/torchdynamo/issues/332

### DIFF
--- a/tests/test_torchinductor.py
+++ b/tests/test_torchinductor.py
@@ -453,6 +453,14 @@ class CommonTemplate:
 
         self.common(fn, (torch.randn(1, 17, 8, 9),))
 
+    def test_multilayer_low_prec(self):
+        def fn(a):
+            return torch.mean(a)
+
+        self.common(
+            fn, ((torch.rand((10, 3, 352, 352), dtype=torch.float16, device="cuda"),))
+        )
+
     def test_min_max_reduction(self):
         def fn(a, b):
             return ((a + b).max(), (a + b).min(), torch.amax(a + 1, keepdim=True))
--- a/torchinductor/ir.py
+++ b/torchinductor/ir.py
@@ -403,9 +403,15 @@ class Reduction(Loops):
             else:
                 return body()
 
+        # triton will automatically compute reductions in fp32 if reducing over fp16/bf16
+        # within the kernel. keep the intermediate in fp32 so as to keep the whole reduction
+        # in fp32 and not reduce precision by breaking up the kernel into multiple layers
+        intermediate_dtype = (
+            dtype if dtype not in (torch.float16, torch.bfloat16) else torch.float
+        )
         intermediate = Reduction.create(
             device,
-            dtype,
+            intermediate_dtype,
             wrapper_fn,
             [*ranges, split],
             [block_size],


jansel (approved):

torchinductor/lowering.py
```diff
@@ -1598,11 +1598,15 @@ def mean(x, axis=None, keepdim=False, *, dtype=None):
         x = to_dtype(x, dtype)
     size = x.get_size()
     axis = _validate_reduction_axis(x, axis)
+    # compute in higher-precision until end of mean lowering
+    output_dtype = x.get_dtype()
+    if output_dtype in (torch.float16, torch.bfloat16):
+        x = to_dtype(x, torch.float)
```
ngimel, line 1604: is this needed? Triton codegen automatically upcasts on read, so no low precision numbers should be present in the kernel

jansel: I think you are right, this should happen automatically.

eellison: Without this, you get a cast back to float16 before the division, and the test fails

ngimel: There should not be cast backs other than at the kernel boundaries, and division should happen in the same kernel.

eellison: yea, a related issue is that a separate kernel is launched for just the divide.. not sure yet ATM why it is not being fused
```

from ctypes import c_void_p, c_long
import torch
from torch import empty_strided, as_strided
from torchinductor.codecache import CppCodeCache, TritonCodeCache

aten = torch.ops.aten

import triton
import triton.language as tl

from torchinductor.triton_ops.autotune import pointwise_heuristics
from torchinductor.triton_ops.autotune import reduction_heuristics
from torchinductor.triton_ops.autotune import grid

@reduction_heuristics(size_hints=[128, 32768])
@triton.jit
def kernel0(in_ptr0, out_ptr0, xnumel, rnumel, XBLOCK : tl.constexpr, RBLOCK : tl.constexpr):
    xindex = xoffset + tl.reshape(tl.arange(0, XBLOCK), [XBLOCK, 1])
    xmask = xindex < xnumel
    rbase = tl.reshape(tl.arange(0, RBLOCK), [1, RBLOCK])
    x0 = xindex
    _tmp1 = tl.zeros([XBLOCK, RBLOCK], tl.float32) + 0
    for roffset in range(0, rnumel, RBLOCK):
        rindex = roffset + rbase
        rmask = rindex < rnumel
        r1 = rindex
        tmp0 = tl.load(in_ptr0 + (r1 + (29040*x0)) % 3717120, xmask & rmask).to(tl.float32)
        _tmp1 = tl.where(xmask & rmask, _tmp1 + tmp0, _tmp1)
    tmp1 = tl.reshape(tl.sum(_tmp1, 1), [XBLOCK, 1])
    tl.store(out_ptr0 + x0, tmp1, xmask)

@reduction_heuristics(size_hints=[1, 128])
@triton.jit
def kernel1(in_ptr0, out_ptr0, xnumel, rnumel, XBLOCK : tl.constexpr, RBLOCK : tl.constexpr):
    xindex = xoffset + tl.reshape(tl.arange(0, XBLOCK), [XBLOCK, 1])
    xmask = xindex < xnumel
    rbase = tl.reshape(tl.arange(0, RBLOCK), [1, RBLOCK])
    _tmp1 = tl.zeros([XBLOCK, RBLOCK], tl.float32) + 0
    for roffset in range(0, rnumel, RBLOCK):
        rindex = roffset + rbase
        rmask = rindex < rnumel
        r0 = rindex
        tmp0 = tl.load(in_ptr0 + r0, rmask)
        _tmp1 = tl.where(xmask & rmask, _tmp1 + tmp0, _tmp1)
    tmp1 = tl.reshape(tl.sum(_tmp1, 1), [XBLOCK, 1])
    tl.store(out_ptr0 + 0 + tl.zeros(tmp1.shape, tl.int32), tmp1, xmask)

@pointwise_heuristics(size_hints=[1])
@triton.jit
def kernel2(in_ptr0, out_ptr0, xnumel, XBLOCK : tl.constexpr):
    xmask = xindex < xnumel
    tmp0 = tl.load(in_ptr0 + 0, None).to(tl.float32)
    tmp1 = 3717120
    tmp2 = tmp0 / tmp1
    tl.store(out_ptr0 + 0 + tl.zeros(tmp2.shape, tl.int32), tmp2, xmask)


def call(arg0):
    arg0_size = arg0.size()
    s0 = arg0_size[0]
    s1 = arg0_size[1]
    s2 = arg0_size[2]
    buf0 = empty_strided((128, ), (1, ), device='cuda', dtype=torch.float32)
    kernel0[grid(128)](arg0, buf0, 128, 29040)
    buf1 = empty_strided((), (), device='cuda', dtype=torch.float16)
    kernel1[grid(1)](buf0, buf1, 1, 128)
    del buf0
    buf2 = empty_strided((), (), device='cuda', dtype=torch.float16)
    kernel2[grid(1)](buf1, buf2, 1)
    return (buf2, )


if __name__ == "__main__":
    from microbenchmarks.microbench import print_performance
    arg0 = rand_strided((10, 3, 352, 352), (371712, 123904, 352, 1), device='cuda', dtype=torch.float16)
    print_performance(lambda: call(arg0))
```

ngimel: Hm that's a problem if that's the case, I hoped #446 would help with that but probably it requires setting the number of tiles to work. Then all reduction lowerings, not just mean, will need to be fixed for this. 
Primtorch decomps handle casting to/from higher datatypes, so alternatively we can rely on primtorch traces inserting the necessary casts.

eellison: Well, primtorch has the same issue of converting back to fp16 before the division

ngimel: For `mean` primtorch keeps reduction result in promoted type, and converts only after division https://github.com/pytorch/pytorch/blob/d6cd130e0b791a0556ebc7bbd6871840b65ad35c/torch/_refs/__init__.py#L1753

PR #488, opened 2022-06-29
[RFC] Add a backwards compatible strict_mode to dynamo to fail aggressively on verify_correctness 

voznesenskym:
I don't love this name, and I am torn on it being in `.optimize()`

Alternatives:

- remove from `.optimize()` add a config value to verify_corectness, like mode=strict
- give it a better name than strict_mode, but keep it in `.optimize()`

### DIFF
--- a/tests/test_verify_correctness.py
+++ b/tests/test_verify_correctness.py
@@ -97,7 +97,7 @@ class TestVerifyCorrectness(torchdynamo.testing.TestCase):
         d = 4
         r1 = None
         r2 = fn(a, (b, c), d)
-        with torchdynamo.optimize_assert(compiler_fn):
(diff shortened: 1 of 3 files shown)


ezyang:
If we are afraid of confusion with the existing config, we could do something like `assert_correctness`. BTW, the corresponding input argument in torch.jit.trace is `check_trace` which is annoyingly inappropriate for dynamo haha

jansel (changes requested):
I'd suggest just making `strict_mode` always true and removing the flag.  We already have the `verify_correctness` config, so I think also having `strict_mode` is just redundant.  If people set verify_correctness (not the default), they should get exceptions.

PR #496, opened 2022-06-30
Improve logging for failed 'invalidated' guard

wconstab:
Previously:
torchdynamo hit config.cache_size_limit (64)
   function: 'toy_example' (example.py:8)
   reasons:  ['___guarded_code.valid']

Now:
torchdynamo hit config.cache_size_limit (64)
   function: 'toy_example' (example.py:8)
   reasons:  ["Object 140487837021376 of type <class '__main__.TestMod'> is weakref'd by this code, but has been finalized"]

### DIFF
--- a/torchdynamo/guards.py
+++ b/torchdynamo/guards.py
@@ -346,7 +346,9 @@ class CheckFunctionManager:
         f_globals: Optional[Dict] = None,
     ):
         self.valid = True
+        self.invalidated_reason = None
         self._weakrefs = []
+        self._weakref_invalidate_reasons = {}
         self._seen_ids = set()
 
         local_builder = GuardBuilder(self.id_ref, f_locals, self, renames=True)
@@ -371,7 +373,9 @@ class CheckFunctionManager:
         )
         # TODO(whc) maybe only the 'check_tensors' one is ambiguous? if so we can be less general..
         verbose_code_parts = (
-            ["___guarded_code.valid"] + local_builder.code + global_builder.code
+            ["__guarded_code_invalid_reason()"]
+            + local_builder.code
+            + global_builder.code
         )
 
         tensor_check_names = (
@@ -403,6 +407,7 @@ class CheckFunctionManager:
                 ("___check_tensors", check_tensors_fn),
                 ("___check_tensors_verbose", check_tensors_verbose_fn),
                 ("tensor_check_names", tensor_check_names),
+                ("__guarded_code_invalid_reason", lambda: self.invalidated_reason),
             ]
         )
         closure_vars.update(CLOSURE_VARS)
@@ -428,6 +433,8 @@ class CheckFunctionManager:
     def invalidate(self, ref):
         # A weakref is no longer valid, self.check_fn should return false
         self.valid = False
+        reasons = self._weakref_invalidate_reasons[ref]
+        self.invalidated_reason = f"This compiled code holds a weakref to object id({reasons['id']}) of type {reasons['type']}, which has been finalized"
 
     def id_ref(self, obj):
         """add a weakref, return the id"""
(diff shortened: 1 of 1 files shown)


torchdynamo/guards.py
```diff
@@ -428,13 +433,21 @@ def ___make_guard_fn({','.join(closure_vars.keys())}):
     def invalidate(self, ref):
         # A weakref is no longer valid, self.check_fn should return false
         self.valid = False
+        reasons = self._weakref_invalidate_reasons[ref]
+        self.invalidated_reason = f"This compiled code holds a weakref to object id({reasons['id']}) of type {reasons['type']}, which has been finalized"
```
voznesenskym, line 437: Invalidate can be called from anywhere. Instead, I would add another small func `_invalidate_weakref`, and change where we bind a method to `weakref.ref` to take `_invalidate_weakref` which would then call `invalidate(reason)`

torchdynamo/guards.py
```diff
@@ -371,7 +373,9 @@ def compile_check_fn(self, local_builder, global_builder):
         )
         # TODO(whc) maybe only the 'check_tensors' one is ambiguous? if so we can be less general..
         verbose_code_parts = (
-            ["___guarded_code.valid"] + local_builder.code + global_builder.code
+            ["__guarded_code_invalid_reason()"]
```
voznesenskym, line 376: I don't think you need to actually compile anything to do with `reason` into the `check_fn`. I would keep this as `.valid` but then store a .reason alongside where you invalidate.

voznesenskym (changes requested):

torchdynamo/guards.py
wconstab: well, this is something we could overhaul in general, but right now the way I did the verbose printing was to construct a duplicate 'check function' which only runs in the verbose log printer and has 1:1 lines with the original check_function.

The original check_function stays 'fast', but if it fails and enters the callback that prints logs, we run the slow/verbose variant which can produce explanation strings.

wconstab: this is pretty hacky.  but how it works is,
1) the fast 'check_fn' fails
2) the callback gets triggered
https://github.com/pytorch/torchdynamo/blob/main/torchdynamo/guards.py#L454
3) the slow 'verbose_check_fn' code parts get iterated, and run one by one
 - True bools or None strings cause iteration to continue, 
 - false bools cause the code part to be printed directly (the legacy fallback)
 - strings returned by verbose codeparts are directly used

wconstab: putting perf considerations aside, it seems like a clean api would be to define a 'guard' obj which includes 2 functions (check() and reason())

the fast path check_fn would consist of something like 
  for g in guards()
       if not g.check():
           FAIL
           
 and then the verbose function could simply run the same loop but instead of doing whatever 'FAIL' means it could call g.reason() and log it.
 
 I went with the hack above to avoid messing with the fast path at all, but it should be rethought.

torchdynamo/guards.py
wconstab: i guess so, but it looks like invalidate(self, ref) is a callback of the specific signature compatible with weakref, so maybe this func itself should be renamed to weakref_invalidate and the general one can accept a "reason" string?

torchdynamo/guards.py
voznesenskym: Maybe this is too much for this PR, but instead of having to iterate guards over that failed, we should probably produce guards with a side effect of storing their failure, and then instead of re-iterating and re-evaluating guards, you can just print go and see exactly which guards failed. O(1 * failures) instead of O(guards * failures). This would also allow us to not special case "valid" as each guard could have a "reason" field, with fallback to just printing the failed function as executed at time of failure if no "reason" is provided.

But for this PR, I dont think we need to modify what goes into `check_fn`. You can keep it as `invalidated_reason` where you have it, but just make sure to pass it through as a field and add it to `guard_fn` in this same func (`compile_check_fn`). Then in your guard fail hook, you can just read it off the guard_fn.

torchdynamo/guards.py
voznesenskym: Sure, yeah, I missed that, that sounds good.

torchdynamo/guards.py
wconstab: one reason i didn't go this route (for verbose tensor_check, or for other not yet implemented verbose checks) is that i want to be able to format a string including info from both the original (guarded on) value (such as its size, id, etc), and from the new value (such as its size, id, value, etc).  It's a bit restrictive to have to prepare the string at guard creation time, since at that time you only have access to the original value.

By creating a 'verbose_check_fn' and rerunning its parts at fail-hook time, you can format messages with info from both old and new(runtime) values.  I think this is worth the cost, as we shouldn't care about the performance of our guard-fail-hook.

jansel:

PR #498, opened 2022-07-02
[WIP][Discussion] Write out a deeper documentation on how we go from …

voznesenskym:
…user code to producing guards

### DIFF
--- a/documentation/GuardsOverviewPt1.md
+++ b/documentation/GuardsOverviewPt1.md
@@ -39,7 +39,7 @@ with torchdynamo.optimize(my_compiler):
         toy_example(torch.randn(10), torch.randn(10))
 ```
 
-This allows TorchDynamo to capture the interpreted Python frames, grab any and all relevant information, and speed things up wherever it can. The speedup comes from a few places, and can be rather dependent on the backend (my_compiler above) provided, but the one speedup we care about most for today's overview is **caching**.
+This allows TorchDynamo to capture the interpreted Python frames, grab any and all relevant information, and speed things up wherever it can. The speedup comes from a few places, and can be rather dependent on the backend (my_compiler above) provided, but the one speedup we care about most for today's overview is **caching**. Caching itself is not a direct speedup, so much as a critical enablement to allow us to prevent recompilation. We dig a hole with dynamo, and caching allows us to get out. Its a speedup from that perspective, but relatively neutral when all things are considered - however, it enables us to hold perf neutrality while then enabling backends - the true source of our speedups.
 
 With even a pass-through no-op backend provided:
 ```py
@@ -50,7 +50,7 @@ We can see TorchDynamo speeding up Python execution quite a bit, even on regular
 
 ## Caching and Guards Overview
 
-TorchDynamo operates through caching modified (by TorchDynamo) user bytecode. When we receive a frame for evaluation, we check if the **objects referenced in the frame have changed** in certain ways, and if not, we read the modified user bytecode out and evaluate it.  The detail of how we do this will be saved for a later writeup. Instead, we will focus on how we can identify wether or not the **objects referenced in the frame have changed**. This is a critical piece of functionality in TorchDynamo, because it drives the entire invalidation lifecycle. We refer to this functionality as **guards**.
+TorchDynamo operates through caching modified (by TorchDynamo) user bytecode. When we receive a frame for evaluation, we check if the **objects referenced in the frame have changed** in certain ways, and if not, we read the previously (or already) modified user bytecode to evaluate it.  The detail of how we do this will be saved for a later writeup. Instead, we will focus on how we can identify wether or not the **objects referenced in the frame have changed**. This is a critical piece of functionality in TorchDynamo, because it drives the entire invalidation lifecycle. We refer to this functionality as **guards**.
 
 At a very high level, the vastly oversimplified TLDR flow is this:
 
--- a/documentation/GuardsOverviewPt1.md
+++ b/documentation/GuardsOverviewPt1.md
@@ -1,0 +1,308 @@
+# How TorchDynamo Works - Pt1 - From user code to guards
+
+## Welcome
+Welcome to the first entry in a series of technical documents outlining how TorchDynamo works.
+
+We assume the reader already has a working knowledge of PyTorch, and Python, and a basic understanding of TorchDynamo's role in the PyTorch Ecosystem. The target audience is engineers and researchers looking to learn more about how TorchDynamo operates under the hood.
+
+## Using Dynamo
+From a UX perspective, TorchDynamo is very easy to use. The user invokes `torchdynamo.optimize`, either as a context:
+```py
+with torchdynamo.optimize(my_compiler):
+```
+or as an annotation:
+```py
+@torchdynamo.optimize(my_compiler)
+def fn_foo(bar):
+```
+
+Where a complete example looks like this:
+
+```py
+from typing import List
+import torch
+import torchdynamo
+
+def toy_example(a, b):
+    x = a / (torch.abs(a) + 1)
+    if b.sum() < 0:
+        b = b * -1
+    return x * b
+
+def my_compiler(gm: torch.fx.GraphModule, example_inputs: List[torch.Tensor]):
+    print("my_compiler() called with FX graph:")
+    gm.graph.print_tabular()
+    return gm.forward  # return a python callable
+
+with torchdynamo.optimize(my_compiler):
+    for _ in range(100):
+        toy_example(torch.randn(10), torch.randn(10))
+```
+
+This allows TorchDynamo to capture the interpreted Python frames, grab any and all relevant information, and speed things up wherever it can. The speedup comes from a few places, and can be rather dependent on the backend (my_compiler above) provided, but the one speedup we care about most for today's overview is **caching**.
+
+With even a pass-through no-op backend provided:
+```py
+def my_compiler(gm: torch.fx.GraphModule, example_inputs: List[torch.Tensor]):
+    return gm.forward
+```
+We can see TorchDynamo speeding up Python execution quite a bit, even on regular Python, not just PyTorch.
+
+## Caching and Guards Overview
+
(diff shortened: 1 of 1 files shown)


jansel:
This is awesome, thanks for writing it!

I also think a big gap is user-facing documentation in addition to developer-facing.

documentation/GuardsOverviewPt1.md
```diff
@@ -0,0 +1,308 @@
...
+    print("my_compiler() called with FX graph:")
+    gm.graph.print_tabular()
+    return gm.forward  # return a python callable
+
+with torchdynamo.optimize(my_compiler):
+    for _ in range(100):
+        toy_example(torch.randn(10), torch.randn(10))
+```
+
+This allows TorchDynamo to capture the interpreted Python frames, grab any and all relevant information, and speed things up wherever it can. The speedup comes from a few places, and can be rather dependent on the backend (my_compiler above) provided, but the one speedup we care about most for today's overview is **caching**.
```
anijain2305, line 42: > the one speedup we care about most for today's overview is caching

It's somewhat unclear about the relationship between caching and speedup. My view is that caching infact makes program slower than eager. Its goal is to avoid recompilation. So we are basically digging a hole, and caching allows us to get out of it. The speedup mostly comes from the backends, and other optimizations.

documentation/GuardsOverviewPt1.md
```diff
@@ -0,0 +1,308 @@
...
+With even a pass-through no-op backend provided:
+```py
+def my_compiler(gm: torch.fx.GraphModule, example_inputs: List[torch.Tensor]):
+    return gm.forward
+```
+We can see TorchDynamo speeding up Python execution quite a bit, even on regular Python, not just PyTorch.
+
+## Caching and Guards Overview
+
+TorchDynamo operates through caching modified (by TorchDynamo) user bytecode. When we receive a frame for evaluation, we check if the **objects referenced in the frame have changed** in certain ways, and if not, we read the modified user bytecode out and evaluate it.  The detail of how we do this will be saved for a later writeup. Instead, we will focus on how we can identify wether or not the **objects referenced in the frame have changed**. This is a critical piece of functionality in TorchDynamo, because it drives the entire invalidation lifecycle. We refer to this functionality as **guards**.
```
anijain2305, line 53: > we read the modified user bytecode to evaluate it

we read the previously (or already) modified user bytecode to evaluate it

documentation/GuardsOverviewPt1.md
voznesenskym: Very good point, will amend it. I was trying to walk a fine line around complexity and readability, but this is a real omission on my end.

documentation/GuardsOverviewPt1.md
```diff
@@ -0,0 +1,308 @@
...
+```
+
+This allows TorchDynamo to capture the interpreted Python frames, grab any and all relevant information, and speed things up wherever it can. The speedup comes from a few places, and can be rather dependent on the backend (my_compiler above) provided, but the one speedup we care about most for today's overview is **caching**. Caching itself is not a direct speedup, so much as a critical enablement to allow us to prevent recompilation. We dig a hole with dynamo, and caching allows us to get out. Its a speedup from that perspective, but relatively neutral when all things are considered - however, it enables us to hold perf neutrality while then enabling backends - the true source of our speedups.
+
+With even a pass-through no-op backend provided:
+```py
+def my_compiler(gm: torch.fx.GraphModule, example_inputs: List[torch.Tensor]):
+    return gm.forward
+```
+We can see TorchDynamo speeding up Python execution quite a bit, even on regular Python, not just PyTorch.
```
anijain2305, line 49: Asking just because curious - Any example of where this happens? is it because we are specializing on ints etc?

documentation/GuardsOverviewPt1.md
```diff
@@ -0,0 +1,308 @@
...
+
+## Caching and Guards Overview
+
+TorchDynamo operates through caching modified (by TorchDynamo) user bytecode. When we receive a frame for evaluation, we check if the **objects referenced in the frame have changed** in certain ways, and if not, we read the previously (or already) modified user bytecode to evaluate it.  The detail of how we do this will be saved for a later writeup. Instead, we will focus on how we can identify wether or not the **objects referenced in the frame have changed**. This is a critical piece of functionality in TorchDynamo, because it drives the entire invalidation lifecycle. We refer to this functionality as **guards**.
```
anijain2305, line 53: > we read the previously (or already) modified user bytecode

Oh I meant use whatever word you like - previously/already :) Just wanted to point out that we are not re-modifiying.

anijain2305, line 53: Is "transformed" a better word than "modified" user bytecode? Generally, compilers have analysis and transformation.

documentation/GuardsOverviewPt1.md
```diff
@@ -0,0 +1,308 @@
...
+
+At a very high level, the vastly oversimplified TLDR flow is this:
+
+1) We receive a python frame
+2) We convert the given frame from (1), passing it through instruction translation
+3) For the objects captured in (2), we create tracking objects that are (a) tracked on an output graph, which is an internal specialization of a torch.fx.Tracer (and the topic of a later writeup), and (b) guards, the topic of this document.
+4) We process the guard objects created in (3), turning them into a generated python function, check_fn, associated with a piece of code.
+5) The check_fn is evaluated whenever we encounter this code a subsequent time - if a check_fn passes and evaluates to True, we know the code in the cache and the code encountered here is the same, and can be safely used. If it fails and evaluates to False, we know the code in the cache is not valid, and can be thrown out in favor of a new entry.
```
anijain2305, line 61: > an be thrown out in favor of a new entry.

Maybe mention that we can recompile, instead of eviction?

documentation/GuardsOverviewPt1.md
```diff
@@ -0,0 +1,308 @@
...
+```py
+def  _convert_frame_assert(frame: types.FrameType, cache_size: int):
+```
+
+So, what does this function do?
+
+Top to bottom, we:
+
+1) Check if we have seen this `code`(see: f_code here https://docs.python.org/3/library/inspect.html) before, and exit early if we have
+2) Check if the code we are looking at is a tricky case we have not added support for yet (The detail of what happens when we skip a frame is a bit out of scope of this writeup, but a later writeup around unimplemented, eager fallback vs whole graph capture, and more of that is planned for later)
```
anijain2305, line 87: > but a later writeup around unimplemented, eager fallback vs whole graph capture, and more of that is planned for later

Nit - First later can be ignored

documentation/GuardsOverviewPt1.md
```diff
@@ -0,0 +1,308 @@
...
+4) We pass the frame, alongside a function that creates an `InstructionTranslator` (more on this later) through bytecode transformation, via `transform_code_object`. A few crucial things happen under the hood here:
+    1)  We produce new code through `transform_code_object`
+
+    2) We produce an fx tracer named `output` through `InstructionTranslator` [*Note: This can be a little confusing, as `InstructionTranslator` is not an fx tracer, but its stored in a variable named tracer, and its output **is** a fx tracer.*]
+
+    3) We produce guards and store them on `output` above
+
+    4) We produce `output_instructions` and store them on `output` above (a bit out of scope for this document)
+
+    5) We  and map the newly produced transformed code to the initial code we read off the frame. (This mapping is worth remembering, we will refer to it much later on below where we cover guard failures).
```
anijain2305, line 98: > We and map

We map

documentation/GuardsOverviewPt1.md
```diff
@@ -0,0 +1,308 @@
...
+elif istype(value, range):
+	guards = self.make_guards(GuardBuilder.EQUALS_MATCH)
+	return RangeVariable(value=value, guards=guards)
+```
+
+Since `source` was set at the construction time of this `VariableTracker`, all that was needed here was to provide the fn, `GuardBuilder.EQUALS_MATCH` to the `create_fn` field.
+
+This `create_fn` must be a method on `GuardBuilder`. The reason for this becomes apparent in our next step. Once we have all the guards created for a frame, we move on to `CheckFunctionManager` and `compile_check_fn`.
+
+Remember that `convert_frame` function way above, in the first section? Before it can produce a `GuardedCode`, it needs to run the `CheckFunctionManager`, with all the guards, to produce a `check_fn` which will then, in turn get passed in alongside the code into `GuardedCode`. This is the same `check_fn` that we store in our cache entry, and the same one we run to know wether or not to retrieve the code stored alongside. For reference, here is that code:
```
anijain2305, line 239: wether -> whether

documentation/GuardsOverviewPt1.md
```diff
@@ -0,0 +1,308 @@
...
+def  _convert_frame_assert(frame: types.FrameType, cache_size: int):
+```
+
+So, what does this function do?
+
+Top to bottom, we:
+
+1) Check if we have seen this `code`(see: f_code here https://docs.python.org/3/library/inspect.html) before, and exit early if we have
+2) Check if the code we are looking at is a tricky case we have not added support for yet (The detail of what happens when we skip a frame is a bit out of scope of this writeup, but a later writeup around unimplemented, eager fallback vs whole graph capture, and more of that is planned for later)
+3) We check if the `cache_size` (second arg above) crosses the limit defined in our config , `cache_size_limit`. If it has, we drop the frame and log out some warnings. This helps us avoid constant recompilation of a frame as it generally means that the frame is hot in an unexpected way, and caching it is producing needless overhead, as it is likely to get blown away the next time we encounter it anyway.
```
anijain2305, line 88: Nit - How about use eviction instead of blown away?

anijain2305 (approved):
Overall looks pretty awesome. Thanks for writing this up :)

I have minor suggestions here and there. Might be good to spellcheck this to avoid any typos.

anijain2305:
cc @wconstab

documentation/GuardsOverviewPt1.md
voznesenskym: Internal docs and code uses both. I am fine with transformed

Issue #473, opened 2022-06-28
Guard Failures in T5 Model


kevinstephano:
In addition to the `torch.finfo` issue, that I verified was fixed, there remains several guard failures.  Are these unique instances or the hierarchy of the failure?

The last guard failure provided the most specific point in the code dealing with the shape of the input linear matmuls of the multihead attention block:

```
        def project(hidden_states, proj_layer, key_value_states, past_key_value):
            """projects hidden states correctly to key/query states"""
            if key_value_states is None:
                # self-attn
                # (batch_size, n_heads, seq_length, dim_per_head)
                hidden_states = shape(proj_layer(hidden_states))
            elif past_key_value is None:
                # cross-attn
                # (batch_size, n_heads, seq_length, dim_per_head)
                hidden_states = shape(proj_layer(key_value_states))

            if past_key_value is not None:
                if key_value_states is None:
                    # self-attn
                    # (batch_size, n_heads, key_length, dim_per_head)
                    hidden_states = torch.cat([past_key_value, hidden_states], dim=2)
                else:
                    # cross-attn
                    hidden_states = past_key_value
            return hidden_states
```

```
[...]

PR #500, opened 2022-07-03
More Huggingface models 

anijain2305:
Most of this is derived from @kevinstephano simple_dl benchmarking infra. With this we have many models running in Dynamo repo, handily available for debug. One motivation for me is to ensure we could easily repro and also track such failures in futures - https://github.com/pytorch/torchdynamo/issues/473

cc @Chillee @ngimel

### DIFF
--- a/benchmarks/huggingface.py
+++ b/benchmarks/huggingface.py
@@ -1,23 +1,54 @@
 #!/usr/bin/env python
+import importlib
 import logging
 import re
+import subprocess
+import sys
 import warnings
+from functools import partial as partial
 
 import torch
 from common import BenchmarkRunner
 from common import main
-from transformers import AutoConfig
-from transformers import AutoModelForCausalLM
-from transformers import AutoModelForMaskedLM
-from transformers import AutoModelForSeq2SeqLM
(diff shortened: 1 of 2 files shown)


benchmarks/huggingface.py
```diff
@@ -1,112 +1,223 @@
 #!/usr/bin/env python
+import importlib
 import logging
 import re
+import subprocess
+import sys
 import warnings
+from functools import partial as partial
```
jansel, line 8:
```suggestion
from functools import partial
```

torchdynamo/utils.py
```diff
@@ -219,6 +219,12 @@ def clone_input(x):
 
 
 def clone_inputs(example_inputs):
+    if isinstance(example_inputs, dict):
+        for key, value in example_inputs.items():
```
jansel, line 223:
```suggestion
        example_inputs = dict(example_inputs)
        for key, value in example_inputs.items():
```

So we don't mutate the caller's copy of `example_inputs`

jansel (approved):
lgtm with some minor nits

PR #501, opened 2022-07-03
Cleanup in ExactWeakKeyDictionary

anijain2305:
I have not really debugged why we need this in the first place. But many new warnings have come up after this fix - https://github.com/pytorch/torchdynamo/issues/477. I have seen this while running pytest and even huggingface.py tests.

cc @ezyang

### DIFF
--- a/torchdynamo/utils.py
+++ b/torchdynamo/utils.py
@@ -74,14 +74,13 @@ class ExactWeakKeyDictionary:
         self.values[idx] = value
 
     def _remove_id(self, idx):
-        if idx in self.values:
-            del self.values[idx]
         if idx in self.refs:
+            del self.values[idx]
             del self.refs[idx]
 
     def clear(self):
-        self.values.clear()
         self.refs.clear()
+        self.values.clear()
 
 
 def istype(obj, allowed_types):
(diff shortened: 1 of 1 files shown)


torchdynamo/utils.py
```diff
@@ -74,8 +74,9 @@ def __setitem__(self, key, value):
         self.values[idx] = value
 
     def _remove_id(self, idx):
-        if idx in self.refs:
+        if idx in self.values:
             del self.values[idx]
+        if idx in self.refs:
             del self.refs[idx]
```
voznesenskym, line 80: Hmm, does this feel like its hiding a bug elsewhere in the class, where these are getting out of sync?

anijain2305: Yeah, I think it is hiding a bug underneath. I did not get a chance to dig deep here.

Let me convert this PR to WIP, and I will take deeper look.

voznesenskym: If you throw up a repro case, I can look early next week too!

jansel: I think this is actually coming from `.clear()` in this same class.  Clear deletes refs then values, when it should do it in opposite order.

anijain2305:
Error is gone after @jansel's suggested fix.

anijain2305:
Closing as https://github.com/pytorch/torchdynamo/pull/475 already covers this.

PR #502, opened 2022-07-03
Reduction deps

ngimel:
Downstream reduction dependencies often expect sizes that are different from the reduction writes sizes, and thus are not correctly marked as fusable even when reduction is complete. 
This PR adds couple more MemoryDeps for reduction ops so that dependencies of downstream ops (such as dividing by number of elements for mean, or computing mean and rstd for layer norm) are satisfied and they can be put in the same kernel with reduction.

### DIFF
--- a/torchinductor/dependencies.py
+++ b/torchinductor/dependencies.py
@@ -20,6 +20,31 @@ class MemoryDep(typing.NamedTuple):
         size = (*self.size, *[x for x in extra_sizes if x != 1])
         return MemoryDep(self.name, self.index, size)
 
+    def maybe_swap_sizes(self):
+        # swap only in simple cases where index is trivial and
+        # there are just 2 sizes
+        if len(self.size) == 2 and len(self.index.args) == 0 and \
+           self.index.name == 'c0':
+            size = (self.size[1], self.size[0])
+            index = self.index.subs({"c0" : "c1"})
+            return MemoryDep(self.name, index, size)
+        else:
+            return self
+
+    def strip_last_size(self):
+        nsizes = len(self.size)
+        if nsizes >= 1 and len(self.index.args) <= nsizes - 1:
+            # make sure last dim index is not used
+            v = sympy.Symbol("_strip_size_tester")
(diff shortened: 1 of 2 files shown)


torchinductor/dependencies.py
```diff
@@ -20,6 +20,34 @@ def broadcast_extend_sizes(self, extra_sizes):
...
+
+    def strip_last_size(self):
+        nsizes = len(self.size)
+        if nsizes >= 1 and len(self.index.args) <= nsizes - 1:
+            # make sure last dim index is not used
+            v = sympy.Symbol("_strip_size_tester")
+            prefix = "c"
+            last_index = f"{prefix}{len(self.size)-1}"
+            expr1 = self.index.subs({last_index: v})
+            if expr1 == self.index:
```
jansel, line 45:
```suggestion
            if last_index not in self.index.free_symbols:
```

jansel (approved):
ltgm

Chillee:
For my own education, can you explain a bit more on what prevented the fusion previously?

ngimel:
> For my own education, can you explain a bit more on what prevented the fusion previously?

Memory dependencies for ops are recorded based on the reads of the corresponding op, so e.g. for division in `mean` memory dep will look like 
```
MemoryDep(name='buf0', index=c0, size=(s0*s2,))
```
However, the "ready" dependencies for upstream ops are updated based on the writes of the upstream op, and writes for reduction would reference the whole domain including reduction dimension. This is arguably wrong, but I don't know if or how we should fix that, so upstream reduction would say that it's satisfying dependence in the form of 
```
MemoryDep(name='buf0', index=c0, size=(s0*s2, s1))
```
where `s1` is reduction dimension. These 2 dependencies won't match, so division wouldn't recognize that its dependence is in fact satisfied. This PR makes it so reduction updates its satisfied dependencies in several forms that downstream might expect (swapping unfortunately doesn't work yet, I'll try to figure out what's wrong with it, so normalization on the dimension other than last would still be split)

PR #505, opened 2022-07-03
[inductor] Support rand()/dropout()

jansel:
Example output:
```
seed_cuda_0 = None  # 94243ad63b02490aa7e207bad45cb26ee512c3d7571c22796c9e2de91622ecb9

@pointwise_heuristics(size_hints=[1024])
@triton.jit
def kernel0(seed0, in_ptr1, out_ptr0, ks0, xnumel, XBLOCK : tl.constexpr):
    xmask = xindex < xnumel
    x0 = xindex
    tmp0 = tl.load(seed0 + 0, None)
    tmp6 = tl.load(in_ptr1 + x0, xmask)
    tmp1 = x0
    tmp2 = tl.rand(tmp0, tmp1)
    tmp3 = 0.5
    tmp4 = tmp2 > tmp3
    tmp5 = tmp4.to(tl.float32)
    tmp7 = tmp5 * tmp6
    tmp8 = 2.0
    tmp9 = tmp7 * tmp8
    tl.store(out_ptr0 + x0, tmp9, xmask)


def call(arg0):
    seed_cuda_0.fill_(random.randrange(2**31))
    arg0_size = arg0.size()
    s0 = arg0_size[0]
    buf0 = empty_strided((s0, ), (1, ), device='cuda', dtype=torch.float32)
    kernel0[grid(s0)](seed_cuda_0, arg0, buf0, s0, s0)
    return (buf0, )

```

### DIFF
--- a/tests/test_torchinductor.py
+++ b/tests/test_torchinductor.py
@@ -3,6 +3,7 @@ import contextlib
 import dataclasses
 import functools
 import importlib
+import random
 import unittest
 from unittest.mock import patch
 
@@ -1969,6 +1970,7 @@ class CommonTemplate:
             check_lowp=False,
         )
 
+    @unittest.skipIf(not config.fallback_random, "requires config.fallback_random")
     def test_bernoulli(self):
         def fn(a):
             b = torch.empty_like(a)
@@ -2032,6 +2034,20 @@ class CommonTemplate:
             return aten.new_empty_strided(a, [1, 128, 128], [16384, 128, 1]).fill_(123)
 
         self.common(fn, [torch.randn(55)])
+
+    @patch.object(torchinductor.config.triton, "cudagraphs", True)
+    def test_dropout(self):
+        random.seed(1234)
+        torch.manual_seed(1234)
+
+        @torchdynamo.optimize("inductor")
+        def fn(a):
+            return torch.nn.functional.dropout(a, 0.5, True)
+
+        x = torch.ones(1000, device=self.device, dtype=torch.float32)
+        result = fn(x)
+        self.assertTrue(400 < result.nonzero().shape[0] < 600)
+        self.assertTrue(0.9 < result.mean().item() < 1.1)
 
 
 if HAS_CPU:
--- a/tests/test_torchinductor.py
+++ b/tests/test_torchinductor.py
@@ -2049,6 +2049,40 @@ class CommonTemplate:
         self.assertTrue(400 < result.nonzero().shape[0] < 600)
         self.assertTrue(0.9 < result.mean().item() < 1.1)
 
+    def test_dropout_deterministic(self):
+        if self.device == "cpu":
+            # TODO(jansel): CPU RNG is not yet deterministic
+            raise unittest.SkipTest("CPU currently nondeterministic")
+
+        @torchdynamo.optimize("inductor")
+        def fn(a):
+            return torch.nn.functional.dropout(a, 0.55, True)
+
+        for cg in (False, True):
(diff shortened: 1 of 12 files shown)


tests/test_torchinductor.py
```diff
@@ -2033,6 +2035,20 @@ def fn(a):
 
         self.common(fn, [torch.randn(55)])
 
+    @patch.object(torchinductor.config.triton, "cudagraphs", True)
+    def test_dropout(self):
+        random.seed(1234)
```
ngimel, line 2040: would be good to test that a function with the same seed produces deterministic results, and that replaying cuda graph for a second time (without setting seed) produces different results

torchinductor/codegen/wrapper.py
```diff
@@ -97,6 +98,8 @@ def __init__(self):
             ["", "", f"def call({', '.join(V.graph.graph_inputs.keys())}):"]
         )
         with self.prefix.indent():
+            for name in V.graph.randomness_seeds:
+                self.prefix.writeline(f"{name}.fill_(random.randrange(2**31))")
```
ngimel, line 102: Is it better to get a seed from `random.randrange` or from torch cpu-side generator? I'm not sure if typical call to `torch.manual_seed` also sets a seed for python-side random generation, so determinism could be lost?
Also, this won't work with cuda graphs, because this argument to `fill_` kernel will be baked in, and will be used on all subsequent calls. Handling of random + cuda graphs on the eager side is a bit delicate, with the state being advanced at each replay, and read in the kernels.

torchinductor/lowering.py
```diff
@@ -981,15 +1009,54 @@ def inner(
...
+        assert layout in (0, torch.strided)
+        assert memory_format in (None, torch.contiguous_format)
+        if dtype is None:
+            dtype = x.get_dtype()
+        else:
+            dtype = decode_dtype(dtype)
+        device = device or x.get_device()
+        size = list(x.get_size())
+        return creation_fn(
+            size,
```
ngimel, line 1041: this won't correctly reproduce *_like behavior, because by default *_like functions preserve permutation of original tensor (e.g. `zeros_like(x)` would return channels-last if `x` was channels last, or transposed if `x` was transposed 2d tensor)

jansel: This is true, though decomps in general get strides wrong, so this won't be the only case like this.

Currently, none of tensor constructors support layout specifiers (they all have asserts they are unset or contiguous).  Adding support for specifying layouts is not yet implemented and future work.

torchinductor/codegen/wrapper.py
jansel: I changed this to use `torch.randint`

tests/test_torchinductor.py
jansel: I added a test for this

torchinductor/codegen/wrapper.py
```diff
@@ -97,6 +98,10 @@ def __init__(self):
             ["", "", f"def call({', '.join(V.graph.graph_inputs.keys())}):"]
         )
         with self.prefix.indent():
+            for name in V.graph.randomness_seeds:
+                self.prefix.writeline(
+                    f"torch.randint(2**31, size=(), dtype=torch.int32, out={name})"
```
ngimel, line 103: my head spins thinking about which random counters are advanced here and how cuda graphs are handling it, but looks like it should work (at the cost of one extra kernel which in future we should be able to avoid).

ngimel (approved):

PR #506, opened 2022-07-03
[inductor] Fix bug with invalidated reuse

jansel:
We were hitting a bug in some training benchmarks where you had:
```
for roffset in range(0, rnumel, RBLOCK):
   ...
   write to bufX
   ...
...
for roffset in range(0, rnumel, RBLOCK):
   ...
   read from bufX
   ...
```

Since these get fused into a single kernel, we would remove bufX, which would result in an error of trying to read from an undefined buffer.


desertfire (approved):
LGTM

PR #507, opened 2022-07-03
[inductor] Handle no-op slice_scatter

jansel:


torchinductor/lowering.py
```diff
@@ -808,6 +808,11 @@ def slice_scatter(x, src, dim=0, start=None, end=None, step=1):
     src = expand(src, src_size)
     src_loader = src.make_loader()
 
+    if start == 0 and end == dim_size and step == 1:
+        # do nothing???
+        assert tuple(x.get_size()) == tuple(src.get_size())
+        return src
```
ngimel, line 814: slice_scatter is part of functionalization and thus has to always return a tensor that's copy of the corresponding slice of `src` (otherwise we could have just used `src[slice]`, so to preserve the intended semantics it should be `x.copy_(src)` ? At least that's what's aten.slice_scatter is doing.

jansel: This is the functional version of `slice_scatter` (not `slice_scatter_`), so we aren't allowed to mutate `x`.  So `x.copy_(src)` would be wrong.

We could return `src.clone()` instead.

ngimel: So `a` argument is completely ignored except to check its size? Yeah those functionalization-only ops have weird API.

jansel: Yes, the first arg is completely ignored in this case.  Perhaps it is the backwards of a no-op slice that selected every element of a tensor?

ngimel (approved):

PR #512, opened 2022-07-04
Enable guard_nn_modules, module change observation

voznesenskym:
Fixes reverted https://github.com/pytorch/torchdynamo/pull/326


torchdynamo/guards.py
```diff
@@ -52,6 +60,87 @@
...
 
+class NNModuleChangeTrackerUtil:
+    original_register_parameter = None
+    original_add_module = None
+    original_load_state_dict = None
+    original_setattr = None
+
+    @classmethod
+    def register_patches_on_torch_nn_module(cls):
+        if getattr(torch.nn.Module, "_NNModuleChangeTrackerUtil__dynamo_module_patch", True):
```
voznesenskym, line 71: @jansel this was the issue - we were patching over and over and not noticing it, because of a name issue on getattr and hasattr (`_NNModuleChangeTrackerUtil__dynamo_module_patch` vs `__dynamo_module_patch`). We tested on smaller models or use cases where this patch was not getting called as much, and so was still creating the issue, but hiding it. We now only set this up once.

torchdynamo/guards.py
```diff
@@ -52,6 +60,85 @@
...
+    original_load_state_dict = None
+    original_setattr = None
+
+    @classmethod
+    def register_patches_on_torch_nn_module(cls):
+        if getattr(
+            torch.nn.Module, "_NNModuleChangeTrackerUtil__dynamo_module_patch", True
+        ):
+            # register_parameter
+            torch.nn.Module.__dynamo_module_patch = False
```
jansel, line 75: Should this be `_NNModuleChangeTrackerUtil__dynamo_module_patch`?

torchdynamo/guards.py
```diff
@@ -52,6 +60,85 @@
...
+
+            @functools.wraps(original_register_parameter)
+            def custom_register_parameter(self, key, value):
+                NNModuleChangeTrackerUtil.invalidate_module(self)
+                return original_register_parameter(self, key, value)
+
+            torch.nn.Module.register_parameter = custom_register_parameter
+
+            # add_module
+            original_add_module = torch.nn.Module.add_module
```
jansel, line 87: Since these should be fixed.  Perhaps we should just read them once in global scope.

torchdynamo/guards.py
voznesenskym: No - this becomes `_NNModuleChangeTrackerUtil__dynamo_module_patch`

torchdynamo/guards.py
voznesenskym: This may get called multiple times, as `de_register_patches_on_torch_nn_module` can get invoked from `reset` or when we exit dynamo context, so it felt more correct to grab them here. No strong opinions.

torchdynamo/guards.py
voznesenskym: Actually, do you know why the name is getting mangled with the calling class like this?

jansel: I have no idea what could be causing this name mangling, or if we can count on the name mangling happening in the future.

voznesenskym: Discussed offline, this is a "feature" of python.

torchdynamo/guards.py
```diff
@@ -52,6 +60,83 @@
...
+
+            torch.nn.Module.__setattr__ = custom_setattr
+            cls.original_register_parameter = original_register_parameter
+            cls.original_add_module = original_add_module
+            cls.original_load_state_dict = original_load_state_dict
+            cls.original_setattr = original_setattr
+
+    @classmethod
+    def de_register_patches_on_torch_nn_module(cls):
+        torch.nn.Module = torch.nn.Module
```
anijain2305, line 120: Is this a typo?

torchdynamo/guards.py
```diff
@@ -298,28 +400,28 @@ def CONSTANT_MATCH(self, guard: Guard):
...
-
-        def setup_guard():
-            assert istype(val.training, bool)
-            self.code.append(f"{ref}.training == {val.training}")
 
-        if hasattr(val, "training"):
-            # There are cases where a monkeypatched object has a guard made between __new__ and __init__
-            setup_guard()
+        val = self.get(guard.name)
+        if hasattr(val, "training") and config.guard_nn_modules:
```
anijain2305, line 405: Just for my curiosity - Why are we checking for training attr here?

anijain2305 (commented):
Pretty nice work! This might help few of the testcases in PyTorch-Dynamo tests as well.

torchdynamo/guards.py
voznesenskym: Good question - context here: https://github.com/pytorch/torchdynamo/pull/322/files

jansel (approved):

voznesenskym:
Attempt #2!

pyjhzwh:
`python benchmarks/torchbench.py -dcuda --inductor-settings --float16 -n50 --inductor --filter drq`
The perf of drq slowed down from 5.13x(previous commit) to 0.867(this commit)

PR #532, opened 2022-07-08
[inductor] Fix a bmm related bug

desertfire:
Summary: The bug is exposed with pytorch_struct training run. We need to
be conservative on the optimization for bs=1 bmm, when any operand is not consecutive.

### DIFF
--- a/torchinductor/ir.py
+++ b/torchinductor/ir.py
@@ -1893,7 +1893,11 @@ class BatchMatrixMultiply(ExternKernelOut):
             size=[b3, m, n],
         ).as_fixed()
 
-        if b3 == 1:
+        if (
+            b3 == 1
+            and is_contiguous_storage_and_layout(a)
+            and is_contiguous_storage_and_layout(b)
+        ):
             # convert to normal mm
             data = MatrixMultiply(
                 layout=output_layout.as_fixed(),
--- a/tests/test_torchinductor.py
+++ b/tests/test_torchinductor.py
@@ -779,7 +779,7 @@ class CommonTemplate:
         )
         self.common(mod, (torch.randn(2, 8),))
 
-    def test_bmm(self):
+    def test_bmm1(self):
         def fn(a, b):
             return (
                 torch.bmm(a, b),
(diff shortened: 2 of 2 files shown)


jansel (approved):

torchinductor/ir.py
```diff
@@ -1893,7 +1893,11 @@ def create(cls, a, b):
...
 
-        if b3 == 1:
+        if (
+            b3 == 1
+            and is_contiguous_storage_and_layout(a)
+            and is_contiguous_storage_and_layout(b)
+        ):
             # convert to normal mm
             data = MatrixMultiply(
                 layout=output_layout.as_fixed(),
```
desertfire, line 1903: @jansel  I guess another option is to generate a ReinterpretView instead of View when the batch size is 1, but I am not sure if that is safe to do here.

torchinductor/ir.py
```diff
@@ -1893,7 +1893,11 @@ def create(cls, a, b):
             size=[b3, m, n],
         ).as_fixed()
 
-        if b3 == 1:
+        if (
+            b3 == 1
+            and is_contiguous_storage_and_layout(a)
```
ngimel, line 1898: this doesn't look right, MatrixMultiply should be able to handle transposed inputs

desertfire: The problem is because `View.create` does not return a `ReinterpretView` because of the checking at https://github.com/pytorch/torchdynamo/blob/044e8bd85757e6cb46ff29cbe4c56227809ecdea/torchinductor/ir.py#L695, and thus later failed an assertion at https://github.com/pytorch/torchdynamo/blob/044e8bd85757e6cb46ff29cbe4c56227809ecdea/torchinductor/ir.py#L1570.

I think making a special case in `View.create` might work. @jansel , any suggestion?

jansel: Hrm, in this case perhaps we should be treating it as contiguous, and ignore size=1 dims in contiguous checks.  The `bs=1` doesn't actually matter.

ngimel: `size=1` is already not checked, but one (or both) of the 2d matrices resulting from removing batch dimension might not be contiguous, and thus `View.create` returns  `View` and not `ReinterpretView`. I think we can return `ReinterpretView` in more cases by specifying proper strides in the `new_layout`

ngimel: I think we should extend SqueezeView.create to accept a specific dimension, and use `Squeezeview.create` instead of `View.create` below.

jansel (approved):
I agree with @ngimel here we should be able to map bmm to mm in the transposed case.  It might be a bit annoying to do that because not all of the view types work with extern kernels without adding a copy.

Accepting because this fixes a bug, and converting `bmm` to `mm` is a minor optimization that it is likely ok to miss.

ngimel:
Here's what I had in mind #548, it still allows bmm -> mm optimization for discontiguous inputs, and won't cause more copies than bmm itself would have caused.

PR #533, opened 2022-07-08
Special support for HFPretrainedConfig class objects

anijain2305:
~~@jansel This is for the HF models __get_attribute__ method~~

image

~~I am trying to inline the function, but then it calls object.__get_attribute__. And I am not sure, how to handle it.~~

Update - I also added special handling for HFPretrainedConfig. I am fine with whatever way is better. 

Further update - After offline discussion, decided to go with HFPretrainedConfig.

### DIFF
--- a/torchdynamo/variables/dicts.py
+++ b/torchdynamo/variables/dicts.py
@@ -279,3 +279,32 @@ class DataClassVariable(ConstDictVariable):
                 assert variables.ConstantVariable.is_literal(defaults[name])
                 return variables.ConstantVariable(defaults[name]).add_options(self)
         super(DataClassVariable, self).var_getattr(tx, name)
+
+
+class HFPretrainedConfigVariable(VariableTracker):
+    """
+    Hack for HuggingFace PretrainedConfig
+    """
+
+    @staticmethod
+    def is_matching_cls(cls):
+        try:
+            from transformers.configuration_utils import PretrainedConfig
+
+            return issubclass(cls, PretrainedConfig)
+        except ImportError:
+            return False
+
+    @classmethod
+    def is_matching_object(cls, obj):
+        return cls.is_matching_cls(type(obj))
+
+    def __init__(self, obj, **kwargs):
+        super(HFPretrainedConfigVariable, self).__init__(**kwargs)
+        self.obj = obj
+        assert self.is_matching_cls(type(obj))
+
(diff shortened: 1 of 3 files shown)


torchdynamo/variables/misc.py
```diff
@@ -63,6 +63,8 @@ def call_method(
         inner_fn = self.const_getattr(self, name)
         if inner_fn is object.__init__:
             return LambdaVariable(identity, **options)
+        elif inner_fn is object.__getattribute__:
+            unimplemented(f"Cant handle object getattribute")
```
anijain2305, line 67: Not sure how to handle this.

voznesenskym: Check out the discussion on https://github.com/pytorch/torchdynamo/pull/454/files?

anijain2305: Added a special handling for PreTrainedConfig

torchdynamo/variables/dicts.py
```diff
@@ -279,3 +279,32 @@ def var_getattr(self, tx, name: str) -> "VariableTracker":
                 assert variables.ConstantVariable.is_literal(defaults[name])
                 return variables.ConstantVariable(defaults[name]).add_options(self)
         super(DataClassVariable, self).var_getattr(tx, name)
+
+
+class HFPretrainedConfigVariable(VariableTracker):
```
anijain2305, line 284: This is an alternate way - special handling.

torchdynamo/variables/user_defined.py
```diff
@@ -281,7 +281,24 @@ def var_getattr(self, tx, name):
         options = VariableTracker.propagate(self)
         value = self.value
         source = AttrSource(self.source, name) if self.source else None
-        self._check_for_getattribute()
+
+        try:
```
anijain2305, line 285: This is a more general way - however with issues that I am not sure how to resolve.

torchdynamo/exc.py
```diff
@@ -67,7 +67,7 @@ def add_to_stats(self, category="unimplemented"):
 
 
 def unimplemented(msg: str):
-    if config.debug:
+    if True or config.debug:
```
jansel, line 70: Revert this?

jansel (approved):
I think special casing pretrained config fine way to handle it.

Handling getattribute more generally would be possible, but I don't suspect it will be very common.

PR #534, opened 2022-07-08
Emit in_spec and out_spec, relax some asserts, intermediate step towards non-modified callable

voznesenskym:


torchdynamo/eval_frame.py
```diff
@@ -279,16 +279,28 @@ def dynamo_normalization_capturing_compiler(
...
+            sig_type = signature_types[idx]
+            in_type = in_types[idx]
+            assert (
+                sig_type == in_type
+            ), "Export produced a graph with mismatched type signature {sig_type} vs expected {in_type} for arg {idx}"
+
+    if len(in_sig.parameters) == len(out_sig.parameters):
+        assert_equal_params(input_types)
+    else:
+        from functorch._src.aot_autograd import pytree
```
Chillee, line 293: pytree is also in `torch.utils._pytree`

voznesenskym: Even better, thanks :)

torchdynamo/eval_frame.py
```diff
@@ -279,16 +279,28 @@ def dynamo_normalization_capturing_compiler(
...
+        flat_args, _ = pytree.tree_flatten(args, kwargs)
+        for arg in flat_args:
+            flat_input_types.append(arg.__class__)
+
+        assert len(flat_input_types) == len(
+            out_sig.parameters
+        ), "Flattened inputs length must match out signature parameter lengths."
+        assert_equal_params(flat_input_types)
 
     return (graph, out_guards)
```
jansel, line 305: How would the user call the flattened graph i we don't return the pytree?

voznesenskym: Good question. Where I integrate it now, we call pytree on the caller side, and pass in the flattened args again.

jansel (approved):

Issue #538, opened 2022-07-10
2 Torchbench models failed when unspec int/float is default 


yanboliang:
I run torchbench with the [unspec int/float features](https://github.com/pytorch/torchdynamo/pull/520) and found some bugs. I'm fixing several easy bugs at https://github.com/pytorch/torchdynamo/pull/539, but there are still two models doesn't work well.

* Super_SloMo 
```
SyntaxError: duplicate argument 'self_trainFlowBackWarp_W' in function definition
```

* vision_maskrcnn

Failed the correctness check

jansel:
@yanboliang we should disable #520 until these bugs are fixed.  TorchBench not working will prevent us from running experiments.

anijain2305:
@yanboliang Can we close this?

PR #539, opened 2022-07-10
Fix several bugs of running torchbench with unspec int/float

yanboliang:
I run torchbench with unspec int/float by default and found several bugs. This PR will fix these easy bugs and add workaround to ensure torchbench is still green. The two models which need work around has been track at [here](https://github.com/pytorch/torchdynamo/issues/538).


torchdynamo/variables/builder.py
```diff
@@ -126,7 +133,12 @@ def _common_constants():
             2048,
             4096,
             0.1,
+            0.01,
             0.5,
+            0.05,
+            352,  # Work around for Super_SloMo
```
yanboliang, line 139: The new added above is common constants from torchbench, the bellow three lines are workaround for the two failed models.

jansel (approved):

jansel (changes requested):
Since this doesn't fix all the bugs lets revert this PR with https://github.com/pytorch/torchdynamo/pull/546

Then merge again once torchbench works.

jansel:
Reopen this when it is ready for review

PR #541, opened 2022-07-10
[inductor] Add aten.avg_pool2d_backward lowering

jansel:


tests/test_torchinductor.py
```diff
@@ -2095,6 +2099,48 @@ def fn(a):
...
+                [0, 0],
+                True,
+                False,
+                None,
+            )
+
+        self.common(
+            fn,
+            [
+                torch.randn([2, 4, 7, 7]),
```
ngimel, line 2118: add non-square inputs for more robust testing?

torchinductor/lowering.py
```diff
@@ -1606,6 +1607,155 @@ def fn(idx):
...
+
+                if gradient is None:
+                    # don't need mask for 0, 0
+                    gradient = part
+                else:
+                    mask = ops.and_(
+                        ops.lt(ph, phend),
+                        ops.lt(pw, pwend),
+                    )
+                    gradient = ops.where(mask, ops.add(gradient, part), gradient)
```
ngimel, line 1747: will this be converted into atomic adds where needed?

ngimel (approved):

torchinductor/lowering.py
jansel: This shouldn't be needed, since we are just adding two registers inside the kernel together.  This basically just unrolls the entire window and does it a single step.

PR #542, opened 2022-07-10
[inductor] Add aten.max_pool2d_with_indices_backward lowering

jansel:


torchinductor/lowering.py
```diff
@@ -1538,6 +1537,107 @@ def fn(idx, return_index):
...
+@register_lowering(aten.max_pool2d_with_indices_backward, type_promote=False)
+def max_pool2d_with_indices_backward(
+    grad_output, x, kernel_size, stride, padding, dilation, ceil_mode, indices
+):
+    assert dilation == 1 or all(d == 1 for d in dilation)
+    assert isinstance(x, TensorBox)
+    assert len(kernel_size) == 2
+    assert len(stride) == 2
+    assert padding == 0 or len(padding) == 2
+    assert len(x.get_size()) in (3, 4)
```
desertfire, line 1549: Do we need to add memory format checking here?

jansel: This should work for all memory formats.  This 3/4 check is actually just for omitting the batch dim.

PR #543, opened 2022-07-10
[inductor] Enable min-cut rematerialization

jansel:


Chillee (approved):
LGTM, assuming that all of the non-AOTAutograd additions are intentional :)

torchinductor/config.py
```diff
@@ -71,7 +71,7 @@ class triton:
 
     # limit tiling dimensions
     # Disable tiling until we figure out how tiling and fusion work together
-    max_tiles = 1
```
pyjhzwh, line 74: Should we keep `max_tiles = 1` for fusion?

jansel: If this makes things faster (or helps your testing) feel free to change it.

PR #553, opened 2022-07-11
[inductor] Add lowering for scatter_add

desertfire:
Summary: As title.
There is a remaining issue for the cpp backend.

### DIFF
--- a/torchinductor/lowering.py
+++ b/torchinductor/lowering.py
@@ -1313,8 +1313,8 @@ def index_select(x, dim, indices):
     )
 
 
-@register_lowering(aten.scatter_add, type_promote=False)
-def scatter_add(x, dim: int, index, src):
+@register_lowering(aten.scatter_add_, type_promote=False)
+def scatter_add_(x, dim: int, index, src):
     assert isinstance(x, TensorBox)
     assert isinstance(dim, int)
     assert "int" in str(index.get_dtype())
@@ -1344,7 +1344,6 @@ def scatter_add(x, dim: int, index, src):
             ops.constant(0, x.get_dtype()),
         )
 
-    result = clone(x)
     scatter = ir.Scatter(
         device=x.get_device(),
         dtype=x.get_dtype(),
@@ -1355,11 +1354,11 @@ def scatter_add(x, dim: int, index, src):
     )
     buffer = ir.ComputedBuffer(
         None,
-        ir.MutationLayout(result),
+        ir.MutationLayout(x),
         scatter,
     )
     buffer.name = V.graph.register_buffer(buffer)
-    return result
+    return x
 
 
 @register_lowering(aten.upsample_nearest2d)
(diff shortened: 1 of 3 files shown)


torchinductor/lowering.py
```diff
@@ -1305,6 +1305,55 @@ def fn(idx):
...
+            src_loader(idx),
+            ops.constant(0, x.get_dtype()),
+        )
+
+    result = clone(x)
+    scatter = ir.Scatter(
+        device=x.get_device(),
+        dtype=x.get_dtype(),
+        inner_fn=fn,
+        ranges=x.get_size(),
```
jansel, line 1344: Should ranges be `index.get_size()`?

If we are iterating over the ranges of `x` I would expect we need to broadcast `index` into the same shape as `x`.  I haven't looked up the semantics of scatter_add.

jansel (changes requested):
What is the error we are getting on CPU?

torchinductor/lowering.py
```diff
@@ -1305,6 +1305,55 @@ def fn(idx):
...
+                condition = boundary_check
+            else:
+                condition = ops.and_(condition, boundary_check)
+        return ops.where(
+            condition,
+            src_loader(idx),
+            ops.constant(0, x.get_dtype()),
+        )
+
+    result = clone(x)
```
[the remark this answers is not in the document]
desertfire: I was thinking about doing that after seeing how `index_put` is done. Let me fix it.

My general question is how is in-place op supported in TorchInductor? I think at some point you mentioned there is a functionalization pass in FX and another one in TorchDynamo. Can you share some code pointers to those (if my memory was correct)? I am thinking to give a talk on that to our study group.

torchinductor/lowering.py
desertfire: https://pytorch.org/docs/stable/generated/torch.Tensor.scatter_add_.html#torch.Tensor.scatter_add_ explicitly says `index` and `src` do not broadcast.

jansel: How many total adds are performed?  If it is `index.numel()` then this line should be `index.get_size()`.

torchinductor/lowering.py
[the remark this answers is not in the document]
jansel: Inductor supports codegen of mutation ops.  This is one example of it.  Another is input mutation, which happens in some benchmarks.

AOT Autograd does not support mutation, so for training we functionalize here:
https://github.com/pytorch/torchdynamo/blob/7e1d440939f1e84d3ea6b44c9ebdec892229e75d/torchdynamo/optimizations/normalize.py#L325
That will soon switch to dispatcher functionalization.

`scatter_add()` might be the product of functionalizing `scatter_add_()`.  Unfortunately, it doesn't really make sense to codegen a functional scatter.  So we de-functionalize it in order to generate code.

torchinductor/lowering.py
desertfire: Yes, you are right. The cpu backend crash was a SEGFAULT caused by this. Fixed in the new commit.

There is still one problem remaining. When `index` only contains one element, the generated Triton code produces incorrect result. I vaguely remember seeing you have filed some related Triton issue. If not, I can file one.

jansel (approved):

PR #552, opened 2022-07-11
Decrease CI coverage threshold to 90%

jansel:
This result: https://app.circleci.com/pipelines/github/pytorch/torchdynamo/54/workflows/67d0a129-eeec-4e9b-a4c4-f0226b4fb7ae/jobs/56

Only it 94% for `pyhpc_isoneutral_mixing`.  Hopefully this will reduce false alarms.

PR #555, opened 2022-07-12
[inductor] Add `aten.detach` to decomposition list.

Chillee:
Should probably just always enable it.

Also prohibits some rematerialization from occurring sometimes.


jansel:
The test error is fixed by https://github.com/pytorch/torchdynamo/pull/552

jansel (approved):

PR #563, opened 2022-07-13
[inductor] layout of ExternKernelAlloc

pyjhzwh:
For ExternKernelAlloc kernels, output layout should not be fixed to FixedLayout when inputs layout are not fixed (still FlexbileLayout). Instead, when `decide_layout()`, call `apply_constraint()` that will fix both output layout and input layout to avoid layout mismatch if input layout changed in pick_loop_order.

### DIFF
--- a/torchinductor/ir.py
+++ b/torchinductor/ir.py
@@ -451,22 +451,24 @@ def is_contiguous_storage_and_layout(x):
         return False
 
 
-def as_storage_and_layout(x, freeze=True, want_contiguous=False):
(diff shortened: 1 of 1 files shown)


jansel (approved):

jansel:
Make sure you tested this on torchbench.

pyjhzwh:
> Make sure you tested this on torchbench.

I have tested on torchbench with `python benchmarks/torchbench.py -dcuda --inductor-settings --float16 -n50 --inductor` and all passed, and no noticeable slowdown

PR #564, opened 2022-07-13
Improve mutation analysis

ezyang:
* #568
* __->__ #564

- Add inputs_only option to test if inputs are mutated only.
  This is helpful stopgap for as long as functionalization
  can't handle mutated inputs correctly.

- Don't use data pointer on storage to test for aliasing; instead
  use the storage.  While this is technically less precise (since you can
  have two distinct storages with the same pointer), it has the strong merit of
  actually working correctly when you run the analysis with fake tensors.
  And if you had two storages with the same data pointer anyway, you aren't
  going to get the mutation analysis right anyway as the version counter
  won't get updated.

### DIFF
--- a/torchdynamo/optimizations/training.py
+++ b/torchdynamo/optimizations/training.py
@@ -60,8 +60,23 @@ class AOTAutogradStrategy(object):
             if node.target == torch._C._set_grad_enabled:
                 has_set_grad_enabled = True
 
+        import functorch._src.config
+
+        if functorch._src.config.use_functionalize:
+            # There are two problematic classes we still exclude for now with
+            # functionalization:
+            #   - data mutation of inputs (fixed when we stop recording the
+            #   copy_ directly into the graph)
+            #   - metadata mutation of inputs (fixed if we do an extra partition
+            #   to avoid AOTAutograd on the mutated inputs, or if we some how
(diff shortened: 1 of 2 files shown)


ezyang:
To clarify, functionalization does happen mutated inputs, but because it adds copy_ nodes to the graph the rest of our passes that assume functional IR don't

jansel (approved):

torchdynamo/optimizations/training.py
```diff
@@ -60,8 +60,23 @@ def __init__(self, gm: torch.fx.GraphModule, example_inputs):
             if node.target == torch._C._set_grad_enabled:
                 has_set_grad_enabled = True
 
+        import functorch._src.config
+
+        if functorch._src.config.use_functionalize:
```
jansel, line 65: If this is set, should we disable the dynamo functionalization in normalize_ir?

ezyang: Hmm, probably. I'll read over normalization and if I agree I can disable it.

ezyang: So, it seems, not all of normalization (only Functionalization), and dropping Functionalization in FX will drop un inplacing module, but I think that's OK.

PR #574, opened 2022-07-13
Move random values generation before any graph output codegen

yanboliang:
Fix this [bug](https://github.com/pytorch/pytorch/pull/81299) found by Pytorch-Dynamo CI.

We should move random values generation out of the [```if... else..```](https://github.com/pytorch/torchdynamo/blob/main/torchdynamo/output_graph.py#L255), as either condition can use random value. The ```if``` condition is where only graph output, ```else``` condition is where graph output + python variable reconstruction + side effects.

### DIFF
--- a/tests/test_misc.py
+++ b/tests/test_misc.py
@@ -1706,8 +1706,6 @@ class MiscTests(torchdynamo.testing.TestCase):
             random.seed(1)
             res2 = fn(shape)
 
-        print(res1)
-        print(res2)
         self.assertTrue(same(res1, res2))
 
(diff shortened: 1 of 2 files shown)


yanboliang:
cc @ezyang

voznesenskym:
Can you add some detail as to why moving this code fixes something? everything in `random_calls_instructions` looks pretty safe to throw in anywhere?

yanboliang:
@voznesenskym I added more detail in the description.

jansel (approved):

PR #575, opened 2022-07-14
make triton generate kernels with the same src code each time

ngimel:
Currently, for some of the args we iterate over python set, which is non-deterministic, and results in a slightly different (but correct) kernel each time. Thus we don't hit in triton precompiled cache next time we run the same benchmark. This diff changes iterating over set to iterating over sorted elements of the set. 
Unfortunately, I don't know how to test this, but in my local runs I'm now always hitting in triton cache for the same kernel.

### DIFF
--- a/torchinductor/codegen/common.py
+++ b/torchinductor/codegen/common.py
@@ -598,7 +598,5 @@ class Kernel(CodeGen):
         index = sympy.expand(index)
         index = sympy.simplify(index.subs(V.graph.sizevars.replacements))
         sorted_symbols = sorted(index.free_symbols, key=lambda s: s.name)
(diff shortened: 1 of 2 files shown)


Chillee:
LGTM, but I wonder whether we want some kind of more structured way to ensure determinism.

jansel (approved):

Chillee:
I think it's reasonable to add a test for this no? Take some list of kernels, generate them multiple times, make sure they output the same string?

ngimel:
I think within one process results are usually deterministic (though I didn't stress-test it), but running in a different process produces different results. I could test that too though.

PR #577, opened 2022-07-14
Add dynamo benchmark README (runbook)

wconstab:


benchmarks/README.md
```diff
@@ -0,0 +1,23 @@
...
+Training benchmarks approximate training by running the model forward, using a .sum().backward() call in place of the native loss function and skips the optimizer entirely.
+
+Inference benchmarks and Training benchmarks measure correctness by comparing dynamo and eager model outputs given fixed inputs and seeds.
+
+## Runbook
+There are a lot of flags in the benchmark runner, and it can be confusing to know which settings to use or what machine to run it on.  In order to support apples-to-apples comparison, we have provided the following 'standard' settings.
+
+We run benchmarks on AWS machines using 8xNVidia A100 64GB cards and (TODO) CPU.  (TODO instance type and OS/cuda versions) 
+
+### Training
```
wconstab, line 17: @anijain2305 @jansel can you paste some commands in a comment and i'll add them here?

benchmarks/README.md
```diff
@@ -0,0 +1,23 @@
...
+* TorchInductor + Triton
+* NVFuser
+
+
+### Inference
```
wconstab, line 22: @jansel also here? Idk if it's important to maintain commands for every backend but whichever ones you think are popular to run

benchmarks/README.md
```diff
@@ -0,0 +1,23 @@
...
+The runner integrates with models from TorchBenchmark, HuggingFace and TIMM suites and covers both training and inference.
+
+Training benchmarks approximate training by running the model forward, using a .sum().backward() call in place of the native loss function and skips the optimizer entirely.
+
+Inference benchmarks and Training benchmarks measure correctness by comparing dynamo and eager model outputs given fixed inputs and seeds.
+
+## Runbook
+There are a lot of flags in the benchmark runner, and it can be confusing to know which settings to use or what machine to run it on.  In order to support apples-to-apples comparison, we have provided the following 'standard' settings.
+
+We run benchmarks on AWS machines using 8xNVidia A100 64GB cards and (TODO) CPU.  (TODO instance type and OS/cuda versions)
```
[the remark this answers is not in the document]
ngimel: 40 GB cards, benchmarks are not distributed, so 8xGPUs doesn't matter. 
cuda is usually 11.6, although we didn't standardize on it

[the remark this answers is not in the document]
wconstab: looks like we use 'p4d.24xlarge' i'll add that

benchmarks/README.md
wconstab: (from offline chat - any objections to these?)
For nvfuser
`--training  --accuracy-aot-ts-mincut --inductor-settings --float16 --isolate --nvfuser`
(won't use cuda graphs though, don't know how to make it)

For inductor
`--training --inductor --float16 --isolate`
whether cudagraphs are used or not is controlled in torchinductor/config.py files, by default it's on.

ngimel: Sure, we can list flags specifically instead of using `--inductor-setting`.

anijain2305: Hi Will, I am working on a PR that generates this commands automatically. I plan to use this to run commands, and then parse the results in some pretty form. The result of the command geeration is this

~~~
# Commands for torchbench for device=cuda, dtype=float32 for training
python benchmarks/torchbench.py --float32 -dcuda --output=ts_nvfuser_torchbench_float32_training_cuda.csv --training --nvfuser --speedup-ts --use-eval-mode --isolate
python benchmarks/torchbench.py --float32 -dcuda --output=aot_nvfuser_torchbench_float32_training_cuda.csv --training --nvfuser --accuracy-aot-ts-mincut --use-eval-mode --isolate

# Commands for huggingface for device=cuda, dtype=float32 for training
python benchmarks/huggingface.py --float32 -dcuda --output=ts_nvfuser_huggingface_float32_training_cuda.csv --training --nvfuser --speedup-ts --use-eval-mode --isolate
python benchmarks/huggingface.py --float32 -dcuda --output=aot_nvfuser_huggingface_float32_training_cuda.csv --training --nvfuser --accuracy-aot-ts-mincut --use-eval-mode --isolate

~~~

benchmarks/README.md
anijain2305: image

This is what I have right now locally.

wconstab: cool. hmm, it almost makes me wonder if we should be putting these strings of flags into the README directly,

or instead we should add a new subparser to torchbench.py where strings like `ts_nnc_traininng` map directly to a set of args, and it gets maintained in the code instead of the README.

Then in the readme we can just point users to run simple commands like `torchbench.py --training ts_nvfuser` or something?

wconstab:
> We should also add instructions on installing torchbench here (it's surprisingly non-trivial, or there are better ways that I just don't know about).

hmm i have run into various issues at times but generally have found that the install procedure works.  I think we should only point users to this link, but if their process/instructions are not good enough it should be fixed.
https://github.com/pytorch/benchmark#installation
cc @xuzhao9

ngimel:
>Note that when building PyTorch from source, torchtext and torchvision must also be built from source to make sure the C APIs match.
This is a step that gets me.

benchmarks/README.md
anijain2305: I think we can do some cleanup. But overall, reducing the total number of args to shorten the command line might be a difficult goal to achieve. We have things options like dtype (float32/float16/amp) and device (cuda/cpu). And such things are likely to change.

benchmarks/README.md
jansel: For inductor:
```
--training --inductor --float16 -dcuda
```

Don't forget `-dcuda`.   Perhaps we should make `-dcuda` the default.

wconstab:
@anijain2305 maybe we just merge this one now and then you update the TODOs with the relevant commands after https://github.com/pytorch/torchdynamo/pull/585 lands?  Need a stamp if you agree.

anijain2305 (approved):

PR #579, opened 2022-07-14
Enable torch.nn.Parameter to be dictionary key

mlazos:
This functionality is needed to trace through optimizers.

Summary of changes
- Update dictionaries to store weakrefs to parameter keys in f_globals
- Update DICT_KEY guard to check that parameter id's match


torchdynamo/variables/dicts.py
```diff
@@ -144,6 +177,31 @@ def unpack_var_sequence(self, tx):
...
+            or isinstance(key, TensorVariable)
+            and key.parameter_value is not None
+        )
+
+    @classmethod
+    def _key_to_var(cls, tx, key, **options):
+        from .builder import VariableBuilder
+
+        if isinstance(key, torch.nn.Parameter):
+            return VariableBuilder(tx, GlobalSource(global_key_name(key)))(key)
```
mlazos, line 200: TODO: Check that this gets reconstructed correctly after changing to weakref

torchdynamo/guards.py
```diff
@@ -351,10 +355,14 @@ def DICT_KEYS(self, guard):
 
         code = list()
         code.append(f"___check_type_id({ref}, {self.id_ref(t)})")
-        code.append(f"{ref}.keys() == {set(value.keys())!r}")
+        code.append(f"___dict_param_key_ids({ref}) == {set(dict_param_key_ids(value))}")
```
jansel, line 358: This might add some overhead.  We check if there are any param keys, and generate the old version in the common case of no params in the dict keys.

torchdynamo/utils.py
```diff
@@ -430,6 +430,18 @@ def tuple_iterator_getitem(it, index):
...
+def dict_param_key_ids(value):
+    return set([id(k) for k in value.keys() if isinstance(k, torch.nn.Parameter)])
+
+
+def dict_const_keys(value):
+    return set(k for k in value.keys() if not isinstance(k, torch.nn.Parameter))
+
+
+def global_key_name(key):
+    return f"__dict_key_{id(key)}"
```
jansel, line 442: This could collide with another dict key that has been deleted.  We should check for that case.

jansel (changes requested):

torchdynamo/variables/builder.py
```diff
@@ -164,19 +166,39 @@ def _wrap(self, value):
             guards = self.make_guards(GuardBuilder.EQUALS_MATCH)
             return RangeVariable(value=value, guards=guards)
         elif istype(value, (dict, collections.OrderedDict)) and all(
-            map(ConstantVariable.is_literal, value.keys())
+            map(
+                lambda k: ConstantVariable.is_literal(k)
```
anijain2305, line 170: use your new function - `is_valid_key`?

anijain2305:
Overall, looks good to me. Once you have optimizers working, would you be interested in adding optimizers in benchmarks/common.py so that we can test torchbench and huggingface models with optimizers?

torchdynamo/variables/tensor.py
```diff
@@ -147,17 +148,21 @@ def context():
                     example_value = fake_wrapper(example_value)
 
         if isinstance(example_value, torch.Tensor):
+            is_parameter = isinstance(example_value, torch.nn.Parameter)
+            parameter_value = initial_example_value if is_parameter else None
```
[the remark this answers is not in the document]
mlazos: So if fake tensors are enabled, the original value is converted to a FakeTensor. We need the original value to be the Parameter key so we can properly extract the id.

torchdynamo/variables/builder.py
mlazos: There's some subtlety here, when creating a dict from an already created dictionary the keys are not TensorVariables yet, they're raw python values.

mlazos:
> Overall, looks good to me. Once you have optimizers working, would you be interested in adding optimizers in benchmarks/common.py so that we can test torchbench and huggingface models with optimizers?



> Overall, looks good to me. Once you have optimizers working, would you be interested in adding optimizers in benchmarks/common.py so that we can test torchbench and huggingface models with optimizers?



> Overall, looks good to me. Once you have optimizers working, would you be interested in adding optimizers in benchmarks/common.py so that we can test torchbench and huggingface models with optimizers?



> Overall, looks good to me. Once you have optimizers working, would you be interested in adding optimizers in benchmarks/common.py so that we can test torchbench and huggingface models with optimizers?

Let me see if there are other higher priority follow ups that are needed from the distributed team but I think this would be a good continuation of this.

jansel (approved):

desertfire:
@mlazos , this PR triggers an assertion in several torchbench tests, e.g. `./benchmarks/torchbench.py -d cuda --backend eager -k hf_Albert`. If it is not an easy fix, we should revert this first.

mlazos:
> 

I'll take a look at the asserts and and either have a followup fix or revert if necessary

PR #580, opened 2022-07-15
add triton sin lowering

yushangdi:


jansel (changes requested):
Let's also add a line to the C++ backend similar to this:
https://github.com/pytorch/torchdynamo/blob/1e4d17e2fc1e61894f5dac08aabdbdc71378bf5f/torchinductor/codegen/cpp.py#L127

I think it will work without that, but we are relying on cos being defined in global scope.

yushangdi:
> Let's also add a line to the C++ backend similar to this:
> 
> https://github.com/pytorch/torchdynamo/blob/1e4d17e2fc1e61894f5dac08aabdbdc71378bf5f/torchinductor/codegen/cpp.py#L127
> 
> I think it will work without that, but we are relying on cos being defined in global scope.

@jansel I think is already there? line 123 right above

jansel (approved):

jansel:
Whoops you are correct, my fault!

PR #581, opened 2022-07-15
enable specifying batch size in the command line

ngimel:


ngimel:
Nvidia team, can you please make sure that default batch sizes make sense for HF models? cc @kevinstephano, @xwang233

anijain2305 (approved):

kevinstephano:
> Nvidia team, can you please make sure that default batch sizes make sense for HF models? cc @kevinstephano, @xwang233

The batch sizes looked good to me.

PR #586, opened 2022-07-15
Allow (certain) builtins without tensor args into the graph

### LINKED ISSUE #576: [Export] List variable elements are not properly captured (due to return usage?)
For a (trimmed down from a prod model) function like this:

```
def pre_attention_state_ops(input_layer, input, mems, state):
    pre_mems = state[0]
    state[0] = torch.cat([state[0], mems], dim=0)[-input_layer.max_memory_size :, :, :]
    lc_key = state[1]
    lc_val = state[2]

    pre_mems = pre_mems[: pre_mems.shape[0] - input_layer.min2(input_layer.memory_offset, pre_mems.shape[0])]
    return pre_mems, state, lc_key, lc_val
```

we would expect graph placeholders of:

`mems` (via direct usage above)
`state_0` (via both reading and writing `state[0]`)
`state_1` (via ` lc_key = state[1]` which then goes to be used in return)
`state_2` (via ` lc_val = state[2]` which then goes to be used in return)

Instead, we get a graph of: 

image

This is incorrect - and, in fact, leads to correctness bugs in export.

If we take the graph, and call it with the *correct* function inputs, it rejects them due to the `forward` function not even expecting state_1 and state_2! If we trim it down to what it expects, it obviously produces incorrect results.

voznesenskym:
Fixes #576

### DIFF
--- a/torchdynamo/variables/builtin.py
+++ b/torchdynamo/variables/builtin.py
@@ -253,10 +253,9 @@ class BuiltinVariable(VariableTracker):
         # but should still be captured for a fuller, more correct graph.
         # For these cases, we defer to self.can_insert_in_graph as good enough, which checks fx.
         # TODO(voz): Remove the allowlist, and maybe just let it all through? Perhaps config driven?
-        should_check_tensor_args = self.fn not in self._allowed_without_tensor_args()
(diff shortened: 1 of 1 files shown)


voznesenskym:
Not ready for review yet :)

PR #587, opened 2022-07-15
Fall back to calling allclose when cosine_similarity returns invalid value

desertfire:
Summary: inductor enables --cosine as the default when running torchbench. Some training runs fail because of this issue.


jansel (approved):
Interesting, I am guessing we are seeing some nans in benchmarks?

desertfire:
> Interesting, I am guessing we are seeing some nans in benchmarks?

Two cases:
1. Tensors with inf in it, cosine_similarity return nan;
2. For two equal all zero tensors, cosine_similarity returns 0 (which technically should be undefined?)

Issue #559, opened 2022-07-12
[inductor] Background_Matting training failure


desertfire:
Failing with RuntimeError: CUDA: Error- invalid value

desertfire:
The problem can be reproduced with the following simplified Triton code,

```
import torch
from torch import empty_strided
import triton
import triton.language as tl

from torchinductor.triton_ops.autotune import pointwise_heuristics
from torchinductor.triton_ops.autotune import grid

@pointwise_heuristics(size_hints=[256, 262144])
@triton.jit
def kernel2(in_ptr0, out_ptr0, xnumel: tl.constexpr, ynumel: tl.constexpr, XBLOCK : tl.constexpr, YBLOCK : tl.constexpr):
    xoffset = tl.program_id(0) * XBLOCK
    xindex = xoffset + tl.reshape(tl.arange(0, XBLOCK), [XBLOCK, 1])
    xmask = xindex < xnumel
    yoffset = tl.program_id(1) * YBLOCK
    yindex = yoffset + tl.reshape(tl.arange(0, YBLOCK), [1, YBLOCK])
    ymask = yindex < ynumel
    y2 = yindex
    x3 = xindex
    tmp0 = tl.load(in_ptr0 + y2 + (262144*x3), xmask & ymask).to(tl.float32)
    tl.store(out_ptr0 + y2 + (262144*x3), tmp0, xmask & ymask)

def call():
    buf1 = empty_strided((3, 64, 512, 512), (16777216, 262144, 512, 1), device='cuda', dtype=torch.float16)
    buf265 = empty_strided((3, 64, 512, 512), (16777216, 262144, 512, 1), device='cuda', dtype=torch.float16)
    kernel2[grid(192, 262144)](buf1, buf265, 192, 262144)
    return (buf265)

if __name__ == "__main__":
    call()
```

jansel:
Is this just a Triton bug?

If so we should remove the `pointwise_heuristics` wrapper and submit a bug to openai/triton.

jansel:
This kernel just looks like a tiled memcpy.  I am surprised it fails.

desertfire:
It has something to do with autotuning config at this line, https://github.com/pytorch/torchdynamo/blob/62e7e1db7ffe6a142dd924e6776772dbe021297f/torchinductor/triton_ops/autotune.py#L134

[...]

PR #588, opened 2022-07-15
[inductor] Explicitly emit None output as None

desertfire:
Summary:
Before:
```
constant0 = None  # 3855eccb97cd9b8c1773347b7d77322fdbdb736f784af24868acc7a6a6c64413

def kernel2(...):
  ...
  return (buf417, buf412, constant0, ...)

```

After:
```
def kernel2(...):
  ...
  return (buf417, buf412, None, ...)
```

The Before code is okay by itself, however, a problem exists when
autograd is evolved and we may see errors like,
```
RuntimeError: Function CompiledFunctionBackward returned an invalid gradient at index 2 - got [0] but expected shape compatible with [64]
```

### DIFF
--- a/torchinductor/ir.py
+++ b/torchinductor/ir.py
@@ -1341,6 +1341,11 @@ class ConstantBuffer(InputBuffer):
         return ConstantBuffer(V.graph.constant_name(self.name, device), self.layout)
 
 
+class NoneAsConstantBuffer(IRNode):
+    def codegen_reference(self):
+        return "None"
+
+
 @dataclasses.dataclass
 class ComputedBuffer(Buffer):
     data: Loops
@@ -1751,7 +1756,7 @@ class ExternKernel(InputsKernel):
     @classmethod
     def realize_input(cls, x):
         if x is None:
-            return V.graph.add_tensor_constant(torch.tensor(()))
+            return NoneAsConstantBuffer()
         if isinstance(x, Constant):
             return V.graph.add_tensor_constant(
                 torch.tensor(x.value, dtype=x.get_dtype(), device=x.get_device())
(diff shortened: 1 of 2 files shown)


desertfire:
For https://github.com/pytorch/torchdynamo/issues/559

torchinductor/ir.py
```diff
@@ -1751,7 +1756,7 @@ def copy_input(x):
     @classmethod
     def realize_input(cls, x):
         if x is None:
-            return V.graph.add_tensor_constant(torch.tensor(()))
+            return V.graph.add_tensor_constant(torch.tensor(()), is_none=True)
```
jansel, line 1759: Could we just return a `ir.NoneAsConstantBuffer()` directly here?  Is the tensor_constant needed?

desertfire: We could, if we also take care of where it is used. Let me update it.

torchinductor/ir.py
```diff
@@ -1751,7 +1756,7 @@ def copy_input(x):
     @classmethod
     def realize_input(cls, x):
         if x is None:
-            return V.graph.add_tensor_constant(torch.tensor(()))
+            return NoneAsConstantBuffer()
```
desertfire, line 1759: We may even simply use `None` here, but I think having NoneAsConstantBuffer can be less confusing.

Chillee:
Interesting. So basically, previously, it was returning a 0-sized tensor constant which caused issues with autograd?

jansel (approved):

jansel:
Yeah, this was my fault.  I figured None's didn't matter and could take a shortcut in implementing them as 0 tensors.  I was wrong.

PR #589, opened 2022-07-15
[benchmark] Add options to dump raw timing metrics from speedup experiment

xwang233:
per title


xwang233:
cc @jansel @ngimel @ptrblck

benchmarks/common.py
```diff
@@ -323,10 +325,12 @@ def speedup_experiment(args, model_iter_fn, model, example_inputs):
     pvalue = ttest_ind(timings[:, 0], timings[:, 1]).pvalue
     median = np.median(timings, axis=0)
     speedup = median[0] / median[1]
+    if args.dump_raw_metrics:
```
ngimel, line 328: why are we saving to numpy file?

xwang233: we can dump to json files if you prefer that way?

xwang233: having raw metric numbers can help us estimate the uncertainties in timing measurements

jansel: Are you planning to write scripts to analyze this?  What analysis are you planning to do?

We could also just add more metrics to the `.csv` already generated.

xwang233: We do have scripts to analyze this. We currently use standard deviation plus vertical error bars to estimate the uncertainty.

We can add those metrics to the .csv file, while I also think to get the raw metrics is also important. Besides, this flag is off by default and doesn't have noticeable effect to the script?

jansel: How would someone else use this if they don't have those scripts?  Could we add the error bars to the output so in the common case we don't need an extra step?

xwang233:
Hi @ngimel , are you interested in merging this PR as its current shape?

ngimel:
We can merge in the current shape, but please share the scripts you are using to process, and ideally we should be outputting results with error bars to the main report as @jansel suggested.

xwang233:
Hi @ngimel , thanks for the update. We currently use median of results as the timing metric and use the standard deviation of the middle 25%~75% values to estimate the error. The plotter is https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.errorbar.html.

I can put the medians and standard deviations to the output csv files. There were two concerns I have
- The current csv files only output 4 digits which is not precise enough. The numpy dump is precise.
- If the distributions of the raw timings change, it could be beneficial to use another way to estimate the error.

jansel (approved):

PR #596, opened 2022-07-18
[Export] First part of dynamo spec capture, spec format, capture output, check in export

voznesenskym:


torchdynamo/output_graph.py
```diff
@@ -279,6 +281,7 @@ def compile_subgraph(self, tx, partial_convert=False):
             and self.side_effects.is_empty()
         ):
             # optimization to generate better code in a common case
+            # TODO(voz): Add spec here too before land
```
voznesenskym, line 284: Self review: This will be added before land, if the rest of the direciton looks fine

torchdynamo/spec.py
```diff
@@ -0,0 +1,174 @@
...
+                    _append(current_collection, prior)
+                    # Now that this list is closed, we can have further additions to the prior collection
+                    # Which is the collection this collection was nested in
+                    current_collection = prior
+                else:
+                    # Closing with nothing in the stack, this is the end
+                    result = current_collection
+                    current_collection = None
+            else:
+                assert False, f"{current_element} is not yet supported"
```
voznesenskym, line 160: Dict supported will be added in a subsequent PR, this one has enough going on.

torchdynamo/codegen.py
```diff
@@ -163,6 +174,13 @@ def __call__(self, value, allow_cache=True):
 
         self.top_of_stack = value
 
+    def _add_to_spec(self, value, add_open):
+        if isinstance(value, (ListVariable, TupleVariable, ConstDictVariable)):
+            if add_open:
```
jansel, line 179: Should these just be two functions? (open and close)

torchdynamo/spec.py
```diff
@@ -0,0 +1,174 @@
+from enum import Enum
+from enum import unique
+from typing import List
+
+import torch
+
+
+class Spec:
```
jansel, line 8: Should use full words for class names, what about `ExportSpecification`?

torchdynamo/spec.py
```diff
@@ -0,0 +1,174 @@
...
+        collection_stack = []
+        current_collection = None
+        idx = 0
+
+        def _append(e, collection):
+            if isinstance(collection, list):
+                collection.append(e)
+            elif isinstance(collection, tuple):
+                # TODO(voz): Optimize with forward scanning, this isnt that cheap
+                collection += (e,)
```
jansel, line 115: tuples are not mutable, so I think this just changes the local var and not the caller's copy.

torchdynamo/spec.py
```diff
@@ -0,0 +1,174 @@
...
+                    current_collection = prior
+                else:
+                    # Closing with nothing in the stack, this is the end
+                    result = current_collection
+                    current_collection = None
+            else:
+                assert False, f"{current_element} is not yet supported"
+
+            if len(reversed_elements) > 0:
+                current_element = reversed_elements.pop()
```
jansel, line 163: when would we expect this to happen?

torchdynamo/variables/lists.py
```diff
@@ -132,7 +133,14 @@ def python_type(self):
 
     def reconstruct(self, codegen):
         codegen.foreach(self.items)
-        return [create_instruction("BUILD_LIST", len(self.items))]
+        list_instructions = [create_instruction("BUILD_LIST", len(self.items))]
```
jansel, line 136: take a pass through to revert unrelated changes

torchdynamo/spec.py
```diff
@@ -0,0 +1,174 @@
...
+        while current_element is not None:
+            if current_element == Spec.Element.TENSOR:
+                if current_collection is None:
+                    # No current collection, make one
+                    current_collection = (tensors[idx],)
+                else:
+                    # Add tensor to current collection
+                    _append(tensors[idx], current_collection)
+                # advance tensor pointer
+                idx += 1
```
jansel, line 130: This seems to assume tensors are always consumed in the same order and exactly once?  Is this always true?

jansel (changes requested):

torchdynamo/codegen.py
voznesenskym: I had it that way, but the isinstance check felt grosser to duplicate, no strong opinions

torchdynamo/spec.py
voznesenskym: Good catch, will add test

torchdynamo/variables/lists.py
voznesenskym: Will do

torchdynamo/spec.py
```diff
@@ -0,0 +1,174 @@
...
+                collection.append(e)
+            elif isinstance(collection, tuple):
+                # TODO(voz): Optimize with forward scanning, this isnt that cheap
+                collection += (e,)
+            else:
+                assert False, f"Unsupported collection type {collection.__class__}"
+
+        result = None
+        current_element = reversed_elements.pop()
+        while current_element is not None:
```
jansel, line 121: At least for what we have so far, this could totally just be a `pytree.TreeSpec()`.  The features supported by a pytees are a superset.

voznesenskym: Does pytree offer the ability to build it up iteratively, as we do with Spec? That's the biggest gap here, so far.

Issue #602, opened 2022-07-18
ValueError: tmpX is not defined"fastNLP_Bert","vision_maskrcnn","maml",


desertfire:
`./torchbench.py -dcuda --inductor --float16 --no-skip -k maml`

```
ValueError: tmp0 is not defined
...

def kernel9(in_ptr0, in_ptr1, in_ptr2, in_ptr3, out_ptr1, xnumel: tl.constexpr, rnumel: tl.constexpr, XBLOCK : tl.constexpr, RBLOCK : tl.constexpr):
    xoffset = tl.program_id(0) * XBLOCK
    xindex = xoffset + tl.reshape(tl.arange(0, XBLOCK), [XBLOCK, 1])
    xmask = xindex < xnumel
    rbase = tl.reshape(tl.arange(0, RBLOCK), [1, RBLOCK])
    tmp2 = tl.load(in_ptr2 + tmp0, None).to(tl.float32)
                             ^
    _tmp9 = tl.zeros([XBLOCK, RBLOCK], tl.float32) + 0
    for roffset in range(0, rnumel, RBLOCK):
        rindex = roffset + rbase
        rmask = rindex < rnumel
        r0 = rindex
        tmp0 = tl.load(in_ptr0 + r0, rmask)
        tmp4 = tl.load(in_ptr3 + r0, rmask).to(tl.float32)
        tmp1 = tl.load(in_ptr1 + tmp0 + (5*r0) + tl.zeros([XBLOCK, RBLOCK], tl.int32), xmask & rmask).to(tl.float32)
        # Move tmp2 assignment here would make it work
        tmp3 = tmp1 + tmp2
        tmp5 = tl.libdevice.log(tmp4) if tmp4.dtype is tl.float64 else tl.log(tmp4)
        tmp6 = tmp3 - tmp5
        tmp7 = -tmp6
        tmp8 = tmp7.to(tl.float32)
        _tmp9 = tl.where(xmask & rmask, _tmp9 + tmp8, _tmp9)
    tmp9 = tl.reshape(tl.sum(_tmp9, 1), [XBLOCK, 1])
    for roffset in range(0, rnumel, RBLOCK):
        rindex = roffset + rbase
        rmask = rindex < rnumel
        tmp10 = 5
        tmp11 = tmp9 / tmp10
        tmp12 = tmp11.to(tl.float32)
        tl.store(out_ptr1 + 0 + tl.zeros(tmp12.shape, tl.int32), tmp12, xmask)
```

[...]

PR #607, opened 2022-07-19
[inductor] Fix a load promotion related bug

desertfire:
Summary: This fixes a tmp* not defined problem seen in maml, see
https://github.com/pytorch/torchdynamo/issues/602

### DIFF
--- a/torchinductor/codegen/triton.py
+++ b/torchinductor/codegen/triton.py
@@ -591,8 +591,11 @@ class TritonKernel(Kernel):
         if upcast:
             line += ".to(tl.float32)"
 
-        if self.inside_reduction and "rmask" not in mask:
+        if self.inside_reduction and "rmask" not in mask and self.loads != self.compute:
             # can lift a common load outside of reduction loop
+            # One exception is when this is invoked inside of an indirect_load,
+            # because there may be dependency, Kernel.indirect_load sets self.loads
+            # to self.compute which we can use as a checking here.
             tmp = self.cse.generate(self.body, line)
         else:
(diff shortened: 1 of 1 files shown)


torchinductor/codegen/triton.py
```diff
@@ -591,8 +591,11 @@ def load(self, name: str, index: sympy.Expr, upcast: bool = False):
         if upcast:
             line += ".to(tl.float32)"
 
-        if self.inside_reduction and "rmask" not in mask:
+        if self.inside_reduction and "rmask" not in mask and self.loads != self.compute:
```
jansel, line 594:
```suggestion
        if self.inside_reduction and "rmask" not in mask and self._load_mask is None:
```

Might be more readable and future-proof.  Should do the same thing.

jansel (changes requested):
Thanks for finding this!  Minor nit.

torchinductor/codegen/triton.py
desertfire: Hmm, I tried your suggestion locally, but it doesn't solve the problem. I guess `self._load_mask` is not a good indictor of whether it is an indirect-load? I can add another flag if that's more readable.

```
tmp0 = tl.load(in_ptr0 + r0, rmask)
tmp2 = tl.load(in_ptr2 + tmp0, None).to(tl.float32)
```

jansel: Ah you are be right.  Let's land this version then.

jansel (approved):

PR #608, opened 2022-07-19
[inductor] Use AOT Autograd for inference

jansel:


torchinductor/lowering.py
```diff
@@ -786,8 +786,9 @@ def arange(
     )
 
 
-@register_lowering(torch.linspace)
-def linspace(start, end, steps, *, dtype=None, device=None):
+@register_lowering([torch.linspace, aten.linspace])
```
Chillee, line 789: When are you seeing `torch.foo` operators btw?

Chillee (approved):
LGTM

torchinductor/lowering.py
jansel: python_key_normalize captures `torch.*` tensor factories.

PR #609, opened 2022-07-19
[inductor] Add upsample_bilinear2d and grid_sampler_2d

jansel:


Chillee (commented):
What's the issue with "indexing not getting traced"?

jansel:
> What's the issue with "indexing not getting traced"?

`torch.arange` becomes a tensor constant in the captured graph.

Chillee:
@jansel So if that was fixed would be able to use the decomposition?

jansel:
> @jansel So if that was fixed would be able to use the decomposition?

Yes, assuming the perf is on par -- which I expect it would be.

There are a few minor things:
- This version has 3 fewer loads than the open PR, but that can be fixed in the decomp.
- We might want to try to find a way to do the `.realize()` call on the inputs from the decomp, but I think that could be done by adding a prim.realize().

Chillee:
@jansel I guess, similar question for ops like `constant_pad_nd`. If we wrote those using indexing expressions, would that be feasible to use the decomp for?

Something like this:

```
def constant_pad_nd(input, padding):
    indexing = torch.arange(input.shape[0] + padding[0] + padding[1])
    input_indexing = torch.clamp(indexing - padding[0] + 1, min=0, max=input.shape[0] - 1)
    return torch.where((indexing > padding[0]) & (indexing < padding[0] + input.shape[0]) , input[input_indexing], 0)
```

The current one in core looks more like

```
def constant_pad_nd(input, padding):
    new_val = torch.zeros(input.shape[0] + padding[0] + padding[1])
    new_val_range = torch.narrow(input, padding[0], input.shape[0] + padding[0])
    new_val_range.copy_(input)
    return new_val
```

jansel:
Yes, in general if we can make decomps have comparable perf we should use decomps.

For constant_pad_nd, your version would not generate a conditional load -- where a lowering version would generate a conditional load.  I'm not sure if that matters, we should benchmark it.

Chillee:
btw, I can confirm that https://github.com/pytorch/pytorch/pull/81598 fixes the `arange` issues.

torchinductor/lowering.py
```diff
@@ -641,71 +611,88 @@ def bernoulli_(x, *args):
...
-make_fallback(aten.grid_sampler_2d)
+make_fallback(aten.grid_sampler_2d_backward)
 make_fallback(aten.native_batch_norm_backward)
 make_fallback(aten.randperm)
 make_fallback(aten.reflection_pad2d_backward)
 make_fallback(aten.sort)
 make_fallback(aten.topk)
 make_fallback(aten.upsample_bilinear2d_backward)
 make_fallback(aten.upsample_nearest2d_backward)
+make_fallback(aten.mse_loss_backward)
```
Chillee, line 695: I think there's a decomposition for this.

Chillee (approved):
LGTM.

PR #612, opened 2022-07-19
Adding TIMM models

anijain2305:
Adding TIMM bechmark suite

* TIMM has close to 700 models. For now, we have chosen one model from each family, resulting in 69 models.
* The current AOT Autograd + NVFuser results look like follows

image

There are several issues here to get good coverage
* Cold start memory spike - Even after using fake tensors, I am seeing many models unable to fit in GPU memory.
* TorchInductor does not pass on any model because of missing lowerings. I think adding a few lowerings can give us a better picture (cc @jansel).

cc @ngimel @Chillee


ngimel:
cc @xwang233

jansel (approved):

ngimel:
@xwang233 and nvidia team, does the list of models and the results we are obtaining look as expected?
@anijain2305 for the chart above, what are the batch sizes and is it fp16?

ngimel (approved):

anijain2305:
@xwang233 Please wait for a day or two. I have few more fixes to get broader coverage.

@ngimel The batch sizes are present in the uploaded file - timm_models_list.txt. These are basically the largest batch sizes that could fit on the GPU with eager. Some of the models, when going through Dynamo-AOT-nvfuser chain, give OOM issues. For them, I am going half the batch size. I will open an issue to track this cold start memory spike.

benchmarks/timm_models.py
```diff
@@ -0,0 +1,284 @@
...
+    filename = "benchmarks/" + filename
+with open(filename, "r") as fh:
+    lines = fh.readlines()
+    lines = [line.rstrip() for line in lines]
+    for line in lines:
+        model_name, batch_size = line.split(" ")
+        TIMM_MODELS[model_name] = int(batch_size)
+
+
+# TODO - Figure out the reason of cold start memory spike
```
anijain2305, line 46: @ngimel These are the models that pass eager, but fail when we go through the Dynamo-AOT-Nfuser pipeline. We have to understand why it fails.

benchmarks/timm_models_list.txt
```diff
@@ -0,0 +1,69 @@
+adv_inception_v3 128
```
anijain2305, line 1: @ngimel These are the batch sizes. This PR also contains a batch size finder, which finds the largest eager batch size that could fit on the GPU. These batch sizes are computed using that finder,

anijain2305:
@xwang233  These are the current results

For channels first
image


For channels last
image

* A lot of channels last issue is Accuracy failures. Will have to look into more detail here. But skipping for this PR.


Will be good if you can see if these numbers look similar to what you have seen.

anijain2305:
@Chillee mentioned that functionalize might resolve the accuracy issues for channel-last models. Just noting down here to further debug this.

benchmarks/common.py
```diff
@@ -872,6 +917,12 @@ def parse_args():
 
     parser.add_argument("--float16", action="store_true", help="cast model to fp16")
     parser.add_argument("--float32", action="store_true", help="cast model to fp32")
+    parser.add_argument(
```
ngimel, line 920: can you please add warning here that this arg is ignored for non-timm models?

xwang233:
Hi @ngimel @anijain2305 , thank you for sharing your benchmark results for the TIMM models! Below are TIMM training benchmark results we tested on A100, using nvfuser+TorchScript and nvfuser+AOT_autograd (from functorch). We don't have a nvfuser+torchdynamo result at this moment but we're working on it. I think the general speedup trends we got are similar as yours.

The numbers appending model names are batch sizes.

cc @csarofeen @ptrblck

scurve_plot_1658774510088298962
scurve_plot_1658774546477268000

ngimel:
@xwang233 please not that torchbench benchmarking for now doesn't include optimizer/zeroing grads, whereas benchmark.py in timm does. 
@anijain2305, @xwang233 can we get results in a spreadsheet so that's easier to compare?

PR #614, opened 2022-07-19
[Export] Add arg input and result output reordering, bypass support

voznesenskym:


torchdynamo/utils.py
```diff
@@ -633,3 +633,26 @@ def recompile_reasons(code):
             rpt += "No cache-limited recompilations detected.\n"
 
         return rpt
+
+
+def recursive_allclose(lhs, rhs, rtol=1e-5, atol=1e-8):
```
jansel, line 638: is this the same as testing.same()?

voznesenskym: Yes, but  (A) its used in non testing code, so I figured we should have a smaller, separate impl w/o the `elif type(a).__name__ in (` stuff and (B) I ripped it off from somewhere else.

I could ostensibly consolidate the testing impl to use this one, but I'd rather not have non-testing code import testing code.

jansel: Lets combine them.

torchdynamo/eval_frame.py
```diff
@@ -256,50 +273,83 @@ def dynamo_normalization_capturing_compiler(
...
+    # graph.recompile()
+
+    # 2) NOTE: This may not be sufficient! In the event that we have args that pass around the graph,
+    # and are elided by dynamo, the reconstruct code above places them into the output correctly.
+    # However, _PyTreeCodeGen does not do that. If we are to ensure our users have a good UX, we
+    # need to do the codegen reconstruction for them.
+    # This gives us a few major paths:
+    #  A) Expose a util function that takes the output of this (graph, specs, etc)
+    # and lets users call it with regular function args, and that function handles the input remapping.
+    #  B) We write out own version of _PyTreeCodeGen that generates the remapping and makes it into the graph.
```
voznesenskym, line 344: B, so far, I think, is the best option here, because if we want to offer the user convenience functions short of rewriting (codegen + recompile) the graph (like a util or wrapper func), we would have to end up calling forward twice - once to capture the 'eager' args, and once on the exported graph, so that we can reconstruct. Codegen + recompile avoids this problem.

voznesenskym: Thanks, did not know about `Transformer` - this looks perfect. Order should be very easy to extract from the functions already in this PR.

torchdynamo/eval_frame.py
```diff
@@ -256,50 +276,76 @@ def dynamo_normalization_capturing_compiler(
...
+            out_spec_export,
+            result_traced,
+        )
 
-    # TODO(voz): A major feature gap here, atm, is that we return a graph with a flat_args signature.
+    # NOTE: This is a lengthy outline of major UX gaps, but the primary goal is to produce
+    # a graph with the same exact input and output signatures as the original traced callable.
+
+    # 1) TODO(voz): call _codegen on the graph to rewrite the graphs input and output
+    # Blocked on bugs in fx. https://github.com/pytorch/pytorch/pull/81177
```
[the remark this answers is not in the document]
voznesenskym: Yes.

torchdynamo/eval_frame.py
```diff
@@ -256,50 +276,76 @@ def dynamo_normalization_capturing_compiler(
...
+    #         out_spec_traced,
+    #     )
+    # )
+    # graph.recompile()
+    # 2) NOTE: This may not be sufficient! In the event that we have args that pass around the graph,
+    # and are elided by dynamo, the reconstruct code above places them into the output correctly.
+    # However, _PyTreeCodeGen does not do that. If we are to ensure our users have a good UX, we
+    # need to do the codegen reconstruction for them.
+
+    # 3) TODO(voz): A major feature gap here, atm, is that we return a graph with a flat_args signature.
```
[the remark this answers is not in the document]
voznesenskym: I was talking about the UX - I'll rewrite this.

torchdynamo/eval_frame.py
```diff
@@ -240,12 +240,23 @@ def toy_example(a, b):
...
     import torch.utils._pytree as pytree
 
     graph = None
     out_guards = None
+    compiler_captured_inputs = None
+
+    def produce_matching(source_args, candidate_args):
+        matched_elements_positions = []
+        for x in candidate_args:
+            matched_elements = [torch.allclose(x, y) for y in source_args]
```
jansel, line 252: This method of matching seems wrong.  `x is y` seems like the correct thing.  

We could have a function that returns two different tensors with the same elements in the example input.  (But not all inputs.)

Perhaps we need to only run the model once so the tensors will pass an `x is y` test?

jansel (changes requested):

torchdynamo/eval_frame.py
voznesenskym: Agreed - but wasn't sure how we could run only once. We need to run it *once* to capture the graph - so that would mean that the second run here, the one where we run the exported graph, is optional. We could, ostensibly, instead of running the second graph, do analysis on the placeholder/input nodes and output of the function... I left a todo here, would you be open to kicking this can down the road?

torchdynamo/eval_frame.py
```diff
@@ -256,50 +267,105 @@ def dynamo_normalization_capturing_compiler(
...
+        matched_output_elements_positions = produce_matching(
+            flat_both, flat_results_traced
+        )
+
+    class ChangeInputOutputSignature(torch.fx.interpreter.Transformer):
+        def __init__(
+            self,
+            m,
+            arg_len,
+            matched_input_elements_positions,
```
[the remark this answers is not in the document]
voznesenskym: Agreed, fixed.

torchdynamo/eval_frame.py
```diff
@@ -256,50 +267,105 @@ def dynamo_normalization_capturing_compiler(
...
+            self.old_args_gen = (
+                self.new_args[i] for i in self.matched_input_elements_positions
+            )
+
+        def placeholder(self, target, args, kwargs):
+            return next(self.old_args_gen)
+
+        def output(self, target, args, kwargs):
+            old_result_flat = args[0]
+            reshaped_new_arg = pytree.tree_unflatten(self.new_args, self.in_spec)
```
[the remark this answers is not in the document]
voznesenskym: Agreed, we do not

torchdynamo/eval_frame.py
voznesenskym: Resolved

tests/test_export.py
```diff
@@ -0,0 +1,429 @@
...
+            real_result = func(torch.tensor([[[1.3737, 0.1]]]))
+
+        torchdynamo.reset()
+
+        exported = torchdynamo.export(func, torch.tensor([[[1.3737, 0.1]]]))
+        out_graph = exported[0]
+
+        dynamo_result = out_graph(torch.tensor([[[1.3737, 0.1]]]))
+
+        print("R:", real_result)
```
jansel, line 69: Remove prints

torchdynamo/eval_frame.py
```diff
@@ -240,12 +240,24 @@ def toy_example(a, b):
...
 
     graph = None
     out_guards = None
+    compiler_captured_inputs = None
+    graph_result_export = None
+
+    def produce_matching(source_args, candidate_args):
+        matched_elements_positions = []
+        for x in candidate_args:
+            matched_elements = [x is y for y in source_args]
```
jansel, line 253: This function reads a bit awkward.  The `x is y` test is very cheap, so no need to cache it.

This is also `O(n^2)`, but could be written as `O(n)` using a `id()`+dictionary .

torchdynamo/eval_frame.py
```diff
@@ -256,50 +268,100 @@ def dynamo_normalization_capturing_compiler(
...
+        nonlocal compiler_captured_inputs
+
         assert graph is None, "whole graph export entails exactly one graph"
         graph = gm
-        return gm.forward
+        compiler_captured_inputs = example_inputs
+
+        def result_capturing_wrapper(*graph_inputs):
+            nonlocal graph_result_export
+            graph_result_export = graph(*graph_inputs)
```
jansel, line 279: I'd expect us to need to be capturing the inputs too, for matching them.  id/is matching might not work otherwise.

torchdynamo/eval_frame.py
```diff
@@ -256,50 +268,100 @@ def dynamo_normalization_capturing_compiler(
...
-
-    # TODO(voz): A major feature gap here, atm, is that we return a graph with a flat_args signature.
-    # There is currently more work that needs to be done on the fx side before we can support the UX we want.
-    # The future UX here will not return a spec, but will rather return a graph with the original signature
-    # and return type as the passed in callable, `f`.
-    return (graph, out_guards, in_spec, out_spec)
+    dynamo_flat_args, dynamo_in_spec = pytree.tree_flatten(compiler_captured_inputs)
+
+    matched_input_elements_positions = []
+    if in_spec != dynamo_in_spec:
```
jansel, line 299: The specs matching doesn't mean the order will match.

torchdynamo/eval_frame.py
```diff
@@ -256,50 +268,100 @@ def dynamo_normalization_capturing_compiler(
...
+        if len(dynamo_flat_args) > 0:
+            matched_input_elements_positions = produce_matching(
+                flat_args, dynamo_flat_args
+            )
+
+    flat_results_traced, out_spec_traced = pytree.tree_flatten(result_traced)
+    flat_results_export, out_spec_export = pytree.tree_flatten(graph_result_export)
```
jansel, line 306: torchdynamo has the invariant that captured graphs output flat lists of tensors.  So no need for tree_flatten here.

torchdynamo/eval_frame.py
```diff
@@ -256,50 +268,100 @@ def dynamo_normalization_capturing_compiler(
...
+
+    flat_inputs_to_exported_graph, flat_inputs_spec = pytree.tree_flatten(
```
jansel, line 308: torchdynamo has the invariant that captured graphs always input flat lists of tensors, so no need for tree_flatten here.

torchdynamo/eval_frame.py
```diff
@@ -256,50 +268,100 @@ def dynamo_normalization_capturing_compiler(
...
+        compiler_captured_inputs
+    )
+
+    matched_output_elements_positions = []
+    if out_spec_export != out_spec_traced:
```
jansel, line 313: specs matching doesn't mean order matches.

torchdynamo/eval_frame.py
```diff
@@ -256,50 +268,100 @@ def dynamo_normalization_capturing_compiler(
...
+        )
+
+    class ChangeInputOutputSignature(torch.fx.interpreter.Transformer):
+        def __init__(
+            self,
+            m,
+        ):
+            super().__init__(m)
+            arg_len = len(flat_args)
+            if arg_len > 0:
```
jansel, line 326: This test shouldn't be needed.  the `True` path works fine in the `args_len==0` case.

torchdynamo/eval_frame.py
```diff
@@ -256,50 +268,100 @@ def dynamo_normalization_capturing_compiler(
...
+
+        def output(self, target, args, kwargs):
+            dynamo_result_flat = args[0]
+            lookup = [*dynamo_result_flat, *self.new_args]
+            out_spec_positions = [
+                i
+                for i in range(
+                    len(matched_output_elements_positions), out_spec_traced.num_leaves
+                )
+                if i not in matched_output_elements_positions
```
jansel, line 351: This seems a bit awkward?  What not just do one matching?

jansel (changes requested):

torchdynamo/eval_frame.py
voznesenskym: Sure.

torchdynamo/eval_frame.py
voznesenskym: We do, L275 right above, but I agree, putting in here might read cleaner.

torchdynamo/eval_frame.py
voznesenskym: we can just remove this, worst case is the matching is identical.

torchdynamo/eval_frame.py
voznesenskym: Sg

voznesenskym:
Thanks for the great feedback, applied, and new test added for a bug I found in applying changes.

torchdynamo/eval_frame.py
```diff
@@ -240,12 +240,26 @@ def toy_example(a, b):
...
+    def produce_matching(source_args, candidate_args):
+        matched_elements_positions = []
+        dict_of_source_args = dict()
+        for i in range(0, len(source_args)):
+            element_id = id(source_args[i])
+            dict_of_source_args[element_id] = i
+
+        for i in range(0, len(candidate_args)):
+            arg = candidate_args[i]
+            if id(arg) in dict_of_source_args:
```
jansel, line 259: What happens if this condition is False?  Does it still work?  Or should this be an error?

jansel (changes requested):

torchdynamo/eval_frame.py
voznesenskym: Let's make it an assert :)

jansel (approved):

PR #615, opened 2022-07-19
[inductor] Force indirect-indexing to use dense_mask

desertfire:
Summary: The triton backend's range-tree based index comparison can not
handle indirect-indexing, which causes mask to incorrectly become None
in some cases. Force mask to be the same as dense_mask in this case.

### DIFF
--- a/torchinductor/codegen/triton.py
+++ b/torchinductor/codegen/triton.py
@@ -502,21 +502,22 @@ class TritonKernel(Kernel):
         itervars = list(itertools.chain(*self.set_ranges(*new_ranges)))
         return [[fn(itervars) for fn in fns] for fns in return_getters_groups]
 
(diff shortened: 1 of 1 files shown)


desertfire:
@jansel , although the current implementation can solve the CUDA error problem in https://github.com/pytorch/torchdynamo/issues/602, it is not practical as it treats all torch.int64 loads as potentially being used as indirect-load indices. So I need to find a way to identify loads that are truly used as indirect-load indices, but it seems that information is only known when the actual indirect-load happens. One possible solution to use something like the DeferredIndentedBuffer for loads, but I wonder if there is any other nicer way to do it. Or, should this be considered as a triton bug?

jansel:
I responded here https://github.com/pytorch/torchdynamo/issues/602#issuecomment-1189741580

jansel (approved):
lgtm!

Issue #611, opened 2022-07-19
Decomposition - cudnn_batch_norm vs native_batch_norm


anijain2305:
For few of the TIMM models, there are accuracy differences when I use AOT Autograd. On further debugging, I found that removing `cudnn_batch_norm` and `cudnn_batch_norm_backward` resolves the issue. 

I am able to repro it, but in somewhat unusual way. I have incpetion model. In the first version, all the batch norm ops are `cudnn_batch_norm` ops. In the second version, I have replaced the very first `cudnn_batch_norm` with `native_batch_norm`. With this, I am able to repro the accuracy difference.

The zip file is here

[inception.tar.gz](https://github.com/pytorch/torchdynamo/files/9145065/inception.tar.gz)

The file - `inception_fail.py` compares the above two versions of the model and shows that the output tensors deviate a lot.

anijain2305:
cc @Chillee @desertfire

anijain2305:
@ptrblck identified the issue here that TF32 matmuls are amplifying the noise coming from the first matmul. There are some minor differences between native and cudnn version of batch_norm, which in isolation are ok, but when amplified through TF32 matmuls cause issues.

So, a better way is to either use FP32 for accuracy comparisons. This might also not be enough, as I see in `inception_v3`, where the accuracy failure is now in the gradient value (compared to forward output when we were using TF32)

So, there is no other way than relaxing the accuracy constraint further. I am going to use cosine similarity for such models.

PR #616, opened 2022-07-20
Disabling TF32 for accuracy comparison

anijain2305:
Some context here - https://github.com/pytorch/torchdynamo/issues/611

Even though this is not enough, it can help reduce the number of spurious models that fail because of TF32 issues.

@jansel @Chillee @ngimel Does this make sense?

### DIFF
--- a/benchmarks/common.py
+++ b/benchmarks/common.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 import argparse
 import collections
+import contextlib
 import copy
 import csv
 import functools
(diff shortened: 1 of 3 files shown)


benchmarks/common.py
```diff
@@ -48,6 +49,26 @@
 output_filename = None
 
 
+# We are primarily interested in tf32 datatype
+torch.backends.cuda.matmul.allow_tf32 = True
```
jansel, line 53: We might want to set this inside `main()`.  A lot of the benchmarks change global state when we import them.

benchmarks/common.py
```diff
@@ -667,53 +688,56 @@ def run_one_model(
...
+                    if not same(correct_result, correct_rerun_result):
+                        print("INCORRECT - Variation in Eager runs itself")
+                        if not skip_accuracy_check:
+                            return sys.exit(-1)
+
+                torch.manual_seed(1337)
+                torchdynamo.reset()
+                if experiment.func is cold_start_experiment:
+                    results = []
+                    results.append(experiment(model, example_inputs, optimize_ctx))
```
jansel, line 719: Won't this disable tf32 for our performance measurements?

jansel (changes requested):

benchmarks/common.py
anijain2305: This one is specific to cold_start_experiment. The real experiment is called later on.
But, I am moving this one out of scope of with no_tf32.

benchmarks/common.py
anijain2305: Good idea.

anijain2305:
Closing as this does not really help.

PR #617, opened 2022-07-20
Resnet50_quantized_qat

anijain2305:
Using cosine similarity for resnet50_quantized_qat

When used with AOTAutograd even with eager backend, there are some accuracy differences. Its unclear to me where they are coming from. Perhaps, it could be a result of quantization. Using cosine_simarity here.

cc @Chillee - Is that ok?

### DIFF
--- a/benchmarks/common.py
+++ b/benchmarks/common.py
@@ -650,7 +650,7 @@ class BenchmarkRunner:
         return set()
 
     @property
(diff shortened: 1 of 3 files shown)


Chillee:
LGTM, assuming cosine similarity fixes the accuracy issues.

Chillee (approved):

PR #621, opened 2022-07-20
Raise AssertionError controlled with a flag

desertfire:
Summary: In https://github.com/pytorch/torchdynamo/pull/579/, the
torchbench CI run hits an ssertion but didn't report failure on that. It
will be helpful to raise AssertionError at this active development
stage.


desertfire:
Hmm, this exposes existing assertions from our unit tests. I wonder if we should clean up them?

jansel (approved):

PR #624, opened 2022-07-20
Fix bug with dataclass codegen

mlazos:
DataclassVariable expects the previous behavior (pre #579) to setup some kw args, overrode the new behavior.

### DIFF
--- a/torchdynamo/variables/dicts.py
+++ b/torchdynamo/variables/dicts.py
@@ -298,7 +298,7 @@ class DataClassVariable(ConstDictVariable):
             codegen(self.items[key])
         return [
             codegen.create_load_const(keys),
-            create_instruction("CALL_FUNCTION_KW", len(keys))
+            create_instruction("CALL_FUNCTION_KW", len(keys)),
         ]
 
     def call_method(
--- a/torchdynamo/variables/dicts.py
+++ b/torchdynamo/variables/dicts.py
@@ -293,10 +293,13 @@ class DataClassVariable(ConstDictVariable):
 
     def reconstruct(self, codegen):
         codegen.extend_output([codegen._create_load_const(self.user_cls)])
-        result = list(super().reconstruct(codegen))
-        assert result[-1].opname == "BUILD_CONST_KEY_MAP"
-        result.append(create_instruction("CALL_FUNCTION_KW", result.pop().argval))
-        return result
+        keys = tuple(self.items.keys())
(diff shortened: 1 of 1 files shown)


voznesenskym:
Since its a fix to a bug, would it be onerous to add a test?

torchdynamo/variables/dicts.py
```diff
@@ -293,10 +293,13 @@ def as_proxy(self):
...
-        assert result[-1].opname == "BUILD_CONST_KEY_MAP"
-        result.append(create_instruction("CALL_FUNCTION_KW", result.pop().argval))
-        return result
+        keys = tuple(self.items.keys())
+        for key in keys:
+            codegen(self.items[key])
+        return [
+            codegen.create_load_const(keys),
+            create_instruction("CALL_FUNCTION_KW", len(keys)),
+        ]
```
voznesenskym, line 302: 1) This looks duped to `ConstDictVariable`'s `reconstruct ` except you create the `CALL_FUNCTION_KW` instruction instead of `BUILD_CONST_KEY_MAP` - can we dedup?

2) There's an optimization around:

```
if len(self.items) == 0:
            return [create_instruction("BUILD_MAP", 0)]
```

Is it worth keeping here too?

mlazos:
> Since its a fix to a bug, would it be onerous to add a test?

So the issue occurred in some of the HF models, I wasn't able to dig deeper and figure out an exact repro. The true solution is to have assertions in the HF models trigger a failure in CI, which Bin has #621 for

torchdynamo/variables/dicts.py
mlazos: We can't dedup, the dict used to expect constant keys but it doesn't anymore, I kept the behavior of DataClassVariable exactly as it was before. It previously had an assertion that BUILD_CONST_KEY_MAP was the first instruction so that implies the length is never zero with a dataclass, although, it's supported in this case anyway.

desertfire:
> > Since its a fix to a bug, would it be onerous to add a test?
> 
> So the issue occurred in some of the HF models, I wasn't able to dig deeper and figure out an exact repro. The true solution is to have assertions in the HF models trigger a failure in CI, which Bin has #621 for

I am ok with not adding another test, but I would like to verify the `ci/circleci: gpu_linux` CI result (which is failing now).

desertfire (approved):

PR #626, opened 2022-07-20
[inductor] triton template incorporate new conv

pyjhzwh:
- incorporate #605 into triton template
- refactor `conv_heuristics()`
- update some tests
- decide whether use triton template using node.kernel info rather than static config.triton.convolution because triton conv don't support many cases

It won't affect the default behavior when tochinductor.triton.convolution == "aten"


torchinductor/codegen/triton_conv_delta_x_hwc.j2
```diff
@@ -0,0 +1,219 @@
...
+        )
+        mask_w = (off_x_crs < CRS)[:, None] & (off_w_k < KERNEL_N)[None, :]
+        # ------ prefetch ------
+        # ------ load x ------
+        matrix_x = tl.load(x_ptrs, mask=mask_x)
+        # ------ load w ------
+        matrix_w = tl.load(w_ptrs, mask=mask_w)
+
+    # add bias if is not None
+    if bias is not None:
```
jansel, line 168: We should be able to drop bias support (we split it out into a separate kernel in lowering)

torchinductor/codegen/triton_conv_delta_x_hwc.j2
```diff
@@ -0,0 +1,219 @@
...
+    mask_y = (
+        (off_y_n < BATCH)[:, None]
+        & (off_y_h < OUT_H + output_padding_h)[:, None]
+        & (off_y_w < OUT_W + output_padding_w)[:, None]
+        & (off_y_k < KERNEL_N)[None, :]
+    )
+
+{% if pointwise_code %}
+{{ pointwise_code | indent(4, true) }}
+    {#
```
jansel, line 206: remove?

torchinductor/ir.py
```diff
@@ -1091,6 +1143,15 @@ def make_indexer(self):
         ), f"convert {type(self).__name__} to FixedLayout first"
         return self.as_fixed().make_indexer()
 
+    def __eq__(self, o) -> bool:
```
jansel, line 1146: o=>other

torchinductor/ir.py
```diff
@@ -2556,26 +2644,35 @@ def map_args(self):
...
+                ),
                 ("OUT_H", f"{self.get_size()[2]}"),
                 ("OUT_W", f"{self.get_size()[3]}"),
-                ("stride_h", f"{const_args[0][0]}"),
-                ("stride_w", f"{const_args[0][1]}"),
-                ("padding_h", f"{const_args[1][0]}"),
-                ("padding_w", f"{const_args[1][1]}"),
-                ("dilation_h", f"{const_args[2][0]}"),
-                ("dilation_w", f"{const_args[2][1]}"),
+                ("stride_h", f"{V.graph.sizevars.size_hint(const_args[0][0])}"),
```
jansel, line 2666: `size_hint` is not safe to use.   It is just a hint and could change.

torchinductor/triton_ops/conv.py
```diff
@@ -297,13 +206,193 @@ def _kernel(
...
+        )[:, None] & (off_x_crs < CRS)[None, :]
+        mask_w = (off_x_crs < CRS)[:, None] & (off_w_k < KERNEL_N)[None, :]
+        # ------ prefetch ------
+        # ------ load x ------
+        matrix_x = tl.load(x_ptrs, mask=mask_x)
+        # ------ load w ------
+        matrix_w = tl.load(w_ptrs, mask=mask_w)
+
+    # add bias if is not None
+    if bias is not None:
```
jansel, line 348: drop bias support

jansel (changes requested):

torchinductor/codegen/triton_conv_delta_x_hwc.j2
pyjhzwh: This is where we put fusable pointwise code in the template.

torchinductor/codegen/triton_conv_delta_x_hwc.j2
pyjhzwh: Removed bias

torchinductor/ir.py
pyjhzwh: Use const_args instead

torchinductor/triton_ops/conv.py
pyjhzwh: Removed bias

jansel (approved):

PR #489, opened 2022-06-29
[Not to Merge] A PoC of Graph Break Compiler

alanwaketan:
Summary:
This is a PoC to break graphs in order to support gradient hooks for the
AOT backend. E2E use cases are composing DDP/FSDP with dynamo.

Test Plan:
WIP.

PR #628, opened 2022-07-21
DDP optimization via graph-breaks in Dynamo

wconstab:
Attempts to interpose between the unmodified DDP program and
user-provided dynamo backend, adding a special 'graph break'
stage that breaks up graphs to be compiled by dynamo according
to the DDP bucketing strategy

Detects DDP by using a new API (get_active_ddp_module) exposed by torch DDP, and only enables itself inside the 'run_ddp_forward' method.  Also grabs bucket size and other metadata from the active DDP module.

TODO
- validate/benchmark

Builds on prototype from @alanwaketan #489

cc @ezyang @mrshenli @anijain2305 @jansel

### DIFF
--- a/torchdynamo/eval_frame.py
+++ b/torchdynamo/eval_frame.py
@@ -633,9 +633,10 @@ class TorchPatcher:
         ]
 
         # disable dynamo for the wrapper that helps give dynamo hints about entering DDP
-        DistributedDataParallel._inside_ddp_forward = skip(
-            DistributedDataParallel._inside_ddp_forward
-        )
+        if hasattr(DistributedDataParallel, "_inside_ddp_forward"):
+            DistributedDataParallel._inside_ddp_forward = skip(
+                DistributedDataParallel._inside_ddp_forward
+            )
 
         # disable profile hook
         for opt in optimizers:
--- a/torchdynamo/eval_frame.py
+++ b/torchdynamo/eval_frame.py
@@ -13,9 +13,11 @@ from unittest.mock import patch
 import torch
 import torch.utils._pytree as pytree
 from torch.fx.experimental.proxy_tensor import make_fx
+from torch.nn.parallel.distributed import DistributedDataParallel
 
 import torchdynamo
 from torchdynamo.debug_utils import wrap_backend_debug
+from torchdynamo.optimizations.distributed import DDPOptimizer
 from torchdynamo.utils import checkpoint_params
 from torchdynamo.utils import clone_inputs
 from torchdynamo.utils import compile_times
@@ -231,6 +233,20 @@ def catch_errors_wrapper(callback):
             ):
                 # nametuple constructor
                 return None
+            if config.optimize_ddp:
+                ddp_module = DistributedDataParallel._get_active_ddp_module()
+                if ddp_module and frame.f_code.co_name == "forward":
+                    with compile_lock:
+                        ddp_optimizer = DDPOptimizer(
+                            bucket_bytes_cap=ddp_module.bucket_bytes_cap,
+                            parameters_to_ignore=ddp_module.parameters_to_ignore,
+                            backend_compile_fn=callback._torchdynamo_orig_callable,
+                        )
+                        hijacked_callback = convert_frame.convert_frame(
+                            ddp_optimizer.compile_fn, guard_export_fn=None
+                        )
+                        return hijacked_callback(frame, cache_size)
+
             with compile_lock:
                 return callback(frame, cache_size)
         except Exception:
@@ -615,6 +631,11 @@ class TorchPatcher:
             for opt in torch.optim.__dict__.values()
             if inspect.isclass(opt) and issubclass(opt, torch.optim.Optimizer)
         ]
+
+        # disable dynamo for the wrapper that helps give dynamo hints about entering DDP
+        DistributedDataParallel._inside_ddp_forward = skip(
+            DistributedDataParallel._inside_ddp_forward
+        )
 
         # disable profile hook
         for opt in optimizers:
--- a/torchdynamo/optimizations/distributed.py
+++ b/torchdynamo/optimizations/distributed.py
@@ -1,0 +1,181 @@
+from typing import Any
+from typing import List
+
+import torch
+import torch.fx.traceback as fx_traceback
+from torch import fx
+from torch.fx.node import Node
+
+
+def args_str(args):
+    # a debug helper
+    if torch.is_tensor(args):
+        return f"T[{args.shape}]"
+    elif isinstance(args, tuple):
+        return f"tuple({', '.join([args_str(x) for x in args])})"
+    elif isinstance(args, list):
+        return f"list({', '.join([args_str(x) for x in args])})"
+    else:
+        return str(args)
+
+
+class DDPOptimizer:
+    def __init__(
+        self,
+        bucket_bytes_cap: int,
+        parameters_to_ignore: List[str],
+        backend_compile_fn,
+        debug=False,
+    ):
+        self.bucket_bytes_cap = bucket_bytes_cap
+        self.parameters_to_ignore = parameters_to_ignore
+        self.backend_compile_fn = backend_compile_fn
+        self.debug = debug
+
+    def compile_fn(self, gm: fx.GraphModule, example_inputs: List[torch.Tensor]):
+        """
+        TODO:
+        - handle params_and_buffers_to_ignore
+        - handle kwargs
+        """
+
+        # 1: compute the partition map according to DDP bucket logic
+        bucket_bytes = 0
+        bucket_actual_sizes = []
+        node_splits = [[]]
+        for node in reversed(gm.graph.nodes):
+            if bucket_bytes >= self.bucket_bytes_cap:
+                bucket_actual_sizes.insert(0, bucket_bytes)
+                bucket_bytes = 0
+                node_splits.insert(0, [])
+
+            if node.op == "output" or node.op == "placeholder":
+                continue
+
+            elif node.op == "call_module":
+                target = gm.get_submodule(node.target)
+                params_size_b = sum(
+                    [
+                        p.storage().nbytes()
+                        for p in target.parameters()
+                        if p.requires_grad
+                    ]
+                )
+                bucket_bytes += params_size_b
+                # print(f"accumulated {params_size_b} b from {node}")
+            else:
+                # TODO(whc) confirm this:
+                # (e.g. call_method, call_function aren't expected to 'have' parameters)
+                pass
+
+            node_splits[0].append(node)
+
+        if len(node_splits) == 1:
+            if self.debug:
+                print(
+                    "DDPOptimizer did not split graphs."
+                    f" Accumulated {bucket_bytes} bytes, and bucket cap is {self.bucket_bytes_cap}"
+                )
+            return self.backend_compile_fn(gm, example_inputs)
+
+        if len(bucket_actual_sizes) < len(node_splits):
+            bucket_actual_sizes.insert(0, bucket_bytes)
+
+        if self.debug:
+            print(
+                f"DDPOptimizer used bucket cap {self.bucket_bytes_cap}"
+                f" and split graphs into parameter sizes {', '.join([str(b) for b in bucket_actual_sizes])}"
+            )
+
+        # 2: partition the graphmodule according to bucket capacity
+        partition_map = {}
+        for p, nodes in enumerate(node_splits):
+            for node in nodes:
+                partition_map[node] = p
+
+        split_gm = fx.passes.split_module.split_module(
+            gm, None, lambda node: partition_map[node]
+        )
+        if self.debug:
+            with open("debug_ddp_optimizer.log", "w") as dump_file:
+                dump_file.write("---orig graph---")
+                dump_file.write(str(gm.graph))
+                dump_file.write("\n---split graph---")
+                dump_file.write(str(split_gm.graph))
+
+        # 3: compile each of the partitioned submodules using the user-provided compiler
+        class SubmodCompiler(torch.fx.interpreter.Interpreter):
+            def __init__(self, module, compiler, debug=False):
+                super().__init__(module)
+                self.compiler = compiler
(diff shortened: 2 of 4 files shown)


torchdynamo/eval_frame.py
```diff
@@ -132,7 +134,29 @@ def __init__(self):
...
+                up = up.f_back
+            return False
+
         try:
+            if _inside_ddp_callstack(frame) and frame.f_code.co_name == "forward":
+                print(f"compile for ddp: {frame.f_code.co_name} {frame.f_code.co_filename}")
+                with compile_lock:
+                    # TODO(whc) find a way to get these parameters from the DDP module
+                    # and configure the backend compiler and convert_frame correctly
+                    ddp_optimizer = DDPOptimizer(bucket_cap_mb = 25, parameters_to_ignore = [], backend_compile_fn = BACKENDS["aot_autograd"])
```
wconstab, line 157: @jansel any ideas how to achieve these few lines worth of stuff in a more sane way?

wconstab: - how to detect 'inside ddp'
- how to access attributes of the DDP module object (such as its 'bucket_cap_mb' attr)
- how to get the user-provided backend compile_fn rather than hardcoding aot_autograd here
- whether the hijacked_callback approach is even correct?

torchdynamo/optimizations/distributed.py
```diff
@@ -0,0 +1,145 @@
...
+        TODO:
+        - handle params_and_buffers_to_ignore
+        - implement call_method/call_forward
+        - handle kwargs
+        """
+
+        print("DDPOptimizer called with FX graph:")
+        gm.graph.print_tabular()
+        print()
+        # 1. Splitting the gm into small graphs, following DDP parameter bucketing
```
ezyang, line 44: That means this logic needs to be synchronized with the DDP bucketing logic, is that right

ezyang:
@wconstab now that you've gotten this working e2e, how long do you think we're going to keep using this code? It looks like it ended up being a bit complicated, and iirc, @jansel proposed a different strategy that specialized on whether or not we were the first use of a parameter or not (though, tbf, I don't remember how this proposal was going to work)

torchdynamo/optimizations/distributed.py
wconstab: right. 
1) ddp first buckets based on reverse order of parameter declaration in nn module structure
2) then after first run, ddp rebuckets (incurs tensor copies) using the actual execution order of first backward, which can be different
3) ddp's actual bucketing logic is pretty simple, given an order of parameters and a set of parameters to exclude, it fills buckets to capacity

wconstab:
> @wconstab now that you've gotten this working e2e, how long do you think we're going to keep using this code? It looks like it ended up being a bit complicated, and iirc, @jansel proposed a different strategy that specialized on whether or not we were the first use of a parameter or not (though, tbf, I don't remember how this proposal was going to work)

@ezyang I'd say this is barely working e2e, it's still a POC and needs some help to be viable (hence the questions on the PR).  I think this approach could have 2 outcomes
 - best case (for it): we don't find much value in larger-graph optimizations and/or it is really hard to fix AotAutograd tracing collectives, so we keep using this
 - worst case (for it): we only use this for a really short time before we find it necessary to do the tracing approach

All of that depends on perf benchmark numbers too.  It may be that the cost/benefit on benchmarks isn't there for investing in the harder approach, so we should check the perf and prioritize accordingly

torchdynamo/optimizations/distributed.py
```diff
@@ -0,0 +1,105 @@
...
+
+class DDPOptimizer:
+    def __init__(
+        self, bucket_bytes_cap: int, parameters_to_ignore: List[str], backend_compile_fn
+    ):
+        self.bucket_bytes_cap = bucket_bytes_cap
+        self.parameters_to_ignore = parameters_to_ignore
+        self.backend_compile_fn = backend_compile_fn
+
+    def compile_fn(self, gm: fx.GraphModule, example_inputs: List[torch.Tensor]):
```
mrshenli, line 16: what happens with dynamic execution orders? Will re-compilation leads to new graph breaks? DDP might not be able to adapt to new bucketization accordingly, as 1) frequent rebucketziation is expensive 2) re-compilation is a one-time thing, but execution might switch between different orders multiple times.

torchdynamo/eval_frame.py
```diff
@@ -133,10 +136,32 @@ def __init__(self):
...
+    gm.graph.print_tabular()
+    return gm.forward
+
+
 def catch_errors_wrapper(callback):
     @functools.wraps(callback)
     def catch_errors(frame, cache_size):
         try:
+            # TODO(whc) move the ddp code below the bailouts. but make sure it doesn't cause DDP to be skipped
+            ddp_module = DistributedDataParallel.get_active_ddp_module()
```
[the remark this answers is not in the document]
wconstab: I added that api in a diff on the pytorch side: https://github.com/pytorch/pytorch/pull/83333

The idea is, I needed to do 2 things and i solved them both with one API.
1) let torchdynamo know if its compiling a frame that is inside a DDP forward context or not
2) give dynamo's DDPOptimizer access to the bucket size, and other state vars for the current DDP module

Returning 'None' for active DDP module from get_active_ddp_module() is the equivalent of #1 being false, and I can then extract the needed values off the module if #1 is true.

torchdynamo/optimizations/distributed.py
wconstab: Let's assume that DDP initialized its buckets with the 'wrong' order, i.e. an order that will not be observed at runtime.

Also note, I don't think it's possible under dynamo for there to be dynamic behavior within one compiled function.  If something causes dynamism, it would force dynamo to recompile anyways.  This could be OK if it happened infrequently but otherwise it would be a perf issue (regardless of DDP) and the user might need to change their model to avoid it if using dynamo.

So, what I think would happen is
1) DDP's initial buckets are chosen. They may be wrong, and are currently ignored by DDPOptimizer.
2) DDPOptimizer partitions the graph into buckets according to the true order of execution _of the forward graph_.  
3) Then it runs AotAutograd and generates a backward graph, but bucket ordering is _not_ updated respecting the backward graph.
4) At runtime, the compiled fwd/bwd graph is run and hooks fire in-between the split subgraphs.  DDP may observe a different execution order and update its bucketing scheme after the first iteration.

If the program is not dynamic, things would be stable after that.  If there is dynamism, then I think each new version would incur dynamo compilation, and _if_ execution order is different enough to make DDP reconfigure buckets, it seems that DDP would do so multiple times.  

This seems to suggest that dynamo and DDP costs under dynamism are orthogonal: in eager, dynamism would potentially cause multiple re-configurations of DDP buckets, and that remains the same under dynamo.  Without DDP, dynamo still incurs expensive recompilation under dynamism.  Does that sound right to you @mrshenli?

torchdynamo/optimizations/distributed.py
```diff
@@ -0,0 +1,105 @@
...
+        - implement call_method/call_forward
+        - handle kwargs
+        """
+
+        # compute the partition map according to DDP bucket logic
+        bucket_bytes = 0
+        bucket_actual_sizes = []
+        node_splits = [[]]
+        for node in reversed(gm.graph.nodes):
+            if bucket_bytes >= self.bucket_bytes_cap:
```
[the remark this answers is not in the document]
wconstab: ok, that might be worth implementing here.  If you have a preference on what the logic would look like let me know.

torchdynamo/optimizations/distributed.py
```diff
@@ -0,0 +1,105 @@
...
+                bucket_actual_sizes.insert(0, bucket_bytes)
```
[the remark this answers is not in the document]
wconstab: I think that is a good idea, just to make the behavior of the 'compiled' program seem more stable/predictable.  But i'd rather treat that as an optimization to add after getting the critical parts working.

torchdynamo/optimizations/distributed.py
```diff
@@ -0,0 +1,105 @@
...
+                bucket_bytes = 0
+                node_splits.append([])
+
+            if node.op == "output" or node.op == "placeholder":
+                continue
+
+            elif node.op == "call_module":
+                target = gm.get_submodule(node.target)
```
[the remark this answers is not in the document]
wconstab: I believe they are guaranteed to be modules at the same level of granularity to each other, since they are produced by the partitioner pass.  In other words, they should be non-leaf modules and they should contain other modules that could be non-leaf or leaf modules.

torchdynamo/optimizations/distributed.py
```diff
@@ -0,0 +1,105 @@
...
+            f"DDPOptimizer used bucket cap {self.bucket_bytes_cap} and split graphs into parameter sizes {', '.join([str(b) for b in bucket_actual_sizes])}"
+        )
+
+        # partition the graphmodule according to bucket capacity
+        partition_map = {}
+        for p, nodes in enumerate(node_splits):
+            for node in nodes:
+                partition_map[node] = p
+        split_gm = fx.passes.split_module.split_module(
+            gm, None, lambda node: partition_map[node]
```
wconstab, line 70: @mrshenli I've just rewritten most of this DDPOptimizer to use fx.passes.split_module pass.  That pass accepts a 'partition function' which takes 'fx.Node' as input and returns int that represents which partition it should be placed in by the pass.

Since I was iterating on Jiewen's POC, I still used his bucketing logic (above) to build a map, then I made my partition function a simple lambda that reads from the map.

I would propose instead that we put a 'DDP partition function' inside pytorch core in the distributed.py, and I can just call that from here and delete the above code.  This way the logic lives close to DDP and can be more easily be maintained by your team.  Wdyt?

torchdynamo/eval_frame.py
```diff
@@ -133,10 +136,32 @@ def __init__(self):
...
+            if ddp_module and frame.f_code.co_name == "forward":
+                print(
+                    f"compile for ddp: {frame.f_code.co_name} {frame.f_code.co_filename}"
+                )
+                with compile_lock:
+                    ddp_optimizer = DDPOptimizer(
+                        bucket_bytes_cap=ddp_module.bucket_bytes_cap,
+                        parameters_to_ignore=ddp_module.parameters_to_ignore,
+                        backend_compile_fn=my_compiler,
```
wconstab, line 159: @anijain2305 @voznesenskym Any idea how I can get the  'my_compiler' function that the user provided in their 'torchdynamo.optimize' call?  I could probably figure out a way to hack it together, but I am hoping for a clean way.

voznesenskym: Not in front of it right now, I can check definitively later, but is that not `callback` in this func?

wconstab: originally i assumed `callback` was the thing, and i used it.  But i have been facing some (possibly unrelated) bugs, and when i read the code I thought callback is actually wrapped in some context which i am not sure i wanted to reuse.

voznesenskym: oh that might be `convert_frame` and friends. Try adding this line to convert_frame before the return

`_convert_frame._compiler_fn = compiler_fn`

And then see if your `callback` has a `_compiler_fn`? I can look in a moment as well.

torchdynamo/optimizations/distributed.py
mrshenli: Yep, this makes sense to me. 

>  in eager, dynamism would potentially cause multiple re-configurations of DDP buckets, and that remains the same under dynamo.

Today's DDP will only reconfigure buckets once after the first iteration. But I agree with your point that, in this case, Dynamo + DDP should be able to offer on-par perf with eager-DDP. 

>  If there is dynamism, then I think each new version would incur dynamo compilation, and if execution order is different enough to make DDP reconfigure buckets, it seems that DDP would do so multiple times.

My original concern was that, suppose there are one dynamic if-else clause, and the app will switch between these two paths every 100 iterations. Would I be correct that dynamo will only trigger re-compile on the 1st and 101st iteration, but not 201st iteration as it is the same as the cached on for the 1st iteration? If that's the case, how can we tell DDP to update its buckets on the 201st iteration. Just OOC, I don't think this is a high-pri to address.

torchdynamo/optimizations/distributed.py
mrshenli: SGTM, it might be easier to keep it consistent with the rest of DDP. 

Question: would I be correct that we are not planning to change existing DDP to use this partition function any time soon though? Because there is no guarantee that we can get an `fx.Graph` from the input module? Besides, today's DDP already has a execution order based solution to re-organize buckets?

wconstab: > we are not planning to change existing DDP to use this partition function any time soon though

if you mean rewriting core DDP to rely on this FX pass, I wasn't planning on it.

> there is no guarantee that we can get an fx.Graph from the input module

right, I think there could be cases where legacy DDP works but FX-DDP couldn't get a single whole graph.  I'm not sure how often this would be in practice, or if we could make due with applying bucketization to partial graphs that were captured in those cases.

torchdynamo/optimizations/distributed.py
wconstab: > Would I be correct that dynamo will only trigger re-compile on the 1st and 101st iteration, but not 201st iteration as it is the same as the cached on for the 1st iteration?

yes, this is correct.  Dynamo would reuse cached programs for the if/else branches after they got compiled once each.

> how can we tell DDP to update its buckets on the 201st iteration.

Well, (1) dynamo's DDPOptimizer could insert a node into its compiled graph which calls into the DDP module and notifies it which bucket strategy it is about to use.  (2) this could happen on every iteration if we could make it fast enough and DDP mostly ignores it or checks its hash against the most recent one; or we could try to make dynamo actually pay attention to when its cached program changed from last cached program and only on those boundary conditions issue the notification to DDP.

I'm not sure how well this would work, but it seems like a possibility.  It might make more sense to invest in a new DDP approach that traces comm ops into the graph or implements DDP in FX though.  I see this graph-break solution as potentially a stop-gap.

anijain2305 (approved):

PR #629, opened 2022-07-21
Alias strings for AOT backends, and No-decomps for AOTAutograd accuracy test

anijain2305:
@Chillee suggested a better UX for using AOT Autograd memory efficient fusion. Suggestion is for `BACKENDS['nvfuser]` to point to `aot_autograd_efficient_strategy`. This way, the user does not have to add this extra import from torchdynamo.

Asking for comments @jansel @ngimel if `nvfuser` is a fine replacement. One other possibility is to use `aot_nvfuser` instead, which is little more explicit about using aot.

### DIFF
--- a/benchmarks/common.py
+++ b/benchmarks/common.py
@@ -30,10 +30,6 @@ from torchdynamo.optimizations.inference import offline_autotuner
 from torchdynamo.optimizations.inference import online_autotuner
 from torchdynamo.optimizations.log_args import conv_args_analysis
 from torchdynamo.optimizations.python_key import python_key
-from torchdynamo.optimizations.training import aot_autograd_debug_strategy1
-from torchdynamo.optimizations.training import aot_autograd_nnc_strategy
-from torchdynamo.optimizations.training import aot_autograd_prims_nvfuser_strategy
-from torchdynamo.optimizations.training import aot_autograd_speedup_strategy
 from torchdynamo.profiler import Profiler
 from torchdynamo.profiler import fx_insert_profiling
 from torchdynamo.testing import dummy_fx_compile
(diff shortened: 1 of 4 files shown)


jansel:
Agree we should add an alias string.  `aot_nvfuser` works for me, but fine with either.

anijain2305:
Changed to `aot_nvfuser`

Chillee (approved):
LGTM, although is the "aot_ts" backend even using NNC now?

anijain2305:
> LGTM, although is the "aot_ts" backend even using NNC now?

Yeah, I realized that even though we were using the word NNC before, it was not really NNC. I just removed the NNC word and replaced it with `ts` to be more clear.

jansel:
> LGTM, although is the "aot_ts" backend even using NNC now?

I think so, because of:
https://github.com/pytorch/torchdynamo/blob/f6fd47b2adb833d5b86a5f8e38017b369a537c34/benchmarks/common.py#L1001

Though we should change `args.nvfuser` to default to True at this point.

benchmarks/common.py
```diff
@@ -698,7 +695,7 @@ def run_one_model(
                 return 0
 
             try:
-                with optimize_ctx:
+                with accuracy_ctx:
```
anijain2305, line 698: Just to emphasize - Using a different ctx for accuracy testing. It is same as `optimize_ctx` unless overridden while handling cmdline args.

For aot_autograd, we choose no-decomps

anijain2305:
@jansel @Chillee Can I request one more review please? I am doing a couple of things in this PR

* Adding alias strings for AOT backend
* Switching off Decomps for AOT Autograd testing.

jansel (approved):

Chillee (approved):
lgtm.

Chillee:
Some documentation about what each "string backend" does would be nice though.

PR #631, opened 2022-07-21
Add Dynamic shapes benchmark for hf_bert

wconstab:
To Run: 

use wconstab/dynamic branch on pytorch/benchmark repo (to be merged to main imminently)
(see https://github.com/pytorch/benchmark/pull/1050)

python benchmarks/torchbench.py --dynamic_shapes --training --nvfuser
--accuracy-aot-ts-mincut --devices cuda --repeat 1 -k hf_bert

About the hf_bert changes:
* uses binning strategy I made up, but can easily replace with
distribution inspired by profiling real bert model

About the dynamic benchmark runner:
* uses a fixed sequence of inputs for each model, given by the model
* optionally: repeats over the whole set of bin inputs for eager, dynamo
to reduce variance
* allows warmup once at the beginning for the first shape seen, as with
other dynamo benchmarks.  does not allow subsequent warmup, and switches
from 'torchdynamo.run' to 'torchdynamo.optimize' inside the benchmark

Other notes:
* currently only tested with hf_bert
* although dynamo runs the benchmark, it reports many warnings along the
way (AOTAutograd unsupported mutations, cache limit, etc)
* benchmark metrics (what metrics to report) are WIP

Sample output for hf_bert:
mean: 0.165, median: 0.007, var: 0.116
Speedups by shape: torch.Size([4, 2]):  0.00453,  0.00289
torch.Size([4, 4]):  0.0048,  0.00323
torch.Size([4, 8]):  0.00483,  0.00299,  0.00441,  0.0422
torch.Size([4, 16]):  0.00491,  0.00269,  0.00739,  0.0273
torch.Size([4, 32]):  0.939,  1.09,  1.07,  0.0103
torch.Size([4, 64]):  0.00724,  0.00294,  0.00727,  0.0532
torch.Size([4, 128]):  0.00509,  0.00317,  0.00435,  0.0102,  0.119,
0.737,  0.0355
torch.Size([4, 256]):  0.00629,  0.00352,  0.14,  0.0432,  0.873

cc @soumith @jansel @ezyang @voznesenskym @anijain2305

### DIFF
--- a/benchmarks/common.py
+++ b/benchmarks/common.py
@@ -347,7 +347,9 @@ def speedup_experiment_ds(args, model_iter_fn, model, example_inputs):
     timings = np.zeros((args.repeat, len(example_inputs), 2), np.float64)
 
     if args.repeat > 5:
-        print(f"\ndynamic shapes experiments are slow, consider setting --repeat less than {args.repeat}\n")
+        print(
+            f"\ndynamic shapes experiments are slow, consider setting --repeat less than {args.repeat}\n"
+        )
 
     nwarmup = 4
     for rep in range(args.repeat):
@@ -377,18 +379,25 @@ def speedup_experiment_ds(args, model_iter_fn, model, example_inputs):
     shapes = [x[0].shape for x in example_inputs]
     shape_keys = sorted(set(shapes))
     shape_speedups = {
-        shape: list(map(lambda it: it[1], filter(lambda it: it[0] == shape, zip(shapes, speedups))))
+        shape: list(
+            map(
+                lambda it: it[1],
+                filter(lambda it: it[0] == shape, zip(shapes, speedups)),
+            )
+        )
         for shape in shape_keys
     }
-    output_str = f"mean: {speedups_mean:.3f}, median: {speedups_median:.3f}, var: {speedups_var:.3f}" + \
-        "\nSpeedups by shape: " + \
-        "\n".join([
-        f"{shape}: " + ", ".join(
-        [
-            f"{speedup: .3g}" for speedup in shape_speedups[shape]
-        ])
-        for shape in shape_keys
-    ])
+    output_str = (
+        f"mean: {speedups_mean:.3f}, median: {speedups_median:.3f}, var: {speedups_var:.3f}"
+        + "\nSpeedups by shape: "
+        + "\n".join(
+            [
+                f"{shape}: "
+                + ", ".join([f"{speedup: .3g}" for speedup in shape_speedups[shape]])
+                for shape in shape_keys
+            ]
+        )
+    )
     output_csv(
         output_filename,
         ("dev", "name", "speedup mean", "speedup median", "speedup var"),
@@ -469,6 +478,7 @@ def baselines(models, model_iter_fn, example_inputs, args):
         [current_device, current_name] + [f"{x:.4f}" for x in speedup],
     )
     return result
+
 
 def try_script(model, example_inputs):
     try:
--- a/benchmarks/common.py
+++ b/benchmarks/common.py
@@ -349,7 +349,7 @@ def speedup_experiment_ds(args, model_iter_fn, model, example_inputs):
     if args.repeat > 5:
         print(f"\ndynamic shapes experiments are slow, consider setting --repeat less than {args.repeat}\n")
 
-    nwarmup = 2
+    nwarmup = 4
     for rep in range(args.repeat):
         # Start each rep fresh, e.g. only warmup on example 0
         torchdynamo.reset()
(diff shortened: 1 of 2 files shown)


wconstab:
@anijain2305 I wonder if you could try running this and look at the dynamo errors. 
Full logs/traces here: https://gist.github.com/wconstab/5364080996629669de5a416a76bca509

anijain2305:
Yeah, I can take a look today.

anijain2305:
@wconstab I dont see any issues after rebasing to the latest `main` - https://gist.github.com/anijain2305/e93ea919c4325d2ef00a16039dd4c0bf

AOT Autograd warnings are known, and unrelated to this.

benchmarks/common.py
```diff
@@ -331,6 +334,66 @@ def speedup_experiment(args, model_iter_fn, model, example_inputs):
     return format_speedup(speedup, pvalue, is_correct=is_correct)
 
 
+def speedup_experiment_ds(args, model_iter_fn, model, example_inputs):
+    """
+    Measure speedups over eager using the autotuning inference backend.  To use this:
```
anijain2305, line 339: Can you change the description before merging?

benchmarks/common.py
```diff
@@ -668,6 +736,7 @@ def run_one_model(
         tolerance, cos_similarity = self.get_tolerance_and_cosine_flag(
             is_training, current_device, name
         )
+        static_inputs = (example_inputs[0][0], *example_inputs[1:])
```
anijain2305, line 739: Worried that it will cause other models to break? For example, hugggingface.py models have example_inputs as dict.

wconstab: ah. yea, this is wrong. i just did it as a hack.  forgot to fix.  hmm, i will need to do something more to integrate this properly.

i also need to fix how 'warmup' works for ds, so maybe the thing to do is just short-circuit all of this in the --dynamic-shapes case, jump straight to my `speedup-experiment-ds`, and do the warmup there.

benchmarks/common.py
```diff
@@ -331,6 +334,69 @@ def speedup_experiment(args, model_iter_fn, model, example_inputs):
...
+    Warms up using the first input example and then iterates the inputs,
+    measuring (and expecting minimal) variance between the runtime for different examples.
+
+    """
+    timings = np.zeros((args.repeat, len(example_inputs), 2), np.float64)
+
+    if args.repeat > 5:
+        print(f"\ndynamic shapes experiments are slow, consider setting --repeat less than {args.repeat}\n")
+
+    nwarmup = 2
```
anijain2305, line 352: Maybe 5? Torchscript requires more than 2 warmup iterations in my experience.

benchmarks/common.py
```diff
@@ -331,6 +334,69 @@ def speedup_experiment(args, model_iter_fn, model, example_inputs):
...
+        # Start each rep fresh, e.g. only warmup on example 0
+        torchdynamo.reset()
+        for _ in range(nwarmup):
+            with optimize_ctx:
+                model_iter_fn(model, example_inputs[0])
+
+        for input_idx, inputs in enumerate(example_inputs):
+            # interleave the runs to handle frequency scaling and load changes
+            timings[rep, input_idx, 0], expected_output = timed(
+                model, model_iter_fn, inputs, return_result=True
```
anijain2305, line 363: return_result= False? I don't think you are using the outputs for any accuracy checking here.

benchmarks/common.py
```diff
@@ -360,6 +426,12 @@ def baselines(models, model_iter_fn, example_inputs, args):
     """
     Common measurement code across all baseline experiments.
     """
+
+    if args.dynamic_shapes:
+        # TODO possibly merge ds logic back to main baselines fn, but kept it separate
+        # during exploration phase
+        return baselines_ds(models, model_iter_fn, example_inputs, args)
```
anijain2305, line 433: Are we missing this function?

benchmarks/common.py
```diff
@@ -674,6 +746,18 @@ def run_one_model(
             sys.stdout.flush()
             for submod in itertools.chain([model], model.modules()):
                 assert not torchdynamo.utils.is_jit_model(submod)
+
+            if dynamic_shapes:
+                # skip correctness check for ds benchmark, becuase example_inputs are not
+                # compatible with the code below, and the same benchmarks can be run in
+                # non-dynamic sshapo mode for correctness checks
```
anijain2305, line 753: sshapo typo

benchmarks/common.py
```diff
@@ -1059,8 +1148,8 @@ def main(runner, original_dir=None):
         runner.skip_models.clear()
 
     experiment = null_experiment
+    global current_name, current_device, output_filename, optimize_ctx
```
anijain2305, line 1151: Do we need optimize_ctx to be global?

anijain2305 (changes requested):

benchmarks/common.py
wconstab: well, i needed it passed into my experiment. i would be ok threading it through instead of making it global, but i wasn't sure which way to go.

benchmarks/common.py
wconstab: oops this was leftover from my first attempt, i am cleaning it up now.

jansel (approved):

benchmarks/common.py
anijain2305: Its ok. We can deal with this in a separate PR.

anijain2305 (approved):

Issue #630, opened 2022-07-21
TorchBench Inference Accuracy Error: "maml"


desertfire:
Repro:
```
python ./benchmarks/torchbench.py -dcuda --inductor --float32 --no-skip  -k maml
```

jansel:
There is a decomp for nll_loss_backward is `torch._decomp`

desertfire:
> There is a decomp for nll_loss_backward is `torch._decomp`

Yes, it's a one-line change I have done locally. The PR for `scatter` is pending. `argmax` is more difficult.

jansel:
Note triton now has upstream support for argmax.

desertfire:
All the missing ops are added. The latest error is an accuracy error,

```
cuda eval  maml                               Similarity score=0.0
INCORRECT
ERROR
```

desertfire:
Latest error msg:
```
Similarity score=0.3764550983905792
INCORRECT
ERROR
```

desertfire:
Close this one, track in https://github.com/pytorch/torchdynamo/issues/1305

PR #632, opened 2022-07-21
[inductor] Support scatter and scatter_reduce

desertfire:
Summary: scatter, scatter_add and scatter_reduce are very similar, so we
can lower scatter and scatter_add to scatter_reduce. There are certain
kwargs, such as reduce in scatter and include_self in scatter_reduce,
that need special cares.

TODO: We need to support more reduction types in scatter_reduce.

### DIFF
--- a/torchinductor/codegen/triton.py
+++ b/torchinductor/codegen/triton.py
@@ -630,7 +630,10 @@ class TritonKernel(Kernel):
         index, mask = self.indexing(index, value)
         if mode is None:
             line = f"tl.store({var} + {index}, {value}, {mask})"
-        elif mode == "atomic_add":
+        elif mode == "atomic_add" or mode == "atomic_add_exclude_self":
+            if mode == "atomic_add_exclude_self":
+                line = f"tl.store({var} + {index}, 0, {mask})"
+                self.stores.writeline(name, line)
             line = f"tl.atomic_add({var} + {index}, {value}, {mask})"
         else:
             raise NotImplementedError(f"store mode={mode}")
--- a/torchinductor/codegen/triton.py
+++ b/torchinductor/codegen/triton.py
@@ -630,10 +630,7 @@ class TritonKernel(Kernel):
         index, mask = self.indexing(index, value)
         if mode is None:
             line = f"tl.store({var} + {index}, {value}, {mask})"
-        elif mode == "atomic_add" or mode == "atomic_add_exclude_self":
-            if mode == "atomic_add_exclude_self":
-                line = f"tl.store({var} + {index}, 0, {mask})"
-                self.stores.writeline(name, line)
+        elif mode == "atomic_add":
             line = f"tl.atomic_add({var} + {index}, {value}, {mask})"
         else:
             raise NotImplementedError(f"store mode={mode}")
--- a/torchinductor/lowering.py
+++ b/torchinductor/lowering.py
@@ -1298,19 +1298,19 @@ def index_select(x, dim, indices):
 
 @register_lowering(aten.scatter_, type_promote=False)
 def scatter_(self, dim: int, index, src, **kwargs):
-    reduction_type = None
-    if kwargs:
-        if kwargs["reduce"] == "add":
-            reduction_type = "sum"
-        elif kwargs["reduce"] == "multiply":
-            reduction_type = "prod"
-        else:
-            assert False
+    reduction_type = kwargs.get("reduce")
+    if reduction_type == "add":
+        reduction_type = "sum"
+    elif reduction_type == "multiply":
+        reduction_type = "prod"
+    else:
+        assert reduction_type is None
     return scatter_reduce_(self, dim, index, src, reduction_type)
 
 
 @register_lowering(aten.scatter_reduce_, type_promote=False)
 def scatter_reduce_(self, dim: int, index, src, reduction_type: str = None, **kwargs):
+    # TODO: Need to support more reduction type
     assert reduction_type is None or reduction_type in {"sum"}
     assert isinstance(self, TensorBox)
     assert "int" in str(index.get_dtype())
@@ -1319,7 +1319,7 @@ def scatter_reduce_(self, dim: int, index, src, reduction_type: str = None, **kw
     self.realize()
     V.graph.realize_users_of(self.get_name())
 
-    include_self = kwargs["include_self"] if kwargs else True
+    include_self = kwargs.get("include_self", True)
     index_loader = index.make_loader()
 
     def output_indexer(idx):
--- a/torchinductor/lowering.py
+++ b/torchinductor/lowering.py
@@ -1298,19 +1298,20 @@ def index_select(x, dim, indices):
 
 @register_lowering(aten.scatter_, type_promote=False)
 def scatter_(self, dim: int, index, src, **kwargs):
+    reduction_type = None
     if kwargs:
-        assert kwargs["reduce"] in {"add", "multiply"}
-    return scatter_reduce_(self, dim, index, src, kwargs["reduce"] if kwargs else None)
-
-
-@register_lowering(aten.scatter_add_, type_promote=False)
-def scatter_add_(self, dim: int, index, src):
-    return scatter_reduce_(self, dim, index, src, "add")
+        if kwargs["reduce"] == "add":
+            reduction_type = "sum"
+        elif kwargs["reduce"] == "multiply":
+            reduction_type = "prod"
+        else:
+            assert False
+    return scatter_reduce_(self, dim, index, src, reduction_type)
 
 
 @register_lowering(aten.scatter_reduce_, type_promote=False)
 def scatter_reduce_(self, dim: int, index, src, reduction_type: str = None, **kwargs):
-    assert reduction_type is None or reduction_type in {"add", "sum"}
+    assert reduction_type is None or reduction_type in {"sum"}
     assert isinstance(self, TensorBox)
     assert "int" in str(index.get_dtype())
     assert 0 <= dim < len(self.get_size())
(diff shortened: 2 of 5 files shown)


desertfire:
For https://github.com/pytorch/torchdynamo/issues/630 and https://github.com/pytorch/torchdynamo/issues/633

torchinductor/lowering.py
```diff
@@ -1313,32 +1313,62 @@ def fn(idx):
     )
 
 
+@register_lowering(aten.scatter_, type_promote=False)
```
ngimel, line 1316: should this (and `scatter_add_`) be a decomposition instead?

torchinductor/lowering.py
```diff
@@ -1313,32 +1313,62 @@ def fn(idx):
...
     assert "int" in str(index.get_dtype())
     assert 0 <= dim < len(self.get_size())
-    src = to_dtype(src, self.get_dtype())
+
     self.realize()
     V.graph.realize_users_of(self.get_name())
 
+    include_self = kwargs["include_self"] if kwargs else True
     index_loader = index.make_loader()
+    if isinstance(src, TensorBox):
```
ngimel, line 1340: better bring it inside `fn`, separating this logic makes it more confusing

torchinductor/lowering.py
desertfire: Fixed `scatter_add_` in the new commit. Not able to decompose `scatter_` into `scatter_reduce_`, because `reduce` can't be None for `scatter_reduce_`.

torchinductor/lowering.py
desertfire: Fixed in the new commit.

torchinductor/lowering.py
ngimel: But you can put the exact code you have below in decomposition to make sure `reduce` is not None?

torchinductor/lowering.py
```diff
@@ -1313,32 +1313,62 @@ def fn(idx):
...
+    assert reduction_type is None or reduction_type in {"sum"}
     assert isinstance(self, TensorBox)
     assert "int" in str(index.get_dtype())
     assert 0 <= dim < len(self.get_size())
-    src = to_dtype(src, self.get_dtype())
+
     self.realize()
     V.graph.realize_users_of(self.get_name())
 
+    include_self = kwargs["include_self"] if kwargs else True
```
ngimel, line 1339: you can potentially have `kwargs` that wouldn't include `include_self`, because other, non-kwarg-only args can be passed as kwargs. Better do `kwargs.get("include_self", True)`

torchinductor/lowering.py
desertfire: I am not sure I understand. `reduce` can be None for `scatter_`.

ngimel: Yes, but you put this function as is in decomposition.py and decorate it with @register_decomposition instead of @register_lowering, would this not work?

desertfire: No, it doesn't work. It will trigger a checking failure from aot_grad/.

torchinductor/codegen/triton.py
```diff
@@ -611,7 +611,10 @@ def store(self, name, index, value, mode=None):
         index, mask = self.indexing(index, value)
         if mode is None:
             line = f"tl.store({var} + {index}, {value}, {mask})"
-        elif mode == "atomic_add":
+        elif mode == "atomic_add" or mode == "atomic_add_exclude_self":
+            if mode == "atomic_add_exclude_self":
```
jansel, line 615: I don't exactly understand this one...  if we zero it out then why do we need the atomic_add? Could this just be a normal store?
```
x=0
x+=y
```
seems the same as
```
x=y
```

torchinductor/lowering.py
jansel: I'm assuming because AOT autograd can't handle mutation.

torchinductor/codegen/triton.py
desertfire: `atomic_add` is needed because of the `index` in `scatter_add` can make it add to the same element multiple times, e.g.
```
>>> torch.scatter_add(torch.zeros(2, 2), 0, torch.tensor([[1],[1]]), torch.ones(2,2))
tensor([[0., 0.],
        [2., 0.]]
```

So even we zero out the destination, we still need an `atomic_add` here.

torchinductor/lowering.py
desertfire: I didn't really trace it but I think what happens here is because my implementation of `scatter_reduce_` is not exactly the same as the semantics of `scatter_reduce_` in PyTorch. The "official" `scatter_reduce_` does not take `None` as an input to the `reduce` parameter. So we can't decompose `scatter_` without `add` into `scatter_reduce_` when we leave the torchinductor land. I could be wrong though.

desertfire: Might be helpful to paste the error msg here,
```
...
../pytorch-release/torch/_ops.py:148: in __call__
    return self._op(*args, **kwargs or {})
../pytorch-release/torch/utils/_python_dispatch.py:74: in wrapped
    return f(self, *args, **kwargs)
../pytorch-release/torch/fx/experimental/proxy_tensor.py:290: in __torch_dispatch__
    return proxy_call(func_overload, args, kwargs)
../pytorch-release/torch/fx/experimental/proxy_tensor.py:78: in proxy_call
    return CURRENT_DECOMPOSITION_TABLE[func_overload](*args, **kwargs)
torchinductor/decomposition.py:376: in scatter
    return self.clone().scatter_(dim, index, src, **kwargs)
../pytorch-release/torch/fx/experimental/proxy_tensor.py:206: in __torch_dispatch__
    return proxy_call(func_overload, args, kwargs)
../pytorch-release/torch/fx/experimental/proxy_tensor.py:78: in proxy_call
    return CURRENT_DECOMPOSITION_TABLE[func_overload](*args, **kwargs)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = ProxyTensor(tensor([[0., 0., 0.],
        [0., 0., 0.]], device='cuda:0'), proxy=Proxy(clone_default)), dim = 0, index = ProxyTensor(tensor([[0]], device='cuda:0'), proxy=Proxy(primals_2))
src = ProxyTensor(tensor([[1., 1., 1.],
        [1., 1., 1.]], device='cuda:0'), proxy=Proxy(primals_3))

    @register_decomposition([aten.scatter_])
    def scatter_(self, dim: int, index, src, *, reduce: str = None):
        if reduce == "add":
            reduce = "sum"
        elif reduce == "multiply":
            reduce = "prod"
        else:
            assert reduce is None
>       return self.scatter_reduce_(dim, index, src, reduce)
E       TypeError: scatter_reduce_(): argument 'reduce' (position 4) must be str, not NoneType
```

torchinductor/lowering.py
```diff
@@ -1313,32 +1313,61 @@ def fn(idx):
...
+    elif reduce == "multiply":
+        reduce = "prod"
+    else:
+        assert reduce is None
+    return scatter_reduce_(self, dim, index, src, reduce)
+
+
+@register_lowering(aten.scatter_reduce_, type_promote=False)
+def scatter_reduce_(
+    self, dim: int, index, src, reduce: str, *, include_self: bool = True
```
ngimel, line 1329: nit: this type annotation seems wrong if reduce can be None?

ngimel (approved):

torchinductor/codegen/triton.py
jansel: how do we know the zeroing happens before the adding?

ngimel: Oh right, good catch, all zeroing will be racing.

jansel (changes requested):
Should do the zeroing in an extra kernel to avoid race.

torchinductor/lowering.py
```diff
@@ -1296,32 +1296,76 @@ def fn(idx):
...
 
 
-@register_lowering(aten.scatter_add_, type_promote=False)
-def scatter_add_(self, dim: int, index, src):
+@register_lowering(aten.scatter_, type_promote=False)
+def scatter_(self, dim: int, index, src, *, reduce: str = None):
+    if reduce == "add":
+        reduce = "sum"
+    elif reduce == "multiply":
+        reduce = "prod"
```
jansel, line 1304: This doesn't exist yet, perhaps just leave an assert/todo.

torchinductor/lowering.py
```diff
@@ -1296,32 +1296,76 @@ def fn(idx):
...
     index_loader = index.make_loader()
 
     def output_indexer(idx):
         indirect_idx = list(idx)
         indirect_idx[dim] = ops.indirect_indexing(index_loader(idx))
         return indirect_idx
 
+    def fn(idx):
+        if isinstance(src, TensorBox):
+            src_loader = src.make_loader()
```
jansel, line 1329: We should call make_loader() outside of fn in order to handle mutation properly.  (make_loader will resolve where StorageBox.data points at the time it is called.  Mutation swings the StorageBox.data pointer to the new value.)

jansel (approved):

jansel:
@desertfire I think test_scatter1_cpu might be flakey, in this run:
https://app.circleci.com/pipelines/github/pytorch/torchdynamo?branch=pull%2F646

For pipeline 446, it failed but worked when I clicked rerun.

desertfire:
> @desertfire I think test_scatter1_cpu might be flakey, in this run: https://app.circleci.com/pipelines/github/pytorch/torchdynamo?branch=pull%2F646
> 
> For pipeline 446, it failed but worked when I clicked rerun.

I deleted its corresponding `@unittest.skip` as it tested ok locally. Let me add it back and open another issue to track it.

PR #635, opened 2022-07-22
[DRAFT] add dump graph utils

yushangdi:
creates a fx graph of the buffers generated by lowering

with this patch you can run e.g.

`python benchmarks/torchbench.py --training --devices=cuda --inductor-dump --isolate -k hf_Bert`

to dump the forward and backward graphs of compute-buffers of hf_Bert. The resulting svg file will be in torchbenchmark/.

If you want to run it for other graphs, change the model_name in compile_fx so the dumped file matches the model name. 

Also the dumping is quite slow, so for large models like hf_Bert, you would want to change the number of layers to a small number.

### DIFF
--- a/benchmarks/common.py
+++ b/benchmarks/common.py
@@ -331,6 +331,30 @@ def speedup_experiment(args, model_iter_fn, model, example_inputs):
     return format_speedup(speedup, pvalue, is_correct=is_correct)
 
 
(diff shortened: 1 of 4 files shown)


jansel (approved):

yushangdi:
add in this PR : https://github.com/pytorch/torchdynamo/pull/677

PR #636, opened 2022-07-22
[RFC] Rewrite rand_cpu to be deterministic, use same algo as triton (philox)

voznesenskym:


torchinductor/codegen/cpp.py
```diff
@@ -75,10 +75,176 @@ def cpp_prefix():
...
             inline double mod(double a, double b) { return std::fmod(a, b); }
 
-            float rand_cpu(unsigned int seed) {
-                static thread_local std::mt19937 gen(seed ^ omp_get_thread_num());
-                static_assert(std::mt19937::min() == 0);
-                return gen() * static_cast<float>(1.0 / (std::mt19937::max() + 1.0));
+            #include <stdint.h>
+            #include <cmath>
+
+            namespace at {
```
jansel, line 81: Could we just find the location of this on disk and `#include` it?

torchinductor/codegen/cpp.py
```diff
@@ -223,8 +389,8 @@ def logical_or(a, b):
         return f"{a} || {b}"
 
     @staticmethod
-    def rand_cpu(seed: sympy.Expr, dtype):
-        return f"static_cast<{DTYPE_TO_CPP[dtype]}>(rand_cpu({seed}));"
+    def rand_cpu(seed: sympy.Expr, offset: sympy.Expr, dtype):
```
jansel, line 392: We should be able to remove the `rand_cpu` op and just share the same op as triton.

jansel (changes requested):

torchinductor/codegen/cpp.py
voznesenskym: We could, but then we need to change to have `cpp_compile_command(... include_pytorch=True)` as this dep comes from pytorch, and it feels odd to add this as a dep just for a single header.

torchinductor/codegen/cpp.py
voznesenskym: Sorry, can you elaborate?

torchinductor/codegen/cpp.py
voznesenskym: To elaborate a bit: I don't think plumbing to just include a single header without the requisite deps in the library path will work. Even if we make the assumption that the dep is already installed - that header pulls in other headers, and it feels odd to then plumb in a few headers like that without modifying the paths and libs, which in turn feels too heavy for such a small piece of functionality.

torchinductor/codegen/cpp.py
voznesenskym: Or do you literally mean just call this op `rand` and do the cuda check in here?

jansel: Because of this line:
https://github.com/jansel/torchdynamo/blob/f6fd47b2adb833d5b86a5f8e38017b369a537c34/torchinductor/lowering.py#L634

We generate `ops.rand()` in triton and `ops.rand_cpu()` on CPU.  

I want them to both be `ops.rand()`.

torchinductor/codegen/cpp.py
jansel: I think we could just splat in the absolute path with something like:
```py
philox_h=os.path.join([x for x in torch.utils.cpp_extension.include_paths() if x.endswith("/torch/include")][0], "ATen/core/PhiloxRNGEngine.h")

cpp_string = f"""
...
#include "{philox_h}"
...
"""
```

That assumes it is a header-only dep, and we don't need to link any symbols.

voznesenskym: I assume this is a demonstrative snippet, and you mean doing that for all the dependent paths too? As we won't have those headers in the search paths, we'd have to absolute-path include them too. I ask, because although I have not tested it yet, but I think if we just do what we have in your snippet here, you're going to get a file not found for the first dep not in the search paths.

jansel: No, just for this one file we need from the pytorch source tree.  This should be the only one.

All the existing includes are C++ standard libraries.

voznesenskym: That's because I rewrote it to be like that.

This is the real list of includes:

```
// define constants like M_PI and C keywords for MSVC
#ifdef _MSC_VER
#define _USE_MATH_DEFINES
#include <math.h>
#endif

#include <stdint.h>

#ifdef __CUDACC__
#include <cuda.h>
#endif

#include <ATen/core/Array.h>
#include <c10/macros/Macros.h>
#include <c10/util/Exception.h>
#include <c10/util/Half.h>
#include <cmath>
```

And those pull in others, etc etc. At its core, if we trim it down, we don't really need most of those, but if we import it as is we have to also pull them in or get a file not found.

Tested it, also, and confirming it is file not found unless we grab the deps as well.

torchinductor/codegen/cpp.py
jansel: Let's just add the needed `-I` to the `g++` command then.

torchinductor/codegen/cpp.py
jansel: Let's just add the needed -I to the g++ command then.

voznesenskym: SGTM, easy change.

torchinductor/codegen/cpp.py
```diff
@@ -68,17 +68,21 @@ def cpp_prefix():
             #include <random>
             #include <omp.h>
```
voznesenskym, line 69: Let me check which of these deps we still need, probably not random, and I am not sure where we use iostream?

voznesenskym: @jansel I see we use `std::atomic_ref` from <atomic>, I don't see any usage of iostream, do you recall why this is here?

jansel: No I don't.  If it passes tests+torchbench we can remove it.

torchinductor/codegen/cpp.py
```diff
@@ -63,22 +63,24 @@ def cpp_prefix():
...
-            float rand_cpu(unsigned int seed) {
-                static thread_local std::mt19937 gen(seed ^ omp_get_thread_num());
-                static_assert(std::mt19937::min() == 0);
-                return gen() * static_cast<float>(1.0 / (std::mt19937::max() + 1.0));
+            constexpr float normalize (uint32_t value) {
+                return value * static_cast<float>(1.0 / std::numeric_limits<uint32_t>::max());
+            }
+
+            float normalized_rand_cpu(uint64_t seed, uint64_t offset) {
+                return normalize(at::Philox4_32_10(seed, omp_get_thread_num(), offset)());
```
jansel, line 83: Why do we need `omp_get_thread_num()` here?  Deterministic random should not depend on openmp scheduler.

torchinductor/codegen/cpp.py
```diff
@@ -63,22 +63,24 @@ def cpp_prefix():
...
 
-            float rand_cpu(unsigned int seed) {
-                static thread_local std::mt19937 gen(seed ^ omp_get_thread_num());
-                static_assert(std::mt19937::min() == 0);
-                return gen() * static_cast<float>(1.0 / (std::mt19937::max() + 1.0));
+            constexpr float normalize (uint32_t value) {
+                return value * static_cast<float>(1.0 / std::numeric_limits<uint32_t>::max());
+            }
+
+            float normalized_rand_cpu(uint64_t seed, uint64_t offset) {
```
jansel, line 82: Seed/offset is defined as uint32_t at the call point.

torchinductor/lowering.py
```diff
@@ -631,14 +631,15 @@ def rand_or_randn(
...
 
+            random_pos = ir.FixedLayout(
+                device,
+                dtype,
+                size,
+                ir.FlexibleLayout.contiguous_strides(size),
+                offset=offset,
+            ).make_indexer()
+
             if device.type == "cuda":
```
jansel, line 642: Can we remove this condition entirely?

jansel (changes requested):

torchinductor/lowering.py
voznesenskym: Good question! I am actually ignorant of the logic here. Let me try. What I don't know is - can we use the same seed for both? ATM, the cuda seed is from `ops.load(V.graph.random_seed_buffer(device), sympy.Integer(0))` and the cpu seed is from `ops.index_expr(V.graph.sizevars.seed(), torch.int32)`. Ostensibly, yes, since `random_seed_buffer` looks like it should just work with a dtype. 

Let me try it.

jansel (approved):

jansel:
@voznesenskym with this PR might be able to disable this skip:
https://github.com/pytorch/torchdynamo/blob/6639e0f3945e92765f24ff3f037e362df5faefb6/tests/test_torchinductor.py#L2231

voznesenskym:
> raise unittest.SkipTest("CPU currently nondeterministic")

https://github.com/pytorch/torchdynamo/pull/650 Sadly, this just moves the error to: `RuntimeError: "bernoulli_scalar_cpu_" not implemented for 'Half'`

I took a peek at `bernoulli_scalar_cpu_`... we could try changing it to AT_DISPATCH_ALL_TYPES_AND3 with Half, but I am not sure where that would break. I can take a look sometime later, if this becomes blocking or important later on.

PR #653, opened 2022-07-25
Add training/AotAutograd example to README

wconstab:
Adds a minimal description and example of using AotAutograd with dynamo
- doesn't actually run, maybe I should add the code for a real model
- omits further discussion about 'AOT' family of backends and which ones to use, probably should add this too

cc @jiayisunx


jansel (approved):

wconstab:
* doesn't actually run, maybe I should add the code for a real model
* omits further discussion about 'AOT' family of backends and which ones to use, probably should add this too

PR #655, opened 2022-07-26
RFC - pass result back to avoid making proxies for torch.* methods

voznesenskym:
Given:

```
def foo(x):
    y = torch.empty(3)
    return x * y

torchdynamo.reset()
graph = None
def capturing_compiler(c, i):
    global graph
    graph = c
    print(c.graph.print_tabular())
    return graph.forward

with torchdynamo.optimize(capturing_compiler, nopython=True):
    z = foo(torch.randn([1, 1]))
```

and

```
outgm, wrap = python_key_normalize(
    graph,
    torch.randn([1, 1])
)
```

Before this change, we would return:

image

After this change, we return:

image

Longer term, we should probably switch back to only a single impl of this, using the one in fx ProxyTensor (cc: @Chillee)

### DIFF
--- a/torchdynamo/optimizations/python_key.py
+++ b/torchdynamo/optimizations/python_key.py
@@ -6,6 +6,7 @@ from itertools import chain
 import torch
 from torch.fx import GraphModule
 from torch.fx import Node
+from torch.fx.experimental.proxy_tensor import ProxyTorchDispatchMode
 from torch.nn.utils import _stateless
 
 from ..allowed_functions import torch_get_name
@@ -96,6 +97,8 @@ def python_key_normalize(
     params_len = len(params_flat)
     nargs = params_len + len(example_inputs)
 
+    tracer = None
+
     class PatchingInterpreter(torch.fx.Interpreter):
         def run_node(self, n: torch.fx.Node):
             try:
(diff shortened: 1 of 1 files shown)


jansel (changes requested):
Suppose you have

```
@torchdynamo.optimize(...)
def foo():
   return torch.empty(10)


a=foo()
a.fill_(3)
b=foo()
b.fill_(4)

print(a)
print(b)
```


With our current behavior, `a` and `b` will be separate tensors.

With your example graph `a` and `b` will alias each other and produce wrong results.

voznesenskym:
> Suppose you have
> 
> ```
> @torchdynamo.optimize(...)
> def foo():
>    return torch.empty(10)
> 
> 
> a=foo()
> a.fill_(3)
> b=foo()
> b.fill_(4)
> 
> print(a)
> print(b)
> ```
> 
> With our current behavior, `a` and `b` will be separate tensors.
> 
> With your example graph `a` and `b` will alias each other and produce wrong results.

Interesting. I suppose we need something the preserves the behavior of 'make a new tensor like this'. Ideally it wouldn't be a placeholder, but a call_function to `aten::zeros`. Let me see what that might look like.

Chillee:
A short term fix would be to run the op while enabling `ProxyTorchDispatchMode` (https://github.com/pytorch/pytorch/blob/master/torch/fx/experimental/proxy_tensor.py#L365)

Something like...
```
with proxy_mode:
   result = super().run_node(n)
```
might work?

voznesenskym:
> A short term fix would be to run the op while enabling `ProxyTorchDispatchMode` (https://github.com/pytorch/pytorch/blob/master/torch/fx/experimental/proxy_tensor.py#L365)
> 
> Something like...
> 
> ```
> with proxy_mode:
>    result = super().run_node(n)
> ```
> 
> might work?

Funny that you mention ProxyTorchDispatchMode - I was just reading that because I was trolling around the codebase to see how we handle the 'subsequent-operation-modifies-a-constant' as another thing that could go wrong by just capturing an input constant and passing it around naively. The solution there around lift_fresh vs lift_fresh_copy seems pretty neat. 

Let me give this a try.

voznesenskym:
After latest commit and proxy dispatch mode, it looks like this 

image

Which matches: "Ideally it wouldn't be a placeholder, but a call_function to aten::empty." 

Still RFC though - open to other approaches as well.

jansel (approved):

PR #657, opened 2022-07-26
Trace through Adam optimizer step()

mlazos:
* Add torch.optim to skipfiles
* Don't allow torch.optim* in the graph
* Unwrap profiling functions from optimizer.step() calls
* Add FakeItemVariable to handle calls to item() on FakeTensors
* Added a test to verify number of frames compiled from Adam.step

Adam optimizer for #328


torchdynamo/allowed_functions.py
```diff
@@ -116,12 +116,17 @@ def _allowed_function_ids():
...
         allowed_modules = ("torch", "math")
+        disallowed_modules = "torch.optim."
         allowed_modules_dot = tuple([x + "." for x in allowed_modules])
         module = inspect.getmodule(obj)
         if module is None:
             return False
 
         mod_name = module.__name__
+
+        if mod_name.startswith(disallowed_modules):
```
voznesenskym, line 127: nit - move `allowed_modules` and `allowed_modules_dot` lines below these checks?

torchdynamo/config.py
```diff
@@ -95,4 +92,4 @@
 
 # Not all backends support scalars. Some calls on torch.Tensor (like .item()) return a scalar type.
 # When this flag is set to False, we introduce a graph break instead of capturing.
-capture_scalar_outputs = False
+capture_scalar_outputs = True
```
voznesenskym, line 95: I would make sure this is an okay default - I am not sure every backend is alright with this. If it is, we should just pull the config, as I made it config driven out of an abundance of caution w/r/t backends.

Chillee:
Why do we need to capture scalar outputs?

torchdynamo/eval_frame.py
```diff
@@ -360,6 +361,24 @@ def patch():
...
+        for opt in optimizers:
+            opt._cuda_graph_capture_health_check = disable(
+                opt._cuda_graph_capture_health_check
+            )
+            # disable future hooking
+            setattr(opt.step, "hooked", True)
+            # disable any currently set hooks
+            unwrapped_step = getattr(opt.step, "__wrapped__", None)
+            if unwrapped_step:
+                opt.step = unwrapped_step
```
voznesenskym, line 380: Do we nee to clean this up when we exit dynamo?

torchdynamo/skipfiles.py
```diff
@@ -104,6 +104,16 @@ def _module_dir(m: types.ModuleType):
 FILENAME_ALLOWLIST = {
     torch.nn.Sequential.__init__.__code__.co_filename,
 }
+
+# Include optimizer code for tracing
+FILENAME_ALLOWLIST |= set(
+    [
+        inspect.getfile(obj)
+        for obj in torch.optim.__dict__.values()
+        if inspect.isclass(obj)
```
voznesenskym, line 113: nit - you do this iteration twice, make a helper?

mlazos:
> Why do we need to capture scalar outputs?

I don't need to change the default here, I can just change it for the test. This is because of a call to `item()` in Adam.  Alternatively, I discussed with Jason that Adam has a `capturable` parameter, which when enabled works around the `item()` call. I'll need to PR a change to Pytorch to fix some asserts around the usage of `capturable` though.

torchdynamo/config.py
mlazos: Yeah I don't think I need to make this the default. I can leave it up to the user to decide whether they want to use this

torchdynamo/eval_frame.py
mlazos: Agreed, I can change this to on enter/exit

torchdynamo/skipfiles.py
mlazos: The second iteration only contains 8 values, I don't think this is worth it, it's also only run once when this file is imported.

torchdynamo/eval_frame.py
mlazos: I chatted with Jason about this a little more, and he said it's okay for now as longer term we will upstream these patches. Upstreaming this one will be trickier, but I think I can do it with a decorator. The reason I asked was because patching and unpatching on every enter and exit will have a tiny perf hit. In the end, the only person who will encounter this issue is someone who imports dynamo and then tries to interleave using dynamo with using `torch.autograd.profiler.record_function`.

voznesenskym: Yeah, its not a crazy important case - more about leaving things as we found them. I am find with us kicking this can down the road if you and Jason are.

jansel (approved):

Issue #633, opened 2022-07-21
size_hint assertion: "hf_Reformer"


desertfire:
Repro:
```
./torchbench.py -dcuda --inductor --float16 --no-skip -k hf_Reformer
```

desertfire:
The latest error,
```
Traceback (most recent call last):
  File "/fsx/users/binbao/torchdynamo-tip/torchinductor/graph.py", line 221, in call_function
    return lowerings[target](*args, **kwargs)
  File "/fsx/users/binbao/torchdynamo-tip/torchinductor/lowering.py", line 139, in wrapped
    return decomp_fn(*args, **kwargs)
  File "/fsx/users/binbao/torchdynamo-tip/torchinductor/lowering.py", line 1347, in index
    output_size[j] = V.graph.sizevars.guard_equals(
  File "/fsx/users/binbao/torchdynamo-tip/torchinductor/sizevars.py", line 127, in guard_equals
    assert self.size_hint(expr) == 0, (expr, self.size_hint(expr))
AssertionError: (-4999, -4999)

torchinductor.exc.LoweringException: AssertionError: (-4999, -4999)
  target: aten.index.Tensor
  args[0]: TensorBox(
    ReinterpretView(
      StorageBox(
        InputBuffer(name='primals_2', layout=FixedLayout('cuda', torch.float32, size=[1438452, 1], stride=[1, 1]))
      ),
      FixedLayout('cuda', torch.float32, size=[4, 359613], stride=[359613, 1])
    )
  )
  args[1]: [TensorBox(
    View(
      StorageBox(
        Pointwise(
          'cuda',
          torch.float32,
          index_expr(i0, torch.float32),
          ranges=[4]
        )
      ),
      size=(4, 1),
      reindex=lambda i0, i1: [i0]
    )
  ), TensorBox(StorageBox(
    ConcatKernel(name='buf31', layout=FixedLayout('cuda', torch.int64, size=[4, 5000], stride=[5000, 1]), inputs=[ComputedBuffer(name='buf26', layout=AliasedLayout('cuda', torch.int64, size=[4, 1000], stride=[5000, 1]), data=Pointwise(
      'cuda',
      torch.int64,
      load(buf9, 1000*i0 + i1, False) + constant(0, torch.int64),
      ranges=[4, 1000]
    )), ComputedBuffer(name='buf27', layout=AliasedLayout('cuda', torch.int64, size=[4, 1000], stride=[5000, 1]), data=Pointwise(
[...]

PR #658, opened 2022-07-26
[inductor] Support argmax and argmin

desertfire:


desertfire:
For https://github.com/pytorch/torchdynamo/issues/630, https://github.com/pytorch/torchdynamo/issues/633

desertfire:
For the C++ backend, we generate the following kernel with a user-defined OMP reduction operation,
```
                       float* __restrict__ out_ptr0,
                       const long ks0,
                       const long ks1)
{
    {
        struct IndexValue_1 {size_t index; float value;};
        IndexValue_1 tmp1(0, std::numeric_limits<float>::min());
        #pragma omp declare reduction(argmax : struct IndexValue_1 :\
            omp_out.value = omp_in.value < omp_out.value ? omp_out.value : omp_in.value,\
            omp_out.index = omp_in.value < omp_out.value ? omp_out.index : omp_in.index)\
                initializer(omp_priv = {0, std::numeric_limits<float>::min()})
        #pragma omp parallel num_threads(4)
        {
            #pragma omp for nowait reduction(argmax:tmp1)
            for(long i0=0; i0<ks0*ks1; ++i0)
            {
                auto tmp0 = in_ptr0[i0];
                if (tmp1.value < tmp0) {
                    tmp1.index = i0; tmp1.value = tmp0;
                }
            }
        }
        out_ptr0[0] = tmp1.index;
    }
}
```

jansel (approved):

jansel:
> This is the generated Trition kernel for a simple argmax. Thanks to @jansel for the idea on how to map the result of tl.argmax back to the orginal index using masked tl.where + tl.sum.
> 
> ```
> @reduction_heuristics(size_hints=[1, 8192])
> @triton.jit
> def kernel0(in_ptr0, out_ptr0, ks0, ks1, xnumel, rnumel, XBLOCK : tl.constexpr, RBLOCK : tl.constexpr):
>     xoffset = tl.program_id(0) * XBLOCK
>     xindex = xoffset + tl.reshape(tl.arange(0, XBLOCK), [XBLOCK, 1])
>     xmask = xindex < xnumel
>     rbase = tl.reshape(tl.arange(0, RBLOCK), [1, RBLOCK])
>     _tmp1 = tl.zeros([XBLOCK, RBLOCK], tl.float32) + float("-inf")
>     _tmp1_index = tl.zeros([XBLOCK, RBLOCK], tl.int32)
>     for roffset in range(0, rnumel, RBLOCK):
>         rindex = roffset + rbase
>         rmask = rindex < rnumel
>         r0 = rindex
>         tmp0 = tl.load(in_ptr0 + r0, rmask, eviction_policy='evict_last').to(tl.float32)
>         _tmp1_index = tl.where(xmask & rmask & (_tmp1 < tmp0),  rindex, _tmp1_index)
>         _tmp1 = tl.where(xmask & rmask & (_tmp1 < tmp0), tmp0, _tmp1)
>     _tmp1_index_reduce = tl.reshape(tl.argmax(_tmp1, 1), [XBLOCK, 1]).to(tl.int32)
>     _tmp1_index_mask = (tl.arange(0, RBLOCK)[None, :] == _tmp1_index_reduce)
>     tmp1 = tl.reshape(tl.sum(tl.where(_tmp1_index_mask, _tmp1_index, 0), 1), [XBLOCK, 1])
>     tl.store(out_ptr0 + 0 + tl.zeros(tmp1.shape, tl.int32), tmp1, xmask)
> ```

@ptillet @Jokeren I am curious about your thoughts for this generated code for `torch.argmax`.

Because we have a loop (to handle large sizes), we end up doing two Triton reductions:
1) `tl.argmax` to figure out which thread had found the max (in a thread-local accumulator)
2) `tl.sum` where all but one thread in the block contributes a zero, to select the "true" index of the max.  This step could also be done via storing to a temporary buffer and doing an indirect read. 

This pattern is a little awkward, and I am guessing the extra `tl.sum()` introduces barriers in the generated code.

One thing that would simplify this is if we had a 2-arg version of `tl.argmax()`, that took both the values and the indices.  In this example it would be
```
tl.argmax(_tmp1, _tmp1_index)
```
which would return `_tmp1_index[i]` for the `i` that maximizes `_tmp1[i]`.

ngimel:
alternatively, do we have something like `tl.gather` or something to indirectly address `_tmp1_index` with indices returned by `tl.argmax`?

Jokeren:
1. Would the following alternative implementation suffice your needs?

https://github.com/Jokeren/triton-samples/blob/main/argmax.py

2. Supporting tl.gather (and tl.scatter) seems like a good idea. I do think we will encounter many use cases when implementing GNNs, though not a major target for now.

3. Since the above code is generated for 1d cases only, calculating the xmask seems a little bit redundant. I'll suggest getting rid of the xmask related code for 1d cases to seem if you've got any speedup.

ngimel:
3. The above code is not limited to 1d cases, and is supposed to support multi-d, including outer reductions. In the implementation you linked to looks like a single value only will be stored `out_ptr0 + tl.where(_tmp1_index_mask, 0, 0)` - the only address is out_ptr0 + 0, but XBLOCK will often be larger than 1.

Jokeren:
@ngimel Please see my updated code above, which is supposed to support 2d cases. (I don't have the original 2d code, so just executed in my mind 😃 ).

ngimel:
Oh nice, using masking to emulate gather!

jansel:
@Jokeren that is a super cool trick.

It will work for the case of no fusion, but in some cases we don't actually store the output of the reduction and instead pass it directly into another fused op.

Jokeren:
Thank you folks! I'll also take a note on the gather/scatter implementation.

ptillet:
FWIW, I don't think the `gather` implementation will be significantly better than the `sum` implementation, since it would likely also require thread synchronization (i.e., write tensor to shared memory + sync + have thread 0 read desired index and broadcast to other threads). Changing the signature of the argmax could allow us to re-use the synchronization barriers used in the argmax computation though, so there could be a win there. But honestly, I wouldn't worry about the cost of synchronization barriers in bandwidth-bound kernels

Jokeren:
Thank you Phil!

It does need an extra barrier for the gather operation. I'm was thinking that it could be useful in certain cases and it is worth implementing it in the future. 

Regarding the interface of the argmax function, my assumption is that keeping the interface similar to pytorch might be beneficial for users. Meta folks can correct me if I were wrong :).

ngimel:
@jokeren user-visible interface indeed returns just the index, but internal implementation (which is closer to the level that triton kernel operates on) keeps track of both index and value: https://github.com/pytorch/pytorch/blob/72e4aab74b927c1ba5c3963cb17b4c0dce6e56bf/aten/src/ATen/native/SharedReduceOps.h#L447-L466

Jokeren:
Thanks for the information! I'll think a bit more about the interface.

torchinductor/codegen/cpp.py
```diff
@@ -39,9 +39,9 @@ def reduction_init(reduction_type, dtype):
     if reduction_type in ("sum", "any"):
         return 0
     # TODO(jansel): infinity for floats?
-    if reduction_type == "max":
+    if reduction_type == "max" or reduction_type == "argmax":
         return f"std::numeric_limits<{DTYPE_TO_CPP[dtype]}>::min()"
-    if reduction_type == "min":
+    if reduction_type == "min" or reduction_type == "argmin":
```
agentydragon, line 44: nit: `reduction_type in {"min", "argmin"}`

desertfire:
I will commit this now and come back to enhance it if there is any updates on the Triton API.

Jokeren:
Sounds good, will let you know once I push any updates. We are busy with MLIR rewriting at the this point, so probably will get back to you a month or two later.

PR #659, opened 2022-07-26
simplify ModularIndexing even if base is an expr

ngimel:
Per title, Modular indexing is simplified as 
            `ModularIndexing(2 * i2 + r3, 1, 64) => 2 * i2 + r3` with `i2=32` and `r3=2`

### DIFF
--- a/torchinductor/sizevars.py
+++ b/torchinductor/sizevars.py
@@ -58,6 +58,8 @@ class SizeVarAllocator(object):
         return sympy.expand(expr).subs(self.replacements)
 
     def simplify_with_ranges(self, expr, var_ranges):
+        if isinstance(var_ranges, tuple):
+            var_ranges = var_ranges[0]
         """
         Simplify indexing expression with knowledge of the ranges of
         iteration variables.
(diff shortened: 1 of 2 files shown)


torchinductor/sizevars.py
```diff
@@ -58,6 +58,8 @@ def simplify(self, expr):
         return sympy.expand(expr).subs(self.replacements)
 
     def simplify_with_ranges(self, expr, var_ranges):
+        if isinstance(var_ranges, tuple):
```
jansel, line 61: Move this after the docstring (python won't fill `__doc__` unless the docstring is first line)

Perhaps we should also `assert len(var_ranges)==1`

torchinductor/sizevars.py
```diff
@@ -83,8 +85,14 @@ def remove_zero_terms(base, divisor):
         def visit_indexing_div(base, divisor):
             return IndexingDiv(remove_zero_terms(base, divisor), divisor)
 
-        def visit_moduler_indexing(base, divisor, modulus):
+        def visit_modular_indexing(base, divisor, modulus):
             base = remove_zero_terms(base, divisor)
+            if isinstance(base, sympy.Expr):
```
jansel, line 90:
```suggestion
            if isinstance(base, (sympy.Expr, sympy.Symbol)):
```
I think with this change we should be able to delete lines 96 through 102.

jansel (changes requested):

torchinductor/sizevars.py
jansel: Actually, we might be able to just remove this condition entirely and run this always?  I think the only other case is a `sympy.Integer`, and this should work for that fine too.

jansel (approved):

PR #660, opened 2022-07-26
Makefile - Add option for fast pytorch build

anijain2305:
Adding an option for fast pytorch build. Reduces the biuld time to half.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -92,7 +92,7 @@ pull-deps:
 	(cd ../detectron2     && git pull && git submodule update --init --recursive)
 	(cd ../torchbenchmark && git pull && git submodule update --init --recursive)
 
-build-deps: clone-deps
+define conda_deps
 	# conda env remove --name torchdynamo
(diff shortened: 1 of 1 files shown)


Chillee:
@anijain2305 You wanna add this into pytorch core?

anijain2305:
@Chillee Yeah, I can do that. Any particular place where you would want this? 

I would also want to have this in Dynamo repo. All the other deps are also in this Makefile. So, its convenient for me.

Chillee:
@anijain2305 Ed was proposing that we just add this as a flag to the PyTorch build options. So `BUILD_FAST=1 python setup.py develop` would work.

anijain2305:
Ah I see. Thats a good idea. Let me get this in PyTorch then

jansel (approved):

anijain2305:
Not needed. We can do this in PyTorch with BUILD_FAST - https://github.com/pytorch/pytorch/pull/82292

Issue #661, opened 2022-07-26
[inductor] fails with index-simplification error


Chillee:
```
import torchinductor
import torchinductor.config
from torchinductor.compile_fx import compile_fx
import torch
import torch.nn as nn
import torch.fx as fx

import time

# torchinductor.config.debug = True

inps = [torch.randn(512, 256, device='cuda')]

class Mod(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.layers = nn.ModuleList([nn.Linear(256, 256) for _ in range(5)])

    def forward(self, x):
        for layer in self.layers:
            x = layer(x).cos().cos()
        return (x,)


mod = fx.symbolic_trace(Mod()).cuda()

new_mod = compile_fx(mod, inps)
new_mod(*inps)[0].sum().backward()
iters = 100

begin = time.time()
torch.cuda.synchronize()
for _ in range(100):
    new_mod(*inps)[0].sum().backward()
torch.cuda.synchronize()
print((time.time()-begin)/iters)
```
```
  File "/scratch/chilli/fresh/torchdynamo/torchinductor/graph.py", line 260, in codegen
    self.scheduler.codegen()
  File "/scratch/chilli/fresh/torchdynamo/torchinductor/scheduler.py", line 767, in codegen
    self.get_backend(device).codegen(*group)
  File "/scratch/chilli/fresh/torchdynamo/torchinductor/codegen/triton.py", line 883, in codegen
    node.run(*kernel.split_and_set_ranges(node.get_ranges()))
  File "/scratch/chilli/fresh/torchdynamo/torchinductor/codegen/triton.py", line 513, in split_and_set_ranges
    assert all(s == 1 for s in remaining)
AssertionError
```

jansel:
There is a fix for this one in my branch.

jansel:
Fixed in https://github.com/pytorch/torchdynamo/pull/662

PR #662, opened 2022-07-27
[inductor] Add realize_hint and small fixes

jansel:


torchinductor/codegen/triton.py
```diff
@@ -510,7 +510,9 @@ def getter(flat_vars):
 
             return_getters_groups.append(return_getters)
 
-        assert all(s == 1 for s in remaining)
+        assert all(
+            V.graph.sizevars.size_hint(s) == 1 for s in remaining
+        ), f"failed to set ranges {remaining} {lengths}"
```
desertfire, line 515: This is already in the tip. Is this the main fix for https://github.com/pytorch/torchdynamo/issues/661, or is `realize_hint` the main fix?

torchinductor/ir.py
```diff
@@ -2653,27 +2656,43 @@ def realize(self):
         self.data.name = V.graph.register_buffer(self.data)
         return self.data.name
 
+    def realize_hint(self):
+        """
+        Called on buffers we expect to be forced to realize later.
+        """
+        if self.num_reads() > 1:
```
desertfire, line 2663: I am trying to relate this to some recent discussion in our group chat. Does this change mean we are not leveraging DSE anymore? Is there any performance implication here?

desertfire: Oh, I see you are replacing certain `realize` with `realize_hint`, so this should only make perf better.

torchinductor/ir.py
```diff
@@ -2653,27 +2656,43 @@ def realize(self):
...
+                num_reads > config.realize_reads_threshold
                 or len(self.inner_fn_str()) > config.realize_bytes_threshold
             ):
                 self.realize()
 
+    def num_reads(self):
+        if isinstance(
+            self.data, (ComputedBuffer, InputsKernel, InputBuffer, ReinterpretView)
+        ):
+            return 1
```
desertfire, line 2683: Why return 1 for `ComputedBuffer`? Should we recursively call `num_reads` on it?

jansel: Fixed

desertfire (approved):

PR #665, opened 2022-07-27
Dump graph env

yushangdi:
creates a fx graph of the buffers generated by lowering

with this patch you can run e.g.

`INDUCTOR_SCHEDULER_GRAPH=1 python benchmarks/torchbench.py --training --devices=cuda --inductor --skip-accuracy-check -n 1 --isolate -k hf_Bert`

to dump the forward and backward graphs of compute-buffers of hf_Bert. The resulting svg file will be in torchbenchmark/. 

The dumped files' names are in the format of compute_buffer_{num}.svg.

Also the dumping is quite slow, so for large models like hf_Bert, you would want to change the number of layers to a small number.

The graph dumped looks like this: https://www.svgviewer.dev/s/PpmacAjw

image


yushangdi:
changed to be gated by a config flag instead of using a whole different path per request.
A different PR than https://github.com/pytorch/torchdynamo/pull/635

torchinductor/scheduler.py
```diff
@@ -4,12 +4,17 @@
...
 import itertools
 import logging
+import os
 from typing import Any
 from typing import Dict
 from typing import List
 
 import numpy as np
 import torch
+from functorch._src.partitioners import draw_graph
```
Chillee, line 14: Might want to move this import into the function.

torchinductor/scheduler.py
```diff
@@ -458,6 +559,13 @@ def __init__(self, nodes):
                 assert False, node
         self.name_to_node = {node.get_name(): node for node in self.nodes}
 
+        if bool(os.environ.get("INDUCTOR_DEBUG", False)):
```
Chillee, line 562: Rename this to `INDUCTOR_SCHEDULER_GRAPH=1` or something like that for now.

We'll need a better system eventually (cc: @jansel ).

torchinductor/scheduler.py
```diff
@@ -458,6 +559,13 @@ def __init__(self, nodes):
...
+            global graph_dump_index
```
Chillee, line 563: So... we only create a single scheduler per graph we pass to inductor. I'd propose instead that we make a global variable that anybody can access for the purposes of logging.

Chillee (commented):
Mostly LGTM, although I defer to Jason if he has any strong opinions on how debug logging should be structured.

torchinductor/scheduler.py
yushangdi: changed, we also need the PR here: https://github.com/pytorch/pytorch/pull/82368

torchinductor/scheduler.py
yushangdi: renamed

torchinductor/scheduler.py
yushangdi: moved

torchinductor/scheduler.py
```diff
@@ -418,6 +422,105 @@ def get_name(self):
...
 
+def get_fake_func(name):
+    def func1(*args):
+        return 0
+
+    func1.__name__ = name
+    return func1
+
+
+def create_fx_graph(nodes, fname, print_graph=False):
```
Chillee, line 433: By the way, let's move this code to `utils.py`

Chillee (approved):
Other than one final note, LGTM.

Not sure if anybody else has any opinions :)

(might want to wait until other folks chime in)

cc: @anijain2305 @ngimel @desertfire

torchinductor/scheduler.py
yushangdi: resolved offline. Will keep this here because moving it to utils.py might cause circular dependency in the future.

torchinductor/scheduler.py
```diff
@@ -458,6 +561,12 @@ def __init__(self, nodes):
                 assert False, node
         self.name_to_node = {node.get_name(): node for node in self.nodes}
 
+        if INDUCTOR_SCHEDULER_GRAPH:
+            from functorch._src.aot_autograd import get_graph_being_compiled
+
+            graph_name = get_graph_being_compiled()
+            create_fx_graph(self.nodes, graph_name, print_graph=True)
```
ngimel, line 568: not sure you want `print_graph=True` here?

torchinductor/scheduler.py
```diff
@@ -418,6 +422,105 @@ def get_name(self):
...
+
+    outputs = []
+    for _, v in name_to_fx_node.items():
+        if len(v.users) == 0:
+            outputs.append(v)
+    graph.output(outputs[0] if len(outputs) == 1 else tuple(outputs))
+
+    if print_graph:
+        print(graph)
+    print("starting creating module")
```
ngimel, line 516: stray prints?

Chillee: This is within `create_fx_graph`, which is for debugging purposes, so i think it's fine.

torchinductor/scheduler.py
Chillee: Also, let's rename this to `create_fx_from_buffers`

yushangdi: renamed

torchinductor/scheduler.py
yushangdi: as @Chillee said below, this line is only called when we are in debugging mode, so I think printing out the graph is probably helpful?

PR #666, opened 2022-07-27
Moving string to compiler_fn logic to eval_frame

anijain2305:
Helps with checking if a backend has any special context manager, like `aot_nvfuser` sets up torch.jit.fuser("fuser2") ctx mananger.

@jansel @pyjhzwh @ngimel

### DIFF
--- a/torchdynamo/eval_frame.py
+++ b/torchdynamo/eval_frame.py
@@ -245,12 +245,10 @@ def optimize(backend, nopython=False):
     backend = get_compiler_fn(backend)
 
     # Find if backend has any extra context manager
-    backend_ctx_ctor = null_context
-    if hasattr(backend, "backend_ctx_ctor"):
-        backend_ctx_ctor = getattr(backend, "backend_ctx_ctor")
+    backend_ctx_ctor = getattr(backend, "backend_ctx_ctor", null_context)
 
     if nopython:
-        return optimize_assert(backend, backend_ctx_ctor, guard_export_fn=None)
+        return optimize_assert(backend, guard_export_fn=None)
     return _optimize_catch_errors(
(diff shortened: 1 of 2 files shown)


ngimel:
Why would `torch.jit.fuser("fuser2") ` be needed if `--nvfuser` that's set as part of `--aot_nvfuser` is supposed to enabled it?

anijain2305:
> Why would `torch.jit.fuser("fuser2") ` be needed if `--nvfuser` that's set as part of `--aot_nvfuser` is supposed to enabled it?

We are not setting `torch.jit.fuser("fuser2")` explicitly. TorchDynamo supports backend-specific context manager here - https://github.com/pytorch/torchdynamo/blob/main/torchdynamo/eval_frame.py#L72. For aot_nvfuser, we have set the backend context manager here - https://github.com/pytorch/torchdynamo/blob/main/torchdynamo/optimizations/training.py#L178


This was all working fine until we added a string. The string gets converted to a compiler_fn object later in the flow. So, I basically, pre-poned the conversion of string to the compiler_fn, so that we have access to the backend-specific constructor.

So, you would not need `--nvfuser` while calling benchmarking infra after this PR.

torchdynamo/eval_frame.py
```diff
@@ -228,6 +242,9 @@ def toy_example(a, b):
         with torchdynamo.optimize(my_compiler):
            ...
     """
+    backend = get_compiler_fn(backend)
```
jansel, line 245: Do we need this patch in optimize_assert too?

anijain2305: Yes, its there further below.

jansel (approved):

PR #669, opened 2022-07-28
Integrate gradual types into TorchDynamo

migeed-z:
- Propagate gradual type annotations 
- Evaluate conditionals and return constants when possible
- Evaluate on two small examples


torchdynamo/variables/builder.py
```diff
@@ -351,10 +351,13 @@ def wrap_tensor(self, value: torch.Tensor):
...
             # user code.
             with torch._C.DisableTorchFunction():
+                current_type = type(value)
+                if self.name in self.tx.new_annotations:
+                    current_type = self.tx.new_annotations[self.name]
                 tensor_variable = TensorVariable.create(
                     tx=self.tx,
                     proxy=self.tx.output.create_graph_input(
-                        re.sub(r"[^a-zA-Z0-9]+", "_", self.name), type(value)
+                        re.sub(r"[^a-zA-Z0-9]+", "_", self.name), current_type
```
jansel, line 360: I'd suggest keeping the annotation separate from the type.  Annotations can be wrong.

torchdynamo/variables/tensor.py
```diff
@@ -41,6 +46,17 @@
 from .lists import SizeVariable
 
 
+class FakeRootModule(torch.nn.Module):
```
jansel, line 49: Is this the same as the other one?  Can we just reuse it (or move it to utils)?

jansel (commented):
A few comments.  I'd also suggest putting this behind a config.

tests/test_gradual_types.py
```diff
@@ -0,0 +1,80 @@
...
+    import z3  # noqa
+    from torch.fx.experimental.migrate_gradual_types.transform_to_z3 import (  # noqa
+        evaluate_conditional_with_constraints,
+    )
+
+    HAS_Z3 = True
+except ImportError:
+    HAS_Z3 = False
+
+skipIfNoZ3 = unittest.skipIf(not HAS_Z3, "no z3")
```
jansel, line 19:
```suggestion
skip_if_missing_z3 = unittest.skipIf(not HAS_Z3, "no z3")
```

torchdynamo/symbolic_convert.py
```diff
@@ -1225,6 +1225,17 @@ def __init__(
             symbolic_globals=collections.OrderedDict(),
             f_code=f_code,
         )
+
+        self.new_annotations = {}
+
+        if "self" in f_locals:
```
jansel, line 1231: I actually liked your previous trick a bit better.  (The one using `gc.get_referrers`.)

I hadn't thought of it when I suggested this one.  This one is worse.

torchdynamo/variables/tensor.py
```diff
@@ -41,6 +44,16 @@
 from .lists import ShapeVariable
 from .lists import SizeVariable
 
+try:
+    import z3  # type: ignore[import]
```
jansel, line 48: This code sounds familiar.  Perhaps refactor into:
`torchdynamo.utils.has_z3()`

jansel (changes requested):

jansel:

Issue #603, opened 2022-07-18
Inference error tracking: "timm_efficientdet"


desertfire:
Jason reported out of memory on RTX3090.

desertfire:
New error:
```
    kernel28[grid(24000)](buf74, buf76, buf725, 24000, 6400)
NameError: name 'buf76' is not defined
```

jansel:
> Running on A100, and hitting
> 
> ```
> RuntimeError: gather(): Expected dtype int64 for index
> ```

Hitting same thing as today

PR #674, opened 2022-07-28
[inductor] Add type conversion for div when both operands are integers

desertfire:
Summary: The problem is exposed in https://github.com/pytorch/torchdynamo/issues/603,
`RuntimeError: gather(): Expected dtype int64 for index`.

There are two problems here,
1) The return types of floordiv and truediv are different when both
operands are integer type.
2) Decomposing floordiv into div+floor gives a different result on
triton from eager because of CUDA fdiv mode choice differences, see
the discussion at https://github.com/openai/triton/issues/605.

### DIFF
--- a/torchinductor/codegen/triton.py
+++ b/torchinductor/codegen/triton.py
@@ -201,11 +201,23 @@ class TritonOverrides(OpOverrides):
             return f"tl.where({tmp}>{x}, {tmp}-1, {tmp})"
 
     @staticmethod
+    def floordiv(a, b):
+        # See the comment in lowering.div_mode
+        div = f"tl.fdiv({a}.to(tl.float32), {b}.to(tl.float32), True)"
+        return f"{ops.floor(div)}"
+
+    @staticmethod
     def trunc(x):
         if has_triton_libdevice():
             return f"tl.libdevice.trunc({x})"
         else:
             return f"{x}.to(tl.int32).to(tl.float32)"
+
+    @staticmethod
+    def truncdiv(a, b):
+        # See the comment in lowering.div_mode
+        div = f"tl.fdiv({a}.to(tl.float32), {b}.to(tl.float32), True)"
+        return f"{ops.trunc(div)}"
 
     @staticmethod
     def ceil(x):
--- a/torchinductor/codegen/triton.py
+++ b/torchinductor/codegen/triton.py
@@ -202,9 +202,12 @@ class TritonOverrides(OpOverrides):
 
     @staticmethod
     def floordiv(a, b):
-        # See the comment in lowering.div_mode
-        div = f"tl.fdiv({a}.to(tl.float32), {b}.to(tl.float32), True)"
-        return f"{ops.floor(div)}"
+        # See the comment in lowering.div_mode. a and b are integer type.
+        # Similar to div_floor_kernel_cuda in pytorch core.
+        # Notice that // in triton behaves as truncdiv instead of floordiv
+        quot = f"{a} // {b}"
+        rem = f"{a} % {b}"
+        return f"tl.where(({a} < 0) != ({b} < 0), tl.where({rem} != 0, {quot} - 1, {quot}), {quot})"
 
     @staticmethod
     def trunc(x):
@@ -215,9 +218,9 @@ class TritonOverrides(OpOverrides):
 
     @staticmethod
     def truncdiv(a, b):
-        # See the comment in lowering.div_mode
-        div = f"tl.fdiv({a}.to(tl.float32), {b}.to(tl.float32), True)"
-        return f"{ops.trunc(div)}"
+        # See the comment in lowering.div_mode. a and b are integer type.
+        # Notice that // in triton behaves as truncdiv instead of floordiv
+        return f"{a} // {b}"
 
     @staticmethod
     def ceil(x):
--- a/torchinductor/lowering.py
+++ b/torchinductor/lowering.py
@@ -98,6 +98,13 @@ def is_integer_type(x):
         return isinstance(x, int)
 
 
+def is_boolean_type(x):
+    if isinstance(x, TensorBox):
+        return is_boolean_dtype(x.get_dtype())
+    else:
+        return isinstance(x, bool)
+
+
 def decode_device(device):
     if device is None:
         return torch.tensor(0.0).device  # default device
@@ -2623,13 +2630,16 @@ def truncdiv(a, b):
 @register_lowering(aten.div.Tensor_mode)
 def div_mode(a, b, rounding_mode=None):
     both_integer = is_integer_type(a) and is_integer_type(b)
+    both_boolean = is_boolean_type(a) and is_boolean_type(b)
     on_triton = is_triton(a) or is_triton(b)
 
     # floordiv and truncdiv need special handling for integer tensors on Triton,
     # see the discussion at https://github.com/openai/triton/issues/605
     if rounding_mode == "floor":
+        assert not both_boolean, "floordiv operands can not be boolean at the same time"
         return floordiv(a, b) if (both_integer and on_triton) else floor(div_mode(a, b))
     if rounding_mode == "trunc":
+        assert not both_boolean, "truncdiv operands can not be boolean at the same time"
         return truncdiv(a, b) if (both_integer and on_triton) else trunc(div_mode(a, b))
     if both_integer:
         # truediv produces a float tensor even if both operands are integer types
--- a/torchinductor/lowering.py
+++ b/torchinductor/lowering.py
@@ -18,6 +18,7 @@ from .ir import Reduction
 from .ir import SqueezeView
 from .ir import TensorBox
 from .ir import View
+from .ir import is_triton
 from .utils import has_torchvision_roi_align
 from .utils import sympy_product
 from .virtualized import V
@@ -86,6 +87,23 @@ def decode_dtype(dtype: int):
     assert dtype in DTYPE_ID_LOOKUP, f"id {dtype} missing from DTYPE_ID_LOOKUP"
     dtype = DTYPE_ID_LOOKUP[dtype]
     return dtype
+
+
+def is_integer_type(x):
+    if isinstance(x, TensorBox):
+        return x.get_dtype() in {
+            torch.uint8,
+            torch.int8,
+            torch.int16,
+            torch.int32,
+            torch.int64,
+        }
+    else:
+        return isinstance(x, int)
+
+
+def is_triton_device(x):
+    return isinstance(x, TensorBox) and is_triton(x.get_device())
 
 
 def decode_device(device):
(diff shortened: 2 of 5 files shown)


torchinductor/codegen/triton.py
```diff
@@ -200,13 +200,25 @@ def floor(x):
             tmp = ops.trunc(x)
             return f"tl.where({tmp}>{x}, {tmp}-1, {tmp})"
 
+    @staticmethod
+    def floordiv(a, b):
+        # See the comment in lowering.div_mode
+        div = f"tl.fdiv({a}.to(tl.float32), {b}.to(tl.float32), True)"
+        return f"{ops.floor(div)}"
```
jansel, line 207:
```suggestion
        return ops.floor(div)
```

torchinductor/codegen/triton.py
```diff
@@ -200,13 +200,25 @@ def floor(x):
...
         if has_triton_libdevice():
             return f"tl.libdevice.trunc({x})"
         else:
             return f"{x}.to(tl.int32).to(tl.float32)"
 
+    @staticmethod
+    def truncdiv(a, b):
+        # See the comment in lowering.div_mode
+        div = f"tl.fdiv({a}.to(tl.float32), {b}.to(tl.float32), True)"
+        return f"{ops.trunc(div)}"
```
jansel, line 220:
```suggestion
        return ops.trunc(div)
```

torchinductor/lowering.py
```diff
@@ -88,6 +89,23 @@ def decode_dtype(dtype: int):
     return dtype
 
 
+def is_integer_type(x):
+    if isinstance(x, TensorBox):
+        return x.get_dtype() in {
```
jansel, line 94: perhaps 
```
return not x.get_dtype().is_floating_point
```

torchinductor/lowering.py
```diff
@@ -88,6 +89,23 @@ def decode_dtype(dtype: int):
...
+            torch.int16,
+            torch.int32,
+            torch.int64,
+        }
+    else:
+        return isinstance(x, int)
+
+
+def is_triton_device(x):
+    return isinstance(x, TensorBox) and is_triton(x.get_device())
```
jansel, line 106: perhaps move this logic into ir.is_triton?

torchinductor/codegen/triton.py
ngimel: Converting integer numbers to float32 is wrong (you are losing precision). Please look at how integer floordiv is done in the eager mode without converting to fp32 type:
```
            if (c10::signs_differ(a, b)) {
              // Subtracts one from the results of truncation division if the
              // divisor and dividend have different sign(bit)s and the
              // remainder of the division is nonzero
              const auto quot = a / b;
              const auto rem = a % b;
              return rem ? quot - 1 : quot;
            }
```
that's roughly what you need to do here.

torchinductor/codegen/triton.py
ngimel: Converting integer numbers to float32 is wrong (you are losing precision), and, since you only ever send integer arguments here, you can just do return f"{{a} // {b}}, triton supports that for integral args.
Please add tests that would catch that precision loss from fp32, you'd need numbers more than 2**24 for this.
Note (this is confusing), triton's `//` is actually `trunc` mode, equivalent to c-style integer division, not to python `//`

torchinductor/lowering.py
ngimel: this won't be future proof, as we'd add `complex` types at some point. But perhaps you could use `torch._prims_common.is_integer_dtype`.

torchinductor/lowering.py
```diff
@@ -2522,6 +2540,36 @@ def copy_(dst, src, non_blocking=False):
...
+    # floordiv and truncdiv need special handling for integer tensors on Triton,
+    # see the discussion at https://github.com/openai/triton/issues/605
+    if rounding_mode == "floor":
+        return floordiv(a, b) if (both_integer and on_triton) else floor(div_mode(a, b))
+    if rounding_mode == "trunc":
+        return truncdiv(a, b) if (both_integer and on_triton) else trunc(div_mode(a, b))
+    if both_integer:
+        # truediv produces a float tensor even if both operands are integer types
+        return div(
+            to_dtype(a, torch.float32) if isinstance(a, TensorBox) else a,
```
ngimel, line 2567: float32 should not be hardcoded here, it's `torch.get_default_dtype()`, users can set it to fp64

torchinductor/lowering.py
```diff
@@ -2522,6 +2540,36 @@ def copy_(dst, src, non_blocking=False):
...
+
+
+@make_pointwise
+def truncdiv(a, b):
+    return ops.truncdiv(a, b)
+
+
+@register_lowering(aten.div.Tensor_mode)
+def div_mode(a, b, rounding_mode=None):
+    both_integer = is_integer_type(a) and is_integer_type(b)
```
ngimel, line 2555: this doesn't correctly handle situations when one of the inputs is integer and one is boolean, and trunc or floor mode is set (pytorch produces integer result in this case)

ngimel (changes requested):

torchinductor/codegen/triton.py
desertfire: No wonder my earlier try of using `//` for floordiv didn't work. Isn't this a triton bug then, given triton still lives in the Python world?

ngimel: Yeah that's triton bug, we should file triton issue for that.

torchinductor/lowering.py
desertfire: Should be solved by including `is_boolean_dtype` checking in `is_integer_type`.

ngimel: If both inputs are boolean, pytorch errors out

torchinductor/codegen/triton.py
ngimel: https://github.com/openai/triton/issues/619

torchinductor/lowering.py
desertfire: Is there a spec somewhere for this kind of typing rules? Or is it basically trying and getting the results? For this specific both boolean problem, I wonder if we can relax the implementation in inductor? After all, the front end should error our before reaching here, and there are other ops currently implemented without very rigid checking.

desertfire: If I write the following test,
```
    def test_div6(self):
        def fn(a, b):
            return (
                aten.div(a, b, rounding_mode=None),
                aten.div(a, b, rounding_mode="floor"),
                aten.div(a, b, rounding_mode="trunc"),
            )

        # treat boolean as integer
        self.common(
            fn, (torch.ones([8, 8], dtype=torch.bool), torch.ones([8, 8], dtype=torch.bool))
        )
```
it does error out before even reaching my changed code with,
```
E       RuntimeError: "div_floor_cpu" not implemented for 'Bool'

../pytorch-release/torch/_ops.py:175: RuntimeError
```

@ngimel , what do you think?

ngimel: Yeah it errors out with eager error. However, imagine in future we would be tracing with fake tensors instead of real ones, then this error wouldn't happen and it would be on inductor to produce correct behavior, so it still makes sense to have necessary checks in inductor, even if we can't test them.

torchinductor/lowering.py
```diff
@@ -2534,12 +2574,14 @@ def copy_(dst, src, non_blocking=False):
 add = register_pointwise(aten.add)
 div = register_pointwise(aten.div)
```
ngimel, line 2575: this PR looks good now, but I think aten.div would still not promote both integer inputs, like eager does? Can we add it in this PR, or in a follow up?

torchinductor/lowering.py
```diff
@@ -2526,6 +2543,52 @@ def copy_(dst, src, non_blocking=False):
...
+def div(a, b):
+    def fn(*args):
+        return ops.div(*args)
+
+    both_integer = is_integer_type(a) and is_integer_type(b)
+    # truediv produces a float tensor even if both operands are integer types
+    dtype = (
+        torch.get_default_dtype()
+        if both_integer
+        else torch.promote_types(a.get_dtype(), b.get_dtype())
```
ngimel, line 2583: Better use `torch.result_type(a, b)` here, that would solve test errors and also get correct result when one of the inputs is 0-dim tensor (promote_types wouldn't)

torchinductor/lowering.py
```diff
@@ -2526,6 +2543,52 @@ def copy_(dst, src, non_blocking=False):
...
+        return truncdiv(a, b) if (both_integer and on_triton) else trunc(div(a, b))
+    return div(a, b)
+
+
+@register_lowering([aten.div, prims.div])
+def div(a, b):
+    def fn(*args):
+        return ops.div(*args)
+
+    dtype = torch.result_type(
```
ngimel, line 2578: no, this is still wrong. Are `a` and `b` not necessarily tensors or numbers so you cannot call `torch.result_type` directly? Type promotion rules differ for 0d and 1d or more tensors, and here you are creating a 0d tensor instead of whatever tensor actually was, you need faithful tensor models to send to `result_type`

Chillee:
Is there any reason we can't just use primtorch `div` ref and then lower the resulting ops instead? https://github.com/pytorch/pytorch/blob/master/torch/_refs/__init__.py#L930

torchinductor/lowering.py
desertfire: Oh, here `a` and `b` are either scalars or IRs (TensorBox, etc.). They are not really torch.tensor at this point. An alternative solution is to do something similar to https://github.com/pytorch/torchdynamo/blob/cf0d110d58e46f75904c07eea2a2f539490aeed7/torchinductor/lowering.py#L119, what do you think?

ngimel: The existing dtype computation is also wrong, and we really should be fixing it everywhere. You can't just reduce the list of dtypes, because dtypes priority varies on whether the argument is a numer, a 0d tensor or a 1d-or-more tensor. Is it possible to get dimensionality of TensorBox?

ngimel:
@chillee most of this PR is lowering resulting ops, sure, we can use primtorch's div to route to them.

Chillee:
@ngimel It seems like there's a lot of discussion about type promotion that I think the refs already handle, no?

ngimel:
`div` reference doesn't handle type promotion, it redispatches to `true_div`, `floor_div` and `trunc_div`. The latter 2 would definitely require lowering and not using refs, I don't know, maybe we could use `true_div`.

Chillee:
> The latter 2 would definitely require lowering and not using refs

If they would definitely require lowering, why aren't they prims 🤔

ngimel:
lol prims div was fixed to return correct type just 4 days ago #82949, trunc_div/floor_div reference were fixed to do the right thing just now #83288 (but again, I do think lowerings and not references for those are better).

desertfire:
> > The latter 2 would definitely require lowering and not using refs
> 
> If they would definitely require lowering, why aren't they prims 🤔

Maybe this PR title is a little misleading. One of the problems I am trying to solve here is getting the right result for `floor_div` and `trunc_div` of integers on Trition, and we need some special handlings in lowering and codgen because of that. I will try the idea of using ref for `true_div`.

ngimel:
`true_divide` is just a wrapper around prim, and generally we want to avoid those (because they hardcode broadcasting), so I guess we need to solve type promotion problems here (similarly for other binary and unary ops in lowering.py)

desertfire:
> `true_divide` is just a wrapper around prim, and generally we want to avoid those (because they hardcode broadcasting), so I guess we need to solve type promotion problems here (similarly for other binary and unary ops in lowering.py)

'avoid those' meaning avoid using ref ops? It will be really nice to leverage type promotion from there, so that we can get consistent type promotion rules and other benefits.

ngimel:
We should be able to use refs, but for some elementary refs (that go directly to prims) it's hard to use type promotion and not hardcoding dimensions that comes from broadcast_in_dim. Getting consistent type promotion is nice though, so we should find a way to do that.

torchinductor/lowering.py
```diff
@@ -96,6 +114,20 @@ def decode_device(device):
...
+    def construct_input(inp):
+        if isinstance(inp, Number):
+            return inp
+        else:
+            assert hasattr(inp, "get_dtype")
+            dim = len(inp.get_size())
+            # construct a tmp tensor to feed into torch.result_type
+            return torch.zeros([1] * dim, dtype=inp.get_dtype())
+
+    return functools.reduce(torch.result_type, [construct_input(arg) for arg in args])
```
ngimel, line 128: Unfortunately, this won't be correct for more than 2 args (because type promotion is not associative), but we expect to call it for 2 args only in lowerings, so that should be ok.

ngimel (approved):
Awesome, thanks!

torchinductor/lowering.py
```diff
@@ -2526,6 +2558,48 @@ def copy_(dst, src, non_blocking=False):
...
+def div_mode(a, b, rounding_mode=None):
+    both_integer = is_integer_type(a) and is_integer_type(b)
+    both_boolean = is_boolean_type(a) and is_boolean_type(b)
+    on_triton = is_triton(a) or is_triton(b)
+
+    # floordiv and truncdiv need special handling for integer tensors on Triton,
+    # see the discussion at https://github.com/openai/triton/issues/605
+    if rounding_mode == "floor":
+        assert not both_boolean, "floordiv operands can not be boolean at the same time"
+        return floordiv(a, b) if (both_integer and on_triton) else floor(div(a, b))
```
ngimel, line 2581: wait, you still need to promote types to the higher one even if both are integer?

PR #676, opened 2022-07-28
More HF models

anijain2305:
The script is modified from - https://gist.github.com/jamesr66a/1e65d3b396d4c1a50d1384d7f4eeeb1c (thanks @jamesr66a )

Batch sizes = Half of the largest batch size that could fit on GPU with eager.

dtype = float32

image

Could not run inductor for some reason. Will run later.

@ngimel @Chillee


benchmarks/huggingface.py
```diff
@@ -28,45 +27,240 @@ def pip_install(package):
...
+HF_FX_SUPPORTED_MODELS = dict()
+
+REFRESH_MODEL_NAMES = False
+
+USE_HALF_BATCH_SIZE = True
+
+filename = "huggingface_models_list.txt"
+if os.path.exists("benchmarks"):
+    filename = "benchmarks/" + filename
+with open(filename, "r") as fh:
```
ngimel, line 73: you should check that file exists, and that you read more than 0 lines from it.

benchmarks/huggingface.py
```diff
@@ -266,18 +335,45 @@ def load_model(
...
             # Handle auto classes
             model = model_cls.from_config(config).to(device, dtype=dtype)
         else:
             model = model_cls(config).to(device, dtype=dtype)
 
+        batch_size = batch_size or batch_size_default
+
+        if model_name in USE_SMALL_BATCH_SIZE:
+            batch_size = USE_SMALL_BATCH_SIZE[model_name]
+        elif USE_HALF_BATCH_SIZE:
```
ngimel, line 370: you should warn here, rather than silently change user specified batch size

benchmarks/huggingface.py
```diff
@@ -266,18 +335,45 @@ def load_model(
...
         if "auto" in model_cls.__module__:
             # Handle auto classes
             model = model_cls.from_config(config).to(device, dtype=dtype)
         else:
             model = model_cls(config).to(device, dtype=dtype)
 
+        batch_size = batch_size or batch_size_default
+
+        if model_name in USE_SMALL_BATCH_SIZE:
+            batch_size = USE_SMALL_BATCH_SIZE[model_name]
```
ngimel, line 369: and similarly, a warning here

benchmarks/huggingface.py
```diff
@@ -354,7 +444,87 @@ def forward_and_backward_pass(self, mod, inputs, collect_outputs=True):
...
+            )
+        except subprocess.SubprocessError:
+            print("ERROR")
+
+
 if __name__ == "__main__":
-    logging.basicConfig(level=logging.WARNING)
-    warnings.filterwarnings("ignore")
-    main(HuggingfaceRunner())
+    if REFRESH_MODEL_NAMES and "--find-batch-sizes" not in sys.argv:
```
ngimel, line 525: How does this ever get triggered? It seems like REFRESH_MODEL_NAMES is hardcoded above.

benchmarks/huggingface.py
```diff
@@ -354,7 +444,87 @@ def forward_and_backward_pass(self, mod, inputs, collect_outputs=True):
         return None
 
 
+def refresh_model_names():
```
ngimel, line 447: can you leave a brief comment what this function does?

benchmarks/huggingface.py
```diff
@@ -354,7 +444,87 @@ def forward_and_backward_pass(self, mod, inputs, collect_outputs=True):
...
+    for members in family.values():
+        chosen_models.update(set(members))
+
+    chosen_models.update(set(EXTRA_MODELS.keys()))
+    filename = "huggingface_models_list.txt"
+    if os.path.exists("benchmarks"):
+        filename = "benchmarks/" + filename
+    for model_name in sorted(chosen_models):
+        try:
+            subprocess.check_call(
```
ngimel, line 513: it seems like this function is doing more than refreshing model names?

benchmarks/huggingface.py
```diff
@@ -354,7 +444,87 @@ def forward_and_backward_pass(self, mod, inputs, collect_outputs=True):
...
+                [sys.executable]
+                + sys.argv
+                + ["--find-batch-sizes"]
+                + [f"--only={model_name}"]
+                + [f"--output={filename}"]
+            )
+        except subprocess.SubprocessError:
+            print("ERROR")
```
ngimel, line 521: more detailed logging, e.g. which model failed?

benchmarks/huggingface.py
```diff
@@ -28,45 +27,240 @@ def pip_install(package):
...
+    from transformers import SwinModel
+    from transformers import ViTForImageClassification
+    from transformers import ViTForMaskedImageModeling
+    from transformers import ViTModel
     from transformers import XLNetConfig
     from transformers import XLNetLMHeadModel
 
 
-# We are primarily interested in tf32 datatype
-torch.backends.cuda.matmul.allow_tf32 = True
```
ngimel, line 55: where did this go? We still want to benchmark with tf32.

anijain2305: Ah thanks! I thought I had it in common.py. Added it in common.py now.

ngimel (approved):

PR #678, opened 2022-07-29
Append stack trace to fx node

SherlockNoMad:
For debugging purpose, user commonly wish to trace back from an fx.Node to the user code. 

image
Sample output: 

```
placeholder x
  File "/fsx/users/bahuang/repos/torchdynamo/test.py", line 21, in forward
    def forward(self, x):


call_function <built-in method sqrt of type object at 0x7f4c3f8d8360>
  File "/fsx/users/bahuang/repos/torchdynamo/test.py", line 22, in forward
    x = self.n(x)
  File "/fsx/users/bahuang/repos/torchdynamo/test.py", line 12, in forward
    a = torch.sqrt(q)


call_method relu
  File "/fsx/users/bahuang/repos/torchdynamo/test.py", line 22, in forward
    x = self.n(x)
  File "/fsx/users/bahuang/repos/torchdynamo/test.py", line 13, in forward
    return f(a)
  File "/fsx/users/bahuang/repos/torchdynamo/test.py", line 8, in f
    return x.relu()


call_module self_sigmoid
  File "/fsx/users/bahuang/repos/torchdynamo/test.py", line 23, in forward
    x = self.sigmoid(x)


output output
None
```

### DIFF
--- a/torchdynamo/output_graph.py
+++ b/torchdynamo/output_graph.py
@@ -433,3 +433,12 @@ class OutputGraph(fx.Tracer):
         for node in self.graph.nodes:
             if "example_value" in node.meta:
                 del node.meta["example_value"]
+
+    def create_proxy(self, kind, target, args, kwargs, name=None, type_expr=None, proxy_factory_fn=None):
+        rv = super().create_proxy(kind, target, args, kwargs, name, type_expr, proxy_factory_fn)
+
+        # append stack trace to fx node
+        frame_summary = self.root_tx.frame_summary()
+        rv.node.stack_trace = f"{frame_summary}\n  {frame_summary.line}"
+
+        return rv
--- a/torchdynamo/output_graph.py
+++ b/torchdynamo/output_graph.py
@@ -451,7 +451,12 @@ class OutputGraph(fx.Tracer):
 
         # append stack trace to fx node
         tx = current_tx if current_tx else self.root_tx
-        frame_summary = tx.frame_summary()
-        rv.node.stack_trace = f"{frame_summary}\n  {frame_summary.line}"
+        frame_summaries: List[traceback.FrameSummary] = []
+        while tx:
+            frame_summaries.append(tx.frame_summary())
+            tx = getattr(tx, "parent", None)
+
+        msgs = reversed(traceback.StackSummary.from_list(frame_summaries).format())
+        rv.node.stack_trace = "".join(msgs)
 
         return rv
--- a/torchdynamo/output_graph.py
+++ b/torchdynamo/output_graph.py
@@ -434,11 +434,24 @@ class OutputGraph(fx.Tracer):
             if "example_value" in node.meta:
                 del node.meta["example_value"]
 
-    def create_proxy(self, kind, target, args, kwargs, name=None, type_expr=None, proxy_factory_fn=None):
-        rv = super().create_proxy(kind, target, args, kwargs, name, type_expr, proxy_factory_fn)
+    def create_proxy(
+        self,
+        kind,
+        target,
+        args,
+        kwargs,
+        name=None,
+        type_expr=None,
+        proxy_factory_fn=None,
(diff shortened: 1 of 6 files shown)


torchdynamo/output_graph.py
```diff
@@ -433,3 +433,12 @@ def cleanup(self):
         for node in self.graph.nodes:
             if "example_value" in node.meta:
                 del node.meta["example_value"]
+
+    def create_proxy(self, kind, target, args, kwargs, name=None, type_expr=None, proxy_factory_fn=None):
+        rv = super().create_proxy(kind, target, args, kwargs, name, type_expr, proxy_factory_fn)
+
+        # append stack trace to fx node
```
SherlockNoMad, line 440: Do we need some debug flag to enable this？

torchdynamo/output_graph.py
```diff
@@ -433,3 +433,12 @@ def cleanup(self):
...
+        frame_summary = self.root_tx.frame_summary()
```
SherlockNoMad, line 441: hmm... using self.root_tx doesn't handle nested functions. 

I still need to handle nested InliningInstructionTranslator ...

voznesenskym:
J/w Why do we want to append stack trace to fx nodes?

SherlockNoMad:
> J/w Why do we want to append stack trace to fx nodes?

For debugging purpose... it's a common request to correlate back to the user code from an FX node.

torchdynamo/output_graph.py
SherlockNoMad: fixed.

torchdynamo/output_graph.py
jansel: I'm assuming this track traces will be small, so I don't see a strong need.

torchdynamo/output_graph.py
```diff
@@ -433,3 +433,25 @@ def cleanup(self):
...
+        current_tx=None,
+    ):
+        rv = super().create_proxy(
+            kind, target, args, kwargs, name, type_expr, proxy_factory_fn
+        )
+
+        # append stack trace to fx node
+        tx = current_tx if current_tx else self.root_tx
+        frame_summary = tx.frame_summary()
+        rv.node.stack_trace = f"{frame_summary}\n  {frame_summary.line}"
```
jansel, line 455: If you want to print a `traceback.FrameSummary` you could use code like:
https://github.com/pytorch/torchdynamo/blob/9ecbdaed5c1afe2f8443e0f13e081191cc29b0d3/torchdynamo/exc.py#L55
which I'd expect to look a bit nicer.

The other option would be to leave it as a `traceback.FrameSummary` object, which will allow users to query file/line more easily.

torchdynamo/output_graph.py
```diff
@@ -433,3 +433,25 @@ def cleanup(self):
...
+    def create_proxy(
+        self,
+        kind,
+        target,
+        args,
+        kwargs,
+        name=None,
+        type_expr=None,
+        proxy_factory_fn=None,
+        current_tx=None,
```
anijain2305, line 446: Soft suggestion - feel free to ignore - replace `current_tx` to just `tx`?

anijain2305 (commented):
Thanks! Glad that the utility was already there. It turned out to be exactly what we needed :)

torchdynamo/output_graph.py
SherlockNoMad: ok.

torchdynamo/output_graph.py
SherlockNoMad: I will keep it as current_tx, as it's a stack of txs....

torchdynamo/output_graph.py
SherlockNoMad: done.

anijain2305 (approved):

PR #680, opened 2022-07-29
[inductor] Per-buffer merging of contiguous dims

jansel:
This improves the logic for merging contiguous dims by reusing the kernel-level simplification on a per-buffer level.

### DIFF
--- a/torchinductor/codegen/triton.py
+++ b/torchinductor/codegen/triton.py
@@ -15,6 +15,7 @@ from .. import codecache
 from .. import config
 from .. import ir
 from ..utils import has_triton_libdevice
+from ..utils import sympy_dot
 from ..utils import sympy_product
 from ..virtualized import V
 from ..virtualized import ops
@@ -235,6 +236,12 @@ class RangeTree:
         self.prefix = prefix
         self.depth = depth
         self.length = length
+        self.recursive_names = {name}
+
+    def compute_recursive_names(self):
+        for c in self.children.values():
+            c.compute_recursive_names()
+            self.recursive_names.update(c.recursive_names)
 
     def cache_clear(self):
         for child in self.children.values():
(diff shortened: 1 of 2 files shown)


torchinductor/codegen/triton.py
```diff
@@ -274,17 +281,44 @@ def __init__(
...
         itervars = []
         for sv in reversed(lengths):
             node = node.child_node(sv)
             itervars.append(node.symbol())
+        self.compute_recursive_names()
         return list(reversed(itervars))
 
+    def vars_and_sizes(self, index: sympy.Expr):
+        """Figure out vars from this tree used in index"""
+        symbol_names = set(map(str, index.free_symbols)) & self.recursive_names
```
ngimel, line 295: can you sort this set so that kernels still have deterministic source?

torchinductor/codegen/triton.py
```diff
@@ -571,23 +604,11 @@ def var_ranges(self):
...
-            for sym in sorted_expr_symbols
-            if sym in self.range_tree_nodes
-        ]
-        if nodes:
-            nodes.sort(key=lambda x: x.depth)
-            expr = nodes[-1].simplify(expr)
-            sorted_expr_symbols = sorted(expr.free_symbols, key=lambda s: s.name)
-
-        for sym in sorted_expr_symbols:
+        for sym in expr.free_symbols:
```
ngimel, line 609: and here, expr.free_symbols is a set, so it will be iterated in arbitrary order

torchinductor/codegen/triton.py
jansel: This one is only used for `set.union` calls, so I think we want to keep it a set.  The order comes from the tree structure, not from this variable.  I'll change the other one though.

torchinductor/codegen/triton.py
jansel: fixed

ngimel (approved):

PR #683, opened 2022-07-30
Added ability for torchbench to read from a file

Chillee:
`time ./benchmarks/torchbench.py -dcuda --float16 --inductor --training --batch_size_file /scratch/chilli/fresh/torchdynamo/benchmarks/torchbench_models_list.txt`


benchmarks/torchbench.py
```diff
@@ -272,6 +273,17 @@ def load_model(
         benchmark_cls = getattr(module, "Model", None)
         if not hasattr(benchmark_cls, "name"):
             benchmark_cls.name = model_name
+        batch_size = 1
+        lines = open('/fsx/users/chilli/model_sizes.txt', 'r').readlines()
```
anijain2305, line 277: Can we put the file in torchdynamo repo itself?

ngimel:
What would be the way to run with default batch sizes?

Chillee:
@ngimel Hmm... no way here. I guess I'll add an additional flag.

Chillee:
Added `--batch_size_file` flag, which allows torchbench to read the batch size from a file.

benchmarks/common.py
```diff
@@ -1449,14 +1452,35 @@ def main(runner, original_dir=None):
     experiment = functools.partial(experiment, args, model_iter_fn)
 
     if args.only:
```
ngimel, line 1454: how does this work when args.only is false?

Chillee: Well, if you're running with `isolate` (which I think we always should), it just ends up calling `--only`.

ngimel: Can we check then that if `--batch-size-file` is set, either `--isolate` or `--only` is set too?

Chillee: Updated.

ngimel (approved):

PR #684, opened 2022-07-30
[WIP] [RFC] - Gathering feedback on approach - Limited control flow operator

voznesenskym:
This RFC exposes a limited control flow operator, with limitations when compared to supporting all of the stuff we currently handle (mostly via unimplemented) through generic_jump. 

Explicitly chosen limitations:
- Explicitly User authored
- Tensor only inputs
- Tensor only outputs
- Input and output schema in the true/false control flow branches must match 

I am looking for feedback on UX here, more so than implementation at first. This impl has some hacks in it, but those are not worth fixing if folks don't like the direction and the UX. 

Alternatively, we could see if we can apply those limitations and use some of this logic in supporting `if` control flow at the operator level via `dis` (`POP_JUMP_IF_TRUE`, `POP_JUMP_IF_FALSE`, etc). The benefits of that are around UX, as we get it for free (no user side changes needed), but worse in terms of complexity of support and edge cases, as well as performance (We probably do not want to just blindly grab and embed every control flow, making it user driven makes it explicit)


torchdynamo/variables/functions.py
```diff
@@ -325,3 +325,101 @@ def reconstruct(self, codegen):
...
+        from torchdynamo.source import LocalSource
+        from torchdynamo.source import NNModuleSource
+
+        true_source = LocalSource(out_true_graph)
+        true_source.local_name = "true_graph"
+        false_source = LocalSource(out_false_graph)
+        false_source.local_name = "false_graph"
+
+        tx.output.add_submodule(
+            out_false_graph, "false_graph", source=NNModuleSource(false_source)
```
jansel, line 394: This name could collide if we have multiple

tests/test_control_flow.py
```diff
@@ -0,0 +1,134 @@
...
+        @torchdynamo.optimize("eager", nopython=True)
+        def foo():
+            nonlocal from_true
+            nonlocal from_false
+
+            def compute(pred, z):
+                def when_true(x):
+                    return x + x
+
+                def when_false(x):
```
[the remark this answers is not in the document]
voznesenskym: Yeah - I was considering this. I kind of like the idea of forcing when_true and when_false to have identical contracts w/r/t to input and return. No strong opinions, and it would not be overly onerous to add at all - do you have a preference?

torchdynamo/variables/functions.py
voznesenskym: Yeah, it totally will, and the naming component is entirely self contained to this file. We can generate a random string.

tests/test_control_flow.py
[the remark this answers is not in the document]
jansel: We should understand why exir was designed that way.  In some cases (e.g. `if len(x) % 2 == 0`) a single input can't go down both sides of the control flow.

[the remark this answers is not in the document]
voznesenskym: That's a great point. I can extend it to cover both T and F flows.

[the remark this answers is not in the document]
voznesenskym: Added

torchdynamo/variables/functions.py
voznesenskym: Fixed

anijain2305 (commented):
Thanks for the PR. I learnt a couple of new things.

I like that this is user controlled. This enables model developers to opt in and change their model to use `torchdynamo.control_flow.cond` operation.

* I am thinking do we need to support different args for True or False, or is there a way to do `union` of both, and ignore unused ones in True or False branch. Just wondering if it simplifies anything.
* I can be way off-base with this question - Do we care about backward pass here? And can gradients flow correctly here?

Regarding making it more generic, I think we can find if the constraints are satisfied during symbolic_conversion and then introduce the new `cond` operation, but I would really hide it behind a flag. And one of the goals of Dynamo is to make backend job easier, and most of the backends will be sad to see a control flow in the graph.

voznesenskym:
> Thanks for the PR. I learnt a couple of new things.
> 
> I like that this is user controlled. This enables model developers to opt in and change their model to use `torchdynamo.control_flow.cond` operation.
> 
> * I am thinking do we need to support different args for True or False, or is there a way to do `union` of both, and ignore unused ones in True or False branch. Just wondering if it simplifies anything.

we can do a map, or tuple, but making it explicit feels better - if we are going to have them be separate, let them be truly separate I say. 

> * I can be way off-base with this question - Do we care about backward pass here? And can gradients flow correctly here?

Good question - we an add some inductor tests on graphs with cond to make sure, if we care about lowering the control flow to backend.

> Regarding making it more generic, I think we can find if the constraints are satisfied during symbolic_conversion and then introduce the new `cond` operation, but I would really hide it behind a flag. And one of the goals of Dynamo is to make backend job easier, and most of the backends will be sad to see a control flow in the graph.

Yes, I feel like this adds complexity and potential points of friction to backend. This was a motivation for making it user driven - just supporting control flow already found in graphs feels like it could cause the backends to do the wrong thing w/r/t optimizations, removals, etc.

Chillee:
IMO, I think control flow operators need to be implemented at the dispatcher level. Otherwise, we're gonna run into a bunch of limitations/hacks when we have various transforms that we want to capture over control flow operators.

voznesenskym:
> IMO, I think control flow operators need to be implemented at the dispatcher level. Otherwise, we're gonna run into a bunch of limitations/hacks when we have various transforms that we want to capture over control flow operators.

Great question. Discussed offline a bit, but atm, my hypothesis is that if we land a good UX / signature with the right limitations in place, we can eventually support autodiff/autograd on this kind of op, as the signature suffices for it. Even though its only for the export usecase atm, nothing is stopping us from eventually handling it correctly in full in tracing. My $0.02 is that as long as the contract is right, and the UX for calling `cond` stays the same, we can slowly open up more features for it. 

Note opinions above are my own, not necessarily confirmed or refuted by @Chillee :P

jansel (changes requested):
As I see it there are two main open issues:

1) Reconcile differences with EXIR cond (example inputs needed or not?).  We should look to real use cases to figure this out.

2) I think a new op belongs in PyTorch core not in dynamo.  I don't want to see a world where we accumulate a bunch of dynamo-only ops that without dynamo.  We should register it as a dispatcher op so that we can trace it with `__torch_dispatch__`/`__torch_function__`.

voznesenskym:
> As I see it there are two main open issues:
> 
> 1. Reconcile differences with EXIR cond (example inputs needed or not?).  We should look to real use cases to figure this out.

Sounds good. 

> 2. I think a new op belongs in PyTorch core not in dynamo.  I don't want to see a world where we accumulate a bunch of dynamo-only ops that don't work without dynamo. 

So this op works without dynamo, there's a test for that - but one would still need to have dynamo as a dep to import it. However, it works without being a dynamo context.

> We should register it as a dispatcher op so that we can trace it with `__torch_dispatch__`/`__torch_function__`.

Sounds good. I'll check with @Chillee on this one too.

tests/test_control_flow.py
```diff
@@ -0,0 +1,208 @@
...
+                    return x * x
```
zou3519, line 25: Does it still work if `when_true` or `when_false` capture values that are not directly passed into the function?

voznesenskym: No - we reject closures in dynamo as unsupported.

jansel: In general dynamo does support closures, but not for export.

voznesenskym: Yes. Anything unsupported becomes a graph break, falling back to eager. In export, we do whole capture or nothing. Currently in this PR the true and false fn are captured with export.

torchdynamo/variables/functions.py
```diff
@@ -325,3 +325,121 @@ def reconstruct(self, codegen):
...
+
+        tx.output.add_submodule(
+            out_true_graph, true_name, source=NNModuleSource(true_source)
+        )
+        tx.output.add_submodule(
+            out_false_graph, false_name, source=NNModuleSource(false_source)
+        )
+
+        true_graph_node = tx.output.create_proxy(
+            "get_attr",
```
tugsbayasgalan, line 410: Why is this get_attr instead of call_module? Aren't true and false branch just submodules on the output graph

voznesenskym:
Done elsewhere

PR #686, opened 2022-07-31
Small improvements to benchmarking

ngimel:
don't recompile if optimize_ctx is same as accuracy_ctx, option to dump kineto profile.


anijain2305 (approved):
Thanks. The compile time latency jumped for me as well.

ngimel:
You mean, jumped up or down? For me it came down.

PR #688, opened 2022-08-01
[inductor] Tile after fusion

jansel:


torchinductor/triton_ops/autotune.py
```diff
@@ -147,13 +176,28 @@ def pointwise_heuristics(size_hints):
...
             ],
             key=["xnumel", "ynumel"],
         )
     if len(size_hints) == 3:
-        return apply_triton_config(triton_config(size_hints, 16, 16, 16))
+        if not config.triton.autotune:
+            return apply_triton_config(triton_config(size_hints, 16, 16, 16))
+        return autotune(
+            [
+                triton_config(size_hints, 16, 16, 16),
```
Chillee, line 191: these look pretty suspicious lol.

jansel: What would you suggest?

ngimel: Getting size_hints == 3 should be pretty unusual, do we know under which conditions we are getting it?

jansel: `config.max_tiles=2` right now, so never.

I was experimenting a bit with `max_tiles=3` (3D) tiling, for the case where you read from 3 different tensors each with different stride=1 dimensions

torchinductor/triton_ops/autotune.py
```diff
@@ -147,13 +176,28 @@ def pointwise_heuristics(size_hints):
         return autotune(
             [
                 triton_config(size_hints, 64, 64),
```
ngimel, line 178: replace with 32 x 32?

ngimel (approved):

torchinductor/triton_ops/autotune.py
jansel: will fix in next pr, merging to unblock @pyjhzwh

PR #689, opened 2022-08-01
Misc Hugginface benchmarking script fixes

anijain2305:
Issues found during automatic benchmark testing infra check (example - https://github.com/pytorch/torchdynamo/issues/681#issuecomment-1200557280)

### DIFF
--- a/benchmarks/huggingface.py
+++ b/benchmarks/huggingface.py
@@ -20,6 +20,9 @@ def pip_install(package):
     subprocess.check_call([sys.executable, "-m", "pip", "install", package])
 
 
+# Disable the flake warnings for the imports. Flake8 does not provide a way to
+# disable just warning for the entire file. Disabling flake8 entirely.
+# flake8: noqa
 imports = [
     "AlbertForPreTraining",
     "AutoConfig",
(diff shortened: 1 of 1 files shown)


benchmarks/huggingface.py
```diff
@@ -21,61 +20,77 @@ def pip_install(package):
...
+# These models contain the models present in huggingface_models_list. It is a
+# combination of models supported by HF Fx parser and some manually supplied
+# models. For these models, we already know the largest batch size that can fit
+# on A100 GPUs - 40 GB.
+BATCH_SIZE_KNOWN_MODELS = dict()
+
+
 # Get the list of models and their batch sizes
-filename = "huggingface_models_list.txt"
+MODELS_FILENAME = "huggingface_models_list.txt"
```
Chillee, line 82: Are we going to check this file in?

ngimel: It should be checked in already?

Chillee: ah didn't realize that.

benchmarks/huggingface.py
```diff
@@ -21,61 +20,77 @@ def pip_install(package):
...
 if os.path.exists("benchmarks"):
-    filename = "benchmarks/" + filename
-assert os.path.exists(filename)
-with open(filename, "r") as fh:
+    MODELS_FILENAME = "benchmarks/" + MODELS_FILENAME
```
jansel, line 84: `os.path.join`

jansel (approved):

PR #691, opened 2022-08-01
Fix the copy of requires_grad

wschin:
When generating example inputs in `torchdynamo/torchdynamo/optimizations/training.py` via calling `self.example_inputs()`, the `requires_grad` attribute is not carried over, so we fix the copy function. This error is propagated all the way to `compile_and_call_fx_graph()` and then custom compiler and the custom compiler won't set `requires_grad` correctly. If loss function is compiled, the training script fails because the generated loss has `requires_grad=False`.


torchdynamo/utils.py
```diff
@@ -245,28 +258,31 @@ def clone_input(x):
             (x.data_ptr() - buffer.data_ptr()) % 32
         ) // x.element_size()
         result = torch.as_strided(buffer, x.size(), x.stride(), cache_line_offset)
+        set_tensor_attributes(result, x, is_train)
```
anijain2305, line 261: How about just doing

`result.requires_grad_(x.requires_grad)` on this line  and then remove the need of additional `is_train` argument.

torchdynamo/utils.py
```diff
@@ -245,28 +258,31 @@ def clone_input(x):
...
         try:
             result.copy_(x.clone())
         except RuntimeError:
             # RuntimeError: unsupported operation: more than one element of the written-to
             # tensor refers to a single memory location. Please clone() the tensor before
             # performing the operation.
-            return torch.clone(x)
+            result = torch.clone(x)
+            set_tensor_attributes(result, x, is_train)
```
anijain2305, line 269: `torch.clone` copies the requires_grad field as well. So, no need to explicitly set this.

anijain2305 (changes requested):

torchdynamo/utils.py
wschin: OK

torchdynamo/utils.py
wschin: Ok.

anijain2305 (approved):

PR #693, opened 2022-08-02
Prefuse (pointwise) nodes before scheduling

Chillee:
Very much WIP, although we gain a small amount on one of the synthetic benchmarks I have :)

Before (this is bad since buf2 and buf7 will be horizontally fused, thus preventing the re-use of the load between buf7 and buf10):

image

After: (note has buf10 has disappeared now).

image

We get a small speedup on this example (from 3000us to 2900 us).

Some thoughts:

1. I think to actually do this we need to refactor `SchedulerNode` quite a bit - specifically, we don't have a "ir.Node" object that we can put on the object. Currently I just fake it.

### DIFF
--- a/torchinductor/scheduler.py
+++ b/torchinductor/scheduler.py
@@ -384,7 +384,14 @@ class SchedulerNode(BaseSchedulerNode):
 
 
 class FusedSchedulerNode(BaseSchedulerNode):
+    """
+    This is a "fake" scheduler node that represents a group of scheduler nodes
+    that are meant to be fused together. The way it does this is by maintaining
+    its unmet dependencies as the union of its constituent nodes.
+    """
+
     def __init__(self, scheduler: "Scheduler", snodes):
+        # NB: No need to call super().__init__() because we don't need to re-use any of its logic.
         self.snodes = snodes
         self.scheduler = scheduler
         self.node = None
@@ -392,18 +399,19 @@ class FusedSchedulerNode(BaseSchedulerNode):
         self.inverse_users = []
         node = snodes[0].node
         self.group = snodes[0].group
-        self.unmet_dependencies = self.snodes[0].read_writes.reads
+        reads = set(self.snodes[0].read_writes.reads)
         for node in snodes[1:]:
-            self.unmet_dependencies |= node.read_writes.reads
+            reads |= node.read_writes.reads
             assert self.group == node.group
         node_bufs = set([node.get_name() for node in snodes])
-        self.unmet_dependencies = [
-            read for read in self.unmet_dependencies if read.name not in node_bufs
-        ]
+        reads = [read for read in reads if read.name not in node_bufs]
+        # self.read_writes field shouldn't really be needed for much, but it's
+        # useful for visualization to track what buffers it reads from
+        self.set_read_writes(dependencies.ReadWrites(reads, None, None, None))
         self.prune_deps()
 
     def get_name(self):
-        return ",".join([x.get_name() for x in self.snodes])
+        return "_".join([x.get_name() for x in self.snodes])
 
     def __repr__(self):
         return f"{type(self).__name__}(nodes={self.get_name()})"
@@ -428,9 +436,6 @@ class FusedSchedulerNode(BaseSchedulerNode):
         raise NotImplementedError
 
     def get_mutations(self):
-        raise NotImplementedError
-
-    def set_read_writes(self, rw):
         raise NotImplementedError
 
     def get_device(self):
@@ -563,6 +568,12 @@ def draw_buffers(nodes, fname, print_graph=False):
 
 
 class DisjointSetUnion:
+    """
+    Maintains a disjoint set union data structure, with union-by-size (log(n) queries).
+
+    See https://cp-algorithms.com/data_structures/disjoint_set_union.html for a reference
+    """
+
     def __init__(self, n):
         self.parent = list(range(n))
         self.size = [1] * n
@@ -607,7 +618,6 @@ def create_fx_from_snodes(snodes: List[BaseSchedulerNode]) -> fx.Graph:
     outputs = []
     # create call_function node for each Buffer and Kernel
     for snode in snodes:
-        node = snode.node
         if isinstance(snode, ExternKernelSchedulerNode):
             node_type = "extern"
             group = node_type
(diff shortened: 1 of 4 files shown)


Chillee:
I implemented a greedy-ish strategy that's relatively simple.

1. I loop through all nodes and consider two nodes to be a fusion opportunity if there is overlap among their reads. (currently N^2).
2. I order all fusion opportunities in descending order of how many reads overlap.
3. I iterate down this list (once), merging the two nodes into "mega-nodes" if they're fusible. Two nodes `a` and `b` are fusible if there is no path from `a` to `b` (other than a direct connection).

torchinductor/scheduler.py
```diff
@@ -362,6 +362,60 @@ def get_priority(self):
...
+        for node in snodes[1:]:
+            read_writes = dependencies.merge_read_writes(read_writes, node.read_writes)
+        node_bufs = set([node.get_name() for node in snodes])
+        read_writes.reads = [i for i in read_writes.reads if i.name not in node_bufs]
+        read_writes.writes = [i for i in read_writes.writes if i.name not in node_bufs]
+        self.set_read_writes(read_writes)
+
+    # SchedulerNode overrides
+    def can_remove_buffer(self, check_group):
+        return any([node.can_remove_buffer(check_group) for node in self.snodes])
```
jansel, line 385: This doesn't look right.  We should assert False here, then have the caller iterate over `self.snodes`.

torchinductor/scheduler.py
```diff
@@ -382,6 +383,78 @@ def get_priority(self):
             return len(self.group) - 1
 
 
+class FusedSchedulerNode(BaseSchedulerNode):
+    def __init__(self, scheduler: "Scheduler", snodes):
```
jansel, line 387: Call `super().__init__()` or add a comment explaining why you're not calling it.

torchinductor/scheduler.py
```diff
@@ -382,6 +383,78 @@ def get_priority(self):
...
+        self.snodes = snodes
+        self.scheduler = scheduler
+        self.node = None
+        self.users = None
+        self.inverse_users = []
+        node = snodes[0].node
+        self.group = snodes[0].group
+        self.unmet_dependencies = self.snodes[0].read_writes.reads
```
jansel, line 395:
```suggestion
        self.unmet_dependencies = set(self.snodes[0].read_writes.reads)
```
Create a copy since you are mutating it below.  We want to leave the original one intact.

torchinductor/scheduler.py
```diff
@@ -459,87 +537,237 @@ def create_fx_from_buffers(nodes, fname, print_graph=False):
...
-            stride = node.get_stride()
-        except AttributeError:
-            stride = None
+        metadata = TensorMetadata(group, dtype, None, None, None, None, None)
+        node.meta["tensor_meta"] = metadata
 
-        layout = type(node.layout)
+    if print_graph:
+        print(graph)
+    print("starting creating module")
```
jansel, line 557: Remove prints before landing, or convert to logging calls.

torchinductor/scheduler.py
```diff
@@ -459,87 +537,237 @@ def create_fx_from_buffers(nodes, fname, print_graph=False):
...
-        if isinstance(snode, NopKernelSchedulerNode):
-            group = "nop"
-        elif isinstance(snode, ExternKernelSchedulerNode):
-            if should_use_template(node):
-                group = snode.group[1]
-            else:
-                group = "extern"
-        else:  # SchedulerNode
+
+class DisjointSetUnion:
```
jansel, line 565: Add docstring

torchinductor/scheduler.py
```diff
@@ -459,87 +537,237 @@ def create_fx_from_buffers(nodes, fname, print_graph=False):
...
+    def is_fusible(src, dst):
+        # Finds whether there's a path from src to dst that isn't a direct edge
+        cur_nodes = collections.deque(
+            [user for user in src.users if user != dst]
+            + [user for user in dst.users if user != src]
+        )
+        vis = set()
+        while len(cur_nodes) > 0:
+            cur = cur_nodes.popleft()
+            if cur in vis:
```
jansel, line 710: Small perf optimization that uses topological sorting.
```suggestion
            if cur in vis or cur.order > max(src.order, dst.order):
```

torchinductor/scheduler.py
```diff
@@ -459,87 +537,237 @@ def create_fx_from_buffers(nodes, fname, print_graph=False):
...
+            for user in cur.users:
+                cur_nodes.append(user)
+        return True
+
+    def fuse_nodes(a: fx.Node, b: fx.Node):
+        a.args = tuple(set(a.args) | set(b.args) - set([a, b]))
+        b.replace_all_uses_with(a)
+        a.meta["fusion_meta"].snodes.extend(b.meta["fusion_meta"].snodes)
+        a.meta["fusion_meta"]._replace(type="fused")
+        a.name = a.name + "_" + b.name
```
jansel, line 724:
```suggestion
        a.name = a.name + "_" + b.name
        a.order = max(a.order, b.order)
```
For the optimization above to work.

jansel (changes requested):

torchinductor/scheduler.py
Chillee: I actually had this perf optimization before, but removed it due to some complexity around correctness.

I’m not actually sure your suggestion maintains correctness. Lemme comment more in that part.

torchinductor/scheduler.py
Chillee: I’m not sure this is sufficient. The problem is that after merging two nodes a and b, where a.order < b.order, the new node cannot necessarily be located at *either* a or b.

a = f(input)
c = f(a)
d = f(input)
b = f(a, d)

a and b can be fused, but the resulting node cannot be at either a’s position nor b’s position.

I think it might still be possible to maintain some kind of structure that allows for more efficient queries, but it’s non-trivial I think.

jansel: I agree "the resulting node cannot be at either a’s position nor b’s position."

Though for this optimization I think that is fine.  The combined node should never need to be placed *after* `max(a.order, b.order)`, so it is fine to short circuit for nodes beyond that point.

Chillee: hmm, I agree with that. Will add a note about that.

torchinductor/scheduler.py
Chillee: I'm not calling it since I don't think any of the things done in the baseschedulernode class are needed? Is that a valid reason :)

torchinductor/scheduler.py
Chillee: Actually, I'm still not sure this works.

Say...

1. fuse(buf_1, buf_5) (`buf1_buf5` has order = 5)
2. is_fusible(buf_0, buf_2), where there's a path `buf_0 => buf_1 => buf_2`.
3. In `is_fusible`, we check if `buf1_buf5.order > max(buf_0.order, buf_2.order)`, which it is, since that's set as `max(buf1_buf5)`.

I think I'd prefer not to do this optimization for now. I think there has to be a better way of optimizing this query.

jansel: Ah you are right.  We might need `.min_order`/`.max_order` to do it safely.  I am ok skipping.

torchinductor/scheduler.py
jansel: Yeah, just document that to it is clear that this is intentional.

jansel (approved):

torchinductor/scheduler.py
Chillee: These only show up in debug code (i.e. pass `INDUCTOR_SCHEDULER_GRAPH=1`).

Issue #687, opened 2022-07-31
Slow eager compilation time with TorchDynamo


anijain2305:
TorchDynamo takes roughly 95 seconds to compile `desnenet121` timm model even with `optimize("eager")`. This is because of aggressive inlining. We inline and therefore recompile everytime we see a common function.

`python benchmarks/timm_models.py --devices=cuda --only=densenet121`

If I disable the Inlining, the recompilation latency goes down to just 5 seconds. However, this is obviously not the right thing to do, because it causes graph breaks and also cache size limit violations.

This effect is even more pronounced for `dpn107` model with roughly 1400 seconds of compilation time, that gets reduced to a mere 5 seconds when inlining is disabled.

@jansel Do you have any suggestions on any re-design that we can do here?

soumith:
cc: @wconstab 

Yea, this doesn't sound right from any UX perspective. Curious to see what's planned next.

jansel:
This is clearly unacceptable.  Hopefully it is an easy fix.

If inlining is the root cause, we could cache the result of inlining to avoid re-tracing the same function.  Though I suspect there might be an easier fix just by optimizing the tracing operation to make inlining fast enough we don't need to worry about it.

The first step would be to add a call to `cProfile` in `_convert_frame_assert` to get some data on where the time is being spent.

voznesenskym:
@anijain2305 have we run a profile to see where time is spent with/without inlining?

anijain2305:
@jansel @voznesenskym Will do a profile and update later today.

anijain2305:
Update - The slowdown is not really due to Inlining. It is due to `RestartAnalysis` - https://github.com/pytorch/torchdynamo/blob/main/torchdynamo/convert_frame.py#L291
There are roughly 50 attempts for one frame, causing time to shoot up.

I am going to debug this further tomorrow.

voznesenskym:
> Update - The slowdown is not really due to Inlining. It is due to `RestartAnalysis` - https://github.com/pytorch/torchdynamo/blob/main/torchdynamo/convert_frame.py#L291 There are roughly 50 attempts for one frame, causing time to shoot up.
> 
> I am going to debug this further tomorrow.

This only comes from `convert_to_unspecialized`, which in turn is only invoked via `setattr` (through our builtin handling). This is of course downstream of our symbolic convert's `STORE_ATTR`.

[...]

PR #697, opened 2022-08-02
Tag the module class Dynamic instead of just instances

anijain2305:
Densenet had one module that was instantiated > 50 times. This module was changing the state of the module. Currently, Dynamo marks the instance of this module as UnSpecialized and then restarts Analysis.

Since, the module was instantiated many times, we were stuck in these restarts.

Solution (which was already present and commented) is to mark the class dynamic, so there will be just one restart and then all the instances will be of UnSpecialized type.

With this, the compile time latency reduced from 94 to 9 seconds for densenet121, and 640 to 19 seconds for dpn107.

xRef - https://github.com/pytorch/torchdynamo/issues/687

### DIFF
--- a/torchdynamo/variables/nn_module.py
+++ b/torchdynamo/variables/nn_module.py
@@ -84,7 +84,10 @@ class NNModuleVariable(VariableTracker):
         """Restart analysis treating this module as an UnspecializedNNModuleVariable"""
         mod = tx.output.get_submodule(self.module_key)
         GenerationTracker.tag(mod)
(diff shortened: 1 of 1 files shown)


anijain2305:
Oh, forgot to mention. Jason said that he commented the line earlier there because there could be a setattr in `__init__` function but not in `forward` and we could turn that class as UnSpecialized.

I could guard it on the __init__ name by plumbing the frame name details through the tx.

jansel (approved):

PR #701, opened 2022-08-03
Add `masked_fill.Tensor`

Chillee:
Not sure why, but this missing decomposition seems to be causing some issues.

### DIFF
--- a/torchinductor/decomposition.py
+++ b/torchinductor/decomposition.py
@@ -201,7 +201,7 @@ def rsub(a, b):
     return b - a
 
 
-@register_decomposition([aten.masked_fill.Scalar])
+@register_decomposition([aten.masked_fill.Scalar, aten.masked_fill.Tensor])
 def masked_fill(value, mask, other):
     if isinstance(other, numbers.Number):
         other = torch.tensor(other, dtype=value.dtype, device=value.device)


torchinductor/decomposition.py
```diff
@@ -201,7 +201,7 @@ def rsub(a, b):
     return b - a
 
 
-@register_decomposition([aten.masked_fill.Scalar])
+@register_decomposition([aten.masked_fill.Scalar, aten.masked_fill.Tensor])
```
ngimel, line 204: fine, but this decomp is too permissive. `masked_fill` supports only 0-d other tensor, and here as long as it's broadcastable with value and mask it'll work.

ngimel (approved):

torchinductor/decomposition.py
jansel: I hit this same issue in another PR and tried the ref.

The ref has issues with creating CPU-tensors for constants, which `aten.where` then refuses to implicitly move to the GPU during tracing.

We should fix this ref so we can delete this.

ngimel: I don't know how to fix the ref so that it can use 0d cpu tensors. `where` doesn't support it, so we need to either convert 0d to python number (that's `.item` call that would break the graph) or copy the tensor to gpu which is synchronizing. ref currently uses `item` but there's no satisfactory solution.

jansel: For inductor calling `.to()` might result in better code, that is what the latest version of this function in main does.  Inductor has optimizations for `constant0.to("cuda")`, where the constant gets moved at compile time (rather than run time).  This only works for constants though.

jansel:
Fixed in #728

PR #702, opened 2022-08-03
Add CUDA graphs CI to torchbench CI job

ezyang:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,8 +9,7 @@ CLANG_FORMAT ?= clang-format-10
 PIP ?= python -m pip
 
 # versions used in CI
-PYTORCH_VERSION ?= dev20220710
-FUNCTORCH_VERSION ?= 9b96f14e65ffdf64a28416054fe536ed14297fe9
+PYTORCH_VERSION ?= dev20220803
 TRITON_VERSION ?= 5b04331dd2efdd23f4475823761fa975de60a514
 
 
@@ -52,7 +51,7 @@ setup:
 setup_nightly:
 	$(PIP) install ninja
 	$(PIP) install --pre torch==1.13.0.$(PYTORCH_VERSION) --extra-index-url https://download.pytorch.org/whl/nightly/cpu
-	$(PIP) install -v "git+https://github.com/pytorch/functorch.git@$(FUNCTORCH_VERSION)"
+	$(PIP) install -v "git+https://github.com/pytorch/pytorch.git@`python -c "import torch.version; print(torch.version.git_version)"`#subdirectory=functorch"
 	$(PIP) install -r requirements.txt
 	python setup.py develop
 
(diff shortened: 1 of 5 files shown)


jansel (changes requested):
I think you need to update the pytorch version pin:
https://github.com/pytorch/torchdynamo/blob/4f5a24e566a0b1d231aaca110f4ee75eaf8dc935/Makefile#L12

ezyang:
this is no longer mergeable as is, need https://github.com/pytorch/torchdynamo/pull/704 first. I hacked out the broken bits for now

ezyang:
this is not mergeable as is, but if people have opinions about the color of the bikeshed (e.g., run cuda graphs ci alongside coverage; do it as a parallel job; etc) please let me know

torchdynamo/optimizations/python_key.py
```diff
@@ -140,16 +136,17 @@ def run_node(self, n: torch.fx.Node):
...
-            python_tensor_cls(elem, proxy)
-            for elem, proxy in zip(chain(params_flat, example_inputs), proxy_args)
-        ]
-        with _stateless.reparametrize_module(
-            gm, pytree.tree_unflatten(args[:params_len], params_spec)
-        ):
-            out = PatchingInterpreter(gm).run(*args[params_len:])
-
-        assert isinstance(out, (tuple, list)), "graph must output tuple()"
+        with proxy_mode:
```
msaroufim, line 139: Maybe it's late but this context manager is defined as `None` above so shouldn't have an `__enter__` defined?

Makefile
```diff
@@ -110,7 +107,7 @@ build-deps: clone-deps
 	(cd ../torchtext   && python setup.py clean && python setup.py develop)
 	(cd ../torchaudio  && python setup.py clean && python setup.py develop)
 	(cd ../detectron2  && python setup.py clean && python setup.py develop)
-	(cd ../functorch   && python setup.py clean && python setup.py develop)
+	(cd ../pytorch/functorch   && python setup.py clean && python setup.py develop)
```
msaroufim, line 110: So people still need to install functorch even though it's back in core?

torchdynamo/optimizations/python_key.py
```diff
@@ -160,6 +157,7 @@ def unpack(x):
 
     with pythonkey_decompose({**aot_autograd_decompositions, **decompositions}):
         tracer: torch.fx.Tracer = PythonKeyTracer()
+        proxy_mode = ProxyTorchDispatchMode(tracer)
```
msaroufim, line 160: does this change the state of the tracer?

Makefile
ezyang: we don't have consolidated build yet, so yes you still have to explicitly build it

torchdynamo/optimizations/python_key.py
ezyang: these are from the version update PR, go put your comments in that one

PR #703, opened 2022-08-03
CI run of accuracy-aot-nop on torchbench

ezyang:

### DIFF
--- a/.circleci/check_aot.py
+++ b/.circleci/check_aot.py
@@ -9,7 +9,7 @@ def check_aot_nop():
     Basic accuracy checking. If a model fails, the speedup will be 0.
     """
 
-    actual = pd.read_csv("aot_nop.csv")
+    actual = pd.read_csv("aot_eager.csv")
 
     failed = []
     for model_name in actual["name"]:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 version: 2.1
 
 jobs:
-  gpu_linux:
+  coverage:
     machine:
       # https://circleci.com/docs/2.0/configuration-reference/#available-linux-gpu-images
       image: ubuntu-2004-cuda-11.4:202110-01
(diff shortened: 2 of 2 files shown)


ezyang:
this is not intended to merge, but I'm curious if this torchbench config actually passes

anijain2305:
Haha, we might need to -x more models here. One thing we could try is --isolate to run it through and collect the failures and then -x them.

ezyang:
Yeah, we should probably stem the bleeding by -x'ing to start.

ezyang:
This also takes kind of a long time to run, so I wonder if we can lower the batch size to make it quicker.

anijain2305:
Ed, I will take over this one

anijain2305:
cc @jansel @ezyang For review

The main question is - Should we parallelize the AOT run? Currently, the job takes 47 min in total.

ezyang:
If you can parallelize it easily, it's probably worth it. But I kind of wonder why it takes so long; we also have a coverage test and that one doesn't take nearly as long, IIUC?

anijain2305:
> If you can parallelize it easily, it's probably worth it. But I kind of wonder why it takes so long; we also have a coverage test and that one doesn't take nearly as long, IIUC?

image

Coverage takes around 5 min. Coverage tests inference. `aot_nop` is about training. So, I think 15 min is expected for AOT eager runs.  Rest of the time 45 min goes in Torchbench install and Dynamo tests on GPUs.

One thing we could skip in `aot_nop` is speedup comparison atleast for this PR. But, I don't think its gonna move the needle in any major way.

anijain2305:
Regarding parallelization, I dont think we can do much parallelization in one job. AFAICT CircleCI supports only 1 GPU machines - https://circleci.com/docs/using-gpu

Other way is to spin up another job for `aot_nop`. There is some wastage because we will have to install Torchbench model in both jobs, but everything else could be parallelized.

ezyang:
20 min for torchbench testing seems reasonable. I'm more concerned that it takes 45min to install torchbench XD

anijain2305:
> 20 min for torchbench testing seems reasonable. I'm more concerned that it takes 45min to install torchbench XD

Oh, I meant that rest of time (out of total 45 min) goes in Torchbench install and setup. More details 

image

Current breakdown
1) setup (conda installs + torchbench) = 17 min
2) Dynamo tests on GPU = 9 min
3) Torchbench coverage = 5 min
4) Torchbench AOT = 15 min
Total = 45 min


I just tried spinning up one more CircleCI job and it seems to work so far. For now,
Job 1 will have (1), (2) and (3).
Job 2 will have (1) and (4).
So, roughly 30 min each.

We then can load balance between two machines as we add more backends (like aot_cudagraphs and inductor) to test. How does that sound?

voznesenskym:
> Current breakdown
> 
> 1. setup (conda installs + torchbench) = 17 min

Relevant: https://github.com/pytorch/torchdynamo/issues/377 - might be able to help us get this down

voznesenskym:
> > Current breakdown
> > 
> > 1. setup (conda installs + torchbench) = 17 min
> 
> Relevant: #377 - might be able to help us get this down

https://circleci.com/docs/caching for circle CI runners

anijain2305:
> https://circleci.com/docs/caching for circle CI runners

This sounds quite useful. We might also be able to share the artifacts of Torchbench installation, and reduce wastage. 
Let's work on that in a separate PR.

jansel:
The pytorch+torchbench setup should totally be cached.  It does the same thing for every PR.

There is also a terrible hack that installs torchbench from a tarball because git-lfs wouldn't work right.

anijain2305:
@jansel Can we tackle caching pytorch+torchbench thing in a separate PR, unless its a blocker?

I think this PR is a step up from where we are today and can catch regressions.

jansel:
Seperate PR is fine with me

jansel (approved):

PR #707, opened 2022-08-03
tl.load other=0 to ensure correctness of fp16

pyjhzwh:
for masked tl.load(), other=0.0/ 0 to ensure correctness. Otherwise, fp16 will have wrong outputs

### DIFF
--- a/torchinductor/triton_ops/conv.py
+++ b/torchinductor/triton_ops/conv.py
@@ -119,9 +119,9 @@ def _kernel_delta_x_hwc(
     mask_w = (off_x_crs < CRS)[:, None] & (off_w_k < KERNEL_N)[None, :]
 
     # ------ load x ------
-    matrix_x = tl.load(x_ptrs, mask=mask_x)
+    matrix_x = tl.load(x_ptrs, mask=mask_x, other=0.0)
     # ------ load w ------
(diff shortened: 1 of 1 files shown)


torchinductor/triton_ops/conv.py
```diff
@@ -138,9 +138,9 @@ def _kernel_delta_x_hwc(
             delta_xh_ptrs += BLOCK_K
             delta_xw_ptrs += BLOCK_K
             delta_xc_ptrs += BLOCK_K
-            delta_xh = tl.load(delta_xh_ptrs, mask=off_x_crs < CRS)
-            delta_xw = tl.load(delta_xw_ptrs, mask=off_x_crs < CRS)
-            delta_xc = tl.load(delta_xc_ptrs, mask=off_x_crs < CRS)
+            delta_xh = tl.load(delta_xh_ptrs, mask=off_x_crs < CRS, other=0)
```
ngimel, line 141: I'm surprised you need to set `other` for offsets - data with invalid offsets should never be loaded, it should be masked.

pyjhzwh: I found it's weird, everthing works fine for float32. But when I only change dtype to float16, some tests would failed the correctness checking

ngimel (approved):

PR #711, opened 2022-08-04
Add support for tracing rest of optimizers

mlazos:
* Add support for tracing remaining optimizers (caveat: most require capturing scalar outputs due to `item()` calls, there are workarounds though, see below)
* 10/13 optimizers can be fully captured
SparseAdam is not supported
LBFGS and RAdam have graph breaks due to data-dependent control

Next steps:
* Fix pytorch assert in Adam which will allow users to set `capturable` to true and avoid calls to `item()`
* Fix optimization in ASGD which introduces data-dependent control, these ops will get fused anyway

#328


mlazos:
FYI, the test failing needs a FakeTensor fix in Pytorch latest, should be ready for review.

jansel (approved):

PR #712, opened 2022-08-04
Push node associated stack trace one level deeper in dynamo->inductor

voznesenskym:
This sets it up so that the graph going into fx.Transformer has the right stack_trace associations on it, node by node

### DIFF
--- a/torchdynamo/optimizations/normalize.py
+++ b/torchdynamo/optimizations/normalize.py
@@ -388,7 +388,7 @@ class Functionalization(Transformer):
 
 
 def swap_node(graph, old_node, new_node):
-    assert(new_node.stack_trace is None), "New node must not have a "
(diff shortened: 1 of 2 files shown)


torchdynamo/optimizations/normalize.py
```diff
@@ -388,6 +388,8 @@ def run_node(self, n: torch.fx.Node):
 
 
 def swap_node(graph, old_node, new_node):
+    assert new_node.stack_trace is None, "New node must not have a stack_trace"
+    new_node.stack_trace = old_node.stack_trace
```
SherlockNoMad, line 392: copy the meta?

SherlockNoMad (approved):

voznesenskym:
Done elsewhere

PR #714, opened 2022-08-04
Add optional unspecializing int/float and heuristic

yanboliang:
This is re-submit of https://github.com/pytorch/torchdynamo/pull/520.

Major changes:
* Add config ```specialize_int_float``` to let users choose spec or unspec int/float.
* Implement [the heuristic 1/2/3](https://github.com/pytorch/torchdynamo/issues/70#issue-1173745001) to decide when to specialize and when not to.
* Add function ```convert_to_constant``` and ```specialize_args_kwargs``` to support switch from unspec to spec if it goes into some operations, e.g, control flow, list index, etc.
* Refactor unit tests to cover both spec and unspec int/float test cases.

Tests:
* All unit tests passed.
* All torchbench models passed with both spec and unspec config.
  * ```vision_maskrcnn``` has to specialize the dividend when float divide by tensor due to precision difference, and constant the float argument of ```torch.clamp``` which can't support arguments on different devices.

Fix https://github.com/pytorch/torchdynamo/issues/538


torchdynamo/symbolic_convert.py
```diff
@@ -545,6 +546,10 @@ def FOR_ITER(self, inst):
 
     def COMPARE_OP(self, inst):
         left, right = self.popn(2)
+        if isinstance(left, UnspecializedPythonVariable):
+            left = left.convert_to_constant(self)
```
yanboliang, line 550: For everyplace calling ```convert_to_constant```, potentially there is chance this variable has been used before this, which makes the same variable has multiple different copies. But this never occurs for all torchbench models, so I leave it as is. The future plan is: I have another PR which use guard failures to determinate switch between spec/unspec, there is a global data structure to store all local variables and distinct values occurred. When it's ready to post, I'll update this function to first check if this variable exists in the global data structure and:  1/ convert to constant if not; 2/ raise ```RestartAnalysis``` and mark this as _must spec_ if yes.

torchdynamo/eval_frame.py
```diff
@@ -272,10 +272,23 @@ def produce_matching(source_args, candidate_args):
 
         for i in range(0, len(candidate_args)):
             arg = candidate_args[i]
-            assert (
-                id(arg) in dict_of_source_args
-            ), "Dynamo input and output is a strict subset of traced input/output"
-            matched_elements_positions.append(dict_of_source_args[id(arg)])
+            # 1-element tensor arg can be unspec int/float
+            if isinstance(arg, torch.Tensor) and isinstance(arg.tolist(), (int, float)):
```
voznesenskym, line 276: 1) Why does `tolist()` make an int or float? is this how we check for it being a single size?
2) `.item()` here will throw if this is not a single element tensor
3) This is 2n where n is size of dict_of_source_args, wonder if we can make it 1n?

torchdynamo/config.py
```diff
@@ -33,6 +33,9 @@
 # disable (for a function) when cache reaches this size
 cache_size_limit = 64
 
+# specializing int/float by default
+specialize_int_float = True
```
voznesenskym, line 37: How would you like this to work with `capture_scalar_outputs`?

torchdynamo/variables/tensor.py
```diff
@@ -671,6 +676,17 @@ def from_tensor_variable(cls, tensor_variable, raw_value, need_unwrap=True):
             need_unwrap=need_unwrap,
         )
 
+    def convert_to_constant(self, tx):
+        for graph_arg in tx.output.graphargs:
```
voznesenskym, line 680: this feels rather heavy - matching on source like this for convert_to_constant - for example, when you do slice, you end up doing k (where k is size of slice) * g (where g is number of graphargs) operations. 

You could instead just make a source->grapharg mapping, right? and then do something like `tx.output.erase(self.source)` which would do `src_to_grapharg[self.source].erase() under the hood` and remove the key?

torchdynamo/variables/tensor.py
```diff
@@ -671,6 +676,17 @@ def from_tensor_variable(cls, tensor_variable, raw_value, need_unwrap=True):
...
+            if graph_arg.source is self.source:
+                graph_arg.erase()
+
+        for g in self.guards:
+            if g is self.volatile_guard:
```
voznesenskym, line 685: what are volatile guards?

torchdynamo/variables/tensor.py
```diff
@@ -671,6 +676,17 @@ def from_tensor_variable(cls, tensor_variable, raw_value, need_unwrap=True):
...
+                g.create_fn = GuardBuilder.CONSTANT_MATCH
```
voznesenskym, line 686: This feels a little sus to me... why repurpose a guard with a new create_fn?

torchdynamo/variables/tensor.py
```diff
@@ -671,6 +676,17 @@ def from_tensor_variable(cls, tensor_variable, raw_value, need_unwrap=True):
...
+
+        return ConstantVariable(value=self.raw_value, guards=self.guards)
```
voznesenskym, line 688: hmm - is this safe? do the old guards made for the last object apply here? I ask because VariableBuilder thinks that `CONSTANT_MATCH` suffices for a constant variable, maybe just have that one guard?

torchdynamo/utils.py
```diff
@@ -416,6 +416,64 @@ def check_constant_args(args, kwargs):
     return all(x.is_python_constant() for x in itertools.chain(args, kwargs.values()))
 
 
+def check_unspec_python_args(args, kwargs):
+    from .variables.constant import ConstantVariable
+    from .variables.tensor import FakeItemVariable
+    from .variables.tensor import UnspecializedPythonVariable
+
+    return (
```
voznesenskym, line 424: Is there a way to write this without iterating over args + kwargs 3 times?

torchdynamo/utils.py
```diff
@@ -416,6 +416,64 @@ def check_constant_args(args, kwargs):
...
+        if isinstance(
+            x,
+            (
+                UnspecializedNumpyVariable,
+                UnspecializedPythonVariable,
+            ),
+        ):
+            specialized_args.append(x.convert_to_constant(tx))
+        else:
+            specialized_args.append(x)
```
voznesenskym, line 462: I am a little confused by this - if args are unspecialized, we record them as constants, but otherwise we record them as is? If this is to provide safety for upstream of places where we check for `isinstance(foo, variables.ConstantVariable)`, we may be better off making a helper function like `is_constant_convertible(foo)` and  `to_constant(foo)`?

torchdynamo/eval_frame.py
yanboliang: 1, ```tolist()``` returns list(if it's not 1-element tensor) or number(if it's 1-element tensor), so it meets our need to check if it's 1-element tensor and return scalar value.
2, Yes, so we can't use ```item()```.
3, Actually the worst case is 2n, average is better than 2n, but I don't think this is bottleneck. Here the # of arguments is not very large usually and this is not a critical C/C++ code. If you have better idea, I'm glad to update it.

torchdynamo/config.py
yanboliang: It works well, I have [unit tests](https://github.com/pytorch/torchdynamo/pull/714/files#diff-48a9157fdd44237aa7399726b4e554f625e74de37b4c9c95acf7cdb9f2072ba4R46) covered.

torchdynamo/variables/tensor.py
yanboliang: I think slice doesn't work as you described, because:
* If we support slice as unspec, 3 graphargs(start, stop, step) at most.
* Actually we don't support slice as unspec.

To have a source->grapharg mapping makes sense, this is aligned with my next step, see my [comments here](https://github.com/pytorch/torchdynamo/pull/714#discussion_r938293298).

torchdynamo/variables/tensor.py
yanboliang: ```volatile_guard``` is a mark to identify this guard may be changed later, I use it when initializing ```UnspecializedPythonVariable``` coz it's possible to convert to ```ConstantVariable``` when it encounter some operations, e.g, control flow, list index, etc.

torchdynamo/variables/tensor.py
yanboliang: When convert to constant, we have to replace the old ```TYPE_MATCH``` guard to ```CONSTANT_MATCH```. We can't replace the whole guard object easily, because it has been propagated to down streams during the symbolic converting.

torchdynamo/variables/tensor.py
yanboliang: L686 updated the match method, so it's safe.

torchdynamo/utils.py
yanboliang: This is a fair point, but I had addressed in this [comment](https://github.com/pytorch/torchdynamo/pull/714#discussion_r938293298). The short answer is: 1, it's rare case,  never occurs for all torchbench models, so I keep it as is to make this PR not too big to review; 2, I'll have another PR which adding a global data structure to store all local variables <-> source <-> spec/unspec, and make a helper function like you pointed out.

torchdynamo/utils.py
yanboliang: There is, but I don't think this is a bottleneck. There are many other places we are using this way, so maybe we can do a sweep together if we find this is bottleneck.

torchdynamo/eval_frame.py
voznesenskym: 1. Maybe instead of to-listing a large tensor, we can just check numel, (thats faster and iirc what .item() does)?
2. Ok
3. Its not a bottleneck, just wondering if we cant keep it down to a single loop.

torchdynamo/config.py
voznesenskym: Maybe a comment on what the spec/unspec + capture/no-capture does? Not super important.

torchdynamo/variables/tensor.py
voznesenskym:
```
 [
                    x.convert_to_constant(self)
                    if isinstance(x, UnspecializedPythonVariable)
                    else x
                    for x in items
                ],
```

Looks like we get a potential to call `convert_to_constant` on every item, which in turn does this heavy iteraiton. Would you mind putting the mapping in this PR, or even as a PR before this one?

torchdynamo/variables/tensor.py
voznesenskym: I understand, the question is more around, why not just make a whole new set of guards? If the reason is to remove the guards, we should be explicit, and remove the old guards and add new ones, instead of repurposing.

torchdynamo/variables/tensor.py
voznesenskym: This is an extension of my previous question - why make >1 `constant_match`? why even have volatile guards? why not just make all new guards here? 

Maybe something much more explicit like:

```
self.guards.remove(type_match)
self.guards.remove(x)
return ConstantVariable(value=self.raw_value, guards=guards(source, name, CONSTANT_MATCH)
```

That way, we are much clearer, it stays simpler, and we do not need to introduce a new notion of volatile guards. 

Also... what about guards that are not create_fn assigned, like tensor_guards? will this still invoke `___check_tensors`?

torchdynamo/utils.py
voznesenskym: We don't need it to be global, I think, and this is actually a slightly different question from mapping altogether.

The question is, why not make some helper methods that make it so that we don't need to do all this conversion ahead of time? Why not just let places that take `ConstVariable` also take `UnspecializedNumpyVariable` and `UnspecializedPythonVariable`, and then just give them a `to_const_variable() -> ConstVariable` type?

torchdynamo/variables/tensor.py
yanboliang:
```
class SliceVariable(BaseListVariable):
  def __init__(self, items, **kwargs):
          start, stop, step = [variables.ConstantVariable(None)] * 3
          if len(items) == 1:
              (stop,) = items
          elif len(items) == 2:
              start, stop = items
          elif len(items) == 3:
              start, stop, step = items
          else:
              assert False
          super().__init__([start, stop, step], **kwargs)
```
Look at ```SliceVariable``` constructor function, ```len(items)``` should always <= 3.

torchdynamo/variables/tensor.py
yanboliang: Because the guard has been propagated to many down streams, it's very heavy operation to trigger re-analysis and update the whole set of guards.

torchdynamo/variables/tensor.py
yanboliang: Still the same, the guards on this variable have been propagated to many down streams. If we do sth like you mentioned, we only update the guard on this variable. But if down stream variables depends on this variable's guard as well, it still uses the old guards(type_match), which cause errors. I think most of your comment here is about this guard replacement, I'm happy to schedule a VC chat to discuss if it can help.

torchdynamo/utils.py
yanboliang: Sorry I can't catch your point, can you elaborate more or do you have example?

torchdynamo/eval_frame.py
yanboliang: I don't know if ```numel``` is faster than ```tolist```, but I'm ok to use ```numel```.

torchdynamo/variables/tensor.py
yanboliang: BTW, ```volatile_guard``` is used to help us find out which guard need to be replaced, because there are possible multiple guards on the same variable, we need to identify which is exactly the unspec guard.

torchdynamo/variables/tensor.py
voznesenskym: Still seems wasteful when it doesn't really need to for loop, but I wont fight you on it.

torchdynamo/variables/tensor.py
voznesenskym: A whole new set of guards just for this object - why do we need to keep the old guards if the type is changing from tensor to constant? and I have > 1 volatile_guard, does that mean I have > 1 CONSTANT_MATCH for the same object?

torchdynamo/variables/tensor.py
voznesenskym: > BTW, volatile_guard is used to help us find out which guard need to be replaced, because there are possible multiple guards on the same variable, we need to identify which is exactly the unspec guard.

I understand what you are doing. I just think it is too complicated. 

You only have a single volatile guard I see you make here, from `GuardBuilder.TYPE_MATCH` - why do we need a whole new concept of `volatile_guard` vertically integrated through so many layers? Why not just find the `TYPE_MATCH` guard and remove it explicitly?

torchdynamo/variables/tensor.py
voznesenskym: See my comment below - I think this system is too complicated for ultimately just removing a single type of guard `TYPE_MATCH`. Let's just remove all the volatile guard notions and explicitly remove the `TYPE_MATCH` and replace it with `CONSTANT_MATCH ` in this flow?

torchdynamo/utils.py
voznesenskym: Sure thing! The pattern we usually follow is not `x.convert_to_foo()` its `x.as_foo()` where `as_foo` is not destructive - it does not change the original object, but produces a new one. What's cool about that is we can have

```
class ConstantVariable :
def as_constant(tx) -> ConstantVariable:
     # identity
    return self
    
...
```

and 
```
UnspecializedVariable
def as_constant(tx) -> ConstantVariable:
    # your existing logic for making a ConstantVariable
    return ConstantVariable(...)
...
```

What's cool about this is you now dont need to do any iteration or pre-processing to turn out `ConstantVariable` from your `Unspecialized` objects, you can just pass around `Unspecialized` everywhere, and as long as you call `.as_constant()` right before you use them, it will be safe, and easier to read, and easier to extend if we get another type that needs to support `.as_constant()`

torchdynamo/variables/tensor.py
yanboliang: > Why not just find the TYPE_MATCH guard and remove it explicitly?

That's exactly what we want to do. A variable can have ___multiple___ guards of ```TYPE_MATCH```, what we can know which guard is the one we want to replace?

torchdynamo/variables/tensor.py
yanboliang: See this [comment](https://github.com/pytorch/torchdynamo/pull/714#discussion_r939038623)

torchdynamo/variables/tensor.py
yanboliang: See this https://github.com/pytorch/torchdynamo/pull/714#discussion_r939038623

torchdynamo/utils.py
yanboliang: Fair point, it can remove the pre-processing and always call ```as_constant```.

torchdynamo/variables/tensor.py
voznesenskym: 1) Would you not remove all the TYPE_MATCH in this case? When would you only remove some? 
2) If you do only need some, so even if you need partial removal, you don't need to make a new volatile_guard notion and store it in kwargs, or `self` - you can just use TYPE_MATCH itself, right? 

instead of:

```
for g in self.guards:
       if g is self.volatile_guard:
```

Can you store in on TYPE_MATCH?

```
for g in self.guards
    if g.is_volatile()
```

And then you are marking the guard itself, instead of having the object store its guard

torchdynamo/utils.py
voznesenskym: Let's write things efficiently if possible the first time around - its not a bottleneck, but if we can turn 3 iterations over the same thing into one, why not?

torchdynamo/utils.py
voznesenskym: Wonderful, thank you sir, sorry it took me so long to communicate what I meant clearly.

torchdynamo/variables/tensor.py
yanboliang: 1, We don't remove all ```TYPE_MATCH```, only remove the one binding to the specific unspec variable.
2, ```TYPE_MATCH``` is a function, so we can't store it on ```TYPE_MATCH```.

voznesenskym: 1) Understood
2) Why not?

```
def foo(x, y, z):
  return x + y + z

f = foo
f.volatile = True
print(f.volatile)
```

voznesenskym: Python is a ~~godless~~ dynamic and exciting language, you can even do:

```
def foo(x, y, z):
  return x + y + z

foo.foo = foo
foo.foo(1, 2, 3)
```

yanboliang: It seems it works well, but if we can put it as an attribute of class/object, why instead we put it as an attribute of function? I think the former is easy to read and understand.

voznesenskym: A function is also an instance of an object here. The benefit here is that its cleaner - what's volatile? the guard is. Instead of parallel bookkeeping, just keep it on the fn or on the guard - this also means that you can handle volatile guards anywhere, you do not need to track the object itself - for this case its the same, but there may be future cases where you just want to iterate over or manage volatile guards without having a reference to their originating object. 

TLDR is, guard is volatile, so store volatile on the guard.

voznesenskym: Thank you for making the change!

jansel (approved):

voznesenskym (approved):

PR #725, opened 2022-08-07
[inductor] Add aten.roll lowering

jansel:
Needed by swin_base_patch4_window7_224

### DIFF
--- a/tests/test_torchinductor.py
+++ b/tests/test_torchinductor.py
@@ -2574,6 +2574,20 @@ class CommonTemplate:
             expected_kernel = 6
         self.assertEqual(torchinductor.metrics.generated_kernel_count, expected_kernel)
 
+    def test_roll(self):
+        def fn(a):
+            return (
+                aten.roll(a, [-3, 10], [1, 2]),
+                aten.roll(a, [5]),
+            )
+
+        self.common(
+            fn,
+            [
+                torch.randn([2, 56, 56, 16]),
+            ],
+        )
+
 
 if HAS_CPU:
 
--- a/torchinductor/lowering.py
+++ b/torchinductor/lowering.py
@@ -485,6 +485,9 @@ def roll(a, shifts, dims=tuple()):
         dims = (dims,)
     dims = [_validate_dim(a, d) for d in dims]
 
+    if sympy_product(a.get_size()) == 0:
+        return clone(a)
+
     len_shifts = len(shifts)
     len_dims = len(dims)
     if len_shifts != 1 or len_dims != 1:
--- a/torchinductor/lowering.py
+++ b/torchinductor/lowering.py
@@ -1,6 +1,7 @@
 import functools
 import itertools
 import logging
+from collections.abc import Iterable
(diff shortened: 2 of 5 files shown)


anijain2305 (approved):
Understood the high level. Approving. Don't have enough context for the details.

torchinductor/lowering.py
```diff
@@ -469,6 +470,62 @@ def slice_(x, dim, start, end, step=1):
...
+            )
+        assert len_dims > 1
+        tail_shifts = shifts[1:]
+        tail_dims = dims[1:]
+        first_dim_rolled = roll(a, shifts[0], dims[0])
+        return roll(first_dim_rolled, tail_shifts, tail_dims)
+
+    (dim,) = dims
+    size = V.graph.sizevars.guard_static_shape(a.get_size()[dim])
+    start = (size - shifts[0]) % size
```
voznesenskym, line 511: Will this be a `ZeroDivisionError` if size is 0? 

The aten op does this:

```
if (self.numel() == 0) {
    return self.clone(at::MemoryFormat::Preserve);
  }
```
Does something upstream protect us from size being 0?

tests/test_torchinductor.py
```diff
@@ -2574,6 +2574,20 @@ def run(x):
             expected_kernel = 6
         self.assertEqual(torchinductor.metrics.generated_kernel_count, expected_kernel)
 
+    def test_roll(self):
```
IvanYashchuk, line 2577: It's possible to use to use `torch.testing._internal.common_methods_invocations.sample_inputs_roll` to get the arguments for testing.
Then it would match what's being tested in PyTorch in `test/test_ops.py -k roll` tests.

torchinductor/lowering.py
jansel: I don't believe size 0 is supported properly yet in inductor. I can add this though.

tests/test_torchinductor.py
jansel: I like this idea.  I created an issue #727

torchinductor/lowering.py
```diff
@@ -469,6 +470,65 @@ def slice_(x, dim, start, end, step=1):
...
+        # By default, the tensor is flattened before shifting, after which the original shape is restored
+        if len_dims == 0 and len_shifts == 1:
+            flat = view(a, [sympy_product(a.get_size())])
+            rolled = roll(flat, shifts, 0)
+            return view(rolled, list(a.get_size()))
+        if len_shifts != len_dims:
+            raise RuntimeError(
+                f"shifts and dimensions must align. shifts: {len_shifts}, dims: {len_dims}"
+            )
+        assert len_dims > 1
```
ngimel, line 506: nit: this is not needed, previous checks take care of len_dims==0 case

torchinductor/lowering.py
ngimel: I guess we could have a shortcut for zero-element tensor to return itself.

Issue #716, opened 2022-08-06
Large spike in peak memory with different backends


anijain2305:
Following is the peak memory footprint for huggingface models. I ran few warmups and then measured peak memory (after resetting it earlier).

baseline = usual way (w/o any TorchDynamo)
eager = TorchDynamo + eager (to check if Dynamo alone causes spikes)
aot_nop = TorchDynamo + AOT Autograd + eager compiler (or no-op compiler)
aot_nvfuser = TorchDynamo + AOT Autograd + ts-nvfuser

<img width="836" alt="image" src="https://user-images.githubusercontent.com/13822661/183237220-71b466e3-8eb7-42ab-9cde-a5227573d01a.png">

The peak memory footprint creates bad UX, as the batch size that fits in GPU with eager, might not be able to fit anymore during the Dynamo + backend tracing/optimization.

cc @Chillee for the issues for AOTAutograd
cc @bdhirsh for similar issue that I saw earlier with functionalization (here functionalization is turned off)

WIP
* PR to regenerate the numbers
* Inductor numbers

soumith:
is aot_inductor going to be similar to aot_nop or aot_ts_nvfuser?

soumith:
whoops, just saw that the Inductor numbers are WIP -- just ignore my comment

PR #726, opened 2022-08-07
Memory measurement and Compilation Time

anijain2305:
Code for https://github.com/pytorch/torchdynamo/issues/716
~~~
python benchmarks/huggingface.py --float32 -dcuda --training --use-eval-mode --profile-backend=pytorch --isolate --output=hf_pytorch.csv |& tee log1.txt
python benchmarks/huggingface.py --float32 -dcuda --training --use-eval-mode --profile-backend=eager --isolate --output=hf_eager.csv |& tee log2.txt
python benchmarks/huggingface.py --float32 -dcuda --training --use-eval-mode --profile-backend=aot_nop  --isolate --output=hf_aot_nop.csv |& tee log3.txt
python benchmarks/huggingface.py --float32 -dcuda --training --use-eval-mode --profile-backend=aot_nvfuser --isolate --output=hf_aot_nvfuser.csv |& tee log4.txt
python benchmarks/huggingface.py --float32 -dcuda --training --use-eval-mode --profile-backend=inductor_cudagraphs --isolate --output=hf_inductor_cudagraphs.csv |& tee log5.txt
~~~

### DIFF
--- a/benchmarks/common.py
+++ b/benchmarks/common.py
@@ -653,6 +653,61 @@ def speedup_experiment_trt(args, model_iter_fn, model, example_inputs):
     )
 
 
+def read_batch_size_from_file(args, filename, model_name):
+    batch_size = None
+    if os.path.exists("benchmarks"):
+        filename = os.path.join("benchmarks", filename)
+    assert os.path.exists(filename), filename
+    with open(filename, "r") as f:
+        lines = f.readlines()
+        lines = [i.split(",") for i in lines if len(i.strip()) > 0]
+        for val in lines:
+            cur_name, b = val
+            if model_name == cur_name:
+                batch_size = int(b)
+    if batch_size is None:
+        raise RuntimeError(
+            f"Batch size could not be found for {model_name} in {args.batch_size_file}"
+        )
+    print(f"batch size: {batch_size}")
+    return batch_size
+
+
+def compilation_profiling_experiment(
+    model_iter_fn, model, example_inputs, backend="pytorch"
+):
+    # Get the context
+    if backend == "pytorch":
+        ctx = NullContext()
+    else:
+        ctx = torchdynamo.optimize(backend)
+
+    def get_peak_memory():
+        return torch.cuda.max_memory_allocated() / 10**9
+
+    try:
+        # Reset and warmup
+        torchdynamo.reset()
+        torch.cuda.empty_cache()
+        t0 = time.perf_counter()
+        with ctx:
+            model_iter_fn(model, example_inputs)
+            model_iter_fn(model, example_inputs)
+            model_iter_fn(model, example_inputs)
+        t1 = time.perf_counter()
+        compilation_latency = t1 - t0
+
+        # Measure memory
+        torch.cuda.reset_peak_memory_stats()
+        with ctx:
+            model_iter_fn(model, example_inputs)
+        peak_memory = get_peak_memory()
+    except Exception:
+        compilation_latency = 0
+        peak_memory = 0
+    return compilation_latency, peak_memory
+
+
 def null_experiment(args, model_iter_fn, model, example_inputs):
     """
     A no-op experiment useful for making sure TorchBenchark alone works properly.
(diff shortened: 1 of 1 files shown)


benchmarks/common.py
```diff
@@ -792,6 +792,71 @@ def decay_batch_exp(self, batch_size, factor=0.5, divisor=2):
             out_batch_size = batch_size - 1
         return max(0, int(out_batch_size))
 
+    def measure_latency_and_memory(self, device, model_name, model_iter_fn, backends):
```
voznesenskym, line 795: Minor, but we should make it clear in the API that this is GPU memory, it could easily get conflated/confused/added CPU memory.

benchmarks/common.py
```diff
@@ -792,6 +792,71 @@ def decay_batch_exp(self, batch_size, factor=0.5, divisor=2):
...
+                    ctx = torchdynamo.optimize(backend)
+
+                # Reset and warmup
+                torchdynamo.reset()
+                torch.cuda.empty_cache()
+                t0 = time.perf_counter()
+                with ctx:
+                    model_iter_fn(model, example_inputs)
+                    model_iter_fn(model, example_inputs)
+                    model_iter_fn(model, example_inputs)
```
voznesenskym, line 817: Why do we run thrice between t0 and t1?

benchmarks/common.py
```diff
@@ -792,6 +792,71 @@ def decay_batch_exp(self, batch_size, factor=0.5, divisor=2):
             out_batch_size = batch_size - 1
         return max(0, int(out_batch_size))
 
+    def measure_latency_and_memory(self, device, model_name, model_iter_fn, backends):
+        get_peak_memory = lambda: torch.cuda.max_memory_allocated() / 10**9
```
voznesenskym, line 796: Device is passed into `measure_latency_and_memory` but not `max_memory_allocated` which could lead to confusion. Is this intentional?

benchmarks/common.py
```diff
@@ -792,6 +792,71 @@ def decay_batch_exp(self, batch_size, factor=0.5, divisor=2):
...
+                model_name,
+                self._args.training,
+                self._args.use_eval_mode,
+            )
+        except NotImplementedError:
+            logging.warn(f"{model_name} failed to load")
+
+        # Collect the compile and memory latency for backend
+        for backend in backends:
+            mem_experiment(backend)
```
voznesenskym, line 846: I don't know CUDA memory well, but retreating back to CPU memory / RAM, I would consider giving each backend its own process. What would happen here / how would our numbers look if a backend leaked?

benchmarks/common.py
```diff
@@ -792,6 +792,71 @@ def decay_batch_exp(self, batch_size, factor=0.5, divisor=2):
...
+                    model_iter_fn(model, example_inputs)
+                peak_memory = get_peak_memory()
+            except:
+                latency = 0
+                peak_memory = 0
+
+            latencies.append(latency)
+            mems.append(peak_memory)
+
+        assert device == "cuda"
```
voznesenskym, line 833: nit: Move precondition way to the top of the func?

benchmarks/common.py
anijain2305: Yeah, thats good point. Its relevant to GPU memory as well. Will run each backend in isolation.

benchmarks/common.py
anijain2305: Nah, I am just bad at naming :)

Will try to name things better.

benchmarks/common.py
anijain2305: This is just warmup. t0 and t1 are there to measure compilation cost. Torchscript can take a couple of iterations before it stop recompiling. 

This cost is approximation for compilation latency. It is approximation because it includes model iteration time as well. However, model iteration time typically is in order of ms. Therefore, t1 - t0 is only for flagging really long compilation times (>10 seconds).

yushangdi:
the current code doesn't take `--batch-size-file` because it is run before the batch-size-file is processed. Maybe we want to  make `--peak-memory-for-backend` work with  `--batch-size-file` ?

anijain2305:
> the current code doesn't take `--batch-size-file` because it is run before the batch-size-file is processed. Maybe we want to make `--peak-memory-for-backend` work with `--batch-size-file` ?

Good catch! Yes, thats correct. Do you want to make take over this PR? You can do `git fetch` and then `git checkout peak-mem` in your torchdynamo repo, and you will have access to this PR.

anijain2305:
@voznesenskym @yushangdi Can you please review again?

voznesenskym (approved):

PR #728, opened 2022-08-07
[inductor] Small fixes for huggingface models

jansel:


ngimel (approved):

torchinductor/sizevars.py
```diff
@@ -11,6 +12,8 @@
 
 from .virtualized import V
 
+log = logging.getLogger(__name__)
```
desertfire, line 15: Used logging and removed for this PR?

jansel: Yeah, I think it is fine to leave this line for the future though.

desertfire (approved):

torchinductor/lowering.py
```diff
@@ -271,12 +271,26 @@ def _to_copy(
...
+    layout=None,
+):
     assert not memory_format, "TODO"
+    assert layout in (None, torch.strided)
     if isinstance(device_or_dtype, torch.dtype):
         return to_dtype(x, device_or_dtype)
     if isinstance(device_or_dtype, torch.device):
         return to_device(x, device_or_dtype)
+    if device is not None:
+        return to_device(x, device)
```
ngimel, line 291: this will be wrong if both device and dtype is specified?

jansel: Good catch, will update.

PR #736, opened 2022-08-08
[inductor] randn ops (cpp, triton)

voznesenskym:


torchinductor/codegen/cpp.py
```diff
@@ -82,7 +82,12 @@ def cpp_prefix():
             }
 
             float normalized_rand_cpu(uint32_t seed, uint32_t offset) {
-                return uint32_to_uniform_float(at::Philox4_32_10(seed, 0, offset)());
+                return uint32_to_uniform_float(at::Philox4_32(seed, 0, offset)());
```
voznesenskym, line 85: Note: `uint32_to_uniform_float` could use a dedup but it is private in `at::Philox4_32` as it does not belong in the contract of a random number engine. If we want to dedup, we can make `uint32_to_uniform_float` a top level free function in some aten utils, and then use that.

ngimel (approved):

PR #740, opened 2022-08-08
Readme - Remove the context manager and more information about backends

anijain2305:
cc @jansel @ezyang 

This is to remove the confusion that arises from the usage of Context manager.

Rendered - https://github.com/pytorch/torchdynamo/blob/remove-ctx-1/README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,26 +60,86 @@ python setup.py develop
 
 ## Usage Example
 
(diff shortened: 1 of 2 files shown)


jansel (approved):

ezyang:
thanks!

PR #742, opened 2022-08-08
[inductor] Create implicit fallbacks for unhandled ops

### LINKED ISSUE #722: torchinductor should default generating fallback for all custom ops
torchinductor has no reason to know about arbitrary custom ops (well, at least until we let people define lowerings for torchinductor custom ops). Thus, even for developers, the default of erroring on unsupported ops is wrong; we should just always fallback in these cases. Custom ops are also irritating to write fallback for because you have to do the "test if you can import" song and dance that we do for torchvision)

### LINKED ISSUE #724: Getting MissingOperator one-by-one is a pain
@jansel please add your "automatically add fallback and keep going" mode lol

jansel:
Fixes #722 
Fixes #724


ezyang (approved):

torchinductor/lowering.py
```diff
@@ -783,6 +783,8 @@ def inner_fn(index):
 # TODO(jansel): we should implement decomps for these
 # https://github.com/pytorch/torchdynamo/issues/327
 make_fallback(aten._adaptive_avg_pool2d_backward)
+make_fallback(aten.argmax)
```
desertfire, line 786: @jansel , is there any specific reason you included argmax here, given that I have added argmax a while ago?

jansel: No.  I think this was a mistake, feel free to remove it.

PR #364, opened 2022-06-09
autotune conv kernels

pyjhzwh:
In torchinductor/ir.py, code_gen() generate `tuned_conv` which will choose candidate kernels(aten.convolution, triton_ops.conv, triton_ops.conv1x1) according to benchmarking results. Store best kernels in global cache with key as ("conv", layer params, input strides, etc.) to avoid re-benchmarking next time.
Benchmark:
`torchinductor/triton_ops/bench/bench_autotune_conv.py`
Test:
`pytest tests/test_torchinductor.py::GpuTests::test_conv_autotune_cuda`

Currently, the tuned_conv overhead is noticeable. Any idea how to reduce the overhead?
And any suggestions for better code style for Autotuner or tuned_conv to make it more generalizable?

PR #745, opened 2022-08-08
Autotune framework for conv and mm

pyjhzwh:
Similar to #364 
But decide best kernels at compile time because fusion is done at runtime. Use `triton.testing.do_bench` to benchmarking on given input shapes/ layer parameters and choose the best kernels with smallest execution time.

### DIFF
--- a/torchinductor/codegen/autotuner.py
+++ b/torchinductor/codegen/autotuner.py
@@ -207,9 +207,9 @@ def tuned_mm(
         # bench_end = time.time()
         # bench_time = bench_end - bench_start
         autotune.cache[key] = builtins.min(timings, key=timings.get)
-        # if torchinductor.config.debug:
-        print("for key = ", key)
-        print("timing", timings)
-        print("best_kernel", autotune.cache[key])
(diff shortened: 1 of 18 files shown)


torchinductor/codegen/autotuner.py
```diff
@@ -58,50 +92,124 @@ def tuned_conv(
...
+                run_kwargs = {"out": c}
+            timing, _, _ = autotune._bench(runnable_kernel, *run_args, **run_kwargs)
+            if "triton_ops" in kernel:
+                timing = timing * adjust_triton
+            timings[kernel] = timing
+        # bench_end = time.time()
+        # bench_time = bench_end - bench_start
+        autotune.cache[key] = builtins.min(timings, key=timings.get)
+        # if torchinductor.config.debug:
+        print("for key = ", key)
```
jansel, line 211: remove prints

jansel (approved):

torchinductor/codegen/autotuner.py
pyjhzwh: Removed

PR #751, opened 2022-08-09
[WIP] Tried supporting reduction + pw prefusion

Chillee:
lots of question marks here.


Chillee:
Subsumed by https://github.com/pytorch/torchdynamo/pull/949

PR #757, opened 2022-08-09
Move aot_cudagraphs backend here

ezyang:
Previously it was in pytorch/pytorch but it depends on torchdynamo
code more closely, so this seems like the logical place.

Previously at https://github.com/pytorch/pytorch/pull/80566

### DIFF
--- a/tests/test_aot_cudagraphs.py
+++ b/tests/test_aot_cudagraphs.py
@@ -1,14 +1,10 @@
 # Owner(s): ["module: cuda graphs"]
 
 import functools
-import sys
 import unittest
 from unittest.mock import patch
 
-import functorch
 import torch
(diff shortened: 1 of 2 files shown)


jansel (approved):

anijain2305:
@ezyang Should we also add a `aot_nvfuser_cudagraphs` i.e. AOTAutograd + Ts-nvfuser with Cudagraphs

ezyang:
We have aot nvfuser, but I'm not sure how the cudagraphs integration with nvfuser works. @Chillee might know more, he said he hacked in inductor's cuda graphs support to nvfuser. Note that cuda graphs is antipattern for dynamic shapes, and also there is quite a bit of logic needed to get it right (which I think is being duped between this backend and inductors' cudagraphs support)

PR #763, opened 2022-08-10
Remove context manager - 2/n; support nn.Modules, run

anijain2305:
* Adds support for nn.Module objects
* In case of nesting of torchdynamo.optimize/run/disable etc, we recursively find the innnermost fn and apply the outermost Dynamo call on it.

### DIFF
--- a/torchdynamo/eval_frame.py
+++ b/torchdynamo/eval_frame.py
@@ -51,6 +51,14 @@ unset = object()
 compile_lock = threading.Lock()
 
 
+def find_unaltered_fn(fn):
+    unaltered_fn = fn
+    while hasattr(unaltered_fn, "_torchdyname_orig_callable"):
+        unaltered_fn = getattr(unaltered_fn, "_torchdynamo_orig_callable")
+        assert callable(unaltered_fn)
+    return unaltered_fn
+
+
 class _TorchDynamoContext:
     def __init__(
         self,
@@ -79,6 +87,7 @@ class _TorchDynamoContext:
         self.backend_ctx.__exit__(exc_type, exc_val, exc_tb)
 
     def __call__(self, fn):
+        fn = find_unaltered_fn(fn)
         # Optimize the forward method of torch.nn.Module object
         if isinstance(fn, torch.nn.Module):
             mod = fn
@@ -95,6 +104,9 @@ class _TorchDynamoContext:
                     return self.forward(*args, **kwargs)
 
             new_mod = Wrapper()
+            # Save the function pointer to find the original callable while nesting
+            # of decorators.
+            new_mod._torchdynamo_orig_callable = mod
             return new_mod
 
         assert callable(fn)
@@ -119,6 +131,10 @@ class _TorchDynamoContext:
             _fn._torchdynamo_disable = True
         else:
             _fn._torchdynamo_inline = fn
+
+        # Save the function pointer to find the original callable while nesting
+        # of decorators.
+        _fn._torchdynamo_orig_callable = fn
 
         # If the function is called with torchdynamo.optimize decorator, we
         # should prevent any type of skipping.
@@ -402,18 +418,16 @@ def optimize_assert(backend, guard_export_fn=None):
 def run(fn=None):
     """Don't do any dynamic compiles, just use prior optimizations"""
     if fn is not None:
+        fn = find_unaltered_fn(fn)
         assert callable(fn)
-        callable_fn = fn
-        # Check if the function is a result of torchdynamo.optimize decorator.
-        if inspect.getattr_static(fn, "_torchdynamo_inline", False):
-            callable_fn = fn._torchdynamo_inline
-        return RunOnlyContext()(callable_fn)
+        return RunOnlyContext()(fn)
     return RunOnlyContext()
 
 
 def disable(fn=None):
     """Decorator and context manager to disable TorchDynamo"""
     if fn is not None:
+        fn = find_unaltered_fn(fn)
         assert callable(fn)
         return DisableContext()(fn)
     return DisableContext()
@@ -426,6 +440,7 @@ def skip(fn=None):
     """
     if fn is None:
         return skip
+    fn = find_unaltered_fn(fn)
     assert callable(fn)
     skip_code(fn.__code__)
     fn._torchdynamo_disable = True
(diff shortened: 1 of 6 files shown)


torchdynamo/eval_frame.py
```diff
@@ -79,6 +87,28 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+                    return optimized_forward(*args, **kwargs)
+
+                def __call__(self, *args, **kwargs):
+                    return self.forward(*args, **kwargs)
+
+            new_mod = Wrapper()
+            # Save the function pointer to find the original callable while nesting
+            # of decorators.
+            new_mod._torchdynamo_orig_callable = mod
+            return new_mod
```
ezyang, line 110: I don't think this is necessary. If you are passed a bound method (e.g., `module.forward`), it is still associated with the original Module object, so you can wrap it and it will still end up operating on the original module. This is what Dmytro was doing in the VHD model, btw

anijain2305: This is for the case like this

`optimized_mod = torchdynamo.optimize("eager")(mod)`

In this case, we just want to optimized the `forward` method and for everything else, we want to re-direct to original module.

Do you mean something else? I am gonna look in VHD model to be sure.

ezyang: I don't think users should expect `optimize` to work this way. The thing expects a callable, you give a callable back, there isn't any extra structure you should expect.

I suppose we could have a `torchdynamo.optimize_module` in analogy to AOTAutograd's module-level equivalent, but I don't really see the point; dynamo has no need to see the module, it can figure it all out even if the module is a free variable to the function in question.

anijain2305: I see. I am ok with that. The question then is - what does UX looks like?

For functions, its good
```
@torchdynamo.optimize("eager")
def fn():
   .....

or
optimized_fn = torchdynamo.optimize("eager")(fn)
```

What about nn.Modules? The following seems sketchy (mutating the original mod).

```
mod.forward = torchdynamo.optimize("eager")(mod.forward)
```

Or do we say that wrap the mod call in another function and then optimize that function? Something like

```
@torchdynamo.optimize("eager")
def fn(....):
    return mod(...)
```

voznesenskym: `mod` here is callable too, you would just have to trace it as `mod(args)` and not `mod.forward(args)`. That's not the same, because iirc while the module callable does call forward, it also does more than that with hooks and such. 

> `mod.forward = torchdynamo.optimize("eager")(mod.forward)`

I also wanted to ask, as in both your cases you are taking out the `optimized_fn ` to use beyond the Dynamo context. When would you need to do that? Generally, if you want to take the result of `.optimize`, you would use `.export`

`mod.forward = torchdynamo.export(mod.forward, args)`

Which should work fine, I believe.

voznesenskym: Oh, I misunderstood, I think, you are not calling to trace, you are calling to store a dynamo wrapped function.

In that case, what is wrong with something like 

```
@torchdynamo.optimize("eager")
def fn(....):
    return mod(...)
``` ?

Comment (anijain2305) [reply torchdynamo/eval_frame.py:110]: Yes, I am calling to store a dynamo wrapper function. It will get traced later on the first invocation. Background is that we are trying to remove the TorchDynamo context manager, because it confuses users.

I am totally ok with the following wrapping. TorchDynamo expects a callable, and returns a callable. Simple and clear.

```
@torchdynamo.optimize("eager")
def fn(....):
    return mod(...)
```

Here, the expectation from user is to encapsulate the code that they want to be optimized in another function, and then call `torchdynamo.optimize` on that function. So, just confirming, if that's ok.

Comment (anijain2305) [reply torchdynamo/eval_frame.py:110]: cc @jansel

Comment (ezyang) [reply torchdynamo/eval_frame.py:110]: I just want to point out that the wordy version here is equivalent to `torchdynamo.optimize("eager")(mod)` (or, perhaps the eta expansion matters, in which case it's equivalent to `torchdynamo.optimize("eager")(lambda *args, **kwargs: mod(*args, **kwargs))`).

Comment (ezyang) [reply torchdynamo/eval_frame.py:115]: However, I have also noticed that this behaves differently to using the context manager, so it feels like there is some sort of latent bug somewhere around here. I have a repro in fbcode if that would help.

Comment (jansel) [reply torchdynamo/eval_frame.py:115]: The main argument I see forwarding `nn.Module()` methods is that is what `fx.symbolic_trace()` and `torch.jit.script()` do, so users of those systems might expect it.

**Comment by anijain2305** [comment]
The current decision is to support torchdynamo.optimize on `nn.Module` objects and return an object that optimizes `forward` and redirects everything else to the original module. The return type is not `nn.Module`.

@jansel Please take a look when you get a chance.

File: torchdynamo/eval_frame.py [hunk 2]
```diff
@@ -51,6 +51,19 @@ def nothing():
 compile_lock = threading.Lock()
 
 
+def innermost_fn(fn):
+    """
+    In case of nesting of _TorchDynamoContext calls, find the innermost
+    function. TorchDynamo caches on fn.__code__ object, so its necessary to find
```
ezyang, line 57: @jansel, does this imply that if a user generates a closure, all separate instances of the closure will share the same cache? That seems suboptimal... the problem is especially prominent here, where `innermost_fn` is seemingly necessary but IMO it shouldn't be...

anijain2305:
@jansel @ezyang Please review when you get a chance.

torchdynamo/eval_frame.py
jansel: Yes, multiple closures with the same `__code__` have the same cache.

jansel (approved):

torchdynamo/eval_frame.py
anijain2305: @jansel Ping if you missed this message.

jansel: Perhaps we should emit a warning/error if you call `torchdynamo.optimize()` with two different compilers without calling `torchdynamo.reset()`.

ezyang:
This breaks `PYTORCH_TEST_WITH_DYNAMO=1 python test/test_nn.py -k test_module_backward_global_hook_writeable` on PyTorch master. I'll file an issue for this regression. This blocks updating the dynamo hash.

PR #768, opened 2022-08-10
Delete python key mode

ezyang:
AOTAutograd subsumes this mode (in particular, if AOTAutograd
gets a graph with no requires grad, or with no_grad on, it should not bother to
wrap the function in a custom autograd.Function); so if you have feature
requests for this, send them to AOTAutograd.  If you need a cheap and
cheerful lowering, you can use make_fx().


Chillee:
There are some ... lingering issues, mainly related to in-place ops I believe.

ezyang:
yes well we should fix them...

Chillee:
I think this should fix them: https://github.com/pytorch/pytorch/pull/82520

But I know @voznesenskym is also using this file somehow?

jansel (approved):

voznesenskym:
> I think this should fix them: https://github.com/pytorch/pytorch/pull/82520
> 
> 
> 
> But I know @voznesenskym is also using this file somehow?

I was, but I worked around it.

anijain2305 (approved):

voznesenskym (approved):

ezyang:
This is blocked on https://github.com/pytorch/pytorch/pull/82520

ezyang:
This PR identifies some known soundness issues with AOTAutograd and mutation.

Chillee:
I've verified that this PR + https://github.com/pytorch/pytorch/pull/83361 pass torchinductor tests.

side note: One of my recent PRs (https://github.com/pytorch/pytorch/pull/83248) is causing failures due to moving compilation out of autograd.Function and into a context that potentially has grad set.

https://github.com/pytorch/pytorch/pull/83361 fixes it, but you can also fix it by patching in the changes in `aot_autograd.py` in the above PR (https://github.com/pytorch/pytorch/pull/83361/files#diff-49334b500179bfa6b103db889c6bb8fe6984f6baa5fe73cef6b8ea59bbc0775bR265).

cc: @jansel

voznesenskym:
Not having this committed puts our fbcode integration in a bad state. I can resolve it manually, but we should get this in.

This class is the last place where we import PythonTensor. Not noticed in dynamo, but fbcode evaluates imports differently and hits this assertion.

@ezyang if you dont have time, I can take over getting through the merge conflicts and getting it in.

Chillee:
@voznesenskym I was actually looking at this, but unfortunately, even after my landed PRs, it doesn't fix all tests (due to tracing in-place modifying inputs). I can take another look at landing it.

voznesenskym:
> @voznesenskym I was actually looking at this, but unfortunately, even after my landed PRs, it doesn't fix all tests (due to tracing in-place modifying inputs). I can take another look at landing it.

Ah rats - alright, no problem. LMK how I can help here. This is probably higher than the broadcast / tl.where bugs I am looking at cause its making internal integration a bit of a pain...

Chillee:
Gonna close this one in favor of a new one, as quite a bit has changed.

Chillee:
In favor of https://github.com/pytorch/torchdynamo/pull/939

Issue #762, opened 2022-08-10
[VDD] Dynamo can produce FX graph with the same argument multiple times


ezyang:
To reproduce:

```
jf get --update --version 164037553
buck2 run @mode/opt -c python.package_style=inplace //hpc/torchrec/models/feed/benchmark:vdd_benchmark -- --iters 20 --compile True --cudagraphs True --pad_seq_embs=true  --dynamo True
```

Sample logs: 

```
  File "/data/sandcastle/boxes/fbsource/buck-out/v2/gen/fbcode/d5479947763d4841/hpc/torchrec/models/feed/benchmark/__vdd_benchmark__/vdd_benchmark#link-tree/torch/fx/graph_module.py", line 77, in _forward_from_src
    _exec_with_source(src, globals_copy)
  File "/data/sandcastle/boxes/fbsource/buck-out/v2/gen/fbcode/d5479947763d4841/hpc/torchrec/models/feed/benchmark/__vdd_benchmark__/vdd_benchmark#link-tree/torch/fx/graph_module.py", line 71, in _exec_with_source
    exec(compile(src, key, 'exec'), globals)
  File "<eval_with_key>.1231", line 4
SyntaxError: duplicate argument 'unary_features_jt_dict_POSITION_ID_values' in function definition
```

full logs: https://www.internalfb.com/intern/paste/P521398402/

If there are more logs you would like to see I encourage you to add them to rage mode ;)

ezyang:
@anijain2305 has the repro and is looking into it

anijain2305:
Finally a repro

~~~
import torchdynamo
import torch

def my_compiler(gm, args):
    print(gm)
    return gm

@torchdynamo.optimize(my_compiler)
def fn(a, b, d):
    x = torch.sin(a)
    keys = list(d._jt_dict.keys())
    print("Graph break")

    p = torch.cos(d._jt_dict[keys[0]]._value)
    q = torch.sin(b)
    r = torch.sigmoid(d._jt_dict[keys[0]]._value)
    return p + q + r

class Value:
    def __init__(self):
        self._value = torch.randn(4)

class Sample:
    def __init__(self):
        self._jt_dict = {}
        self._jt_dict["POSITION_ID"] = Value()

fn(torch.randn(4), torch.randn(4), Sample())
~~~

PR #772, opened 2022-08-11
Explicitly set placeholder name while creating FxGraph node

anijain2305:
Fixes https://github.com/pytorch/torchdynamo/issues/762

@ezyang

### DIFF
--- a/tests/test_repros.py
+++ b/tests/test_repros.py
@@ -1437,3 +1437,30 @@ class ReproTests(torchdynamo.testing.TestCase):
             m.forward(listy)
 
         self.assertEqual(cnt.frame_count, 1)
+
+    def test_vdd_duplicate_error(self):
+        def fn(a, dt):
+            keys = list(dt._jt_dict.keys())
+            p = torch.cos(dt._jt_dict[keys[0]]._value)
+            q = torch.sin(a)
+            r = torch.sigmoid(dt._jt_dict[keys[0]]._value)
+            return p + q + r
+
+        class Value:
+            def __init__(self):
+                self._value = torch.randn(4)
+
+        class Sample:
+            def __init__(self):
+                self._jt_dict = {}
+                self._jt_dict["POSITION_ID"] = Value()
+
+        a = torch.randn(4)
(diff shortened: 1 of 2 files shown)


voznesenskym (approved):

ezyang:
I'd love to know why this works. Is it because we were putting illegal characters in the name before?

voznesenskym:
> I'd love to know why this works. Is it because we were putting illegal characters in the name before?

I think its cause of the line(s) below:

```
 used_names = {n.name for n in placeholders}
        if name in used_names:
            for i in itertools.count():
                if f"{name}_{i}" not in used_names:
                    name = f"{name}_{i}"
                    break
```

anijain2305:
Yes, @voznesenskym has it right. Basically, the name that we send to Fx `create_node` API is different from the name that's actually assigned to the placeholder node. So, our unique name generation logic is not able to handle this difference.

ezyang:
Tracer.create_proxy calls create_node calls Graph.create_node, which says

```
            name (Optional[str]): an optional string name for the ``Node``.
                This will influence the name of the value assigned to in the
                Python generated code.
```

in other words, `create_node` is supposed to guarantee that the generated name is unique. If this is not working, there is some deeper problem with FX itself or our use of the API.

anijain2305:
> Tracer.create_proxy calls create_node calls Graph.create_node, which says
> 
> ```
>             name (Optional[str]): an optional string name for the ``Node``.
>                 This will influence the name of the value assigned to in the
>                 Python generated code.
> ```
> 
> in other words, `create_node` is supposed to guarantee that the generated name is unique. If this is not working, there is some deeper problem with FX itself or our use of the API.

Thanks for encouraging me to look into this. We were not passing name to `create_node`. Instead, it was relying on `target` to come up with a candidate  here - https://github.com/pytorch/pytorch/blob/master/torch/fx/graph.py#L728

After you suggestion, I explicitly setup `name` arg. I checked the new fix on VDD as well, and its passing.

ezyang (approved):

PR #778, opened 2022-08-11
[inductor] New tiling heuristic

jansel:


torchinductor/codegen/triton.py
```diff
@@ -1135,45 +1155,81 @@ def select_tiling(node_schedule, numel, reduction_numel):
...
         for node in EnableReduction.filter(node_schedule):
-            tiled_groups = TritonScheduling.select_node_tiling(node)
-            if tiled_groups:
-                candidate_tiles[tiled_groups] += 1
-
-        # TODO(jansel): join two 2D tiles into a 3D tile
-        # TODO(jansel): add a cost function for tiling instead of most_common
-        for tiled_groups, count in candidate_tiles.most_common():
+            for tiling in cls.candidate_tilings(node):
+                if tiling.name in seen_names:
```
ngimel, line 1162: nit: if you send seen_names as an arg to candidate_tilings, you could bail out earlier for deps you've already seen

ngimel (approved):
This looks good, scoring according to the number of distinct memory locations will automatically prioritize non-broadcasted tensors, which is what we want.

torchinductor/codegen/triton.py
jansel: This would defeat the `lru_cache` on `candidate_tilings`.  We end up calling it multiple times due to `is_better_tiling_available()`.

PR #781, opened 2022-08-11
allow_tff32=True for triton_ops.matmul

pyjhzwh:
allow_tff32=True for triton_ops.matmul
Otherwise, triton_ops.mamul_out is way slower than aten.mm.out in torchinductor. Because in torchinductor, `
torch.backends.cuda.matmul.allow_tf32 = True
torch.backends.cudnn.allow_tf32 = True`

### DIFF
--- a/torchinductor/triton_ops/matmul.py
+++ b/torchinductor/triton_ops/matmul.py
@@ -58,7 +58,7 @@ def _kernel(
         else:
             a = tl.load(A, mask=rk[None, :] < k, other=0.0)
             b = tl.load(B, mask=rk[:, None] < k, other=0.0)
-        acc += tl.dot(a, b, allow_tf32=False)
+        acc += tl.dot(a, b)
         A += BLOCK_K * SPLIT_K * stride_ak
         B += BLOCK_K * SPLIT_K * stride_bk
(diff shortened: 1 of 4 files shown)


torchinductor/triton_ops/matmul.py
```diff
@@ -58,7 +58,7 @@ def _kernel(
         else:
             a = tl.load(A, mask=rk[None, :] < k, other=0.0)
             b = tl.load(B, mask=rk[:, None] < k, other=0.0)
-        acc += tl.dot(a, b, allow_tf32=False)
+        acc += tl.dot(a, b)
```
ngimel, line 61: can you add allow_tf32 argument to the kernel, it should be set by querying current pytorch global precision setting, so that comparisons are always fair? 
Same should be done for template that's used to generate matmul/conv.

ngimel (approved):
Approving to unblock. This should be an option to the template so that precision of the generated kernel matches the global precision set in pytorch.

torchinductor/triton_ops/matmul.py
pyjhzwh: Sure. Thanks for the suggestions. I will do that in later PR

PR #782, opened 2022-08-11
add allow_tf32 option in torchinductor.triton_ops.matmul

pyjhzwh:
add allow_tf32 option in torchinductor.triton_ops.matmul, allow_tf=torch.backends.cuda.matmul.allow_tf32
Do not guard_static_shape in strides otherwise canonicalize of ExternKernel buffer Dependency won't match the following fusable nodes

### DIFF
--- a/torchinductor/triton_ops/matmul.py
+++ b/torchinductor/triton_ops/matmul.py
@@ -58,7 +58,7 @@ def _kernel(
         else:
             a = tl.load(A, mask=rk[None, :] < k, other=0.0)
             b = tl.load(B, mask=rk[:, None] < k, other=0.0)
-        acc += tl.dot(a, b, allow_tf32=False)
+        acc += tl.dot(a, b)
         A += BLOCK_K * SPLIT_K * stride_ak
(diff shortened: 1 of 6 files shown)


torchinductor/triton_ops/matmul.py
```diff
@@ -121,13 +122,14 @@ def grid(META):
...
             c.stride(1),
+            allow_tf32=allow_tf32,
             GROUP_M=8,
             ACC_TYPE=ACC_TYPE,
         )
 
     @staticmethod
-    def forward(a, b, out):
-        return _matmul_out._call(a, b, out)
+    def forward(a, b, out, allow_tf32=torch.backends.cuda.matmul.allow_tf32):
```
jansel, line 131: This will read the value at `import` time, not at runtime.   Default values get evaluated when a function is defined.

jansel (changes requested):

torchinductor/triton_ops/matmul.py
pyjhzwh: That makes sense. set default as True

jansel (approved):

Issue #327, opened 2022-06-06
TorchInductor missing ops tracker


jansel:
The following ops are using `ir.FallbackKernel`, but we should rewrite them to us decomps:
- [ ] aten._embedding_bag
- [ ] aten.nll_loss_forward
- [ ] torchvision.roi_align
- [ ] aten.grid_sampler_2d (could use the version [from here](https://github.com/pytorch/pytorch/issues/34704)
- [ ] aten.upsample_bilinear2d
- [ ] aten._cudnn_rnn (might need to hold of on this if perf doesn't match)

The following ops are needed for training support, note this is just the first failure from each benchmark so there could be more after this:
- [ ] aten._adaptive_avg_pool2d_backward
- [ ] aten.avg_pool2d_backward
- [ ] aten.expand_as
- [ ] aten._fused_moving_avg_obs_fq_helper
- [ ] aten.glu_backward
- [ ] aten.masked_fill_
- [ ] aten.max_pool2d_with_indices_backward
- [ ] aten.native_batch_norm_backward
- [ ] aten.reflection_pad2d_backward
- [ ] aten.select_scatter
- [ ] aten.convolution_backward (might need to hold of on this if perf doesn't match)

Chillee:
> aten.convolution_backward

Unfortunately, there's no good way of decomposing this today. The gradient (wrt the weights iirc?) cannot be expressed with a forwards convolution.

> aten.nll_loss_forward

This one is WIP: https://github.com/pytorch/pytorch/pull/78491

ezyang:
> Unfortunately, there's no good way of decomposing this today. The gradient (wrt the weights iirc?) cannot be expressed with a forwards convolution.

Well, to be more precise, if we have a suitably generalized convolution that has output shape and transposition that would be sufficient for all convolutions forwards/backwards.

soumith:
> The gradient (wrt the weights iirc?) cannot be expressed with a forwards convolution.

It can. Where you have to pad the output image appropriately, and the convolve it with a transposed kernel weight matrix.

See the last diagram in http://soumith.ch/ex/pages/2014/08/07/why-rotate-weights-convolution-gradient/

Chillee:
> aten.log1p

btw, this is currently a prim in Primtorch. So there won't be a further decomposition for it.

jansel:
> > aten.expand_as
> 
> Also, this is a CompositeImplicitAutograd op - where are you seeing this today?

[...]

PR #783, opened 2022-08-11
[inductor] Lowering aten.reflection_pad2d_backward

sangongs:
This PR adds lowering for `aten.reflection_pad2d_backward` as suggested in issue #327.

### DIFF
--- a/torchinductor/lowering.py
+++ b/torchinductor/lowering.py
@@ -1838,16 +1838,19 @@ def reflection_pad2d_backward(grad_output, x, padding):
             i = ops.index_expr(i, torch.int32)
             return ops.and_(ops.ge(i, lb), ops.le(i, ub))
 
-        def accumulate(out_x, out_y, *index_ranges):
+        def accumulate(out_x, out_y, index_range1, index_range2=None):
             nonlocal grad
-            cond = index_range_condition(index_ranges[0])
-            for r in index_ranges[1:]:
-                cond = ops.and_(cond, index_range_condition(r))
-            g = ops.where(
-                cond,
-                load_from_output(out_x, out_y),
-                ops.constant(0.0, grad_output.get_dtype()),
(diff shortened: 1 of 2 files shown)


torchinductor/lowering.py
```diff
@@ -1816,6 +1815,78 @@ def fn(idx):
...
+            i, lb, ub = index_range
+            i = ops.index_expr(i, torch.int32)
+            return ops.and_(ops.ge(i, lb), ops.le(i, ub))
+
+        def accumulate(out_x, out_y, *index_ranges):
+            nonlocal grad
+            cond = index_range_condition(index_ranges[0])
+            for r in index_ranges[1:]:
+                cond = ops.and_(cond, index_range_condition(r))
+            g = ops.where(
```
jansel, line 1846: This might be better as 
```suggestion
            g = ops.masked(cond, lambda: load_from_output(out_x, out_y), 0.0)
```

Which won't actually do the read unless we are at the boundary.

sangongs: Thank you for catching this!

torchinductor/lowering.py
```diff
@@ -1838,16 +1838,16 @@ def index_range_condition(index_range):
...
-            cond = index_range_condition(index_ranges[0])
-            for r in index_ranges[1:]:
-                cond = ops.and_(cond, index_range_condition(r))
-            g = ops.where(
-                cond,
-                load_from_output(out_x, out_y),
-                ops.constant(0.0, grad_output.get_dtype()),
-            )
+            if index_range1[2] < index_range1[1]:
+                return
```
sangongs, line 1844: If the upper bound is less than the lower bound, we can get rid of one accumulation. This happens when the padding size is zero.

jansel (approved):
Thanks!

PR #785, opened 2022-08-11
Set up per-operator input database, per-operator microbenchmarking

eellison:
Introduce a mode for benchmarking that will run models and serialize operators and their frequency that is toggled with `--log-operator-inputs`. Example usage:

`python benchmarks/runner.py --suites=torchbench --training --dtypes=float16 --output=/scratch/eellison/work/torchdynamo/benchmarks/bench_logs/torchbench_train/ --log-operator-inputs`

The outputs for torchbench, timm, and huggingface have been included in this PR as `.zip` files. 

Here are operators and call count for [torchbench](https://gist.github.com/eellison/47a1a6b40f7d931bed6bd26e405d0423),  [huggingface](https://gist.github.com/eellison/c24495e69837be3ef5126159d38b2a0d), and [timm](https://gist.github.com/eellison/613441515a8bd6c8c945d75aa8375ce7).

Also introduces a microbench script to compare operators to eager/nvfuser:

Example usage (just running single input for now):
`python ./benchmarks/microbenchmarks/operatorbench.py --op=aten.avg_pool2d.default --dtype=float16 --suite=timm`

> INFO torchinductor.scheduler: RUN buf0
INFO torchinductor.codegen.triton: codegen numel=s0*s1*IndexingDiv(s2, 2)**2 reduction_numel=1 nodes=1
INFO torchinductor.scheduler: NEW KERNEL
Perf for aten.avg_pool2d.default torch.float16 w/cudagraphs
JIT NVFuser speedup over aten 1.0
Inductor speedup over aten 1.1039353900823776

Follow-ups: do sweep on operators we are slow on, prioritize [lowerings](https://github.com/pytorch/torchdynamo/issues/327) which are invoked more frequently.


benchmarks/common.py
```diff
@@ -1616,6 +1631,44 @@ def main(runner, original_dir=None):
         print_summary(output_filename)
 
 
+def log_operator_inputs(model, example_inputs, model_iter_fn, name, args):
+    output_split = args.output.split("/")
```
jansel, line 1635: Use [os.path](https://docs.python.org/3/library/os.path.html) for filename manipulation.

benchmarks/common.py
```diff
@@ -1616,6 +1631,44 @@ def main(runner, original_dir=None):
...
+            model_iter_fn(model_fake, example_inputs_fake, collect_outputs=False)
+    except Exception as e:
+        print(f"{name} failed to run with fake tensors, trying real. Exception: {e}")
+        operator_mode = OperatorInputsMode()
+        try:
+            with operator_mode:
+                model_iter_fn(model, example_inputs, collect_outputs=False)
+        except Exception as e2:
+            print(f"{name} failed to run with real. Exception: {e2}")
+            raise e2
```
jansel, line 1666: re-raise exception
```suggestion
            raise
```

benchmarks/microbenchmarks/operator_inp_utils.py
```diff
@@ -0,0 +1,189 @@
...
+from typing import Generator
+from typing import Iterable
+from typing import Tuple
+
+import torch
+from torch.utils._python_dispatch import TorchDispatchMode
+from torch.utils._pytree import tree_flatten
+from torch.utils._pytree import tree_map
+
+OP_INP_DIRECTORY = os.path.dirname(__file__) + "/operator_inp_logs/"
```
jansel, line 20: `os.path.join()`

benchmarks/timm_models.py
```diff
@@ -275,7 +275,8 @@ def _gen_target(self, batch_size, device):
         )
 
     def compute_loss(self, pred):
-        return self.loss(pred, self.target)
+        # calling lift so modes enabled for forward/backward can handle self.target
+        return self.loss(pred, torch.ops.aten.lift_fresh_copy(self.target))
```
jansel, line 279: Will this effect performance measurements?

voznesenskym:
Two questions:

1) Why are we committing .zip files?

2) If we need them for processing/benchmarking - how did we generate them? all the .zip code I see looks like loading code.

ezyang:
I didn't really read the PR, but you could also unzip the zip before checking it in lol

eellison:
> Thought about this a little more - instead of storing the .zip files, can we regenerate the contents of the .zip from the runner each time?

It takes way too long to generate the inputs, even with fake tensor, for this to really make sense. You don't want to have to wait multiple minutes every time you want to run a script that tests the performance of changing an operator lowering with recorded inputs. 

> Storing the .zip in a git repo makes it a bit of a hard to deal with black box (can't diff by line)

I'm not sure this is really an issue, since no one is going to be line by line comparing the 6 megabytes of operator inputs from TIMM before and after some change. 

>  but you could also unzip the zip before checking it in lol

This would be over 10MB for the three files - seems kind of wasteful when the whole repro size is ~3 MB ( as opposed to ~.4 MB compressed)

I don't know if anyone else has any strong thoughts here - the changes talked about here are pretty minimal so we can always land and make changes as we use this more..

benchmarks/microbenchmarks/operator_inp_utils.py
```diff
@@ -0,0 +1,189 @@
...
+    return False
+
+
+def skip_args(elems):
+    for i in tree_flatten(elems)[0]:
+        # only shows up in constructors and ops like that
+        if isinstance(i, (torch.memory_format, torch.storage.UntypedStorage)):
+            return True
+        # TODO: serialize/deserialize sparse arguments
+        if isinstance(i, torch.Tensor) and i.is_sparse:
```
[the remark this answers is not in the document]
ezyang: Well, given that you wrote the json format yourself, pretty easy. You'll need to say how many sparse and dense dims, and maybe nnz and coalesced if you want to get frisky

ezyang:
The gists are too big for github to load 😂

ezyang:
Elias, something that's not clear from the PR description: the input database is *metadata* only, right? If so, I think we should design a compact text format for describing this sort of metadata; something like Python code you could eval() to inflate the tensors would be a pretty good start. Then we should feel pretty comfortable with checking these in as plaintext; they're basically like OpInfo sample inputs but machine generated.

benchmarks/microbenchmarks/operator_inp_logs/hf_train/AlbertForMaskedLM_training.txt
```diff
@@ -0,0 +1,115 @@
+Operator: aten._log_softmax.default
+cnt: 1, '((T([2048, 30000], th.float16), 1, False), {})'
```
ezyang, line 2: Since I'm not going to get a chance to bikeshed this later, I'm going to bikeshed this some more. How about...

```
F(T(2048, 30000, f16), 1, False)  # count: 1
```

Main changes:
* Use actual function call syntax, so you don't have to waste characters on arg tuple and kwarg dict. If you like, we can do it in a wordy way aka aten._log_softmax.default(T(2048, 30000, f16), 1, False). One benefit for the wordy format is you only one test per line, rather than having to deal with the hierarchical structure you currently have
* Support omitting the list/tuple in the tensor constructor, like our real constructors, saves two characters
* Use the shortened dtype names in https://github.com/pytorch/pytorch/blob/71141c30232d436ad61d2931af8808e4451f8cde/test/test_meta.py#L51-L65 They're used in Rust and I'm fond of the shortening
```

File: benchmarks/microbenchmarks/operator_inp_logs/hf_train/AlbertForMaskedLM_training.txt [hunk 7]
```diff
@@ -0,0 +1,115 @@
...
+Operator: aten._log_softmax_backward_data.default
+cnt: 1, '((T([2048, 30000], th.float16), T([2048, 30000], th.float16), 1, th.float16), {})'
+Operator: aten._softmax.default
+cnt: 12, '((T([4, 64, 512, 512], th.float16), -1, False), {})'
+Operator: aten._softmax_backward_data.default
+cnt: 12, '((T([4, 64, 512, 512], th.float16), T([4, 64, 512, 512], th.float16), -1, th.float16), {})'
+Operator: aten._to_copy.default
+cnt: 1, "((T([4, 1, 1, 512], th.float32),), {'dtype': th.float16})"
+Operator: aten._unsafe_view.default
+cnt: 36, '((T([4, 64, 512, 64], th.float16), [256, 512, 64]), {})'
```
ezyang, line 12: and then when there are strides, I'd suggest we spend the characters calling it out explicitly: `stride=[256, 512, 64]`

ezyang:
It would be good to get PR feedback from the folks who would also be using the microbenchmarks.

jansel:
It would be useful to see some data generate from this.  What ops are we the slowest on?

It also might make sense to filter out view ops.  Views should be represented in the strides of inputs to other ops.

benchmarks/microbenchmarks/operatorbench.py
```diff
@@ -0,0 +1,88 @@
...
+
+def microbenchmark(target, args, kwargs, dtype):
+    gm, gm_args = gen_gm_and_inputs(target, args, kwargs)
+    compiled_fn = compile_fx(gm, gm_args)
+    cudagraphs_eager = cudagraphs_inner(gm, gm_args, copy_outputs=False)
+    cudagraphs_jit = cudagraphs_inner(
+        torch.jit.trace(gm, gm_args), gm_args, copy_outputs=False
+    )
+
+    repeats = 3
```
desertfire, line 42: Can we just have a fast correctness checking mode? It will improve our improve test coverage, and may also help us to identify if any of those model accuracy error is real.

Chillee:
Might get useful to get an operator count minus the ones we're already decomposing.

benchmarks/microbenchmarks/operator_inp_logs/hf_train/AlbertForMaskedLM_training.txt
eellison: Done

benchmarks/microbenchmarks/operator_inp_logs/hf_train/AlbertForMaskedLM_training.txt
eellison: Used shortened dtype names..

eellison:
> It would be useful to see some data generate from this. What ops are we the slowest on?

Just opened https://github.com/pytorch/torchdynamo/issues/922 and https://github.com/pytorch/torchdynamo/issues/921 (still need to benchmark torchbench ops).

> It also might make sense to filter out view ops

Yea I filter those out and constructors in `non_compute_operator`.

benchmarks/microbenchmarks/operatorbench.py
eellison: Will do as follow up, when I looked into this previously I think there are are a lot of nan-handling errors, which I think the details of are a little bit tricky what pytorch guarantees (will do non-empty input generation to avoid nan errors).

ngimel:
For benchmarking we'd also need strides, not just sizes - different strides can result in completely different perf

eellison:
@ngimel those are recorded - see the output

ngimel:
> @ngimel those are being recorded - see the output. when the tensors are contiguous we omit serializing the strides, otherwise we serialize them: `cnt: 12, ((T([128, 512, 512], f16, stride=(262144, 1, 512)), T([128, 512, 64], f16)), {})`

Cool, I've seen a few cases w/o strides so didn't notice that they were recorded when needed.

torchinductor/compile_fx.py
```diff
@@ -40,26 +41,11 @@ def call_function(self, target, args, kwargs):
...
-            else:
-                g_args.append(arg)
-        assert all(not isinstance(x, torch.Tensor) for x in kwargs.values())
-        node = g.call_function(target, tuple(g_args), kwargs)
-        if isinstance(expected, torch.Tensor):
-            node = (node,)
-        g.output(node)
-
-        gm = torch.fx.GraphModule({}, g)
+        gm, gm_inps = gen_gm_and_inputs(target, args, kwargs)
```
desertfire, line 44: CheckEachNode is called with python_key which is being removed. You need to update this if we want to re-generate the data in the future. I am ok with leaving it as is for this commit.

desertfire (approved):
LGTM. I already saw you filing issues found through this approach, so it would be valuable to have the PR in soon and work on follow-ups if needed.

PR #786, opened 2022-08-12
improve splits heuristics to handle outer reductions

ngimel:
This improves transformer models in torchbench
hf_Bert 1.05 -> 1.13x
timm_vision_transformer 1.09 -> 1.12x
hf_Bart 0.96 -> 1.06x

I'm not happy that autotuning configs leak into picking the split, but we can refactor it later (or, ideally, figure out how to make splits runtime and not compile time, that will be needed for dynamic shapes. Currently split reduction are done with static shapes).

### DIFF
--- a/benchmarks/torchbench_models_list.txt
+++ b/benchmarks/torchbench_models_list.txt
@@ -18,6 +18,7 @@ resnet50,128
 resnext50_32x4d,128
 shufflenet_v2_x1_0,512
 squeezenet1_1,512
+timm_nfnet,256
 timm_efficientnet,128
 timm_regnet,128
 timm_resnest,256
--- a/torchinductor/ir.py
+++ b/torchinductor/ir.py
@@ -382,6 +382,149 @@ class Reduction(Loops):
             self.reduction_type,
         )
 
+    @staticmethod
+    def num_splits(
+        device,
+        dtype,
+        inner_fn,
+        ranges,
+        reduction_ranges,
+        reduction_type,
+        reduction_numel,
+    ):
+        num_sm = torch.cuda.get_device_properties(device).multi_processor_count
+        min_elements_per_thread = 32
+        max_elements_per_thread = 512
+        threads_per_sm = 2048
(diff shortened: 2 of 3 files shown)


jansel (approved):
Awesome! Perhaps we could write an autotuner for something like this in the future.

eellison:
Just curious - where is the equivalent reduction heuristics in eager ? And what values did you autotune for ?

Chillee:
Could you add a "before and after" example? (and then maybe add a link to that example in the code)

ngimel:
@eellison Somewhat similar logic is in Reduce.cuh (which is not really repeated verbatim here). The main idea is that each thread has to have an optimum number of items to work on (in eager it's 16-512), so we try to launch the grid that has maximum number of blocks where this conditions are still preserved. In practice that means that for small tensors each thread works on 16 items, and the number of blocks increases with the size of the tensor until the gpu is filled, then, in the medium regime, the number of work items per thread increases while the number of blocks stays constant, and finally, when device is full and each thread would have more than max items for really large tensors, we start increasing number of blocks again. 
There are also differences in how many blocks you are launching depending on whether this is an inner or outer reduction (for inner reduction roughly a threadblock works on a single output, for outer it has to be at least 16 outputs, turns out for some complicated reductions + broadcasts 128 is better)
Finally, eager is relatively good at handling uneven splits, while inductor generates a significantly more complicated indexing for uneven splits, resulting in worse kernels, so I do some accommodations to do even splits when possible. 
The heuristics now are pretty basic, mostly aimed at existing models to become better (and for standalone layer norm kernels for typical sizes to be reasonable), improvements will come later.
@chillee there is almost no difference in the generated code, except an additional kernel is added every time reduction is split to reduce partial values. (8192, 768).reduction(0) would be split into 64 pieces along 0th axis, and then those intermediate results would be combined by a second kernel. It's possible (and probably slightly better) to do it in a single kernel with atomic counters, and we'll get to it later (but that would probably make inductor reductions incompatible with pre-volta hardware, it looks like triton supports atomics for Volta+ only).

Issue #748, opened 2022-08-09
Investigate "skipping cudagraphs due to multple devices" issue


jansel:
```
./benchmarks/torchbench.py -dcuda --float16 --inductor --training -k hf_Bart
```

```
cuda train hf_Bart                            WARNING:root:using triton random, expect difference from eager
WARNING torchinductor.compile_fx: skipping cudagraphs due to multple devices
WARNING torchinductor.compile_fx: skipping cudagraphs due to multple devices
WARNING:root:using triton random, expect difference from eager
WARNING:root:using triton random, expect difference from eager
WARNING:root:using triton random, expect difference from eager
WARNING:root:using triton random, expect difference from eager
WARNING:root:using triton random, expect difference from eager
WARNING:root:using triton random, expect difference from eager
WARNING:root:using triton random, expect difference from eager
WARNING torchinductor.compile_fx: skipping cudagraphs due to multple devices
WARNING torchinductor.compile_fx: skipping cudagraphs due to multple devices
WARNING:root:using triton random, expect difference from eager
WARNING torchinductor.compile_fx: skipping cudagraphs due to multple devices
WARNING torchinductor.compile_fx: skipping cudagraphs due to multple devices
WARNING:root:using triton random, expect difference from eager
WARNING torchinductor.compile_fx: skipping cudagraphs due to multple devices
WARNING torchinductor.compile_fx: skipping cudagraphs due to multple devices
WARNING:root:using triton random, expect difference from eager
WARNING torchinductor.compile_fx: skipping cudagraphs due to multple devices
WARNING torchinductor.compile_fx: skipping cudagraphs due to multple devices
WARNING:root:using triton random, expect difference from eager
WARNING torchinductor.compile_fx: skipping cudagraphs due to multple devices
WARNING torchinductor.compile_fx: skipping cudagraphs due to multple devices
WARNING:root:using triton random, expect difference from eager
WARNING torchinductor.compile_fx: skipping cudagraphs due to multple devices
WARNING torchinductor.compile_fx: skipping cudagraphs due to multple devices
WARNING:root:using triton random, expect difference from eager
```

These warnings about multiple devices are a bit odd, and may be slowing things down by preventing use of cudagraphs.
1) If these issues aren't real (perhaps due to constants which we can move at compile tile), we should enable cudagraphs in these cases and make the multi-device check less sensitive.
[...]

PR #787, opened 2022-08-12
[inductor] Don't update device_types for optimized-away constants

jansel:
Fixes #748

PR #788, opened 2022-08-12
[inductor] Make factory ops create tensor on the same device as graph inputs

desertfire:
Summary: To solve the "skipping cudagraphs due to multple devices"
issue, see https://github.com/pytorch/torchdynamo/issues/748

### DIFF
--- a/torchinductor/config.py
+++ b/torchinductor/config.py
@@ -44,6 +44,9 @@ implicit_fallbacks = True
 # Enables a fusion pass that groups nodes together before the scheduler
 prefuse_nodes = True
 
+# Use the graph inputs' device (should be unique) as factory ops' output device
+use_graph_device_for_factory_ops = True
+
 
 # config specific to codegen/cpp.pp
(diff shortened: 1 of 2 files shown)


ngimel (approved):

jansel:
I don't think this is needed:
```py
def fn(x):
    return torch.arange(len(x), device="cpu").to(x.device)
```

Results in the output code:
```py
@pointwise_heuristics(size_hints=[32])
@triton.jit
def kernel0(out_ptr0, xnumel, XBLOCK : tl.constexpr):
    xmask = xindex < xnumel
    x0 = xindex
    tmp0 = x0
    tmp1 = tmp0.to(tl.int64)
    tl.store(out_ptr0 + x0, tmp1, xmask)
```

TorchInductor is smart enough to move the `arange()` at compile time.

desertfire:
Already solved in https://github.com/pytorch/torchdynamo/pull/787

PR #791, opened 2022-08-12
Add model training loss convergence benchmark

yanboliang:
This example follows how users write their own end to end model training/evaluation pipeline using _huggingface_. The loss function is in convergence w/o Dynamo:
```
Loss history : [1.638656200170517, 1.6167153787612916, 1.591713446378708, 1.5690117526054381, 1.5178297901153563, 1.4247627103328704, 1.325632695555687, 1.23151354432106, 1.2123524838685988, 1.1726598739624023, 1.0923227965831757, 1.1367704308032989, 1.0908222246170043, 1.1158152759075164, 1.0315754282474519, 1.06554756462574, 1.0479815247654916, 1.0996541720628739, 1.0450058096647263, 1.0318933933973313, 1.0339357393980027, 1.0286797326803208, 1.052253406047821, 1.0163359659910203, 0.9940995579957962, 0.988142160475254, 0.9932016897201538, 0.9409900230169296, 0.9778001809120178, 1.0251611429452896, 0.9572644853591918, 1.013562046289444, 0.9348490306735039, 0.9323300030827523, 1.0125887167453766, 0.9674993979930878]
```  
There are several bugs if we enable Dynamo, I'm working on fixing them now. Posting it before bug-free is to collect feedback on if this framework is what users usually does and then we can optimize Dynamo to provide the best performance and experience with least onboarding cost.

cc @anijain2305

### DIFF
--- a/benchmarks/training_loss.py
+++ b/benchmarks/training_loss.py
@@ -1,7 +1,8 @@
 import argparse
-import math
 import os
 import sys
+import time
+from datetime import timedelta
 
 import torch
 from datasets import load_dataset
@@ -94,13 +95,11 @@ def model_training_evaluation(
 def check_loss(ref_loss, res_loss):
     assert len(ref_loss) == len(res_loss)
     length = len(ref_loss)
-    for i in range(math.ceil(length / 10)):
-        end = i + 10 if i + 10 <= length else length
-        x = sum(ref_loss[i:end]) / 10
-        y = sum(res_loss[i:end]) / 10
-        if not math.isclose(x, y, rel_tol=1e-1):
-            return False
-    return True
+    x = min(length, 10)
+    if sum(res_loss[-x:]) / 10 <= sum(ref_loss[-x:]) / 10 + 1e-1:
+        return True
+    else:
+        return False
 
 
 def parse_args():
@@ -150,18 +149,28 @@ def main():
     )
     optimizer_cls = getattr(sys.modules["torch.optim"], args.optimizer)
     optimizer = optimizer_cls(model.parameters(), lr=args.lr)
+    native_start = time.time()
     ref_loss, _ = model_training_evaluation(
         None, train_dataloader, eval_dataloader, model, optimizer, args.epochs
     )
+    native_end = time.time()
     res_loss, _ = model_training_evaluation(
         args.backend, train_dataloader, eval_dataloader, model, optimizer, args.epochs
     )
+    dynamo_end = time.time()
     if check_loss(ref_loss, res_loss):
-        print("[PASSED] TorchDynamo end to end training passed!")
+        print(
+            "[PASSED] TorchDynamo end to end training loss is less than or equal to native PyTorch"
+        )
     else:
         print(
-            "[FAILED] TorchDynamo end to end training failed due to loss not close to native Pytorch"
+            "[FAILED] TorchDynamo end to end training loss is greater than native Pytorch"
         )
+    native_elapsed = native_end - native_start
+    dynamo_elapsed = dynamo_end - native_end
+    print(f"Train model on {args.epochs} epochs:")
+    print(f"PyTorch spent {timedelta(seconds=native_elapsed)}")
+    print(f"TorchDynamo spent {timedelta(seconds=dynamo_elapsed)}")
 
 
 if __name__ == "__main__":
--- a/benchmarks/training_loss.py
+++ b/benchmarks/training_loss.py
@@ -1,26 +1,25 @@
+import argparse
 import os
+import sys
 
 import torch
 from datasets import load_dataset
 from datasets import load_metric
-from torch.optim import SGD
 from torch.utils.data import DataLoader
 from transformers import AutoModelForSequenceClassification
 from transformers import AutoTokenizer
 
 import torchdynamo
 
-torchdynamo.config.fake_tensor_propagation = False
+# torchdynamo.config.fake_tensor_propagation = False
 
 # You will download around 84G dataset if you run this end to end training/evaluation example.
 
 os.environ["TOKENIZERS_PARALLELISM"] = "false"
 device = torch.device("cuda") if torch.cuda.is_available() else torch.device("cpu")
-num_samples = 10000
-num_epochs = 3
 
 
-def data_processing(num_samples):
+def data_processing(num_samples, batch_size):
     dataset = load_dataset("yelp_review_full")
     tokenizer = AutoTokenizer.from_pretrained("bert-base-cased")
 
@@ -40,8 +39,10 @@ def data_processing(num_samples):
         tokenized_datasets["test"].shuffle(seed=42).select(range(num_samples))
     )
 
-    train_dataloader = DataLoader(small_train_dataset, shuffle=True, batch_size=8)
-    eval_dataloader = DataLoader(small_eval_dataset, batch_size=8)
+    train_dataloader = DataLoader(
+        small_train_dataset, shuffle=True, batch_size=batch_size
+    )
+    eval_dataloader = DataLoader(small_eval_dataset, batch_size=batch_size)
 
     return train_dataloader, eval_dataloader
 
@@ -56,13 +57,13 @@ def training_iter_fn(batch, model, optimizer):
 
 
 def model_training_evaluation(
-    train_dataloader, eval_dataloader, model, optimizer, num_epochs
+    backend, train_dataloader, eval_dataloader, model, optimizer, num_epochs
 ):
     model.to(device)
     model.train()
     loss_history = []
     # Support backends: eager, aot_nop and aot_nvfuser
-    opt_training_iter_fn = torchdynamo.optimize("eager")(training_iter_fn)
+    opt_training_iter_fn = torchdynamo.optimize(backend)(training_iter_fn)
     for epoch in range(num_epochs):
         running_loss = 0.0
         for i, batch in enumerate(train_dataloader, 0):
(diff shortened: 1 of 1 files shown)


yanboliang:
Update: This training benchmark works well and ready for review now.

benchmarks/training_loss.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+def model_training_evaluation(
+    train_dataloader, eval_dataloader, model, optimizer, num_epochs
+):
+    model.to(device)
+    model.train()
+    loss_history = []
+    # Support backends: eager, aot_nop and aot_nvfuser
+    opt_training_iter_fn = torchdynamo.optimize("eager")(training_iter_fn)
```
anijain2305, line 65: pass `backend` and support it in arg.parser?

benchmarks/training_loss.py
```diff
@@ -0,0 +1,100 @@
...
+from datasets import load_dataset
+from datasets import load_metric
+from torch.optim import SGD
+from torch.utils.data import DataLoader
+from transformers import AutoModelForSequenceClassification
+from transformers import AutoTokenizer
+
+import torchdynamo
+
+torchdynamo.config.fake_tensor_propagation = False
```
anijain2305, line 13: Can we turn this on? Or list down the issues that we need to fix to turn it on?

benchmarks/training_loss.py
```diff
@@ -0,0 +1,100 @@
...
+        predictions = torch.argmax(logits, dim=-1)
+        metric.add_batch(predictions=predictions, references=batch["labels"])
+
+    print("Loss history : " + str(loss_history))
+    print("Accuracy : " + str(metric.compute()))
+
+
+if __name__ == "__main__":
+    train_dataloader, eval_dataloader = data_processing(num_samples)
+    model = AutoModelForSequenceClassification.from_pretrained(
```
anijain2305, line 94: This can go in next PR. But, I have a few suggestions

1) We should provide some glue code, so that we can say test PASSED or FAILED. Like, run native pytorch and optimize("some_backend") have close-enough loss after some iterations.
2) Ability to specify more models.
3) Ability to specify more optimizers.

anijain2305 (changes requested):

benchmarks/training_loss.py
voznesenskym: Can we use @patch to set this, instead of setting it globally?

benchmarks/training_loss.py
```diff
@@ -0,0 +1,100 @@
...
+from torch.optim import SGD
+from torch.utils.data import DataLoader
+from transformers import AutoModelForSequenceClassification
+from transformers import AutoTokenizer
+
+import torchdynamo
+
+torchdynamo.config.fake_tensor_propagation = False
+
+# You will download around 84G dataset if you run this end to end training/evaluation example.
```
voznesenskym, line 15: Is there a way to cache this artifact in CI? Or maybe have a flag to run a smaller version?

voznesenskym:
How useful is printing loss history? How do we use this benchmark to tell if dynamo changes convergence on its own? with a backend? As it is, I am not sure who would run this benchmark and get value from it...

My $0.02 (take it or leave it) Is we already have training benchmarks... maybe we can extend those to report loss / entropy / normalized cross-entropy (w/e you like) at a given epoch (say, 1000?) We can leverage the infra @anijain2305 made to make our awesome "Performance Dashboard" and just add a column / record for $some_loss_measure at $some_k_epochs. Thoughts?

anijain2305:
> How useful is printing loss history? How do we use this benchmark to tell if dynamo changes convergence on its own? with a backend? As it is, I am not sure who would run this benchmark and get value from it...
> 
> My $0.02 (take it or leave it) Is we already have training benchmarks... maybe we can extend those to report loss / entropy / normalized cross-entropy (w/e you like) at a given epoch (say, 1000?) We can leverage the infra @anijain2305 made to make our awesome "Performance Dashboard" and just add a column / record for $some_loss_measure at $some_k_epochs. Thoughts?

We don't do multiple iteration training in the current benchmarking infra (though it would be ideal to put all this one place). We run just one iteration of fwd-bwd there. There are multiple issues behind it - (1) As it stands now, the goal of current benchmarking infra is to aid development, find bugs in single iter, and quickly iterate. (2) There is no easy way to get valid dataloader for all the benchmarks (or even a few) to train to completion.

This created a gap, where we were not testing Dynamo on end-to-end training. So, I suggested Yanbo to create this. This has exposed multiple bugs. The main goal here is to support training for few models + optimizers and convince ourselves that Dynamo+inductor actually work for long training jobs.

voznesenskym:
> > How useful is printing loss history? How do we use this benchmark to tell if dynamo changes convergence on its own? with a backend? As it is, I am not sure who would run this benchmark and get value from it...
> 
> > 
> 
> > My $0.02 (take it or leave it) Is we already have training benchmarks... maybe we can extend those to report loss / entropy / normalized cross-entropy (w/e you like) at a given epoch (say, 1000?) We can leverage the infra @anijain2305 made to make our awesome "Performance Dashboard" and just add a column / record for $some_loss_measure at $some_k_epochs. Thoughts?
> 
> 
> 
> We don't do multiple iteration training in the current benchmarking infra (though it would be ideal to put all this one place). We run just one iteration of fwd-bwd there. There are multiple issues behind it - (1) As it stands now, the goal of current benchmarking infra is to aid development, find bugs in single iter, and quickly iterate. (2) There is no easy way to get valid dataloader for all the benchmarks (or even a few) to train to completion.
> 
> 
> 
> This created a gap, where we were not testing Dynamo on end-to-end training. So, I suggested Yanbo to create this. This has exposed multiple bugs. The main goal here is to support training for few models + optimizers and convince ourselves that Dynamo+inductor actually work for long training jobs. 
> 
> 
> 
> 

Makes sense! We should maybe add a followup to track convergence across backends?

anijain2305:
> We should maybe add a followup to track convergence across backends?

Yes, I agree with that. We should refactor to say PASS or FAIL and ability to try different backends. Otherwise, its too much eyeballing for now. Hopefully, we can use this script in the future to publish numbers like the way Nvidia does it - https://github.com/NVIDIA/DeepLearningExamples/tree/master/PyTorch/LanguageModeling/BERT#results

benchmarks/training_loss.py
yanboliang: > Can we turn this on? Or list down the issues that we need to fix to turn it on?

This issue has been tracked by https://github.com/pytorch/torchdynamo/issues/1076

> Can we use @patch to set this, instead of setting it globally?

I think this is both for benchmark and example, we use this to show how to enable TorchDynamo end to end. So would prefer not to add decorator on fn in case confusing users. And this is a temporary config, will remove it when we fix the above issue.

benchmarks/training_loss.py
yanboliang: Sounds good! I added the ability to specify backend/optimizer and several model training parameters. We can add the ability to compare training loss for native pytorch vs TorchDynamo in follow-ups.

benchmarks/training_loss.py
yanboliang: I was thinking to cache it in CI, but it's too big IMO. I have two concerns to have a smaller version:
* We need to generate the smaller version by ourself, so it's not a standard HF training pipeline any more. This is also used as an example/tutorial to help users onboard their models. We are trying to do minimized changes on the way of using HF.
* A smaller version may can't reveal the training loss convergence sometime.

yanboliang:
> Hopefully, we can use this script in the future to publish numbers like the way Nvidia does it - https://github.com/NVIDIA/DeepLearningExamples/tree/master/PyTorch/LanguageModeling/BERT#results

Agree on what this benchmark would evolve toward!

benchmarks/training_loss.py
voznesenskym: Yeah, that seems reasonable, albeit unfortunate. 

cc @malfet for CI caching ideas?

benchmarks/training_loss.py
```diff
@@ -0,0 +1,142 @@
...
+    for batch in eval_dataloader:
+        batch = {k: v.to(device) for k, v in batch.items()}
+        with torch.no_grad():
+            outputs = opt_model(**batch)
+
+        logits = outputs.logits
+        predictions = torch.argmax(logits, dim=-1)
+        metric.add_batch(predictions=predictions, references=batch["labels"])
+
+    print("Loss history : " + str(loss_history))
```
anijain2305, line 89: Can we add the code to tell if a backend PASSED or FAILED by storing the loss history of pytorch-native and the backend? Basically, this test will be super useful, if you just run from commandline, and it says PASS or FAIL. Otherwise, its too much eyeballing.

anijain2305 (changes requested):

anijain2305 (approved):

benchmarks/training_loss.py
```diff
@@ -0,0 +1,177 @@
...
+    else:
+        print(
+            "[FAILED] TorchDynamo end to end training loss is greater than native Pytorch"
+        )
+    native_elapsed = native_end - native_start
+    dynamo_elapsed = dynamo_end - native_end
+    print(f"Train model on {args.epochs} epochs:")
+    print(f"PyTorch spent {timedelta(seconds=native_elapsed)}")
+    print(f"TorchDynamo spent {timedelta(seconds=dynamo_elapsed)}")
+
```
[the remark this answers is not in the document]
yanboliang: I'll address this together with the following PRs.

PR #793, opened 2022-08-12
Use python logger with specific error messages and log levels

mlazos:
* Provide full stack traces to user code on most errors in addition to a dynamo stack trace
* Log level is configurable through dynamo config and are the following
    DEBUG - print full instr by instr traces and all other levels
    INFO - print compiled functions + graphs and below levels
    WARNING - print graph breaks + and below level
    ERROR - print exceptions only

Samples here:
 [Graph break logs](https://www.internalfb.com/phabricator/paste/view/P521889215)
 [Torch Runtime error logs](https://www.internalfb.com/phabricator/paste/view/P521890006)
 [Assertion error logs](https://www.internalfb.com/phabricator/paste/view/P521890670)


run.sh
```diff
@@ -0,0 +1,9 @@
+# Setup the output directory
+rm -rf timm_logs
+mkdir timm_logs
+
+# Commands for timm_models for device=cuda, dtype=float32 for training
+python benchmarks/timm_models.py --float32 -dcuda --no-skip --output=timm_logs/ts_nvfuser_timm_models_float32_training_cuda.csv --training --nvfuser --speedup-dynamo-ts --use-eval-mode --isolate
```
mlazos, line 6: Remove

torchdynamo/config.py
```diff
@@ -12,14 +13,13 @@
...
-trace = False
-
-# print the torchdynamo internal exceptions
-print_internal_exceptions = True
+# log level (lower levels print what it says + all higher levels)
+# DEBUG print full traces <-- lowest level + print tracing of every instruction
+# INFO print compiled functions + graphs
+# WARN print warnings (including graph breaks)
+# ERROR print exceptions (and what user code was being processed when it occurred)
+log_level = logging.DEBUG
```
mlazos, line 21: Change default to warning

torchdynamo/utils.py
```diff
@@ -65,10 +64,59 @@
...
+        if "convert_frame" in frame.filename:
+            break
+        if "eval_frame" in frame.filename:
+            continue
+        user_stack.append(frame)
+
+    return user_stack
+
+
+def format_graph_tabular(graph):
```
voznesenskym, line 107: when would I use this vs `graph.print_tabular()`?

mlazos: `print_tabular()` only prints unfortunately and doesn't return the string

voznesenskym: Right - but we can make it either `print_tabular(file)` where file is a stream, or:

```
def print_tabular():
   print(tabular_out(self.graph))
```

And we use `fx.tabular_out(graph)` - if that make sense?

torchdynamo/utils.py
```diff
@@ -65,10 +64,59 @@
...
 
-@functools.lru_cache(None)
 def init_logging():
     if "PYTEST_CURRENT_TEST" not in os.environ:
         logging.config.dictConfig(LOGGING_CONFIG)
+        logger = logging.getLogger(config.log_name)
+        logger.setLevel(config.log_level)
+
+
+def get_logger():
```
msaroufim, line 74: What's the benefit of wrapping the logging functions like this? At least within `pytorch/pytorch` I've mostly seen the syntax for the python `logging` module used directly

torchdynamo/symbolic_convert.py
```diff
@@ -293,6 +304,11 @@ def step(self):
             exc.real_stack.append(self.frame_summary())
             if self.empty_checkpoint():
                 raise
+        except Exception as exc:
```
msaroufim, line 307: Should we log the stack here?

`logging` has a nice `logging.exception` function which also shows user the stack trace

torchdynamo/config.py
```diff
@@ -1,3 +1,4 @@
+import logging
```
msaroufim, line 1: I'm curious does `import torch` also bring in its own logger? I'm worried that potentially some PyTorch logs could be disabled if a user enables dynamo logging

torchdynamo/utils.py
```diff
@@ -292,12 +340,12 @@ def torchscript(model, example_inputs, verbose=True):
         return torch.jit.trace(model, example_inputs)
     except Exception:
         if verbose:
-            log.exception("jit error")
+            log_error("jit error")
```
msaroufim, line 343: I feel like an exception is a more appropriate thing to log along with its stack trace

torchdynamo/output_graph.py
```diff
@@ -350,9 +352,13 @@ def compile_and_call_fx_graph(self, tx, rv, root):
...
-        if config.debug:
-            print(f"\n{name} {gm.forward.__code__.co_filename}")
-            self.graph.print_tabular()
+
+        try:
+            log_info(
+                f"TRACED GRAPH\n {name} {gm.forward.__code__.co_filename} {format_graph_tabular(gm.graph)}\n"
+            )
+        except ImportError:
+            pass
```
msaroufim, line 361: Should probably just log the error here too

torchdynamo/convert_frame.py
```diff
@@ -172,16 +176,35 @@ def has_tensor(obj):
...
+def format_exception(exc, frame):
+    msg = os.linesep * 2
+    msg += "=" * 10 + " TorchDynamo Stack Trace " + "=" * 10 + "\n"
+    msg += traceback.format_exc()
+    msg += (
+        "\n"
+        + "=" * 10
+        + " The above exception occurred while processing the following code "
+        + "=" * 10
+        + "\n"
```
msaroufim, line 196: One more line break would make the logs easier to visually parse

torchdynamo/config.py
```diff
@@ -12,14 +13,13 @@
...
-trace = False
-
-# print the torchdynamo internal exceptions
-print_internal_exceptions = True
+# log level (levels print what it says + all levels listed below it)
+# DEBUG print full traces <-- lowest level + print tracing of every instruction
+# INFO print compiled functions + graphs
+# WARN print warnings (including graph breaks)
+# ERROR print exceptions (and what user code was being processed when it occurred)
+log_level = logging.WARNING
```
msaroufim, line 21: It may make sense to pull the log level from a `logging.conf` file instead or read it from `argparse`, otherwise if a user has `pip install torchdynamo` when it does get released on pypi they won't be able to easily change the logging level without build from source

torchdynamo/utils.py
msaroufim: Since we're changing this code can we also include the line numbers associated in the python file to the lines in the fx graph

msaroufim (changes requested):

torchdynamo/utils.py
voznesenskym: @SherlockNoMad already put stack_trace on fx nodes.

torchdynamo/convert_frame.py
```diff
@@ -172,16 +176,35 @@ def has_tensor(obj):
...
+    log_debug(
+        f"skipping because no torch.* {frame.f_code.co_name} \
+            {frame.f_code.co_filename} {frame.f_code.co_firstlineno}"
+    )
+
     return False
 
 
+def format_exception(exc, frame):
+    msg = os.linesep * 2
```
anijain2305, line 188: Can we also print the type of exception - SkipFrame vs TorchDynamoInternal etc

torchdynamo/config.py
anijain2305: iiuc, one can easily change the logging from outside, by simply doing 

`torchdynamo.config.log_level = logging.INFO`

torchdynamo/convert_frame.py
```diff
@@ -294,44 +309,44 @@ def debug_print(prefix):
                 except exc.SkipFrame:
                     return None
```
anijain2305, line 310: Can we add a log here as well? I spent a lot of time once to figure where my frame went.

anijain2305: And similarly a log just before - on the RestartAnalysis

anijain2305 (commented):
Thanks, this looks pretty good to me. I am gonna use this to debug the issues discovered by @msaroufim

voznesenskym (approved):

torchdynamo/utils.py
mlazos: Yeah I'll upstream the refactor to fx, I forgot we own that too :D

torchdynamo/utils.py
mlazos: I wanted to get the log name from the config, instead of it being module-specific.

torchdynamo/symbolic_convert.py
mlazos: No we shouldn't. If we're inlining there may be many nested instruction translators, so we would print the exception multiple times. This is why I only print it at the top level. The other piece of complexity here is there are essentially two stack traces. The torchdynamo stack trace that is useful for us, and then the stack trace through the user code which should point to the line we were processing when the exception happened. (That's the `real_stack` shown here.

torchdynamo/utils.py
jansel: I think it is better to have per-file usage of logging.  At the top of every file say:
```
```
Then call `log.info()`, `log.warning()`, etc.

The big advantage of doing it this way is you can control log levels at a finer granularity.  E.g. you can do:
```
logging.getLogger("torchdynamo.symbolic_convert").setLevel(logging.ERROR)
```
to silence logs from a specific file.

This also lets us point very noisy things (like bytecode dumps) in their own logging stream.

torchdynamo/utils.py
```diff
@@ -65,10 +64,59 @@
...
         logging.config.dictConfig(LOGGING_CONFIG)
+        logger = logging.getLogger(config.log_name)
+        logger.setLevel(config.log_level)
+
+
+def get_logger():
+    return logging.getLogger(config.log_name)
+
+
+def log_warning(msg, *args, **kwargs):
```
jansel, line 78: Let's remove these and have a per-file log object.

torchdynamo/utils.py
jansel, line 343: `log.exception()` includes a stack trace.   Are you removing that intentionally?

jansel (changes requested):

torchdynamo/utils.py
mlazos: No, I missed that this doesn't get raised further, I'll use log.exception

torchdynamo/utils.py
mlazos: I'll upstream this and remove once we have the version of Pytorch that contains the update

torchdynamo/convert_frame.py
mlazos: I believe traceback.format_exc already does this

jansel (approved):

PR #800, opened 2022-08-12
Print error when no graph is captured with one_graph=True

mlazos:


torchdynamo/convert_frame.py
```diff
@@ -292,6 +292,8 @@ def debug_print(prefix):
                     if attempt > 100:
                         unimplemented("100+ RestartAnalysis() calls")
                 except exc.SkipFrame:
+                    if one_graph:
+                        print("ERROR: No graph captured with one_graph=True")
```
voznesenskym, line 296: can you put this behind config.debug? Export already has an assert on this, and having both would leak dynamo impl details into stdout.

voznesenskym (approved):

anijain2305 (approved):

torchdynamo/convert_frame.py
```diff
@@ -292,6 +292,8 @@ def debug_print(prefix):
                     if attempt > 100:
                         unimplemented("100+ RestartAnalysis() calls")
                 except exc.SkipFrame:
+                    if one_graph and config.debug:
+                        print("ERROR: No graph captured with one_graph=True")
```
anijain2305, line 296: Another reason we should add a log here :)

mlazos: Yeah I'm going to handle this in the logging PR.

PR #806, opened 2022-08-13
Fix sth like torch.LongTensor(list(np.int64, np.int64, ...))

yanboliang:
This was reported by @anijain2305 when running torchbench:
```
python benchmarks/torchbench.py --training --accuracy-aot-nop -d cuda --isolate --use-eval-mode --only=fastNLP_Bert
```

The root cause is FX doesn't support numpy int/float as [base type](https://github.com/pytorch/pytorch/blob/master/torch/fx/proxy.py#L154), so throw ```NotImplementedError``` error when there is numpy int/float in ```fx.node.args```.

The fast fix is to add numpy int/float as base types in torch.fx, but we think the goal of torch.fx is not to capture every base type after discussing with @anijain2305 .
So we fix this issue at Dynamo side by converting numpy int/float to python int/float before passing them to ```fx.node.args```.

### DIFF
--- a/torchdynamo/variables/torch.py
+++ b/torchdynamo/variables/torch.py
@@ -4,11 +4,13 @@ import types
 from typing import Dict
 from typing import List
 
+import numpy
 import torch._C
 import torch.nn
 
 from torchdynamo.source import GetItemSource
 from torchdynamo.source import NNModuleSource
+from torchdynamo.variables.lists import ListVariable
 from torchdynamo.variables.lists import TupleVariable
 from torchdynamo.variables.misc import AutocastModeVariable
 from torchdynamo.variables.misc import AutogradProfilerContextWrapperVariable
@@ -23,6 +25,7 @@ from ..utils import istype
 from ..utils import product
 from ..utils import proxy_args_kwargs
 from ..utils import specialize_args_kwargs
+from ..utils import tensortype_to_dtype
 from .base import VariableTracker
 from .tensor import TensorWithTFOverrideVariable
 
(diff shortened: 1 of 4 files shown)


yanboliang:
The error is caused by fake tensor. If I switch ```fake_tensor_propagation``` to ```False```, it works well.

torchdynamo/variables/torch.py
```diff
@@ -204,6 +207,18 @@ def call_function(
             assert len(args) == 2
             return args[1]
         else:
+            # Handle sth like torch.LongTensor(list(np.int64, np.int64, ...)),
```
voznesenskym, line 210: nit spell something lol

torchdynamo/variables/torch.py
```diff
@@ -3,9 +3,11 @@
 from typing import Dict
 from typing import List
 
+import numpy
```
voznesenskym, line 6: I think its not amazing to have numpy creeping in everywhere like this. A dep is a dep, even if it ubiquitous - I'd almost rather we dealt with it by `.__name__` and string comparison. No strong opinions though, just something to consider.

voznesenskym (approved):

torchdynamo/variables/torch.py
jansel: IMO numpy is fine.  We already depend on it.

eellison:
Is adding a guard on the exact value of a 512 element numpy tensor better than graph breaking ?

voznesenskym:
> Is adding a guard on the exact value of a 512 element numpy tensor better than graph breaking ?

Specializing on the exact value of a tensor does not sound like the right thing to do, numpy or otherwise...

PR #807, opened 2022-08-13
don't tile broadcasted tensors, those tiles are not needed

ngimel:
Currently a lot of pw kernels are getting codegened with tiling and are subsequently autotuned, when that's not necessary


Chillee (approved):
LGTM

jansel (approved):

ngimel:
Failing coverage test seems unrelated.

PR #820, opened 2022-08-15
[inductor] Make less consts by relaxing matching

voznesenskym:
`./benchmarks/torchbench.py --inductor --training --no-skip --only Super_SloMo` 

Down to 6 graph constants, from 100, at the time the scheduler is created

### DIFF
--- a/torchinductor/graph.py
+++ b/torchinductor/graph.py
@@ -151,7 +151,11 @@ class GraphLowering(torch.fx.Interpreter):
     def add_tensor_constant(self, data):
         def allocate():
             for name, value in self.constants.items():
-                if data.size() == value.size() and torch.allclose(data, value):
+                if (data.size() == value.size() and 
+                    data.stride() == value.stride() and
+                    data.dtype == value.dtype and
+                    data.device == value.device and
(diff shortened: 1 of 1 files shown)


torchinductor/graph.py
```diff
@@ -151,7 +151,7 @@ def visit(value):
     def add_tensor_constant(self, data):
         def allocate():
             for name, value in self.constants.items():
-                if data is value:
+                if data.size() == value.size() and torch.allclose(data, value):
```
jansel, line 154:
```suggestion
                if (data.size() == value.size() and 
                    data.stride() == value.stride() and
                    data.dtype == value.dtype and
                    data.device == value.device and
                    torch.eq(data, value).all()):
```

voznesenskym: Surprised device matters, I thought we had code like `constant_name()` that handles consts devices, but otherwise this looks like an improvement yea!

jansel (approved):
Test on torchbench before landing.

voznesenskym:
Confirmed no errors or regressions against master

PR #825, opened 2022-08-15
Add a `--diff` flag to benchmarking

voznesenskym:
./benchmarks/torchbench.py --inductor --training --no-skip --only Super_SloMo --diff

```
RUNNING ON BRANCH: voz/diff
cpu  train Super_SloMo                         
0.777x p=0.00
RUNNING ON BRANCH: main
cpu  train Super_SloMo                        

0.798x p=0.00
```


voznesenskym:
Diff is probably a bad name as it does not diff :P Open to bikeshedding on the word

jansel (approved):

desertfire:
> Diff is probably a bad name as it does not diff :P Open to bikeshedding on the word

Nit: I don't mind to be wordy on this option naming. `--diff` sounds too generic to me. Maybe something like `--diff-against-main` or `--diff-against-branch`?

wconstab (approved):

PR #827, opened 2022-08-15
Normalize device objects when passed into constructors

Chillee:
Fixes https://github.com/pytorch/torchdynamo/issues/633

### DIFF
--- a/torchinductor/lowering.py
+++ b/torchinductor/lowering.py
@@ -110,7 +110,9 @@ def decode_device(device):
     if device is None:
         return torch.tensor(0.0).device  # default device
     if isinstance(device, str):
-        return torch.device(device)
+        device = torch.device(device)
+    if device.type == "cuda" and device.index is None:
+        return torch.device("cuda", index=torch.cuda.current_device())
     return device
 
 
--- a/torchinductor/scheduler.py
+++ b/torchinductor/scheduler.py
@@ -554,6 +554,8 @@ def draw_buffers(nodes, fname, print_graph=False):
         if "fusion_meta" not in node.meta:
             continue
         group = node.meta["fusion_meta"].group
+        if isinstance(group, tuple):
+            group = group[1]
 
         # gather meta data
         dtype = None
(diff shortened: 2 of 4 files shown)


torchinductor/scheduler.py
```diff
@@ -554,6 +554,8 @@ def draw_buffers(nodes, fname, print_graph=False):
         if "fusion_meta" not in node.meta:
             continue
         group = node.meta["fusion_meta"].group
+        if isinstance(group, tuple):
+            group = group[1]
```
ngimel, line 558: what's `group[1]`?

ngimel (approved):

torchinductor/scheduler.py
Chillee: It's the "non-device" part of group. Group looks like (device, indexing_info).

torchinductor/lowering.py
```diff
@@ -111,6 +111,8 @@ def decode_device(device):
         return torch.tensor(0.0).device  # default device
     if isinstance(device, str):
         return torch.device(device)
+    if device.type == "cuda" and device.index is None:
+        return torch.device("cuda", index=torch.cuda.current_device())
```
eellison, line 115: do we guard on current device anywhere ?

torchinductor/lowering.py
```diff
@@ -111,6 +111,8 @@ def decode_device(device):
         return torch.tensor(0.0).device  # default device
     if isinstance(device, str):
         return torch.device(device)
```
jansel, line 113: We should normalize this one too:
```
>>> torch.device("cuda")
device(type='cuda')
```

jansel (changes requested):

torchinductor/lowering.py
Chillee: done

torchinductor/lowering.py
jansel: We don't, though seems fine to do that in another PR.

jansel (approved):

Issue #623, opened 2022-07-20
AssertionError and IndexError: "vision_maskrcnn",


desertfire:

Chillee:
Can I get a command for this too :)

Chillee:
I'm getting a
```
torchinductor.exc.LoweringException: AssertionError: (1 - s2, -4999)
  target: aten.index.Tensor
  args[0]: TensorBox(StorageBox(
    InputBuffer(name='arg0_1_1', layout=FixedLayout('cuda', torch.float32, size=[s0, s1, s0], stride=[s0*s1, s0, 1]))
  ))
  args[1]: [TensorBox(StorageBox(
    InputBuffer(name='unsqueeze_default_1', layout=FixedLayout('cuda', torch.int64, size=[s0, 1], stride=[1, 1]))
  )), TensorBox(StorageBox(
    InputBuffer(name='cat_default_1_1', layout=FixedLayout('cuda', torch.int64, size=[s0, s2], stride=[s2, 1]))
  ))]
```
for this.

Minimal repro here:

```
from torchinductor.compile_fx import compile_fx_inner
from torch.fx.experimental.proxy_tensor import make_fx

inps = [(torch.Size([4, 359613, 4]), torch.float32), (torch.Size([4, 5000]), torch.int64), (torch.Size([4, 1]), torch.int64), (torch.Size([4, 5000]), torch.float32)]
inps = [torch.zeros(())] + [torch.ones(shape, dtype=dtype, device='cuda') for (shape, dtype) in inps]



def forward(self, arg0_1, cat_default_1, unsqueeze_default, index_tensor):
    index_tensor_2 = torch.ops.aten.index.Tensor(arg0_1, [unsqueeze_default, cat_default_1]);  arg0_1 = unsqueeze_default = cat_default_1 = None
    return []
forward = make_fx(forward)(*inps)
compile_fx_inner(forward, inps)
```

Chillee:
New error

```
  File "<eval_with_key>.734", line 7, in forward
    to_dtype = ops.to_dtype(load, torch.float16);  load = None
  File "/scratch/chilli/fresh/torchdynamo/torchinductor/codegen/common.py", line 579, in inner
    self.compute, getattr(parent_handler, name)(*args, **kwargs)
  File "/scratch/chilli/fresh/torchdynamo/torchinductor/codegen/cpp.py", line 157, in to_dtype
[...]

PR #831, opened 2022-08-16
Index broadcast

ngimel:
Fixes minimum repro from #623

### DIFF
--- a/torchinductor/lowering.py
+++ b/torchinductor/lowering.py
@@ -1340,14 +1340,22 @@ def embedding(weight, indices, padding_idx=-1, scale_grad_by_freq=False, sparse=
 def check_and_broadcast_indices(indices):
     print(indices)
     print([i.get_dtype() for i in indices if i is not None])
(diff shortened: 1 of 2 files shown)


torchinductor/lowering.py
```diff
@@ -1337,12 +1337,35 @@ def fn(idx):
     )
 
 
+def check_and_broadcast_indices(indices):
+    print(indices)
```
jansel, line 1341: remove prints

torchinductor/lowering.py
```diff
@@ -1337,12 +1337,35 @@ def fn(idx):
...
+    print([i.get_dtype() for i in indices if i is not None])
```
ngimel, line 1342:
```suggestion
```

jansel (approved):

Chillee:
Does `vision_maskrcnn` work now btw?

ngimel:
I don't know, #623 didn't have repro command. I suspect this PR breaks more than it fixes, lol, because indices are so often not int64 due to other bugs.

PR #832, opened 2022-08-16
[inductor] Move print_performance helper to utils

jansel:


torchinductor/utils.py
```diff
@@ -52,3 +55,23 @@ def sympy_dot(seq1, seq2):
...
     return {id(x): x for x in it}.values()
+
+
+def timed(model, example_inputs, times=1):
+    synchronize()
+    torch.manual_seed(1337)
+    t0 = time.perf_counter()
+    for _ in range(times):
+        result = model(*example_inputs)
+        synchronize()
```
Chillee, line 66: Should we synchronize after all runs are done?

Chillee: Also, we could also perhaps just use triton's benchmarking code.

Chillee (approved):
lgtm

torchinductor/utils.py
jansel: It is useful to measure CPU overheads.  Plus this works for the C++ backend.  We use the gpu-only measurement in other places though.

Issue #822, opened 2022-08-15
Investigate TorchInductor peak memory usage


jansel:
Most of the failures from huggingface models are currently out of memory errors.  We should profile memory usage and track down what is causing this issue.

Chillee:
Well.. one of them is that torchinductor needs to take advantage of this now :P

https://github.com/pytorch/pytorch/pull/83137

jansel:
#820 might also help

PR #843, opened 2022-08-16
Reduce huggingface batch sizes

jansel:
Temporary workaround for #822 so we can fix non-memory related issues and get signal on other performance improvements.


anijain2305 (approved):

jansel:
> Should we just have a command line arg to divide the set batch sizes by some number?

That could fix it.  Though for the models without memory issues it might be better to keep the bigger batch sizes.

PR #846, opened 2022-08-16
Correctly pre-initialize FakeTensorMode before using it, so restore works

ezyang:
Needed for https://github.com/pytorch/pytorch/pull/83496


jansel (approved):

eellison (approved):
lgtm, but as an API feel like this should be unnecessary

PR #849, opened 2022-08-16
Require torchdynamo.reset() to change backends

jansel:
This removes a footgun where someone changes backends, but then still uses cached results from the prior backend.


torchdynamo/eval_frame.py
```diff
@@ -157,9 +174,20 @@ def _fn(*args, **kwargs):
...
+            global most_recent_backend
+            if (
+                most_recent_backend is not None
+                and most_recent_backend is not compiler_fn
+            ):
+                raise ResetRequired()
+            most_recent_backend = compiler_fn
+            install_generation_tagging_init()
+
+        compiler_fn = innermost_fn(callback)
```
anijain2305, line 187: Is `compiler_fn` used?

torchdynamo/eval_frame.py
```diff
@@ -157,9 +174,20 @@ def _fn(*args, **kwargs):
 
 class OptimizeContext(_TorchDynamoContext):
     def __init__(self, callback, backend_ctx_ctor):
+        def on_enter():
+            global most_recent_backend
+            if (
+                most_recent_backend is not None
+                and most_recent_backend is not compiler_fn
+            ):
+                raise ResetRequired()
```
anijain2305, line 183: Will it cause issues when ones wants to do something like this?

~~~

optimized_fn1 = torchdynamo.optimized("aot_nvfuser")(fn1)
optimized_fn2 = torchdynamo.optimized("inductor")(fn2)
optimized_fn3 = torchdynamo.optimized("trt")(fn3)

~~~

jansel: Correct, but even that example could be buggy: if fn1 and fn2 call a common helper function, you could have the case where the "inductor" run is actually using "nvfuser" code (if you graph breaked on the helper).

torchdynamo/eval_frame.py
jansel: Yes, line 181 and 184

torchdynamo/eval_frame.py
```diff
@@ -36,19 +40,32 @@
 skip_code = _eval_frame.skip_code
 set_guard_fail_hook = _eval_frame.set_guard_fail_hook
 set_guard_error_hook = _eval_frame.set_guard_error_hook
-
 always_optimize_code_objects = utils.ExactWeakKeyDictionary()
+null_context = contextlib.nullcontext
+unset = object()
+compile_lock = threading.Lock()
+most_recent_backend = None
```
voznesenskym, line 47: nit: underscore to make it clear that this is private?

voznesenskym:
LGTM

voznesenskym (approved):

voznesenskym:
An interesting logical extension is `torchdynamo.optimize()` w/o a backend specified, explicitly for safe nesting without having to plumb around a backend.

jansel:
> An interesting logical extension is `torchdynamo.optimize()` w/o a backend specified, explicitly for safe nesting without having to plumb around a backend.

How is this different from just no decorator at all?

PR #851, opened 2022-08-17
Relax the accuracy checks by comparing against fp64

anijain2305:
Passes accuracy for around 20 models for Torchbench for `float16`

* Uses RMSE to find the error from fp64.

Thanks @Chillee @jansel for helping in finding a suitable error metric.

### DIFF
--- a/benchmarks/common.py
+++ b/benchmarks/common.py
@@ -763,45 +763,32 @@ def null_experiment(args, model_iter_fn, model, example_inputs):
     return []
 
 
-def cast_to_fp16(model, inputs):
+def cast_to(dtype, model, inputs):
     # cast model and inputs to fp16
-    model = model.half()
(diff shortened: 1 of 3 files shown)


benchmarks/common.py
```diff
@@ -742,6 +742,21 @@ def cast_to_fp16(model, inputs):
     return model, inputs
 
 
+def cast_to_fp64(model, inputs):
+    # cast model and inputs to fp16
+    model = model.to(torch.float64)
+
+    inputs = tree_map(
+        lambda x: x.to(torch.float64)
+        if getattr(x, "dtype", None) == torch.float16
```
Chillee, line 751: maybe just do something like
```
if x.is_floating_point()
```
?

jansel (approved):

PR #858, opened 2022-08-17
[Needs more testing] Remove std::atomic_ref in favor of cpu_atomic_add_float. Lower stdlib to 14. Random C++ fixes.

voznesenskym:
Title. Needs testing on torchbench.


jansel (approved):

torchinductor/codecache.py
```diff
@@ -138,6 +138,7 @@ def load(cls, source_code):
                 try:
                     subprocess.check_output(cmd, stderr=subprocess.STDOUT)
                 except subprocess.CalledProcessError as e:
+                    print(f"SUBPROCESS ERROR: {' '.join(cmd)} ")
```
voznesenskym, line 141: Makes it a tad easier to just copy the command out of here.

torchinductor/config.py
```diff
@@ -67,9 +67,6 @@ class cpp:
         "g++-12",
         "g++-11",
         "g++-10",
-        "clang++-12",
-        "clang++-11",
-        "clang++-10",
```
voznesenskym, line 72: This is due to issues running the argmin code we generate for cpp:

```

#include "/tmp/torchinductor_voz/in/cinatgwny34da5ivic2yooram3cqpirmraxdww2tir7kgecb4ef2.h"

struct IndexValue_1 {
    size_t index; float value;
};

                       float* __restrict__ out_ptr0,
                       float* __restrict__ out_ptr1,
                       const long ks0,
                       const long ks1)
{
    {

        IndexValue_1 tmp1{0, 3.0f};

        #pragma omp declare reduction(argmax : struct IndexValue_1 :            omp_out.value = omp_in.value < omp_out.value ? omp_out.value : omp_in.value,            omp_out.index = omp_in.value < omp_out.value ? omp_out.index : omp_in.index)        	initializer(omp_priv = {0, std::numeric_limits<float>::min()})
        struct IndexValue_2 {size_t index; float value;};
        IndexValue_2 tmp2{0, 4.0f};
        #pragma omp declare reduction(argmin : struct IndexValue_2 :            omp_out.value = omp_in.value > omp_out.value ? omp_out.value : omp_in.value,            omp_out.index = omp_in.value > omp_out.value ? omp_out.index : omp_in.index)        	initializer(omp_priv = {0, std::numeric_limits<float>::max()})
        #pragma omp parallel num_threads(8)
        {
            #pragma omp for nowait reduction(argmax:tmp1) reduction(argmin:tmp2)
            for(long i0=0; i0<ks0*(ks1*ks1); ++i0)
            {
                auto tmp0 = in_ptr0[i0];
                if (tmp1.value < tmp0) {
                    tmp1.index = i0; tmp1.value = tmp0;
                }
                if (tmp2.value > tmp0) {
                    tmp2.index = i0; tmp2.value = tmp0;
                }
            }
        }
        out_ptr0[0] = tmp1.index;
        out_ptr1[0] = tmp2.index;
    }
}
```

voznesenskym:
This get rekt on CI, but passes on my machine... we may need to put clang back for CI?

torchinductor/codegen/cpp.py
```diff
@@ -321,14 +326,7 @@ def store(self, name, index, value, mode=None):
...
             else:
-                # Note atomic_ref requires C++20 and certain processor features
-                self.stores.writeline(
-                    name,
-                    "static_assert(std::atomic_ref<"
-                    + f"std::remove_pointer_t<decltype({var})>"
-                    + ">::is_always_lock_free);",
-                )
-                line = f"std::atomic_ref({var}[{cexpr(index)}]) += {value};"
+                line = f"cpu_atomic_add_float(&{var}[{cexpr(index)}], {value});"
```
voznesenskym, line 329: a memory layout and sizeof() assert wouldn't hurt here, or upstreamed into cpu_atomic_add_float as its a little loosey-goosey around some edge cases between the input types AND has float assumptions. If we ever need more than floats here (maybe we do already?) we'll need to wrap this into a template and invoke the right one, and potentially provide our own for int.

voznesenskym:
@jansel  can you re-review? it's changed since you stamped.

torchinductor/codecache.py
jansel: `exc.CppCompileError` will print out the command in its message

torchinductor/codegen/cpp.py
```diff
@@ -62,9 +62,12 @@ def argmax_argmin_prefix(reduction_type, dtype, tmpvar):
     struct_name = f"IndexValue_{index_value_name_counter}"
     index_value_name_counter += 1
 
+    # A small annoyance, due to it being a little cumbersome to just throw {} into strings
+    open_brace = "{"
```
jansel, line 66: String format supports this with `{{`

jansel (changes requested):

jansel:
Looking at cpu_atomic_add_float, it just used std::atomic<unsigned>.  Perhaps we could do:
```
template<T>
T atomic_add(volatile T* addr, T offset) {
    /*
    This usage of std::atomic is explicitly unsupported in the docs.
    Add some static_asserts in the hope it will catch cases where std::atomic can't 
    map this to a single compare and swap (CAS) instruction.
    */
    static_assert(std::atomic<T>::is_always_lock_free);
    static_assert(sizeof(std::atomic<T>) == sizeof(T));
    std::atomic<T>* atomic_addr = reinterpret_cast<std::atomic<T>*>(addr);
    return atomic_addr->fetch_add(offset, std::memory_order::relaxed);
}
```

Please test this and look at the generated assembly to make sure it is not wonky.

voznesenskym:
> Looking at cpu_atomic_add_float, it just used std::atomic<unsigned>.  Perhaps we could do:
> 
> ```
> 
> template<T>
> 
> T atomic_add(volatile T* addr, T offset) {
> 
>     /*
> 
>     This usage of std::atomic is explicitly unsupported in the docs.
> 
>     Add some static_asserts in the hope it will catch cases where std::atomic can't 
> 
>     map this to a single compare and swap (CAS) instruction.
> 
>     */
> 
>     static_assert(std::atomic<T>::is_always_lock_free);
> 
>     static_assert(sizeof(std::atomic<T>) == sizeof(T));
> 
>     std::atomic<T>* atomic_addr = reinterpret_cast<std::atomic<T>*>(addr);
> 
>     return atomic_addr->fetch_add(offset, std::memory_order::relaxed);
> 
> }
> 
> ```
> 
> 
> 
> Please test this and look at the generated assembly to make sure it is not wonky.

Hmmm are you sure? I'm not at a laptop right now but iirc std::atomic<float> is not allowed until cpp 20. We may not be able to do std::atomic<T> above. The aten impl does not either, if I recall.

jansel:
`std::atomic` is C++11
`std::atomic_ref` is C++20

voznesenskym:
> `std::atomic` is C++11
> 
> `std::atomic_ref` is C++20

Even std::atomic<float>? I recall that specialization coming later but I could be mistaken. Will check.

voznesenskym:
Also apologies the first comment for some reason became just the word std::atomic instead of braces float braces because of GH formatting...

voznesenskym:
> > Looking at cpu_atomic_add_float, it just used std::atomic.  Perhaps we could do:
> > ```
> > 
> > template<T>
> > 
> > T atomic_add(volatile T* addr, T offset) {
> > 
> >     /*
> > 
> >     This usage of std::atomic is explicitly unsupported in the docs.
> > 
> >     Add some static_asserts in the hope it will catch cases where std::atomic can't 
> > 
> >     map this to a single compare and swap (CAS) instruction.
> > 
> >     */
> > 
> >     static_assert(std::atomic<T>::is_always_lock_free);
> > 
> >     static_assert(sizeof(std::atomic<T>) == sizeof(T));
> > 
> >     std::atomic<T>* atomic_addr = reinterpret_cast<std::atomic<T>*>(addr);
> > 
> >     return atomic_addr->fetch_add(offset, std::memory_order::relaxed);
> > 
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Please test this and look at the generated assembly to make sure it is not wonky.
> 
> Hmmm are you sure? I'm not at a laptop right now but iirc std::atomic float is not allowed until cpp 20. We may not be able to do std::atomic above. The aten impl does not either, if I recall.
> 
> Edit: GitHub formatting swallowed the word float. For some reason.

https://godbolt.org/z/q6MrT9b4e

voznesenskym:
> > > Looking at cpu_atomic_add_float, it just used std::atomic.  Perhaps we could do:
> > > ```
> > > 
> > > template<T>
> > > 
> > > T atomic_add(volatile T* addr, T offset) {
> > > 
> > >     /*
> > > 
> > >     This usage of std::atomic is explicitly unsupported in the docs.
> > > 
> > >     Add some static_asserts in the hope it will catch cases where std::atomic can't 
> > > 
> > >     map this to a single compare and swap (CAS) instruction.
> > > 
> > >     */
> > > 
> > >     static_assert(std::atomic<T>::is_always_lock_free);
> > > 
> > >     static_assert(sizeof(std::atomic<T>) == sizeof(T));
> > > 
> > >     std::atomic<T>* atomic_addr = reinterpret_cast<std::atomic<T>*>(addr);
> > > 
> > >     return atomic_addr->fetch_add(offset, std::memory_order::relaxed);
> > > 
> > > }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Please test this and look at the generated assembly to make sure it is not wonky.
> > 
> > 
> > Hmmm are you sure? I'm not at a laptop right now but iirc std::atomic float is not allowed until cpp 20. We may not be able to do std::atomic above. The aten impl does not either, if I recall.
> > Edit: GitHub formatting swallowed the word float. For some reason.
> 
> https://godbolt.org/z/q6MrT9b4e

It looks like we cannot trivially just do this for floats, as suspected above. I am happy to make a template version that proxies to either atomic<int> add or the aten float impl. Or just leave it as floats only for now and wait for it to fail then add an easy int impl.

jansel:
```
template <typename T>
struct AsIntegerType { typedef T type; };
template <>
struct AsIntegerType<float> { typedef uint32_t type; };
template <>
struct AsIntegerType<double> { typedef uint64_t type; };

template<typename T>
void atomic_add(volatile T* addr, T offset) {
    typedef typename AsIntegerType<T>::type alt_type;
    static_assert(sizeof(std::atomic<alt_type>) == sizeof(T), "std::atomic issue");
    alt_type expected;
    alt_type desired;
    std::atomic<alt_type>* atomic_addr = (std::atomic<alt_type>*)addr;
    do {
        T val = *addr;
        reinterpret_cast<T*>(&expected)[0] = val;
        reinterpret_cast<T*>(&desired)[0] = val + offset;
    } while(!atomic_addr->compare_exchange_weak(expected, desired, std::memory_order_relaxed));
}

```

voznesenskym:
> ```
> 
> template <typename T>
> 
> struct AsIntegerType { typedef T type; };
> 
> template <>
> 
> struct AsIntegerType<float> { typedef uint32_t type; };
> 
> template <>
> 
> struct AsIntegerType<double> { typedef uint64_t type; };
> 
> 
> 
> template<typename T>
> 
> void atomic_add(volatile T* addr, T offset) {
> 
>     typedef typename AsIntegerType<T>::type alt_type;
> 
>     static_assert(sizeof(std::atomic<alt_type>) == sizeof(T), "std::atomic issue");
> 
>     alt_type expected;
> 
>     alt_type desired;
> 
>     std::atomic<alt_type>* atomic_addr = (std::atomic<alt_type>*)addr;
> 
>     do {
> 
>         T val = *addr;
> 
>         reinterpret_cast<T*>(&expected)[0] = val;
> 
>         reinterpret_cast<T*>(&desired)[0] = val + offset;
> 
>     } while(!atomic_addr->compare_exchange_weak(expected, desired, std::memory_order_relaxed));
> 
> }
> 
> 
> 
> ```
> 
> 

image

test/test_torchinductor.py
```diff
@@ -2458,6 +2458,7 @@ def fn(a, dim, index, b):
                 torch.ones(64, 512),
             ],
         )
+        self.assertTrue(False)
```
jansel, line 2461: ?

torchinductor/codecache.py
```diff
@@ -142,6 +142,9 @@ def load(cls, source_code):
 
             cls.cache[key] = cdll.LoadLibrary(output_path)
             cls.cache[key].key = key
+
+        print("LOADING")
```
jansel, line 146: Remove

torchinductor/codegen/cpp_prefix.h
```diff
@@ -0,0 +1,55 @@
...
+
+    alt_type desired;
+
+    std::atomic<alt_type>* atomic_addr = (std::atomic<alt_type>*)addr;
+    do {
+        T val = *addr;
+        reinterpret_cast<T*>(&expected)[0] = val;
+        reinterpret_cast<T*>(&desired)[0] = val + offset;
+    } while(!atomic_addr->compare_exchange_weak(expected, desired, std::memory_order_relaxed));
+}
```
jansel, line 55: Newline end of file.  Run clang-format on this.

jansel (changes requested):

test/cpp/.gitignore
```diff
@@ -0,0 +1,13 @@
...
+CMakeScripts
+Testing
+Makefile
+cmake_install.cmake
+install_manifest.txt
+compile_commands.json
+CTestTestfile.cmake
+_deps
+lib
+bin
```
jansel, line 13: newline end of file

test/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+
+################################
+# Tests
+################################
+
+# TODO(voz): This is a little assumptive of just this one test, rewrite with real dir includes
+include_directories(${ATEN_INCLUDE})
+add_executable(test_cpp_prefix test_cpp_prefix.cpp ../../torchinductor/codegen/cpp_prefix.h)
+target_link_libraries(test_cpp_prefix gtest gtest_main)
+add_test(NAME test_cpp_prefix COMMAND test_cpp_prefix)
```
jansel, line 47: newline end of file

test/cpp/test.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+set -euo pipefail
+IFS=$'\n\t'
+
+cmake . -DATEN_INCLUDE:PATH=$(python -c "import torch; from torch.utils import cpp_extension; print(cpp_extension.include_paths()[0])")
+make 
+./test/bin/test_cpp_prefix
```
jansel, line 7: newline end of file

test/cpp/test_cpp_prefix.cpp
```diff
@@ -0,0 +1,24 @@
...
+    EXPECT_EQ(x, 100.0f); 
+}
+
+TEST(testCppPrefix, testAtomicAddI64)
+{
+    int64_t x = 0.0;
+    int64_t y = 100.0;
+    atomic_add(&x, y);
+    EXPECT_EQ(x, 100); 
+}
```
jansel, line 24: newline end of file, clang-format

torchinductor/codecache.py
```diff
@@ -142,6 +142,8 @@ def load(cls, source_code):
 
             cls.cache[key] = cdll.LoadLibrary(output_path)
             cls.cache[key].key = key
+
+        print(cls.cache[key], input_path)
```
jansel, line 146: remove print

torchinductor/codegen/cpp.py
```diff
@@ -97,43 +99,12 @@ def argmax_argmin_prefix(reduction_type, src_dtype, tmpvar):
...
-                at::Philox4_32 engine(seed, 0, offset);
-                return engine.randn(10);
-            }
-            """
-        ),
-        "h",
-    )
+    path = Path(__file__).parent / "cpp_prefix.h"
+    with path.open() as f:
+        _, filename = codecache.write(
```
jansel, line 104: Add a comment saying we make a copy so caches get invalidated properly

torchinductor/codegen/cpp_prefix.h
```diff
@@ -0,0 +1,50 @@
...
+  alt_type desired;
+
+  std::atomic<alt_type> *atomic_addr = (std::atomic<alt_type> *)addr;
+  do {
+    T val = *addr;
+    reinterpret_cast<T *>(&expected)[0] = val;
+    reinterpret_cast<T *>(&desired)[0] = val + offset;
+  } while (!atomic_addr->compare_exchange_weak(expected, desired,
+                                               std::memory_order_relaxed));
+}
```
jansel, line 50: Newline end of file, clang-format should fix this

jansel (changes requested):

torchinductor/codegen/cpp.py
voznesenskym: Sorry, can you explain this to me? who is the comment for?

jansel: Ah perhaps you got this right by mistake.  This code basically just copies `cpp_prefix.h` to our cache directory.

Someone reading this might think they could just `#include ".../cpp_prefix.h"` directly.  This would be incorrect because if you modify `cpp_prefix.h` it won't invalidate the code cache.

jansel (approved):

torchinductor/codegen/cpp.py
voznesenskym: > Ah perhaps you got this right by mistake. 

The only way I get things right. 

> This code basically just copies cpp_prefix.h to our cache directory.

The old code wrote the contents of a cpp file, from a string. The new code write the contents of a cpp file, from a file.  

I am not sure what you want me to change here.

voznesenskym:
Thanks for stamp, still havent given it a full torchbench pass, so I wont be pushing for a bit.

torchinductor/codegen/cpp.py
jansel: This code is correct, don't worry about it.  I was just suggesting a comment explaining why we need it this way.

voznesenskym:
Looks fine on bench. Need to import this to fbcode and make sure it doesn't break existing integrations.

voznesenskym:
Looked fine in fbcode. Will rebase and retest and maybe land if nothing else arises.

Issue #834, opened 2022-08-16
Issues with argmax and argmin


ngimel:
#833 revealed that
1) argmax is not codegened, and instead is sent to fallback, see example call function from the test
```
def call(arg0_1):
    arg0_1_size = arg0_1.size()
    s0 = arg0_1_size[0]
    buf0 = aten.argmax(arg0_1, 0)
    assert buf0.size() == (144, )
    assert buf0.stride() == (1, )
    buf2 = aten.argmax(arg0_1, 1)
    assert buf2.size() == (144, )
    assert buf2.stride() == (1, )
    buf1 = empty_strided((144, ), (1, ), device='cuda', dtype=torch.float32)
    buf3 = empty_strided((144, ), (1, ), device='cuda', dtype=torch.float32)
    kernel0[grid(144)](arg0_1, buf1, buf3, 144, 144)
    return (buf0, buf1, buf2, buf3, )
```
2) argmin produces result of the same type as input, whereas it should produce int64 result

cc @desertfire

desertfire:
Close this one, and keep https://github.com/pytorch/torchdynamo/issues/859 open but as low priority.

PR #860, opened 2022-08-17
[inductor] Fix argmax/argmin return type bug

desertfire:
Summary: Fix by tracking both the src dtype and the result dtype.
There is one remaining issue, https://github.com/pytorch/torchdynamo/issues/859

### DIFF
--- a/torchinductor/ir.py
+++ b/torchinductor/ir.py
@@ -359,6 +359,7 @@ class Scatter(Pointwise):
 class Reduction(Loops):
     reduction_ranges: List[Expr]
     reduction_type: str
+    # self.dtype represents the dst dtype
     src_dtype: torch.dtype
 
     def __str__(self):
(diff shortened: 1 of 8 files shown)


desertfire:
For https://github.com/pytorch/torchdynamo/issues/834

jansel (approved):

torchinductor/ir.py
```diff
@@ -596,30 +606,32 @@ def fn(index):
...
                 reduction_ranges,
                 reduction_type,
+                src_dtype,
             )
         )
 
     @staticmethod
     def default_value(reduction_type, dtype):
+        if reduction_type in {"max", "argmax"}:
+            return float("-inf") if is_float_dtype(dtype) else -(2**31)
```
ngimel, line 621: This should depend on the dtype? E.g. for max/argmax on uint8 tensor this will overflow, for long tensor potentially produce wrong results?

desertfire: You are right. I initially used 2**63, but it is better to write a util function for this (let me update in another PR), or is there any existing python util for it?

ngimel: There's `torch.iinfo(dtype).max`

PR #863, opened 2022-08-17
Add lru_cache to some hot spots related to sympy manipulation

Chillee:
Reduces compilation time (assuming no autotuning) by about 20-30%.

For ` time  ./benchmarks/torchbench.py -dcuda --inductor --float16 --training --only <model>`

Resnet18: 40s => 30s
densenet121: 2m32s => 2m
hf_Bert: 1m20 => 55s

### DIFF
--- a/torchinductor/sizevars.py
+++ b/torchinductor/sizevars.py
@@ -10,6 +10,7 @@ from sympy import Expr
 from sympy import Integer
 from sympy import Symbol
 
+from .utils import freeze_inputs
 from .virtualized import V
 
 log = logging.getLogger(__name__)
@@ -60,6 +61,8 @@ class SizeVarAllocator(object):
     def simplify(self, expr):
         return sympy.expand(expr).subs(self.replacements)
 
+    @freeze_inputs
+    @functools.lru_cache(256)
     def simplify_with_ranges(self, expr, var_ranges):
         """
         Simplify indexing expression with knowledge of the ranges of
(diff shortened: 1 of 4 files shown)


torchinductor/sizevars.py
```diff
@@ -348,6 +348,7 @@ def codegen_shape_tuple(self, shape):
         return f"({', '.join(parts)})"
 
 
+@functools.lru_cache(None)
```
desertfire, line 351: Any other hot functions in this file?

Chillee: Not as hot I believe.

We can shave off about another 2-3% by caching on `simplify_sizes_with_ranges`, but it requires a bit more code change due to taking in dicts.

jansel: We should have some limit on the cache size other than `None`, perhaps 
```suggestion
@functools.lru_cache(128)
```

Chillee: Seems fine to me.

With `cache_size=None`, on densenet:

```
CacheInfo(hits=32555, misses=2486, maxsize=None, currsize=2485)
``

With `cache_size=256`

```
CacheInfo(hits=32023, misses=3072, maxsize=256, currsize=256)
```

jansel (approved):

Chillee:
I wrapped some more things in `lru_cache` and added a wrapper to `utils.py` to "freeze" inputs.

Issue #862, opened 2022-08-17
AttributeError: 'ComputedBuffer' object has no attribute 'iter_reordering_reindex'


desertfire:
Repro:
```
benchmarks/timm_models.py --training --float32 --use-eval-mode --inductor -k jx_nest_base
```
Error:
```
Traceback (most recent call last):
  File "/fsx/users/binbao/torchdynamo-tip/benchmarks/common.py", line 973, in run_one_model
    new_result = accuracy_model_iter_fn(model, example_inputs)
  File "/fsx/users/binbao/torchdynamo-tip/torchdynamo/eval_frame.py", line 152, in _fn
    return fn(*args, **kwargs)
  File "benchmarks/timm_models.py", line 304, in forward_and_backward_pass
    def forward_and_backward_pass(self, mod, inputs, collect_outputs=True):
  File "benchmarks/timm_models.py", line 304, in forward_and_backward_pass
    def forward_and_backward_pass(self, mod, inputs, collect_outputs=True):
  File "benchmarks/timm_models.py", line 308, in forward_and_backward_pass
    pred = mod(*cloned_inputs)
  File "/fsx/users/binbao/pytorch-release/torch/nn/modules/module.py", line 1185, in _call_impl
    return forward_call(*input, **kwargs)
  File "/fsx/users/binbao/conda/envs/release/lib/python3.8/site-packages/timm/models/nest.py", line 371, in forward
    def forward(self, x):
  File "/fsx/users/binbao/torchdynamo-tip/torchdynamo/eval_frame.py", line 152, in _fn
    return fn(*args, **kwargs)
  File "/fsx/users/binbao/pytorch-release/functorch/functorch/_src/aot_autograd.py", line 795, in forward
    return compiled_f(
  File "/fsx/users/binbao/pytorch-release/functorch/functorch/_src/aot_autograd.py", line 781, in new_func
    compiled_fn = create_aot_dispatcher_function(
  File "/fsx/users/binbao/pytorch-release/functorch/functorch/_src/aot_autograd.py", line 360, in create_aot_dispatcher_function
    aot_dispatch_autograd(flat_fn, fake_flat_tensor_args, aot_config)
  File "/fsx/users/binbao/pytorch-release/functorch/functorch/_src/aot_autograd.py", line 259, in aot_dispatch_autograd
    compiled_fw = aot_config.fw_compiler(fw_module, flat_args)
[...]

PR #864, opened 2022-08-17
[inductor] Fix for fill_(x, aten.lift_fresh(0.0))

jansel:
`aten.lift_fresh()` was causing the fill_value to be a non-constant.

### DIFF
--- a/torchinductor/ir.py
+++ b/torchinductor/ir.py
@@ -1783,9 +1783,6 @@ class ComputedBuffer(Buffer):
 
         # the reordering_reindex in reads' simplify_reorder_and_tile
         reordering_reindex = [same_reorder(range(len(index_vars)))] * len(memory_addrs)
-        for i, reads_buf in enumerate(reads_bufs):
-            if isinstance(reads_buf, ComputedBuffer):
-                reordering_reindex[i] = reads_buf.iter_reordering_reindex
 
         def simplify_and_reorder(x_vars, sizes, reordering_reindex=None):
             sizes, reindex0, reindex1 = self._apply_loop_reordering(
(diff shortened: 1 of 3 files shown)


torchinductor/ir.py
```diff
@@ -1783,9 +1783,6 @@ def simplify_and_reorder(self):
 
         # the reordering_reindex in reads' simplify_reorder_and_tile
         reordering_reindex = [same_reorder(range(len(index_vars)))] * len(memory_addrs)
-        for i, reads_buf in enumerate(reads_bufs):
-            if isinstance(reads_buf, ComputedBuffer):
-                reordering_reindex[i] = reads_buf.iter_reordering_reindex
```
jansel, line 1788: @pyjhzwh I was hitting the error:
```
  File "/home/jansel/torchdynamo/torchinductor/scheduler.py", line 846, in __init__
    self.nodes.append(SchedulerNode(self, node, group_fn))
  File "/home/jansel/torchdynamo/torchinductor/scheduler.py", line 254, in __init__
    ) = node.simplify_and_reorder()
  File "/home/jansel/torchdynamo/torchinductor/ir.py", line 1788, in simplify_and_reorder
    reordering_reindex[i] = reads_buf.iter_reordering_reindex
AttributeError: 'ComputedBuffer' object has no attribute 'iter_reordering_reindex'
```

Is this the right fix for that?

jansel: #862 seems the same

pyjhzwh: Add try.. except block would solve this, #862 "AttributeError" would be gone
```
                try: 
                    reordering_reindex[i] = reads_buf.iter_reordering_reindex
                except AttributeError:
                    pass
```

desertfire: The root cause is a use node is iterated before a def node, so should we topo-sort `self.nodes` and keep it in the sorted order? https://github.com/pytorch/torchdynamo/blob/734141f29cf3dcc32a1a30695518063b19a5e4d6/torchinductor/scheduler.py#L838

pyjhzwh: Sort by node's name should be ok. https://github.com/pytorch/torchdynamo/blob/66da4325abe6af7c59841ba0c8889cef65e094df/torchinductor/scheduler.py#L114

desertfire (approved):

PR #865, opened 2022-08-17
Fix type promotion for elementwise lowerings

ngimel:
Per title

### DIFF
--- a/torchinductor/lowering.py
+++ b/torchinductor/lowering.py
@@ -154,7 +154,7 @@ def _register_lowering(aten_fn, decomp_fn, broadcast, type_promote):
         assert not any(isinstance(x, TensorBox) for x in kwargs.values())
 
         if type_promote and indices:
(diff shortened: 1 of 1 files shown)


Chillee (approved):
LGTM

torchinductor/lowering.py
```diff
@@ -149,9 +154,7 @@ def wrapped(*args, **kwargs):
         assert not any(isinstance(x, TensorBox) for x in kwargs.values())
 
         if type_promote and indices:
-            dtype = functools.reduce(
-                torch.promote_types, [args[i].get_dtype() for i in indices]
-            )
+            dtype = get_promoted_dtype(*[args[i] for i in indices])
```
desertfire, line 157: `args` are filtered to include `TensorBox` only at line 152, but we should include scalars as well when computing the promoted type.

jansel (approved):

desertfire (approved):

PR #866, opened 2022-08-17
Add env caching for CircleCI

desertfire:


jansel (approved):

.circleci/config.yml
```diff
@@ -31,6 +26,19 @@ jobs:
             wget -O torchbenchmark.tar.bz2 "https://drive.google.com/u/0/uc?id=1KvYsqipsvvv3pnNkJzME0iTemDZe0buC&export=download&confirm=t"
```
desertfire, line 26: @jansel , is it ok to include `1KvYsqipsvvv3pnNkJzME0iTemDZe0buC` as part of the cache key? Or is it a better way to checksum this `torchbenchmark.tar.bz2`?

voznesenskym: where does this id come from?

jansel: You can cache it.  If it changes we can update the caching code.

PR #867, opened 2022-08-17
Wrap torch.autograd.profiler.* as FakeContextWrappingVariable and do nothing

yanboliang:


torchdynamo/variables/misc.py
```diff
@@ -292,51 +292,18 @@ def fn_name(self):
...
-        if self.entered:
-            return "torch.ops.profiler._record_function_enter"
-        else:
-            return "torch.ops.profiler._record_function_exit"
+        return variables.ConstantVariable(None, **VariableTracker.propagate(self))
 
-    def _initial_value(self):
-        return self.target_value
+    def fn_name(self):
+        return "autograd.profiler.profile"
```
yanboliang, line 306: This is sort of a tricky way as ```FakeContextWrappingVariable``` can generated by all ```with``` functions we want to ignore potentially, so this can be any function generated a ```FakeContextWrappingVariable```. But since it's a fake one, so I think anyone is ok.

jansel (approved):

anijain2305:
@jansel @yanboliang Should we just cause a graph break on `torch.autograd.porfile.*` calls?

PyTorch tests have started failing for code like this

https://github.com/pytorch/pytorch/blob/master/test/test_jit.py#L1868-L1871
image

I could hide this PR behind a flag if necessary. But wondering if we should just graph break and handle this in 2.x

yanboliang:
The major reason of making the decision to ignore ```torch.autograd.profiler.*``` is most backends not supporting this functionality, so it makes sense to skip these Pytorch tests? I think we are always trying to avoid graph break.

jansel:
We should not graph break on profiler ops.

1) Adding profiling annotations should not slow down your program.  Graph break would cause this. 
2) Even if we graph break, the profiler still won't work because we compile away the aten calls.

I'd be fine making an off-by-default flag to capture profiler ops.  Eventually we need to add profiler support.

PR #869, opened 2022-08-17
add some type annotations and fix some mypy errors

Chillee:
Do we eventually want to turn on mypy as a lint?


ezyang (approved):
I think it's a good idea but it's hard to retroactively turn this sort of thing on

Chillee:
@ezyang I think it's ... doable. Especially if you do it one file at at a time.

voznesenskym:
Pyre can do a good amount of this automatically, we should try that.

voznesenskym (approved):

jansel (approved):

PR #877, opened 2022-08-18
Add mypy to lint and add dependencies.py

Chillee:
* #880
* __->__ #877

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,7 @@
 PY_FILES := $(wildcard *.py) $(wildcard torchdynamo/*.py) $(wildcard torchdynamo/*/*.py) \
             $(wildcard test/*.py) $(wildcard torchinductor/*.py) $(wildcard torchinductor/*/*.py) \
             $(wildcard benchmarks/*.py) $(wildcard benchmarks/*/*.py) $(wildcard .circleci/*.py)
+MYPY_FILES := torchinductor/dependencies.py
 C_FILES := $(wildcard torchdynamo/*.c torchdynamo/*.cpp)
 CLANG_TIDY ?= clang-tidy-10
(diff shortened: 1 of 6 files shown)


ezyang (approved):
it suuure would be nice to start using lintrunner cough cough (@suo)

Makefile
```diff
@@ -36,6 +37,7 @@ lint:
 	black --check --diff $(PY_FILES)
 	isort --check --diff $(PY_FILES)
 	flake8 $(PY_FILES)
+	mypy --show-error-codes $(MYPY_FILES)
```
voznesenskym, line 40: Does this pass CI? I could imagine this causes it to fail for a while. Maybe to keep lint passing, add another make target?

Chillee: I only added `dependencies.py` for now.

torchinductor/dependencies.py
```diff
@@ -14,23 +20,25 @@
 
 log = logging.getLogger(__name__)
 
+Dep = Union["MemoryDep", "StarDep"]
```
Chillee, line 23: What do folks think?

MemoryDep and StarDep seem to be meaningfully different (i.e. StarDep does not have the notion of an index or a size), and there's a bunch of places that currently are annotated as "MemoryDep" that I suspect can actually hold StarDep too.

desertfire: It makes sense, but I would prefer to explicitly write `Dep` as a base class.

jansel: +1 to add a base class

jansel (approved):

Chillee:
cc: @ezyang 🤔 

image

PR #884, opened 2022-08-18
Add col2im_backward/im2col_backward decomposition

desertfire:
Summary: decompose col2im_backward/im2col_backward, although the underlining fold/unfold is still not supported.


jansel (approved):
Delete the fallbacks if there are any

Chillee:
col2im_backward and im2col_backward are just `im2col` and `col2dim` respectively :)

desertfire:
> col2im_backward and im2col_backward are just `im2col` and `col2dim` respectively :)

Yep, and also `im2col` and `col2im` are just special `unfold` and `fold` respectively. Should we decompose `im2col` and `col2im` in core as well?

Chillee:
@desertfire I think `im2col` and `col2im` are actually the lower-level ops in core. We should definitely decompose `im2col`/`col2im`, but I think it's not super trivial.

PR #893, opened 2022-08-18
Deal with missing ProxyTensor class

ezyang:


jansel (approved):

ezyang:
This patch is very dodgy and it would be better if you reset the random state at the start of tracing, so you don't have to lazily do it on the inside

PR #895, opened 2022-08-18
Add dynamo.explain() to help explain and fix graph breaks

voznesenskym:
```py
from typing import List

import torch

import torchdynamo


    print("my_compiler() called with FX graph:")
    gm.graph.print_tabular()
    return gm.forward  # return a python callable


def toy_example(a, b):
    x = a / (torch.abs(a) + 1)
    print("woo")
    if b.sum() < 0:
        b = b * -1
    return x * b

explained = torchdynamo.explain(toy_example, torch.randn(10), torch.randn(10))
print(explained[0])

"""
Dynamo produced 3 graphs, with 2 graph break and 6 ops. 
 Break reasons: 

1. call_function BuiltinVariable(print) [ConstantVariable(str)] {} 
   File "t2.py", line 16, in toy_example
    print("woo")
 
2. generic_jump 
   File "t2.py", line 17, in toy_example
    if b.sum() < 0:
 """
 ```

### DIFF
--- a/torchdynamo/eval_frame.py
+++ b/torchdynamo/eval_frame.py
@@ -289,7 +289,7 @@ def get_compiler_fn(compiler_fn):
     return compiler_fn
 
 
-def optimize(backend, nopython=False):
+def optimize(backend, nopython=False, guard_export_fn=None):
     """
     The main entrypoint of TorchDynamo.  Do graph capture and call
     backend() to optimize extracted graphs.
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by wconstab** [comment]
looks good, i think your API is nicer

but it overlaps a bit with CompileProfiler, you might want to remove that and update docs that point to it or make them use some of the same infra
https://github.com/pytorch/torchdynamo/blob/main/torchdynamo/utils.py#L653

**Comment by voznesenskym** [comment]
> looks good, i think your API is nicer
> 
> but it overlaps a bit with CompileProfiler, you might want to remove that and update docs that point to it or make them use some of the same infra https://github.com/pytorch/torchdynamo/blob/main/torchdynamo/utils.py#L653

TIL about CompileProfiler - I can do any deduping in subsequent PR if you don't mind - CompileProfiler exposes `num_recompiles` which I kind of don't want to here, as this is meant to be a tool purely for users to close graph breaks and I'd rather they not think about recompiles for now. 

I think we can make `CompileProfiler` wrap around `torchdynamo.explain()` and offer (1) an API for `.report()` and `.metrics()` as it does now and (2) give richer detail on recompiles.

**Comment by wconstab** [comment]
> I think we can make CompileProfiler wrap around torchdynamo.explain() and offer (1) an API for .report() and .metrics() as it does now and (2) give richer detail on recompiles.

yea that sounds good.  i also wonder if CompileProfiler is harder to use bc of being an object instead of an API, it could be worth making a flag to .explain() that gives more info about recompiles or something.

ultimately i want the tools to be discoverable and intuitive, so its worth playing with the best API combinations and trying to simplify

**Comment by ezyang** [comment]
Cool stuff!

**Comment by ezyang** [comment]
(not qualified to review the logic; but if you want more feedback on the output format I can try)

File: torchdynamo/eval_frame.py [hunk 1]
```diff
@@ -322,12 +322,66 @@ def toy_example(a, b):
...
+        op_count += len(ops)
+        ops_per_graph.append(ops)
+        if gm.compile_subgraph_reason is not None:
+            break_reasons.append(gm.compile_subgraph_reason)
+        return gm.forward
+
+    def guard_export_print(guards):
+        nonlocal out_guards
+        out_guards.append(guards)
+
```
jansel, line 363: Perhaps we should call reset() before and after so this isn't cached.

torchdynamo/eval_frame.py
```diff
@@ -322,12 +322,66 @@ def toy_example(a, b):
...
-        convert_frame.convert_frame(backend, guard_export_fn=None), backend_ctx_ctor
+        convert_frame.convert_frame(backend, guard_export_fn=guard_export_fn),
+        backend_ctx_ctor,
     )
 
 
+def explain(f, *args, **kwargs):
+    import torchdynamo.symbolic_convert as symbolic_convert
+
+    symbolic_convert.explain = True
```
jansel, line 335: Use patch so if we throw an exception you don't leak the old value

voznesenskym: sg

torchdynamo/eval_frame.py
voznesenskym: sure!

anijain2305 (commented):
Really cool. I like the simplicity of the API. 

My only suggestion is for `symbolic_convert.explain`. Maybe we could have a global config somewhere, where we could add this debug flags and start consolidating other debug efforts as well.

voznesenskym:
> Really cool. I like the simplicity of the API.
> 
> My only suggestion is for `symbolic_convert.explain`. Maybe we could have a global config somewhere, where we could add this debug flags and start consolidating other debug efforts as well.

Agreed - we should absolutely consolidate. Do you want this in the config? or elsewhere? I am open to whatever.

jansel (approved):

PR #903, opened 2022-08-19
Support lowering symbolic shapes from tracer.

Chillee:
Relies on https://github.com/pytorch/pytorch/pull/83719

POC.

```
import torchinductor
torchinductor.config.debug = True
torchinductor.config.dynamic_shapes = True

def f(x):
    a = torch.ones(1)
    b = a.expand(x.shape).cos()
    return (b,)

inp = [torch.randn(4, 5)]
res = make_fx(f, tracing_mode="symbolic")(*inp)
res.graph.eliminate_dead_code()
res.recompile()
print(res.code)
compiled = compile_fx_inner(res, inp)

import time
begin = time.time()
for i in range(1, 10):
    out = compiled(torch.ones(i, i))
    print(out)

print(time.time()-begin)
```

### DIFF
--- a/torchinductor/graph.py
+++ b/torchinductor/graph.py
@@ -213,12 +213,6 @@ class GraphLowering(torch.fx.Interpreter):
         if target is operator.getitem and isinstance(args[0], (list, tuple)):
             return super().call_function(target, args, kwargs)
 
-        print(target)
-        if target == torch.ops.aten.size:
-            return args[0].get_size()[args[1]]
-        if target == torch.ops.math.mul:
-            return args[0] * args[1]
-
         if target not in lowerings:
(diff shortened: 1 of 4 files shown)


torchinductor/graph.py
```diff
@@ -208,6 +208,14 @@ def call_function(self, target, args, kwargs):
         if target is operator.getitem and isinstance(args[0], (list, tuple)):
             return super().call_function(target, args, kwargs)
 
+        print(target)
+        if target == torch.ops.math.size:
```
jansel, line 212: I think we could just put these in `lowerings.py` and use the normal lowerings dict.

jansel (approved):

PR #906, opened 2022-08-19
Added support for kwargs in fallbacks (and also a bunch of other QOL stuff)

Chillee:


torchinductor/config.py
```diff
@@ -35,7 +35,7 @@
 realize_bytes_threshold = 2000
 
 # fallback to eager for random/dropout, this is slow but useful for debugging
-fallback_random = False
+fallback_random = True
```
desertfire, line 38: I suppose default this to False has a performance penalty? Have you done any measurement?

jansel: We should leave this off by default

jansel (changes requested):

torchinductor/config.py
Chillee: accident

jansel (approved):

PR #912, opened 2022-08-19
don't decompose addmm

ngimel:
We shouldn't decompose addmm because the eager runs this as a fused operation without intermediate truncation in low precision. In addition to being more accurate, this also improves some benchmarks (timm_vision_transformer 1.03x -> 1.11x, hf_Bert 1.13x -> 1.14x)


pyjhzwh:
If we want to use triton.matmul later, this will prevent it from generating fused triton kernel. (If we properly set triton.mm configs, the speedup seems promising, 10% improvement on fp16 inference). Could we make the decomposition when we use "inductor" as backend and when torchinductor.config.mm != "aten"?

jansel (approved):

torchinductor/ir.py
```diff
@@ -2393,6 +2393,31 @@ def map_args(self):
...
+        a = cls.realize_input(a)
+        b = cls.realize_input(b)
+        a = cls.require_stride1(a)
+        b = cls.require_stride1(b)
+        return MatrixMultiplyAdd(
+            layout=FlexibleLayout(
+                device=a.get_device(),
+                dtype=a.get_dtype(),
+                size=[m] + [n],
+            ),
```
jansel, line 2416:
```suggestion
            ).as_fixed(),
```

PR #913, opened 2022-08-19
[inductor] Add an opt_level config

desertfire:
Summary: Add a config to control inductor optimization levels, and add
a corresponding flag for our benchmarking script.

### DIFF
--- a/torchinductor/utils.py
+++ b/torchinductor/utils.py
@@ -9,6 +9,18 @@ from torch.cuda import synchronize
 from torch.fx.immutable_collections import immutable_dict
 from torch.fx.immutable_collections import immutable_list
 
+from torchinductor import config
+
+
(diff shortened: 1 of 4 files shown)


torchinductor/utils.py
```diff
@@ -9,6 +9,18 @@
 from torch.fx.immutable_collections import immutable_dict
 from torch.fx.immutable_collections import immutable_list
 
+from torchinductor import config
+
+
+def init_opt_level():
```
jansel, line 15: Perhaps we should have `torchinductor.set_optimization_level(0)` instead of `opt_level` as a config.

If the user does:
```
torchinductor.config.opt_level = 0
torchinductor.config.cudagrapha = True
```
They will not get the result they expect.

jansel:

Issue #911, opened 2022-08-19
Using `min` with dynamic shape and a python constant inside  torchdynamo.export fails


tugsbayasgalan:
Below is the repro:
```
torchdynamo.config.dynamic_shapes = True
def f(x, a):
    return min(x.shape[0], a)

traced = torchdynamo.export(f, torch.ones(6), 3)
print(traced)
```
With the error:
```
TypeError: clamp() received an invalid combination of arguments - got (int, max=int), but expected one of:
 * (Tensor input, Tensor min, Tensor max, *, Tensor out)
 * (Tensor input, Number min, Number max, *, Tensor out)
 ```
I think the reason is `min` calls torch.clamp internally (https://github.com/pytorch/torchdynamo/blob/main/torchdynamo/variables/builtin.py#L373). This is problematic in the case of DynamicShapeVariable (which is x right now), because it doesn't resolve to TensorVariable. But the according to pytorch documentation of torch.clamp, the first argument to it should be a tensor (not a number). Adding this line fixes the issue:
```
if isinstance(a, DynamicShapeVariable):
   a = variables.TorchVariable(torch.tensor).call_function(tx, [a], {})
```

cc: @voznesenskym

voznesenskym:
Nice find! Try putting up a PR with the fix and some tests, and if it passes CI, we can see what the team thinks :)

jansel:
This fix seems reasonable to me.  We should do the same for `max`.

PR #917, opened 2022-08-19
Automatically resolve DynamicShapeVariable to TensorVariable when calling torch.clamp

tugsbayasgalan:
Fixes: https://github.com/pytorch/torchdynamo/issues/911

### DIFF
--- a/test/test_misc.py
+++ b/test/test_misc.py
@@ -2079,6 +2079,15 @@ class MiscTests(torchdynamo.testing.TestCase):
         self.assertRaises(torchdynamo.exc.ResetRequired, fn3)
         fn2()
 
+    def test_dynamo_min_operator_with_shape(self):
+        with torchdynamo.optimize("eager", nopython=True):
+
+            def f(x, a):
+                return min(x.shape[0], a)
+
+            result = f(torch.ones(6), 3)
+            self.assertEqual(result, 3)
+
 
 class TestTracer(JitTestCase):
     def test_jit_save(self):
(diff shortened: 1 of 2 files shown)


test/test_export.py
```diff
@@ -472,3 +472,15 @@ def func(a, b, c):
         dynamo_result = out_graph(*flat_input)
 
         self.assertTrue(torchdynamo.utils.same(real_result, dynamo_result))
+
+    def test_min_op(self):
```
voznesenskym, line 476: This is vague, as it uses export as a black box to enforce a guarantee. Instead, check out how our other test are written, in test_misc for example.

voznesenskym (changes requested):

test/test_export.py
```diff
@@ -472,3 +472,15 @@ def func(a, b, c):
...
+        torchdynamo.reset()
+        torchdynamo.config.dynamic_shapes = True
```
voznesenskym, line 478: Dynamic already has a test suite. Write the test for min without this.

voznesenskym:
Nit: pr title should describe the fix

test/test_export.py
tugsbayasgalan: I think Dynamic test suite just runs some test files under dynamic_shape = True mode. Isn't it better if I have actual test case for the min op (say inside test_misc.py). That way it is also tested under dynamic shapes.

voznesenskym: I believe we are saying the same thing.

tugsbayasgalan: addressed

voznesenskym:
Pass lint, otherwise LGTM

voznesenskym (approved):

PR #919, opened 2022-08-19
[inductor] Add torchbench inference and training to CirleCI

desertfire:


ngimel:
How long does CI take with this?

desertfire:
> How long does CI take with this?

`inductor_torchbench_training` takes 31 mins and `inductor_torchbench_inference` takes 26 mins.

PR #931, opened 2022-08-20
FakeTensor should be wrapped as TensorVariable

yanboliang:
This comes from the discussion with @anijain2305 on if ```tensor.grad``` should be a ```TensorVariable```, obviously the answer is YES. However, I found there are many places we wrap tensor as ```UserDefinedObjectVariable```. We dig into this together and found ```FakeTensor``` was not considered as tensor type and wrap as tensor, which should be fixed.

### DIFF
--- a/test/test_misc.py
+++ b/test/test_misc.py
@@ -2052,6 +2052,21 @@ class MiscTests(torchdynamo.testing.TestCase):
             res = fn(x, m)
         self.assertEqual(ref, res)
 
+    def test_tensor_dot_grad_no_graph_break(self):
+        def fn(a, b):
+            y = 3 * a**3 - b**2
+            y.backward(gradient=torch.tensor([1.0, 1.0]))
+            b.grad.zero_()
+            return a.grad, b.grad
+
+        a = torch.tensor([2.0, 3.0], requires_grad=True)
+        b = torch.tensor([6.0, 4.0], requires_grad=True)
+        cnts = torchdynamo.testing.CompileCounter()
+        with torchdynamo.optimize(cnts):
+            _, b_grad = fn(a, b)
+        self.assertTrue(same(b_grad, torch.tensor([0.0, 0.0])))
+        self.assertEqual(cnts.frame_count, 2)
+
     def test_change_backends(self):
         @torchdynamo.optimize("eager", nopython=True)
         def fn1():
--- a/torchdynamo/utils.py
+++ b/torchdynamo/utils.py
@@ -182,9 +182,14 @@ def is_numpy_float_type(value):
 
 def istensor(obj):
     """Check of obj is a tensor"""
-    return istype(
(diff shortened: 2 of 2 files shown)


voznesenskym:
This looks reasonable but I would do a quick double check with @eellison

torchdynamo/utils.py
```diff
@@ -182,9 +182,14 @@ def is_numpy_float_type(value):
 
 def istensor(obj):
     """Check of obj is a tensor"""
-    return istype(
-        obj, (torch.Tensor, torch.nn.Parameter, *config.traceable_tensor_subclasses)
+    tensor_list = (
+        torch.Tensor,
+        torch.nn.Parameter,
```
voznesenskym, line 187: This is a torch tensor subclass - does explicitly listing it if we list tensor do anything if we change the istype to isinstance?

torchdynamo/utils.py
```diff
@@ -182,9 +182,14 @@ def is_numpy_float_type(value):
...
+        *config.traceable_tensor_subclasses,
     )
+    if fake_tensors_available:
+        tensor_list = tensor_list + (torch._subclasses.FakeTensor,)
+    return istype(obj, tensor_list)
```
voznesenskym, line 192: Can we use isinstance? Then we don't need to list all subclasses...

torchdynamo/utils.py
mlazos: I think we want to explicitly register which sublcasses we handle, `isinstance` will mean all subclasses are automatically handled in this way. If people start making a lot of subclasses and we aren't confident we can handle them without tweaks I think an explicit list is better, if not I'm okay with `isinstance`.

yanboliang: Yeah, it depends on if we allow arbitrary subclasses of ```FakeTensor```, I don't have too much context on this. cc @eellison

voznesenskym: It seems like we should support isinstance and then have a reason not to :) do tests and benchmarks pass with just isinstance?

yanboliang: ```istype``` actually calls ```isinstance```

eellison (approved):
LGTM

test/test_misc.py
```diff
@@ -2052,6 +2052,21 @@ def fn(x, m):
             res = fn(x, m)
         self.assertEqual(ref, res)
 
+    def test_tensor_dot_grad_no_graph_break(self):
+        def fn(a, b):
+            y = 3 * a**3 - b**2
+            y.backward(gradient=torch.tensor([1.0, 1.0]))
```
eellison, line 2058: what was failing the is_tensor check here previously?

yanboliang: Actually it's not failure, previously it wraps tensor ```a``` and tensor ```b``` as ```UserDefinedObjectVariable```(the fallback variable), so cause more graph break when accessing ```.grad```.

Issue #759, opened 2022-08-10
[mobilenet_v3_large][prims+nvfuser] ValueError: Received type <class 'NoneType'> that is neither a tensor or a number!


xwang233:
Reproduce:
```python
# python benchmarks/torchbench.py --training -d cuda --fast --prims-nvfuser --skip-accuracy-check --generate-aot-autograd-stats --isolate -k mobilenet_v3_large
cuda train mobilenet_v3_large                 ERROR:root:unhandled error
Traceback (most recent call last):
  File "/opt/pytorch/torchdynamo/benchmarks/common.py", line 879, in run_one_model
    new_result = model_iter_fn(model, example_inputs)
  File "benchmarks/torchbench.py", line 366, in forward_and_backward_pass
    def forward_and_backward_pass(self, mod, inputs, collect_outputs=True):
  File "benchmarks/torchbench.py", line 366, in forward_and_backward_pass
    def forward_and_backward_pass(self, mod, inputs, collect_outputs=True):
  File "benchmarks/torchbench.py", line 370, in forward_and_backward_pass
    pred = mod(*cloned_inputs)
  File "/opt/pytorch/pytorch/torch/nn/modules/module.py", line 1186, in _call_impl
    return forward_call(*input, **kwargs)
  File "/opt/pytorch/vision/torchvision/models/mobilenetv3.py", line 235, in forward
    def forward(self, x: Tensor) -> Tensor:
  File "/opt/pytorch/torchdynamo/torchdynamo/eval_frame.py", line 94, in _fn
    return fn(*args, **kwargs)
  File "/opt/pytorch/pytorch/functorch/functorch/_src/aot_autograd.py", line 696, in forward
    return compiled_f(
  File "/opt/pytorch/torchdynamo/torchdynamo/eval_frame.py", line 94, in _fn
    return fn(*args, **kwargs)
  File "/opt/pytorch/pytorch/functorch/functorch/_src/aot_autograd.py", line 246, in forward
    fw_outs = normalize_as_list(compiled_fw(*flat_tensor_args))
  File "/opt/pytorch/pytorch/torch/fx/graph_module.py", line 655, in call_wrapped
    return self._wrapped_call(self, *args, **kwargs)
  File "/opt/pytorch/pytorch/torch/fx/graph_module.py", line 277, in __call__
    raise e
  File "/opt/pytorch/pytorch/torch/fx/graph_module.py", line 267, in __call__
[...]

PR #932, opened 2022-08-20
Abstracted minifier out to support nvfuser and other compilers.

Chillee:
cc: @jjsjann123 

As an example of how to use this to debug https://github.com/pytorch/torchdynamo/issues/759

1. Add `COMPILER_REPRO_LEVEL=2`, and run `COMPILER_REPRO_LEVEL=2 python benchmarks/torchbench.py --training -d cuda --fast --prims-nvfuser --skip-accuracy-check --generate-aot-autograd-stats -k mobilenet_v3_large`.
2. This dumps out `minifier_launcher.py`. minifier_launcher` should look like this: https://pastebin.com/XL5rqPXf (629 lines!). 
3. Then, run `python minifier_launcher.py`.
image

4. This will kick off the minifier. It uses the function `nvfuser_fails` (https://github.com/pytorch/torchdynamo/pull/932/files#diff-9e5c615b4a4e886c5452657cbd4084e718fc45e25e3cdd1fda1720a140739e9fL162) to determine whether a graph and inputs "fails". This can be modified to say, only allow a graph to fail if you see "ValueError" in the error string.
5. Wait while it reduces the example... In the meantime, it should be writing out checkpoints to `/tmp/minifier_$USER/checkpoints`, with the name being "number of nodes in graph".
image

6. When it finishes, it'll return the FX graph and inputs that cause it to fail. You can also take the repro from the checkpoints.

```
import torch
import torch.fx as fx
from math import inf

# torch version: 1.13.0a0+git079d3f4
# torch cuda version: 11.6
# torch git version: 079d3f4be7b9c5926ead41e7069df50a2c0115f4


# CUDA Info: 
# nvcc: NVIDIA (R) Cuda compiler driver 
# Copyright (c) 2005-2022 NVIDIA Corporation 
# Built on Thu_Feb_10_18:23:41_PST_2022 
# Cuda compilation tools, release 11.6, V11.6.112 
# Build cuda_11.6.r11.6/compiler.30978841_0 

# GPU Hardware Info: 
# NVIDIA A100-SXM4-40GB : 8 

class Repro(torch.nn.Module):



    def forward(self, getitem, mul_tensor):
        add_tensor = torch.ops.aten.add.Tensor(getitem, 3);  getitem = None
        clamp_default = torch.ops.aten.clamp.default(add_tensor, 0);  add_tensor = None
        clamp_default_1 = torch.ops.aten.clamp.default(clamp_default, None, 6);  clamp_default = None
        return (mul_tensor,)
    
args = [((32, 16, 112, 112), (200704, 12544, 112, 1), torch.float32, 'cuda'), ((32, 16, 112, 112), (200704, 12544, 112, 1), torch.float32, 'cuda')]
args = [rand_strided(shape, stride, dtype, device) for shape, stride, dtype, device in args]
mod = make_fx(Repro())(*args)

from torch.fx.passes.backends.nvfuser import NvFuserBackend
nvfuser = NvFuserBackend()

compiled = nvfuser(mod, args)
compiled(*args)
```

And if we run this file, we see the same error.
```
  File "/scratch/chilli/fresh/pytorch/torch/_prims_common/wrappers.py", line 108, in _fn
    promoted_args = {
  File "/scratch/chilli/fresh/pytorch/torch/_prims_common/wrappers.py", line 109, in <dictcomp>
    x: _maybe_convert_to_dtype(bound.arguments[x], compute_dtype)
  File "/scratch/chilli/fresh/pytorch/torch/_prims_common/wrappers.py", line 35, in _maybe_convert_to_dtype
    raise ValueError(
ValueError: Received type <class 'NoneType'> that is neither a tensor or a number
```


torchdynamo/debug_utils.py
```diff
@@ -142,11 +152,18 @@ def isolate_inductor_fails(fx_g, args, env=None):
...
     from torchinductor import config
     from torchinductor.compile_fx import compile_fx_inner
 
-    config.autotune = False
+    config.triton.autotune = False
+
+    try:
+        result = fx_g(*args)
+        assert isinstance(result, (tuple, list))
+        assert not any([isinstance(x, (tuple, list)) for x in result])
```
voznesenskym, line 164: What about sets maps or other collection objects? Maybe instead we can flatten and check that the size stays the same?

torchdynamo/debug_utils.py
```diff
@@ -159,32 +176,70 @@ def inductor_fails(fx_g, args, check_str=None):
...
 
-def dump_to_minify(gm, args):
+def nvfuser_fails(fx_g, args, check_str=None):
+    from torch.fx.passes.backends.nvfuser import NvFuserBackend
+
+    nvfuser = NvFuserBackend()
+
+    try:
+        compile_mod = nvfuser(fx_g, args)
+        compile_mod = compile_mod(*args)
```
voznesenskym, line 186: If args is a single tensor will this unpack it?

torchdynamo/debug_utils.py
```diff
@@ -159,32 +176,70 @@ def inductor_fails(fx_g, args, check_str=None):
...
+    except Exception as e:
+        if check_str is not None and check_str not in repr(e):
```
voznesenskym, line 188: Nit: Maybe add a comment on the API, It's not readily obvious that check str can result in fails returning a false on a mismatch. It almost feels like it maybe should be two APIS, fails/fails_with_message

torchdynamo/debug_utils.py
```diff
@@ -159,32 +176,70 @@ def inductor_fails(fx_g, args, check_str=None):
...
+            return False
+        print(repr(e))
```
voznesenskym, line 190: Use logger?

torchinductor/compile_fx.py
```diff
@@ -106,46 +106,29 @@ def compile_fx_inner(
...
-        raise
+    graph = GraphLowering(gm, num_dynamic_inputs=len(example_inputs))
+    with V.set_graph_handler(graph):
+        wrap(graph.run)(*example_inputs)
+        compiled_fn = wrap(graph.compile_to_fn())
+
+    # make sure it works, causes issues for mutation
+    # compiled_fn(*example_inputs)
+
+    if cudagraphs and set(graph.device_types) == {"cuda"} and not graph.mutated_inputs:
```
voznesenskym, line 117: And cuda is available? Or checked upstream/downstream?

voznesenskym (approved):
Lgtm left some questions / nits

torchdynamo/debug_utils.py
Chillee: This is mimicking the code in TorchInductor's lowering, which assumes the output is a list/tuple of tensors.https://github.com/pytorch/torchdynamo/blob/main/torchinductor/graph.py#L250

torchdynamo/debug_utils.py
Chillee: Similar to above, it's an invariant that args is always a tuple of tensors.

torchdynamo/debug_utils.py
Chillee: Yeah this API is not really that great haha.

torchinductor/compile_fx.py
Chillee: If the graph contains tensors with "cuda types", then cuda is probably available :P

torchdynamo/debug_utils.py
```diff
@@ -159,32 +176,70 @@ def inductor_fails(fx_g, args, check_str=None):
...
-                    isolate_inductor_fails,
-                    dump_state_inductor,
+                from torchdynamo.debug_utils import (
+                    isolate_fails,
+                    dump_state,
                 )
                 from functorch.compile import minifier
 
-                env_variables = {"CUDA_VISIBLE_DEVICES": "1"}
+                env_variables = {{"CUDA_VISIBLE_DEVICES": "1"}}
```
ngimel, line 211: hm, why 1 here?

Chillee: It's useful to run it on a different device to reduce OOM issues (otherwise you're running both the minifier + the checker in the same GPU). In general OOM issues are one thing that we could improve the minifier on.

PR #940, opened 2022-08-22
Fix multi-level import issue

anijain2305:
`__import__` brings in the top level module when fromlist is empty. For example, `__import__` of `scipy.sparse` brings just `scipy`. More details at - https://docs.python.org/3/library/functions.html#import__

However, our handling of `IMPORT_NAME` bytecode did not handle it correctly, and the source was still pointing to `scipy.sparse`.

This should fix one issue in https://github.com/pytorch/pytorch/pull/83575

cc @ezyang

### DIFF
--- a/test/test_repros.py
+++ b/test/test_repros.py
@@ -45,15 +45,6 @@ def has_detectron2():
         return False
 
 
-def has_scipy():
-    try:
-        import scipy.sparse
-
-        return scipy.sparse.coo_matrix is not None
-    except ImportError:
-        return False
-
-
 def _do_paste_mask(masks, boxes, img_h: int, img_w: int, skip_empty: bool = True):
     # from detectron2 mask_ops.py
 
@@ -1499,12 +1490,11 @@ class ReproTests(torchdynamo.testing.TestCase):
 
         self.assertTrue((to_bitmasks(torch.zeros(10)) == torch.ones(10)).all())
 
-    @unittest.skipIf(not has_scipy(), "requires scipy")
-    def test_scipy_import(self):
+    def test_two_level_import(self):
         def fn():
-            import scipy.sparse
+            import torch.fx
 
-            print(scipy.sparse.coo_matrix)
+            print(torch.fx.symbolic_trace)
 
         opt_fn = torchdynamo.optimize("eager")(fn)
         opt_fn()
--- a/test/test_repros.py
+++ b/test/test_repros.py
@@ -1491,13 +1491,40 @@ class ReproTests(torchdynamo.testing.TestCase):
         self.assertTrue((to_bitmasks(torch.zeros(10)) == torch.ones(10)).all())
 
     def test_two_level_import(self):
-        def fn():
+        import logging
+
+        def fn1(x):
+            return torch.sin(x)
+
+        def fn(x):
             import torch.fx
 
-            print(torch.fx.symbolic_trace)
+            _ = torch.fx.symbolic_trace(fn1)
+            return x * 2
 
-        opt_fn = torchdynamo.optimize("eager")(fn)
-        opt_fn()
+        x = torch.randn(10)
+        fn(x)
+        cnt = torchdynamo.testing.CompileCounter()
+        opt_fn = torchdynamo.optimize(cnt)(fn)
+        opt_fn(x)
+        self.assertEqual(cnt.frame_count, 1)
+
+    # TODO - This test fails
+    # def test_two_level_relative_import(self):
+    #     import logging
+    #     def fn1(x):
+    #         return torch.sin(x)
+
+    #     def fn(x):
+    #         from . import test_functions
+    #         return x * 2
+
+    #     x = torch.randn(10)
+    #     fn(x)
+    #     cnt = torchdynamo.testing.CompileCounter()
+    #     opt_fn = torchdynamo.optimize(cnt)(fn)
+    #     opt_fn(x)
+    #     self.assertEqual(cnt.frame_count, 1)
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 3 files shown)


torchdynamo/symbolic_convert.py
```diff
@@ -453,7 +453,12 @@ def IMPORT_NAME(self, inst):
             fromlist=fromlist.as_python_constant(),
             level=level.as_python_constant(),
         )
-        source = self.import_source(module_name)
+        # __import__ returns the top level package name, so the source has to be
+        # set correctly.
```
ezyang, line 457: I... don't understand this comment lol

torchdynamo/symbolic_convert.py
```diff
@@ -453,7 +453,12 @@ def IMPORT_NAME(self, inst):
...
+        if fromlist.as_python_constant() is None and hasattr(value, "__name__"):
```
ezyang, line 458: CSE this with the fromlist above?

anijain2305: Done

torchdynamo/symbolic_convert.py
anijain2305: Added more description here. Copied from here - https://docs.python.org/3/library/functions.html#import__

torchdynamo/symbolic_convert.py
```diff
@@ -447,15 +447,21 @@ def import_source(self, module_name):
...
-        # __import__ returns the top level package name, so the source has to be
-        # set correctly.
-        if fromlist.as_python_constant() is None and hasattr(value, "__name__"):
+
+        # For __import__, when the name variable is of the form package.module,
+        # normally, the top-level package (the name up till the first dot) is
+        # returned, not the module named by module_name. However, when a
+        # non-empty fromlist argument is given, the module named by name is
+        # returned. Therefore, we set the source correctly here.
+        if fromlist is None and hasattr(value, "__name__"):
```
ezyang, line 464: The docs seem to imply this behavior also would occur if `fromlist` was an empty list? While unlikely, hypothetically someone could emit a bytecode that passes an empty list here rather than None?

ezyang: Regarding the hasattr, if we cannot guarantee that the module has the `__name__` attribute set accurately (and it *is* overrideable, e.g., `import b; b.__name__ = "arf"`, it would seem that it would be better to do this via `module_name`.  It is probably best to replicate the logic in CPython's `__import__` implementation:

```
    if not fromlist:
        # Return up to the first dot in 'name'. This is complicated by the fact
        # that 'name' may be relative.
        if level == 0:
            return _gcd_import(name.partition('.')[0])
        elif not name:
            return module
        else:
            # Figure out where to slice the module's name up to the first dot
            # in 'name'.
            cut_off = len(name) - len(name.partition('.')[0])
            # Slice end needs to be positive to alleviate need to special-case
            # when ``'.' not in name``.
            return sys.modules[module.__name__[:len(module.__name__)-cut_off]]
    elif hasattr(module, '__path__'):

```

test/test_repros.py
```diff
@@ -1490,6 +1499,16 @@ def to_bitmasks(boxes):
...
 
+    @unittest.skipIf(not has_scipy(), "requires scipy")
+    def test_scipy_import(self):
+        def fn():
+            import scipy.sparse
+
+            print(scipy.sparse.coo_matrix)
+
+        opt_fn = torchdynamo.optimize("eager")(fn)
+        opt_fn()
```
ezyang, line 1510: You don't need scipy to trigger this bug, no reason to make the test case specialized for scipy

torchdynamo/symbolic_convert.py
anijain2305: For the `fromlist` empty list, I don't think the Python bytecode can have empty list. I tried a couple of examples
~~~
import dis
def fn1():
    from math import sqrt
dis.dis(fn1)


def fn2():
    import math.sqrt
dis.dis(fn2)
~~~
~~~
  3           0 LOAD_CONST               1 (0)
              2 LOAD_CONST               2 (('sqrt',))
              4 IMPORT_NAME              0 (math)
              6 IMPORT_FROM              1 (sqrt)
              8 STORE_FAST               0 (sqrt)
             10 POP_TOP
             12 LOAD_CONST               0 (None)
             14 RETURN_VALUE

  8           0 LOAD_CONST               1 (0)
              2 LOAD_CONST               0 (None)
              4 IMPORT_NAME              0 (math.sqrt)
              6 STORE_FAST               0 (math)
              8 LOAD_CONST               0 (None)
             10 RETURN_VALUE
~~~


For the hasattr, I am going to replicate the logic. Thanks for passing the pointer.

ezyang: I mean, sure, I agree it's unlikely, but the tweak to handle empty list here is so slight (just do a bool-ish test directly on `fromlist`), there's no downside to doing it. (And you will match the test that `__import__` in cpython does too...)

anijain2305:
@ezyang Thanks for the reviews. I have made the changes.

Also, I found that TorchDynamo was not handling relative imports (`from .abc import xyz`) correctly. This manifests in the form of `level`. I have replicated the `__import__` logic to handle level > 0 scenarios and also added a couple of tests.

PTAL.

torchdynamo/symbolic_convert.py
```diff
@@ -442,15 +442,73 @@ def import_source(self, module_name):
...
+        """
+        bits = package.rsplit(".", level - 1)
+        if len(bits) < level:
+            raise ImportError("attempted relative import beyond top-level package")
+        base = bits[0]
+        return "{}.{}".format(base, name) if name else base
+
+    def calc_package(self):
+        """
+        Copied from the Cpython implementation of __import__
```
ezyang, line 458: I'd leave a link to the implementation in a link, makes it easy for people to crossref.

ezyang (approved):

Issue #945, opened 2022-08-22
clone_tensor(example_value) is fishy


ezyang:
This utility function doesn't preserve strides

```
def clone_tensor(x):
    """Clone the tensor and its gradient"""
    y = x.clone().requires_grad_(x.requires_grad)
    if x.is_leaf and x.grad is not None:
        y.grad = x.grad.clone()
    return y
```

However, we use it in situations where we will immediately record the metadata after, e.g., as seen in `torchdynamo/variables/tensor.py`

```
            # tensor subclasses will not be converted to FakeTensors and need to be cloned
            if not use_fake_tensors or not isinstance(example_value, FakeTensor):
                example_value = clone_tensor(example_value)
            proxy.node.meta["example_value"] = example_value
            specialized_props = cls.specialize(example_value)
```

This means we will specialize stride calls incorrectly. I have a repro for this.

Why are we cloning here at all? It seems fishy.

anijain2305:
Why do we clone?
This allows us to specialize the graph depending on the return value of `TorchVariable` during the bytecode analysis. I think with FakeTensors, we can remove this. But, the code is there as a fallback when FakeTensor has an error.

We could use this clone function instead https://github.com/pytorch/torchdynamo/blob/fc52b39596add70e604b96e2da0e26407c94f149/torchdynamo/utils.py#L263-L282

ezyang:
This causes `PYTORCH_TEST_WITH_DYNAMO=1 python test/test_nn.py -k TestNNDeviceTypeCPU.test_grid_sample_large_index_2d` to fail, though I'm not sure why it doesn't fail on master. Maybe related to https://github.com/pytorch/pytorch/pull/83830

ezyang:
ok patch coming

ezyang:
`clone_input` doesn't work, so I'm waiting on more guidance from the original authors for what to do

jansel:
Could we just have one `clone()` function that preserves both strides and grad?

The reason for cloning is to try to prevent mutation from dynamic_propagation/tracers from leaking into user data.

PR #947, opened 2022-08-22
Record correct specialized stride info when fake tensors disabled

ezyang:
Fixes https://github.com/pytorch/torchdynamo/issues/945

### DIFF
--- a/torchdynamo/variables/tensor.py
+++ b/torchdynamo/variables/tensor.py
@@ -30,7 +30,7 @@ from .. import variables
 from ..exc import TorchRuntimeError
 from ..exc import unimplemented
 from ..source import AttrSource
-from ..utils import clone_tensor
+from ..utils import clone_input
 from ..utils import is_lazy_module
 from ..utils import istype
 from ..utils import product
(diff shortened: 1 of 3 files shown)


torchdynamo/variables/tensor.py
```diff
@@ -154,7 +154,8 @@ def context():
 
             # tensor subclasses will not be converted to FakeTensors and need to be cloned
             if not use_fake_tensors or not isinstance(example_value, FakeTensor):
-                example_value = clone_tensor(example_value)
+                # NB: ensure strides are preserved
+                example_value = clone_input(example_value)
```
anijain2305, line 158: `clone_input` does not copy the gradient of the tensor.
I was thinking of changing `clone_tensor` in utils.py to use `clone_input` instead of using `torch.clone()`.

ezyang: Cloning the gradient here seems unnecessary. Here are the uses of the subsequent `example_value`:

```
            proxy.node.meta["example_value"] = example_value
            specialized_props = cls.specialize(example_value)
            if use_fake_tensors and isinstance(example_value, FakeTensor):
                specialized_props["class_type"] = (
                    torch.nn.Parameter if is_parameter else torch.Tensor
                )
```

Specialize does not check the grad field of the tensor. So the only way the grad could matter is if the proxy.node.meta expected the example value to have the grad field accurately set. But I could easily define the contract for this meta that it does not have to accurately represent grad...

ezyang:
It looks like `clone_input` does not work in enough situations. Need a stronger cloner...

ezyang:
I fixed the quantization problem, please review

jansel (approved):

Issue #938, opened 2022-08-21
Explore TorchIndcutor optimization pass to reorder kernel bodies


jansel:
See #934 for more context.   We found an example of a 10% performance difference from very similar kernels, where the biggest difference seemed to be instruction ordering.

Interesting, my first guess looking at the two kernels is just the ordering of ops.  The loads on the faster kernel are "spread out" while the loads in the slow kernel are "bunched up".

Perhaps we should explore a compiler pass that reorders ops within a kernel.

Our current inductor kernels usually look like:
```
<all of the loads>
<all of the compute>
<all of the stores>
````

When you have indirect loads, it moves those indirect loads into the "compute" section, because they must come after the address computation.  Thus allowing that spread out pattern to be generated.

My thinking of doing that ordering was it makes compiler analysis easier for Triton/LLVM.  I may have been wrong there.

This is just one theory though, we should test it.

_Originally posted by @jansel in https://github.com/pytorch/torchdynamo/issues/934#issuecomment-1221583598_

PR #949, opened 2022-08-22
[inductor] Major rewrite of scheduler

jansel:
The original TorchInductor scheduler was an fast `O(n)` online algorithm that performed fusions during codegen.

In #693 @Chillee added a prefusion step that moved some fusions ahead of codegen, creating `FusedSchedulerNode`.  This resulted in good speedup due to better fusion choices, but did not fully replace the online scheduling algorithm and relied on that original algorithm for correctness.

This PR completely removes the original TorchInductor online scheduling algorithm, and makes all fusions happen ahead of time using `FusedSchedulerNode`.  This new fusion algorithm should be simpler to understand and more hackable.  There is now a single function to decide what fusions are legal, and another one to rank the priority fusions happen in.

Selected fp16 training speedups on my RTX 3090 (results on a real GPU will be measured later)
- model C 1.02x => 1.45x (was a regression from #693)
- model D 0.98x => 1.58x (was a regression from #693)
- pytorch_struct 1.48x => 1.62x
- timm_efficientnet 1.03x => 1.12x
- hf_GPT2 1.11x => 1.30x


torchinductor/codegen/triton_template.py
```diff
@@ -274,96 +275,76 @@ def call_kernel(self, wrapper, name: str):
...
+    if snode1.group != snode2.group:
+        return False
+    tiling = snode1.get_nodes()[0].node.get_template_tiling()
+    for node in snode1.get_nodes():
+        if not TritonKernel.is_compatible(tiling, node.get_ranges()):
+            return False
+    return True
+
+
+def template_codegen(scheduler, scheduler_node, epilogue):
```
jansel, line 299: cc @pyjhzwh on changes to templates

torchinductor/scheduler.py
```diff
@@ -1000,188 +761,279 @@ def dead_node_elimination(self):
...
+            return False
+
+        visited = set()
+        combined_names = node1.get_names() | node2.get_names()
+        combined_predecessors = (
+            node1.recursive_predecessors | node2.recursive_predecessors
+        ) - combined_names
+        return any(check(self.name_to_fused_node[n]) for n in combined_predecessors)
+
+    def can_fuse(self, node1: BaseSchedulerNode, node2: BaseSchedulerNode):
```
jansel, line 872: This is the key function that decides what can be fused

torchinductor/scheduler.py
```diff
@@ -1000,188 +761,279 @@ def dead_node_elimination(self):
...
+            dep.name for dep in node2.unmet_dependencies - node1.read_writes.writes
+        }
+        if remaining_deps & node1_names:
+            return False  # MemoryDeps didn't match
+        for name in remaining_deps:
+            if node1_names & self.name_to_fused_node[name].recursive_predecessors:
+                return False
+        return True
+
+    def score_fusion(self, node1: BaseSchedulerNode, node2: BaseSchedulerNode):
```
jansel, line 929: This is the key function that decides what order fusions are done in.  cc @ngimel on if this heuristic makes sense.

torchinductor/scheduler.py
```diff
@@ -1000,188 +761,279 @@ def dead_node_elimination(self):
...
+                visited.add(node)
+                return bool(combined_names & node.recursive_predecessors) or any(
+                    check(self.name_to_fused_node[n])
+                    for n in node.recursive_predecessors - combined_predecessors
                 )
+            return False
+
+        visited = set()
+        combined_names = node1.get_names() | node2.get_names()
+        combined_predecessors = (
```
Chillee, line 867: Why can't you just do?
```
combined_names = node1.get_names() | node2.get_names()
combined_predecessors = node1.recursive_predcessors | node2.recursive_predecessors
return combined_names & combined_predecessors
```
?

torchinductor/scheduler.py
```diff
@@ -1000,188 +761,279 @@ def dead_node_elimination(self):
...
+        Determine if it is possible to combine node1 and node2 into a
+        single fused node.
+        """
+        if node1 is node2:
+            return False
+        if isinstance(node1, (ExternKernelSchedulerNode, NopKernelSchedulerNode)):
+            return False
+        if isinstance(node2, (ExternKernelSchedulerNode, NopKernelSchedulerNode)):
+            return False
+        if node2.get_names() & node1.recursive_predecessors:
```
Chillee, line 883: Is this just a sanity check?

torchinductor/scheduler.py
```diff
@@ -1000,188 +761,279 @@ def dead_node_elimination(self):
...
+
+        We can fuse them if all the reads of node2 either match
+        corresponding writes in node1, or are written by nodes that can
+        be scheduled before the fusion of node1 and node2.
+        """
+        node1_names = node1.get_names()
+        remaining_deps = {
+            dep.name for dep in node2.unmet_dependencies - node1.read_writes.writes
+        }
+        if remaining_deps & node1_names:
```
Chillee, line 922: I think this is a little bit subtle. Probably worth adding a comment that above comparison is on *types* of memory reads, IIUC.

torchinductor/scheduler.py
```diff
@@ -1000,188 +761,279 @@ def dead_node_elimination(self):
...
+            return False  # MemoryDeps didn't match
+        for name in remaining_deps:
```
Chillee, line 924: is this redundant with checking whether fusing node1 and node2 would form a cycle?

torchinductor/codegen/cpp.py
```diff
@@ -497,34 +490,51 @@ def __init__(self, scheduler):
...
-    def codegen(self, *groups):
-        group, reduction_group = groups
-
+    @staticmethod
+    def can_fuse_horizontal(node1, node2):
+        _, (vars1, reduce1) = node1.group
+        _, (vars2, reduce2) = node2.group
+        if vars1 == vars2 and reduce1 == reduce2:
+            return True
+        if reduce1 == () and vars1 == vars2 + reduce2:
```
Chillee, line 499: Should this be a + or a *?

torchinductor/scheduler.py
```diff
@@ -1000,188 +761,279 @@ def dead_node_elimination(self):
...
+        - Estimate of the saved memory operations
+        - Fusions closer together in original order
+        """
+        memory_score = self.score_fusion_memory(node1, node1)
+        proximity_score = -max(
+            abs(node1.min_order - node2.max_order),
+            abs(node2.min_order - node1.max_order),
+        )
+        return (
+            not node1.is_reduction() and not node2.is_reduction() and memory_score > 0,
```
Chillee, line 945: Hrmm... do we want to fuse the pointwise ops first?

torchinductor/scheduler.py
```diff
@@ -1000,188 +761,279 @@ def dead_node_elimination(self):
...
+            memory_score,
+            proximity_score,
+        )
+
+    def score_fusion_memory(self, node1, node2):
+        """
+        The first term in our fusion score that estimates number of saved memory operations.
+        """
+        common_reads = node1.read_writes.reads & (
```
Chillee, line 954: Do we want to add a term for vertical fusion?

Chillee (commented):
Mostly LGTM. Some minor questions around some of the details of the graph algorithm-y stuff, and I haven't taken a close look at the new codegen stuff, will do in a bit.

torchinductor/dependencies.py
```diff
@@ -94,6 +100,9 @@ def rename(self, renames: Dict[str, str]) -> "StarDep":
             return StarDep(renames[self.name])
         return self
 
+    def numel_hint(self):
```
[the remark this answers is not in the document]
jansel: It shouldn't.

torchinductor/scheduler.py
```diff
@@ -1000,188 +761,279 @@ def dead_node_elimination(self):
...
+        name_to_predecessors = {}
+        for node in self.nodes:
+            recursive_predecessors = set()
+            for dep in node.unmet_dependencies:
+                recursive_predecessors.add(dep.name)
+                recursive_predecessors |= name_to_predecessors[dep.name]
+            name_to_predecessors[node.get_name()] = recursive_predecessors
+            node.recursive_predecessors = recursive_predecessors
+
+        for order, node in enumerate(self.nodes):
```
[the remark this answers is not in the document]
jansel: We only want to do this once to preserve "original" min/max orders.  `topological_sort()` is called at the start and after every fusion pass.

torchinductor/scheduler.py
```diff
@@ -1000,188 +761,279 @@ def dead_node_elimination(self):
...
+            node1 = self.name_to_fused_node[node1.get_first_name()]
+            node2 = self.name_to_fused_node[node2.get_first_name()]
+            if self.can_fuse(node1, node2) and not self.will_fusion_create_cycle(
+                node1, node2
+            ):
+                node3 = FusedSchedulerNode.fuse(node1, node2)
+                fused_nodes.remove(node1)
+                fused_nodes.remove(node2)
+                fused_nodes.add(node3)
+                self.name_to_fused_node.update(
```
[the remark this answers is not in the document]
jansel: Agreed, we could have a better data structure here.

torchinductor/scheduler.py
```diff
@@ -1000,188 +761,279 @@ def dead_node_elimination(self):
...
+        """
+        Helper to find all legal fusion opportunities, sorted by self.score_fusion()
+        """
+        possible_fusions = []
+        for node1_index, node1 in enumerate(self.nodes):
+            for node2 in self.nodes[node1_index + 1 :]:
+                if self.can_fuse(node1, node2):
+                    possible_fusions.append((node1, node2))
+                elif self.can_fuse(node2, node1):
+                    # epilogue fusions are order dependent
```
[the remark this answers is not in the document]
jansel: This matters for some horizontal fusions, which are not always symmetric.  Example would be template+epilogue.

torchinductor/scheduler.py
jansel: I think here we *dont* want to check combined_names, we want to exclude them with the set subtraction.  Nodes in the combined fusion can't create a cycle.

torchinductor/scheduler.py
jansel: No, this is required for correctness.  This is the only place we check for normal (non-indirect-via-fusion) dependencies.

torchinductor/scheduler.py
jansel: We only use the cycle check code (which is slow) for indirect cycles, not direct ones.

torchinductor/codegen/triton.py
```diff
@@ -955,129 +961,136 @@ def __init__(self, scheduler):
...
+        # swap args to hit the case above
+        return self.can_fuse_horizontal(node2, node1)
 
-        self.scheduler.enqueue(nodes_to_reschedule)
-        return node_schedule
+    can_fuse_vertical = can_fuse
+    can_fuse_horizontal = can_fuse
 
-    def codegen(self, numel, reduction_numel):
+    def codegen_nodes(self, nodes):
```
Chillee, line 1021: This code seems like it depends heavily on the order of `nodes`. What kind of guarantees do we have on that from the prefusion pass? And are there multiple valid orders that could lead to different performance?

torchinductor/scheduler.py
jansel: Running pointwise fusions first is needed to fix those regressions on internal models.  I'm not totally sure why.  I'd like to here @ngimel's thoughts.

torchinductor/scheduler.py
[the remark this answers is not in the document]
jansel: We talked a bit more in person.  While this line could be faster, its perf is dominated by the constructor for FusedSchedulerNode -- so the extra complexity of another data structure isn't worth it.

torchinductor/codegen/cpp.py
jansel: In the C++ backend the group is a tuple of all the iteration ranges.  We don't flatten to numel.  For example:

C++ group: `((a,b,c), ())`

Triton group: `(a*b*c, 1)`

torchinductor/scheduler.py
jansel: We should definitely come back to this heuristic in the future and try as many thing we can think of.  I only tried 4 things then stopped.

torchinductor/scheduler.py
jansel: You are correct, this is a bug.  Will fix.

torchinductor/codegen/triton.py
jansel: Yes, we depend on fusion pass to construct a legal order of nodes.

Order matters, as seen in #938

torchinductor/scheduler.py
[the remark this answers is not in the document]
Chillee: I mean that if it's an epilogue fusion, then it has to be true that `node2_index > node1_index`. But, above, you're iterating `node` from `nodes[node1_index:]`. So it seems to me like all epilogue fusion cases are covered.

I'm not really against this just as a "it avoids needing to think about it" though.

Chillee (approved):
LGTM.

torchinductor/scheduler.py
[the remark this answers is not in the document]
jansel: No this is a real thing.
```
node C reads D
node A is triton template
node B reads A and D
```

After fusion 1:
```
node C reads D
node A_B reads A and D and is triton template+epilogue
```

Now you need to swap the orders to fuse them.

torchinductor/codegen/triton.py
```diff
@@ -666,6 +666,12 @@ def indexing(
         elif not mask:
             mask = ["None"]
 
+        if mask == ["xmask"] and index == 0 and self.range_trees[0].numel == 1:
+            # This causes a triton error:
+            # https://gist.github.com/jansel/70c4b1c8041f0f27f96ee95e2edca04a
```
ngimel, line 671: @voznesenskym this is the "cannot broadcast" error that you've been encountering, so perhaps this fixes it.

Issue #767, opened 2022-08-10
Redo config modules so we can error if users try to set non-existent configuration values


ezyang:
The current idiom for defining config modules makes it easy for users to set nonexistent config values. This is especially troublesome if we delete configuration values in the future. We should have some alternative idiom for defining config modules.

My preference is we turn the config modules into a dataclass, which is minimal boilerplate, and then add a `__setattr__` to error when you try to set a config value doesn't exist. The main downside is that you need to use `field()` syntax to initialize configs with nontrivial mutable objects. It also means module syntax no longer works. We can also use the module replacement trick to preserve the module syntax.

Does anyone else want to paint the bikeshed a different color?

cc @jansel @anijain2305 @voznesenskym

jansel:
Some black magic that might solve this without any refactoring:
```
>>> import types
>>> class SealedModule(types.ModuleType):
...     def __setattr__(self, name, val):
...         assert name in self.__dict__, "Cannot add new attributes to torchdynamo.config"
...         return super().__setattr__(name, val)
... 
>>> import torchdynamo.config
>>> torchdynamo.config.__class__ = SealedModule
>>> 
>>> torchdynamo.config.dynamic_shapes = True
>>> torchdynamo.config.dynamic_shapes2 = True
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "<stdin>", line 3, in __setattr__
AssertionError: Cannot add new attributes to torchdynamo.config
>>> 
```


That said refactoring how configuration is handled to take more options as kwargs to `torchdynamo.optimize` and remove the need for global mutation seems like a good idea to me.

PR #951, opened 2022-08-22
Show clear error on illegal config value access

voznesenskym:
Fixes https://github.com/pytorch/torchdynamo/issues/767


torchdynamo/config.py
```diff
@@ -18,7 +19,7 @@
 # INFO print compiled functions + graphs
 # WARN print warnings (including graph breaks)
 # ERROR print exceptions (and what user code was being processed when it occurred)
-log_level = logging.WARNING
+log_level = logging.ERROR
```
voznesenskym, line 22: Fun bonus to stem the spam

voznesenskym: nevermind theres some cursed test

ezyang: IMO warning should be on by default, it just needs to be less chatty

torchdynamo/config.py
```diff
@@ -95,3 +96,33 @@
...
+        assert AccessLimitingConfig.config_has_attr(
+            name
+        ), f"Trying to get {name} - this value does not exist in torchdynamo.config"
+        return getattr(sys.modules[f"{__name__}.real"], name)
+
+    def __getattr__(self, name):
+        return self.config_get_attr(name)
+
+    def __setattr__(self, name, value):
+        self.config_set_attr(name, value)
```
ezyang, line 124: You should factor this somewhere so torchinductor.config can get this treatment too

ezyang (approved):

mlazos:
Could we make the config an object at the TorchDynamo module level
`config = TorchDynamoConfig()` and then override `setattr` normally?  It will probably have a lot fewer edge cases. I'm okay if this code works though.

mlazos (approved):

voznesenskym:
> Could we make the config an object at the TorchDynamo module level `config = TorchDynamoConfig()` and then override `setattr` normally? It will probably have a lot fewer edge cases. I'm okay if this code works though.

That's a big change - I started with that, but it changed too many callsites. Patching sys modules felt like the best back-compat option.

mlazos:
> > Could we make the config an object at the TorchDynamo module level `config = TorchDynamoConfig()` and then override `setattr` normally? It will probably have a lot fewer edge cases. I'm okay if this code works though.
> 
> That's a big change - I started with that, but it changed too many callsites. Patching sys modules felt like the best back-compat option.

What callsites does it change though? I think you'd have to remove the import? Is that what you mean? torchdynamo.config.x = y would still work I think?

voznesenskym:
> > > Could we make the config an object at the TorchDynamo module level `config = TorchDynamoConfig()` and then override `setattr` normally? It will probably have a lot fewer edge cases. I'm okay if this code works though.
> > 
> > 
> > That's a big change - I started with that, but it changed too many callsites. Patching sys modules felt like the best back-compat option.
> 
> What callsites does it change though? I think you'd have to remove the import? Is that what you mean? torchdynamo.config.x = y would still work I think?

Hmm where would the class live? if in torchdynamo/config.py, it would have to be torchdynamo.config.config right? or youd have to instantiate an instance of a config?

mlazos:
> > > > Could we make the config an object at the TorchDynamo module level `config = TorchDynamoConfig()` and then override `setattr` normally? It will probably have a lot fewer edge cases. I'm okay if this code works though.
> > > 
> > > 
> > > That's a big change - I started with that, but it changed too many callsites. Patching sys modules felt like the best back-compat option.
> > 
> > 
> > What callsites does it change though? I think you'd have to remove the import? Is that what you mean? torchdynamo.config.x = y would still work I think?
> 
> Hmm where would the class live? if in torchdynamo/config.py, it would have to be torchdynamo.config.config right? or youd have to instantiate an instance of a config?

it would live in root torchdynamo
and just be config = Config()

Issue #918, opened 2022-08-19
ValueError: tmp2 is not defined, timm_efficientdet


desertfire:
Repro:
```
benchmarks/torchbench.py  -d cuda --training --inductor --use-eval-mode  -k timm_efficientdet
```
Error:
```
  File "/fsx/users/binbao/triton/python/triton/code_gen.py", line 118, in get_value
    raise ValueError(f'{name} is not defined')
ValueError: tmp2 is not defined
...

triton.code_gen.CompilationError: at 7:33:
def kernel88(in_ptr0, in_ptr1, in_ptr2, in_ptr3, in_ptr4, in_ptr5, in_ptr6, in_ptr7, in_ptr8, in_ptr9, in_ptr10, in_ptr11, out_ptr0, out_ptr1, out_ptr2, out_ptr3, out_ptr4, out_ptr5, out_ptr6, out_ptr7, out_ptr8, out_ptr9, xnumel: tl.constexpr, rnumel: tl.constexpr, XBLOCK : tl.constexpr, RBLOCK : tl.constexpr):
    xoffset = tl.program_id(0) * XBLOCK
    xindex = xoffset + tl.reshape(tl.arange(0, XBLOCK), [XBLOCK, 1])
    xmask = xindex < xnumel
    rbase = tl.reshape(tl.arange(0, RBLOCK), [1, RBLOCK])
    x0 = xindex
    tmp19 = tl.load(in_ptr3 + 2, tmp2)
                                 ^
```

desertfire:
@Chillee , minifier produces a file of 7441 lines.

Chillee:
@desertfire Can you send me the file the minifier starts with?

Chillee:
Here's a minimal repro of the issue (run with the minifier).

```
import torch
from torch import tensor, device
import torch.fx as fx
from torchdynamo.testing import rand_strided
from math import inf
from torch.fx.experimental.proxy_tensor import make_fx

class Repro(torch.nn.Module):
    def forward(self, arg38_1, arg81_1, getitem_17, new_zeros_default_4):
        div_tensor_7 = torch.ops.aten.div.Tensor(getitem_17, arg81_1);  getitem_17 = arg81_1 = None
        mul_tensor_24 = torch.ops.aten.mul.Tensor(div_tensor_7, arg38_1);  div_tensor_7 = arg38_1 = None
        sum_default_7 = torch.ops.aten.sum.default(mul_tensor_24);  mul_tensor_24 = None
[...]

PR #956, opened 2022-08-22
[inductor] Fix a use before def error

desertfire:
Summary: for https://github.com/pytorch/torchdynamo/issues/918

### DIFF
--- a/torchinductor/codegen/triton.py
+++ b/torchinductor/codegen/triton.py
@@ -704,7 +704,12 @@ class TritonKernel(Kernel):
         if upcast:
             line += ".to(tl.float32)"
 
-        if self.inside_reduction and "rmask" not in mask and not indirect_indexing:
+        if (
+            self.inside_reduction
+            and "rmask" not in mask
+            and "tmp" not in mask
+            and not indirect_indexing
+        ):
             # can lift a common load outside of reduction loop
             # One exception is when this is an indirect_load.
             tmp = self.cse.generate(self.body, line)
--- a/torchinductor/codegen/triton.py
+++ b/torchinductor/codegen/triton.py
@@ -621,7 +621,6 @@ class TritonKernel(Kernel):
         index: sympy.Expr,
         copy_shape=None,
         dense_indexing=False,
-        load_index0=False,
     ):
         """
         Compute the index and mask to pass to tl.load() or tl.store()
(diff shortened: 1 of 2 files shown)


desertfire:
There is still another error where `tl.load` fails for a dim=0 tensor, 
```
tmp4 =  tl.load(in_ptr1 + 0, tmp2)
...
arg1_1 = rand_strided((), (), device='cuda', dtype=torch.float32)
```

jansel (approved):

torchinductor/codegen/triton.py
```diff
@@ -632,6 +638,7 @@ def indexing(self, index: sympy.Expr, copy_shape=None, dense_indexing=False):
             or indirect_indexing
             or self._load_mask is not None
         ) and index != 0
```
jansel, line 640: The `index != 0` here seems related, I think this was related to loading the random mask.

torchinductor/codegen/triton.py
```diff
@@ -646,7 +653,7 @@ def indexing(self, index: sympy.Expr, copy_shape=None, dense_indexing=False):
                 mask.append(f"{tree.prefix}mask")
             dense_mask.append(f"{tree.prefix}mask")
 
-        if need_dense and not have_dense:
+        if (need_dense and not have_dense) or load_index0:
```
jansel, line 656: Why not
```suggestion
        if (need_dense and not have_dense) or index == 0:
```
?

desertfire: I wasn't sure if this is safe to do if we are not calling from `load`, but CI run suggests this is ok.

ngimel:
> There is still another error where `tl.load` fails for a dim=0 tensor,
> 

wdym? How does it fail?

desertfire:
> > There is still another error where `tl.load` fails for a dim=0 tensor,
> 
> wdym? How does it fail?

The error I saw was,
```
>       name, asm, shared_mem = _triton.code_gen.compile_ttir(backend, generator.module, device, num_warps, num_stages, extern_libs)
E       MemoryError: std::bad_alloc

../triton/python/triton/code_gen.py:1320: MemoryError
```

Adding `tl.zeros([XBLOCK, RBLOCK]` to the indexing would make it pass, but as @ngimel pointed out in https://github.com/pytorch/torchdynamo/pull/981, what we really need here is mask as `None`.

PR #958, opened 2022-08-22
Throw a RuntimeError when accuracy comparison fails

desertfire:
Summary: Without a RuntimeError, we still writes speedup numbers to the
output .cvs file, which causes CI not catching the failures, see an
example at
https://app.circleci.com/pipelines/github/pytorch/torchdynamo/1363/workflows/d32a7326-01e4-4a83-a283-3e4cab0fa203/jobs/1848

### DIFF
--- a/benchmarks/common.py
+++ b/benchmarks/common.py
@@ -1062,9 +1062,11 @@ class BenchmarkRunner:
                     copy.deepcopy(model), torchdynamo.utils.clone_inputs(example_inputs)
                 )
                 if not same(correct_result, correct_rerun_result, fp64_outputs):
-                    print("INCORRECT - Variation in Eager runs itself")
                     if not self.args.skip_accuracy_check:
+                        print("INCORRECT - Variation in Eager runs itself")
                         return sys.exit(-1)
+                    else:
+                        raise RuntimeError("INCORRECT - Variation in Eager runs itself")
 
             t0 = time.perf_counter()
             torch.manual_seed(1337)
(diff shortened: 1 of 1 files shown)


benchmarks/common.py
```diff
@@ -1062,9 +1062,11 @@ def run_one_model(
                     copy.deepcopy(model), torchdynamo.utils.clone_inputs(example_inputs)
                 )
                 if not same(correct_result, correct_rerun_result, fp64_outputs):
-                    print("INCORRECT - Variation in Eager runs itself")
                     if not self.args.skip_accuracy_check:
+                        print("INCORRECT - Variation in Eager runs itself")
                         return sys.exit(-1)
+                    else:
+                        raise RuntimeError("INCORRECT - Variation in Eager runs itself")
```
desertfire, line 1069: I not very sure if we want to relax this one.

desertfire:
@anijain2305 , this PR will likely drag down our pass rate and possibly perf number as well.

jansel (changes requested):
Due to different implementation of dropout (and other things) some variations are expected.

We have `self.args.skip_accuracy_check` which already controls this behavior.  If skip_accuracy_check is true we raise an error, otherwise we should ignore the difference.

desertfire:
> Due to different implementation of dropout (and other things) some variations are expected.
> 
> We have `self.args.skip_accuracy_check` which already controls this behavior. If skip_accuracy_check is true we raise an error, otherwise we should ignore the difference.

Ok, I guess we only need this for the inference comparison and it does work without this PR, https://app.circleci.com/pipelines/github/pytorch/torchdynamo/1408/workflows/40ca691c-87bf-47d9-b7c7-018647fa8e36/jobs/1941

Let me close this one.

jansel:
Just set `skip_accuracy_check=False` in CI.

PR #962, opened 2022-08-22
Add lowering for adaptive avg pool

eellison:
The two occurrences in torchbench of this op are both no-ops lol. When you disable the no-op optimization, the op is ~6x faster (again, for those two sets of inputs).

### DIFF
--- a/torchinductor/lowering.py
+++ b/torchinductor/lowering.py
@@ -1,6 +1,7 @@
 import functools
 import itertools
 import logging
+import math
 from collections.abc import Iterable
 from typing import List
 
@@ -1929,6 +1930,7 @@ def constant_boundary_condition_2d(x, fill_value, padding):
 
     def load(index):
         *prefix, ih, iw = index
+
         mask = ops.and_(
             range_mask(ih, h),
             range_mask(iw, w),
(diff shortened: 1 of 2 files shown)


ngimel:
Hm, I would think they reduce to 1x1 output and thus could be replaced by sum, not no-op?

torchinductor/lowering.py
```diff
@@ -2123,6 +2125,107 @@ def fn(idx):
...
+
+    h_in = V.graph.sizevars.guard_static_shape(h_in)
+    w_in = V.graph.sizevars.guard_static_shape(w_in)
+
+    h_out, w_out = output_size
+
+    # no-op if the same input and output
+    if h_in == h_out and w_in == w_out:
+        # TODO: do I need to copy ? _to_copy does not
+        return x
```
ngimel, line 2173: better do a copy, yes (to make sure that previously run functionalization is still valid)

eellison:
```
>>> x = torch.rand(1, 1, 4, 4)
>>> y = torch.ops.aten._adaptive_avg_pool2d(x, [4, 4])
>>> (x - y).nonzero().shape
torch.Size([0, 4])
```

The output size is given, not the kernel size which differs from the other apis... When the output_size == input_size for adaptive average pooling, that means the kernel size will be [1, 1] and you're doing a 1-element average (no-op).

torchinductor/lowering.py
jansel: https://github.com/pytorch/torchdynamo/blob/4b212e1dce0431a2903426dcd57369eb85ff47d5/torchinductor/lowering.py#L915

jansel (commented):
We should add a test for this.

Kind of funny these are no-ops, though not the first time I've seen cases like that.

jansel (approved):

PR #966, opened 2022-08-23
[inductor] Futher reduce HF and TIMM batch sizes

desertfire:
Summary: to work around OOM for those models


bdhirsh:
Not yet - I'll make an issue for it.

I finished a run locally where I ran the entire benchmark using inductor with / without functionalization. And it's a bit hard to tell from staring at the logs, but:

(1) There are several more models that OOM
(2) From the logs, it **looks** like the model's that OOM'd with functionalization enabled previously used to fall back to eager (probably because without functionalization, dynamo tried to normalize their IR and failed, falling back to eager. Definitely worth confirming that is the case.

Hey @desertfire - do you know what the state of buffer re-use in inductor looks like today? I guess what I'd expect naively is that with buffer re-use, we should be able to get back to the same peak-memory footprint when using a compiler (inductor) compared to eager mode.

One thing that seems worth trying is turning on the reinplacing pass with inductor, and confirming whether or not that decreases the total number of OOM's (and also for models that didn't OOM, seeing what the change in perf is)

desertfire:
> Not yet - I'll make an issue for it.
> 
> I finished a run locally where I ran the entire benchmark using inductor with / without functionalization. And it's a bit hard to tell from staring at the logs, but:
> 
> (1) There are several more models that OOM (2) From the logs, it **looks** like the model's that OOM'd with functionalization enabled previously used to fall back to eager (probably because without functionalization, dynamo tried to normalize their IR and failed, falling back to eager. Definitely worth confirming that is the case.
> 
> Hey @desertfire - do you know what the state of buffer re-use in inductor looks like today? I guess what I'd expect naively is that with buffer re-use, we should be able to get back to the same peak-memory footprint when using a compiler (inductor) compared to eager mode.
> 
> One thing that seems worth trying is turning on the reinplacing pass with inductor, and confirming whether or not that decreases the total number of OOM's (and also for models that didn't OOM, seeing what the change in perf is)

Which buffer re-use are you referring to? It does have buffer re-use enabled by default, but I haven't paid close attention to how much headroom there is for improvement.

jansel (approved):

PR #981, opened 2022-08-24
[inductor] Revise #956 to avoid a perf regression

desertfire:


torchinductor/codegen/triton.py
```diff
@@ -704,7 +704,12 @@ def mask_loads(self, mask):
     def load(self, name: str, index: sympy.Expr, upcast: bool = False):
         var = self.args.input(name)
         indirect_indexing = self.is_indirect_indexing(index)
+        is_index_0 = index == 0
         index, mask = self.indexing(index)
+        if is_index_0 and "tl.zeros" not in index:
+            # Need dense_indexing when index == 0
+            index = f"{index} + tl.zeros({self.dense_size_str()}, tl.int32)"
```
ngimel, line 711: why do you need this?

torchinductor/codegen/triton.py
```diff
@@ -652,7 +652,7 @@ def indexing(
                 mask.append(f"{tree.prefix}mask")
             dense_mask.append(f"{tree.prefix}mask")
 
-        if (need_dense and not have_dense) or index == 0:
+        if need_dense and not have_dense:
             mask = dense_mask
             index_str = f"{index_str} + tl.zeros({self.dense_size_str()}, tl.int32)"
```
ngimel, line 657:
```suggestion
        if need_dense and not have_dense or index == 0:            
            index_str = f"{index_str} + tl.zeros({self.dense_size_str()}, tl.int32)"
            if index == 0:
                mask = ["None"]
            else:
                mask = dense_mask   
```
and remove changes to `load`. If index is 0, you don't need mask at all, a single element load is always valid.

ngimel (approved):

PR #986, opened 2022-08-24
CI stuff

ezyang:
* Delete scale config: We have org level runners now, so this config doesn't do anything.
* Run inductor tests on GPU


desertfire (approved):
Thanks for pushing this to work!

Makefile
```diff
@@ -63,7 +63,8 @@ setup_nightly_gpu:
                       torchtext==0.14.0.$(PYTORCH_VERSION) \
                       --extra-index-url https://download.pytorch.org/whl/nightly/cu116
 	$(PIP) install ninja
-	$(PIP) install -v "git+https://github.com/pytorch/pytorch.git@`python -c "import torch.version; print(torch.version.git_version)"`#subdirectory=functorch"
+	git clone --depth 1 -b `python -c "import torch.version; print(torch.version.git_version)"` https://github.com/pytorch/pytorch.git
```
desertfire, line 66: I see the latest failure relates to this. Is there a reason you need to change this? 

I think we are close to getting this PR done. It will be nice if we can push this through and at least having unit tests running on GHA.

ezyang: For some reason, functorch doesn't install correctly with the pip command on GHA. I kind of just want to wait until functorch is packaged in the nightlies and then we don't need this anymore...

jansel:

PR #987, opened 2022-08-24
[inductor] Unroll small reductions

jansel:


torchinductor/ir.py
```diff
@@ -582,6 +657,18 @@ def fn(index):
 
             return Pointwise.create(device, dst_dtype, fn, ranges)
 
+        if (
+            V.graph.sizevars.size_hint(reduction_numel)
```
Chillee, line 661: I wonder if this is a good idea. If the reduction dim suddenly changes with dynamic shapes, it'll probably be quite bad if we unroll a massive reduction.

Perhaps we should only do this if the reduction dim is static?

jansel: This won't happen due to the guard.  

"Is adding guards ok for optimizations?" is a reasonable question though.  We do it in some other places too.

torchinductor/ir.py
```diff
@@ -562,6 +562,81 @@ def outer_reduction_splits(reduction_numel_hint, numel_hint):
...
+                )
+
+        else:
+            raise NotImplementedError(f"unknown reduction_type={reduction_type}")
+
+        def fn(index):
+            return functools.reduce(
+                combine_fn,
+                (
+                    value_fn(index, rindex)
```
Chillee, line 613: I find this ordering a bit confusing, since we're defining `value_fn` after this.

Chillee (approved):
LGTM

torchinductor/ir.py
jansel: Yeah, I had the other way before, but it results in extra control flow which also doesn't look nice.

Issue #974, opened 2022-08-23
vision_maskrcnn is incorrectly hitting cpp backend for inductor


ezyang:
Repro:

```
$ ./benchmarks/torchbench.py --inductor --float16 -dcuda --no-skip -k vision_maskrcnn
```

stack:

```
  File "/raid/ezyang/pytorch-scratch2/functorch/functorch/_src/aot_autograd.py", line 855, in new_func
    compiled_fn = create_aot_dispatcher_function(
  File "/raid/ezyang/pytorch-scratch2/functorch/functorch/_src/aot_autograd.py", line 435, in create_aot_dispatcher_function
    return aot_dispatch_base(flat_fn, fake_flat_tensor_args, aot_config)
  File "/raid/ezyang/pytorch-scratch2/functorch/functorch/_src/aot_autograd.py", line 276, in aot_dispatch_base
    compiled_fw = aot_config.fw_compiler(fw_module, flat_args)
  File "/raid/ezyang/torchdynamo/torchinductor/compile_fx.py", line 220, in fw_compiler
    return compile_fx_inner(
  File "/raid/ezyang/torchdynamo/torchdynamo/debug_utils.py", line 248, in debug_wrapper
    raise e
  File "/raid/ezyang/torchdynamo/torchdynamo/debug_utils.py", line 236, in debug_wrapper
    compiled_fn = compiler(gm, example_inputs, **kwargs)
  File "/raid/ezyang/torchdynamo/torchinductor/compile_fx.py", line 112, in compile_fx_inner
    compiled_fn = wrap(graph.compile_to_fn())
  File "/raid/ezyang/torchdynamo/torchinductor/graph.py", line 299, in compile_to_fn
    code = self.codegen()
  File "/raid/ezyang/torchdynamo/torchinductor/graph.py", line 293, in codegen
    self.scheduler.codegen()
  File "/raid/ezyang/torchdynamo/torchinductor/scheduler.py", line 1205, in codegen
    self.get_backend(device).codegen(*group)
  File "/raid/ezyang/torchdynamo/torchinductor/codegen/cpp.py", line 522, in codegen
    node.run(vars, reduction_vars)
  File "/raid/ezyang/torchdynamo/torchinductor/scheduler.py", line 347, in run
    self.codegen(index_vars)
  File "/raid/ezyang/torchdynamo/torchinductor/scheduler.py", line 367, in codegen
    self._body(*index_vars)
[...]

PR #988, opened 2022-08-24
Debugging improvements for #974

jansel:
Partial fix to get to the next error in #974, plus some extra debugging I used to diagnose the issue.

### DIFF
--- a/torchdynamo/debug_utils.py
+++ b/torchdynamo/debug_utils.py
@@ -194,6 +194,7 @@ def nvfuser_fails(fx_g, args, check_str=None):
 
 
 def dump_to_minify(gm, args, compiler_name: str):
+    favored_device = 1 if torch.cuda.device_count() >= 2 else 0
     with open(
         os.path.join(torchdynamo.config.base_dir, "minifier_launcher.py"), "w"
     ) as fd:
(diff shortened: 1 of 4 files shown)


torchdynamo/debug_utils.py
```diff
@@ -194,6 +194,7 @@ def nvfuser_fails(fx_g, args, check_str=None):
 
 
 def dump_to_minify(gm, args, compiler_name: str):
+    favored_device = 1 if torch.cuda.device_count() >= 2 else 0
```
desertfire, line 197: Yes, I just figured this out yesterday.

desertfire:
A unit test needs to be fixed.

jansel:
> A unit test needs to be fixed.

This doesn't actually fix this issue.  It just gets to the next error (with bettor debug printouts).

torchinductor/compile_fx.py
```diff
@@ -130,6 +131,10 @@ def compile_fx_inner(
         elif graph.mutated_inputs and set(graph.device_types) == {"cuda"}:
             log.warning("skipping cudagraphs due to input mutation")
 
+    if torchdynamo.config.repro_level > 1:
```
Chillee, line 134: Can we put this in the `wrap_debug` utility?

Chillee (approved):

Issue #929, opened 2022-08-20
Inductor Error: timm models fp16 inference failed


pyjhzwh:
On main branch (commit 08aa6cea0673fa97e43f5fa5ec09588e0af23d1b)
`python benchmarks/torchbench.py --inductor-settings -dcuda --float16 -n50 --inductor --filter timm_efficientdet` failed

```
Traceback (most recent call last):                                                                                                                                                            
  File "/scratch/panyj/work/cpy/torchdynamo/benchmarks/torchbench.py", line 363, in <module>                                                                                                  
    main(TorchBenchmarkRunner(), original_dir)                                                                                                                                                
  File "/scratch/panyj/work/cpy/torchdynamo/benchmarks/common.py", line 1708, in main                                                                                                         
    device, name, model, example_inputs, batch_size = runner.load_model(                                                                                                                      
  File "/scratch/panyj/work/cpy/torchdynamo/benchmarks/torchbench.py", line 256, in load_model                                                                                                
    module = importlib.import_module(f"torchbenchmark.models.{model_name}")                                                                                                                   
  File "/fsx/users/panyj/conda/envs/cpy/lib/python3.9/importlib/__init__.py", line 127, in import_module                                                                                      
    return _bootstrap._gcd_import(name[level:], package, level)                                                                                                                               
  File "<frozen importlib._bootstrap>", line 1030, in _gcd_import                                                                                                                             
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load                                                                                                                          
  File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked                                                                                                                  
  File "<frozen importlib._bootstrap>", line 680, in _load_unlocked                                                                                                                           
  File "<frozen importlib._bootstrap_external>", line 850, in exec_module                                                                                                                     
  File "<frozen importlib._bootstrap>", line 228, in _call_with_frames_removed                                                                                                                
  File "/scratch/panyj/work/torchbenchmark/torchbenchmark/models/timm_efficientdet/__init__.py", line 12, in <module>                                                                         
    from effdet import create_model, create_loader                                                                                                                                            
  File "/fsx/users/panyj/conda/envs/cpy/lib/python3.9/site-packages/effdet/__init__.py", line 2, in <module>                                                                                  
    from .bench import DetBenchPredict, DetBenchTrain, unwrap_bench                                                                                                                           
[...]

PR #992, opened 2022-08-24
Copy meta when swapping nodes

SherlockNoMad:


ezyang (approved):

voznesenskym:
We had this PR before, and it failed some benchmarks w/ errors - can you run the same benchmarks it failed with last time? https://github.com/pytorch/torchdynamo/issues/929 (Run command is in there)

SherlockNoMad:
> We had this PR before, and it failed some benchmarks w/ errors - can you run the same benchmarks it failed with last time? #929 (Run command is in there)

Previous failure is coming from this line 
```
assert not new_node.meta, "New node must not have meta populated"
```

voznesenskym:
> > We had this PR before, and it failed some benchmarks w/ errors - can you run the same benchmarks it failed with last time? #929 (Run command is in there)
> 
> Previous failure is coming from this line
> 
> ```
> assert not new_node.meta, "New node must not have meta populated"
> ```

I know, but it wouldn't hurt to test. This is a very critical codepath and we should be careful.

voznesenskym (approved):

ezyang:
we should be running a quick benchmark suite run in ci...

SherlockNoMad:
> > > We had this PR before, and it failed some benchmarks w/ errors - can you run the same benchmarks it failed with last time? #929 (Run command is in there)
> > 
> > 
> > Previous failure is coming from this line
> > ```
> > assert not new_node.meta, "New node must not have meta populated"
> > ```
> 
> I know, but it wouldn't hurt to test. This is a very critical codepath and we should be careful.

Tried on timm_efficientnet, timm_nfnet, timm_regnet, timm_resnest, timm_vision_transformer, timm_vovnet, and passing without problem.

PR #993, opened 2022-08-24
[WIP] [For debugging torchvision atm] naive fp16 support in cpp

voznesenskym:


jansel:
High level this seems like a reasonable way to add half support to CPU.

Though I think our "cuda" benchmarks running on CPU might be just a bug...

voznesenskym:
> High level this seems like a reasonable way to add half support to CPU.
> 
> Though I think our "cuda" benchmarks running on CPU might be just a bug...

Yea - this doesn't really have a reason to exist yet. Was just trying to tease out the buf0 bug in c++. I'd rather shelve this until we actually need it.

jansel (approved):

voznesenskym:
Landing it, but not sure anyone actually needs this yet.

PR #998, opened 2022-08-24
RFC: Don't use flags to specify experiments in torchbench

ezyang:
This PR doesn't work but I want to get some feedback on it before I keep going. The basic idea is instead of having a giant mutually exclusive group of flags, instead you just have to specify the experiment + backend by hand, and people no longer have to add new experiments to an ever growing list of experiment types.

An alternative is to add a `--config` flag that subsumes all of the mutually exclusive group and makes it more obvious that you're manually toggling a set of command line arguments.

### DIFF
--- a/benchmarks/common.py
+++ b/benchmarks/common.py
@@ -1257,139 +1257,6 @@ def parse_args():
         "--quiet", "-q", action="store_true", help="suppress debug printouts"
     )
 
-    group = parser.add_mutually_exclusive_group()
-    group.add_argument(
-        "--coverage", action="store_true", help="(default) " + help(coverage_experiment)
-    )
-    group.add_argument(
-        "--online-autotune", action="store_true", help=help(speedup_experiment)
-    )
-    group.add_argument(
-        "--offline-autotune", action="store_true", help=help(speedup_experiment)
-    )
-    group.add_argument(
-        "--speedup-fixed1",
-        action="store_true",
-        help="speedup using experimental fixed_strategy backend",
-    )
-    group.add_argument(
-        "--speedup-fixed2",
-        action="store_true",
-        help="speedup using experimental fixed_strategy backend",
-    )
-    group.add_argument(
-        "--speedup-ltc",
-        action="store_true",
-        help="speedup using the ltc backend",
-    )
-    group.add_argument(
-        "--speedup-ltc-trivial",
-        action="store_true",
-        help="speedup using the ltc backend without reusing compiled graph",
-    )
-    group.add_argument(
-        "--cold-start", action="store_true", help=help(cold_start_experiment)
-    )
-    group.add_argument(
-        "--overhead", action="store_true", help=help(overhead_experiment)
-    )
-    group.add_argument(
-        "--speedup-ts", action="store_true", help=help(speedup_experiment_ts)
-    )
-    group.add_argument(
-        "--speedup-sr", action="store_true", help=help(speedup_experiment_sr)
-    )
-    group.add_argument(
-        "--speedup-onnx", action="store_true", help=help(speedup_experiment_onnx)
-    )
-    group.add_argument(
-        "--speedup-trt", action="store_true", help=help(speedup_experiment_trt)
-    )
-    group.add_argument(
-        "--speedup-dynamo-ts",
-        action="store_true",
-        help="TorchDynamo frontend with torchscript backend",
-    )
-    group.add_argument("--python-key", action="store_true")
-    group.add_argument(
-        "--speedup-fx2trt", action="store_true", help=help(speedup_experiment_fx2trt)
-    )
-    group.add_argument(
-        "--speedup-fx2trt-fp16",
-        action="store_true",
-        help=help(speedup_experiment_fx2trt),
-    )
-    group.add_argument(
-        "--accuracy-aot-nop",
-        action="store_true",
-        help="Accuracy testing and speedup for AOT vs Eager",
-    )
-    group.add_argument(
-        "--accuracy-aot-ts",
-        action="store_true",
-        help="Accuracy testing and speedup for AOT with Torchscript(NNC/NVFuser) vs Eager",
-    )
-    group.add_argument(
-        "--accuracy-aot-ts-mincut",
-        action="store_true",
-        help="Accuracy testing and speedup for AOT with Torchscript(NNC/NVFuser) with mincut vs Eager",
-    )
-    group.add_argument(
-        "--print-fx",
-        action="store_true",
-        help="Print fx traces captured from model",
-    )
-    group.add_argument(
-        "--print-aten-ops",
-        action="store_true",
-        help="Print traces of aten ops captured by AOT autograd",
-    )
-    group.add_argument(
-        "--accuracy-ts",
-        action="store_true",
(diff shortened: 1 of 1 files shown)


benchmarks/common.py
```diff
@@ -1530,145 +1395,73 @@ def main(runner, original_dir=None):
...
+        "speedup_onnx": speedup_experiment_onnx,
+        "speedup_trt": speedup_experiment_trt,
+        "speedup_fx2trt": speedup_experiment_fx2trt,
+        "recompile_profiler": recompile_profiler_experiment,
+        "coverage": coverage_experiment,
+    }
+
+    frontends = {
+        "optimize": torchdynamo.optimize,
+        "optimize_catch_errors": torchdynamo.eval_frame._optimize_catch_errors,
```
jansel, line 1442: This isn't actually a general frontend.  I'd suggest just deleting anything that uses it (I think just `--nops` which is testing code that IMO is no longer useful.)

benchmarks/common.py
```diff
@@ -1530,145 +1395,73 @@ def main(runner, original_dir=None):
...
+    }
+
+    optimize_ctx = frontends[args.frontend](backends.get(args.backend, args.backend), nopython=args.nopython)
+    experiment = experiments[args.experiment]
+
+    backend_str = "nvfuser" if args.nvfuser else "nnc"
```
jansel, line 1448: The default should be nvfuser to match eager mode.

benchmarks/common.py
```diff
@@ -1530,145 +1395,73 @@ def main(runner, original_dir=None):
...
+        "speedup_trt": speedup_experiment_trt,
+        "speedup_fx2trt": speedup_experiment_fx2trt,
+        "recompile_profiler": recompile_profiler_experiment,
+        "coverage": coverage_experiment,
+    }
+
+    frontends = {
+        "optimize": torchdynamo.optimize,
+        "optimize_catch_errors": torchdynamo.eval_frame._optimize_catch_errors,
+    }
```
jansel, line 1443: This might not capture configuration that use
- eager mode (`--nothing`): useful for debugging issues in torchbench, when it break up stream
- torchscript: baselines to compare against

benchmarks/common.py
```diff
@@ -1257,139 +1257,6 @@ def parse_args():
         "--quiet", "-q", action="store_true", help="suppress debug printouts"
     )
 
-    group = parser.add_mutually_exclusive_group()
```
jansel, line 1260: The main downside of remove this is we lose our `--help` documentation.  We may need to manually generate `--help` if we go this route.

benchmarks/common.py
```diff
@@ -1530,145 +1395,73 @@ def main(runner, original_dir=None):
...
-        experiment = recompile_profiler_experiment
-    else:
-        optimize_ctx = torchdynamo.optimize(fx_insert_profiling, nopython=args.nopython)
-        experiment = coverage_experiment
-        output_filename = "coverage.csv"
+
+    # these names are based on the historical command line flag values
+    backends = {
+        "overhead": dummy_fx_compile,
+        "online_autotune": online_autotuner,
```
jansel, line 1411: remove, it is old

benchmarks/common.py
```diff
@@ -1530,145 +1395,73 @@ def main(runner, original_dir=None):
...
+        "offline_autotune": offline_autotuner,
```
jansel, line 1412: remove, it is old

benchmarks/common.py
```diff
@@ -1530,145 +1395,73 @@ def main(runner, original_dir=None):
...
+        "python_key": python_key,
```
jansel, line 1413: remove, there is a PR that deletes this

benchmarks/common.py
```diff
@@ -1530,145 +1395,73 @@ def main(runner, original_dir=None):
...
+        "ltc": backends.ltc_reuse_graph,
+        "ltc_trivial": backends.ltc_trivial,
+        "fixed1": fixed_strategy1,
```
jansel, line 1416: delete, vague

benchmarks/common.py
```diff
@@ -1530,145 +1395,73 @@ def main(runner, original_dir=None):
...
+        "fixed2": fixed_strategy2,
```
jansel, line 1417: delete, vague

benchmarks/common.py
```diff
@@ -1530,145 +1395,73 @@ def main(runner, original_dir=None):
...
+        "ts": backends.ts,
+        "fx2trt": backends.fx2trt_compiler,
+        "fx2trt_fp16": backends.fx2trt_compiler_fp16,
```
jansel, line 1420: should control this with the --float16 flag not a separate target.

benchmarks/common.py
```diff
@@ -1530,145 +1395,73 @@ def main(runner, original_dir=None):
...
+        "print_fx": print_fx,
+        "print_aten_ops": print_aten_ops,
+        "nops": torchdynamo.testing.debug_insert_nops,
```
jansel, line 1423: remove, it is old

benchmarks/common.py
```diff
@@ -1530,145 +1395,73 @@ def main(runner, original_dir=None):
...
+        "log_conv_args": conv_args_analysis,
+        "fx_insert_profiling": fx_insert_profiling,
+    }
+
+    experiments = {
+        "speedup": speedup_experiment,
+        "cold_start": cold_start_experiment,
+        "speedup_ts": speedup_experiment_ts,
+        "speedup_sr": speedup_experiment_sr,
+        "speedup_onnx": speedup_experiment_onnx,
```
jansel, line 1433: All of these `speedup_XX` things are baselines not using torchdynamo.  Perhaps they belong in another section.

jansel:
Should also update
1) Makefile (used as a macro store)
2) CI scripts
3) Documentation

benchmarks/common.py
ezyang: Do we really care about running against TS? I feel like the only reason we would try it is if we were trying to write a blog post comparing against TS, but this seems like a lot of work for not much publicity gain.

benchmarks/common.py
ezyang: That's kind of inelegant though, because it breaks the invariant that the backend argument maps to a particular torchdynamo.optimize argument; instead, that argument might wobble depending on other flags. Seems better to not do that, especially since this isn't a cowpath backend?

benchmarks/common.py
ezyang: Does anyone... actually read the `torchbench.py --help` documentation though? There's so many flags, it seems hard for someone to find what they're looking for

benchmarks/common.py
jansel: Baselines are useful, and when we release more broadly I bet there will be an audience interested in comparing to TorchScript.

benchmarks/common.py
jansel: I use `--help` when I forget a flag name.

benchmarks/common.py
wenzhe-nrv: a general comment - could you please consider a way to pass more backends other than assuming nvfuser or nnc? In my case, I would like to benchmark with `onednn` backend for both inference and training with AOTAutograd. Thanks.

jansel:

PR #1000, opened 2022-08-24
Doc on codegen from triton template

pyjhzwh:
As title


documentation/InductorTritonTemplateKernel.md
```diff
@@ -0,0 +1,221 @@
...
+
+This tutorial assumes the reader has an overview of how Torchinductor codegen Triton code and a basic understanding of Triton languange. This tutorial is for developers who want to further improve the performance of some operations, like Convolution and MatMul, by manually-written triton kernel.
+
+## Motivation
+For most of the operations, Torchinductor is able to automatically codegen corresponding Triton kernel. However,
+for `ir.ExternKernel` in torchinductor/, Torchinductor usually call extern functions like `aten.convolution` for Convolution, `aten.mm.out` for MatrixMultiply, etc. But if you have better implementation of those operations in Triton, say `triton.ops.matmul`, one way to let Torchindcutor to use it is replace the kernel and codegen function call to `triton.ops.matmul`. But still, this is an ExternKernel call where the Torchinductor optimize the whole graph unware of what's happening inside the function call. Specifically, it could not benefit from epilog fusion that could alleviate the memory-bound bottlenecks. To get additional speedup beside the customized code, user could add an Triton template for Torchinductor as a reference to cedegen the target kernel.
```
jansel, line 7:
```suggestion
for ops with now lowerings, TorchInductor generate a fallback `ir.ExternKernel`, which just calls the original function.  The main examples here are `aten.convolution` for Convolution, `aten.mm.out` for MatrixMultiply, etc. TorchInductor supports alternate implementations of these ops, such as `triton.ops.matmul`, but we could get further benifiet from epilogue fusion that could alleviate the memory-bound bottleneck of pointwise ops that follow the more expensive kernel. To access this additional speedup, one can add a Triton template for Torchinductor that allows these epilogue fusion kernels to be automatically generated.
```

documentation/InductorTritonTemplateKernel.md
```diff
@@ -0,0 +1,221 @@
...
+
+## Overview
+Codegen from custom templates is derived from codegen normal Triton kernels. So in this tutorial, we will only focus on the difference.
+The main python file for codegen triton templates is [triton_template.py](https://github.com/pytorch/torchdynamo/blob/main/torchinductor/codegen/triton_template.py). Other files that need to modify when adding a new implementation as template file for Torchinductor codegen are [scheduler.py](https://github.com/pytorch/torchdynamo/blob/main/torchinductor/scheduler.py), [ir.py](https://github.com/pytorch/torchdynamo/blob/main/torchinductor/ir.py).
+
+Class TemplateSchedulerNode is a subclass of BaseSchedulerNode. During the `init` of Scheduler, it will create TemplateSchedulerNode if the node `should_use_template()`, and append to the `self.nodes` list. And when Sheduler `codegen()`, the TemplateSchedulerNode will call `codege_template_call()`, where call [`template_codegen()`](https://github.com/pytorch/torchdynamo/blob/main/torchinductor/codegen/triton_template.py#L299) inside the function.
```
jansel, line 13:
```suggestion
Class TemplateSchedulerNode is a subclass of BaseSchedulerNode. During the `init` of Scheduler, it will create TemplateSchedulerNode if the node `should_use_template()`, and append to the `self.nodes` list. And when the Scheduler does `codegen()`, the TemplateSchedulerNode will call [`template_codegen()`](https://github.com/pytorch/torchdynamo/blob/main/torchinductor/codegen/triton_template.py#L299) to generate the combined kernel.
```

documentation/InductorTritonTemplateKernel.md
```diff
@@ -0,0 +1,221 @@
...
+* {{ ... }} for Expressions to print to the template output
+* {# ... #} for Comments not included in the template output
+
+The variables (like `kernel_name`, `pointwise_code`) are defined by the dictionary passed to the template. TritonTemplateKernel will load the jinja2 template files according the node type. And after every variable is generated by codegen the fusable nodes, those varaibles are put into a dictionary for the template file to render.
+
+## TemplateSchedulerNode
+### Update dependency type
+The default ExternKernel node's dependency type is `StarDep`. `StarDep("A")` means the node depends on A[*] (all elements of A). The entire A buffer must be fully computed before the node with that dep is scheduled. It prevents A to be fused with other nodes.
+
+But if we want to enable epilog fusion with the ExternKernel that could codegen from triton template, we need to update its dependencies from `StarDep` to `MemoryDep`. Usually if we want to add a triton template for a new ExternKernel operation, we do not need to modify this class.
```
jansel, line 63:
```suggestion
But if we want to enable epilogue fusion with the ExternKernel that could codegen from triton template, we need to update its dependencies from `StarDep` to `MemoryDep`. Usually if we want to add a triton template for a new ExternKernel operation, we do not need to modify this class.
```

jansel (approved):

PR #1004, opened 2022-08-25
[inductor] Support aten.squeeze_

jansel:
Not totally sure why this example needs squeeze_, but it is easy to support.


ezyang:
nb this is a metadata mutating op, so the better long term solution is for functionalization to expunge it. cc @bdhirsh

jansel:
Agreed.  Let's still land this as it is causing models to break.

ezyang (approved):

PR #1008, opened 2022-08-25
Set batch size to 1 when running CI

desertfire:


benchmarks/huggingface.py
```diff
@@ -354,6 +354,10 @@ def load_model(
         else:
             model = model_cls(config).to(device, dtype=dtype)
 
+        if self.args.ci:
+            # Use batch size 1 for faster CI
+            batch_size = 1
```
ezyang, line 359: Why can't we just apply this in benchmarks/common.py?

ezyang:
While we're at this, the default `--inductor` config runs the models too many times; the default number of repetitions 30, but you only really need two (one for warmup, one for steady state.)

benchmarks/common.py
```diff
@@ -945,10 +945,10 @@ def batch_size_finder(
         return 1
 
     def get_benchmark_indices(self, length):
-        start = self._args.partition_id * (length // self._args.total_partitions)
+        start = self.args.partition_id * (length // self.args.total_partitions)
         end = (
-            (self._args.partition_id + 1) * (length // self._args.total_partitions)
-            if self._args.partition_id < self._args.total_partitions - 1
+            (self.args.partition_id + 1) * (length // self.args.total_partitions)
+            if self.args.partition_id < self.args.total_partitions - 1
```
ezyang, line 951: what's going on here?

ezyang (approved):

benchmarks/huggingface.py
desertfire: We could, but I need to add something like an instance variable to pass around the information. I wasn't sure if your cleanup will touch this part, so I just use something quick to test the time difference.

benchmarks/common.py
desertfire: To clean up my last PR to make it look more consistent with other places. Not worth to submit a separate PR.

benchmarks/huggingface.py
ezyang: I guess, naively, I would have expected fiddling around with `args.batch_size` would do it, but maybe not! This is OK then

PR #1013, opened 2022-08-25
don't turn AliasedLayout into fixed for extern kernels

ngimel:
When one of the inputs to `cat` was extern kernel, AliasedLayout was replaced with Fixed, and thus the extern kernel output wasn't written to cat output. 
I don't like this fix very much, because extern kernels can't always produce output in the layout needed for Concat, and Pointwise in Concat doesn't seem to be doing what it's supposed to, but at least some testcase is fixed.

### DIFF
--- a/test/test_torchinductor.py
+++ b/test/test_torchinductor.py
@@ -1622,6 +1622,25 @@ class CommonTemplate:
             (torch.randn([8, 16]),),
         )
 
+    def test_cat_extern_kernel(self):
+        def fn(x1, x2, x3, x4):
+            x = torch.mm(x2, x3)
+            s = torch.narrow(x, 1, 0, 100)
+            x = torch.mm(s, x4)
+            c = torch.cat((x, x1), 1)
+            return (c,)
+
+        self.common(
+            fn,
+            (
+                torch.randn(256, 256),
+                torch.randn(256, 1024),
+                torch.randn(1024, 1600),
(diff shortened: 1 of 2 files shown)


Chillee:
If this was the issue, i'm surprised we didn't see it more often tbh. Shouldn't see this anytime we have a matmul followed by a concat?

jansel (approved):

ngimel:
We might have seen it but 1) accuracy checking is typically turned off, so we wouldn't notice, there's no codegen or runtime error in this case 2) matmul is usually followed by bias add and that saves us. 
The issue with incorrect buffer reuse for slices that actually causes runtime error is still there.

ngimel:
btw, I don't know what would happen with Conv instead of mamul extern kernel, as conv doesn't have an `out` overload and will produce contiguous input, and unclear if remaining codegen will actually insert a necessary copy. Ditto for triton matmul - I don't think it will be so nice as to produce sliced output, so again, whether remaining codegen will fix that or not remains to be seen, I'll try explicitly checking those cases.

jansel:
The copy (if needed) should get added here:
https://github.com/pytorch/torchdynamo/blob/0b8e55e4a8401296a87dc43de00874465c5690f5/torchinductor/ir.py#L1531

It is always possible there are bugs though.

ngimel:
Conv produces the needed copy, that's cool.

PR #1014, opened 2022-08-25
Fix calling isinstance on torch.nn.Parameter

yanboliang:
Fix this bug:
```
def fn(x):
    a = torch.nn.Parameter(torch.rand(2, 3))
    if isinstance(a, torch.Tensor):
        return x + 1
    else:
        return x - 1

x = torch.tensor([2.5])
print(fn(x))

    print(fn(x))
```
Pytorch output: ```[3.5]```
TorchDynamo output: ```[1.5]```

### DIFF
--- a/torchdynamo/variables/tensor.py
+++ b/torchdynamo/variables/tensor.py
@@ -311,7 +311,10 @@ class TensorVariable(VariableTracker):
 
         def check_type(ty):
             if ty not in tensortype_to_dtype:
-                return self.python_type() is ty
+                if self.python_type() is torch.nn.Parameter and ty is torch.Tensor:
+                    return True
+                else:
+                    return self.python_type() is ty
 
             dtypes = tensortype_to_dtype[ty]
             return self.dtype in dtypes
--- a/torchdynamo/variables/tensor.py
+++ b/torchdynamo/variables/tensor.py
@@ -311,10 +311,7 @@ class TensorVariable(VariableTracker):
 
         def check_type(ty):
             if ty not in tensortype_to_dtype:
-                if self.python_type() is torch.nn.Parameter and ty is torch.Tensor:
-                    return True
-                else:
-                    return self.python_type() is ty
+                return issubclass(self.python_type(), ty)
 
             dtypes = tensortype_to_dtype[ty]
             return self.dtype in dtypes
--- a/test/test_misc.py
+++ b/test/test_misc.py
@@ -2075,6 +2075,20 @@ class MiscTests(torchdynamo.testing.TestCase):
         self.assertTrue(same(b_grad, torch.tensor([0.0, 0.0])))
         self.assertEqual(cnts.frame_count, 2)
 
+    def test_torch_nn_parameter_isinstance(self):
+        def fn(x):
+            a = torch.nn.Parameter(torch.rand(2, 3))
+            if isinstance(a, torch.Tensor):
+                return x + 1
+            else:
+                return x - 1
+
+        x = torch.tensor([2.5])
+        ref = fn(x)
+        opt_fn = torchdynamo.optimize("eager")(fn)
+        res = opt_fn(x)
+        self.assertEqual(ref, res)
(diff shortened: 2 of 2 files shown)


torchdynamo/variables/tensor.py
```diff
@@ -311,7 +311,10 @@ def call_isinstance(self, tensor_type):
 
         def check_type(ty):
             if ty not in tensortype_to_dtype:
-                return self.python_type() is ty
+                if self.python_type() is torch.nn.Parameter and ty is torch.Tensor:
```
voznesenskym, line 314: We already have utils is_tensor:

```
def istensor(obj):
    """Check of obj is a tensor"""
    tensor_list = (
        torch.Tensor,
        torch.nn.Parameter,
        *config.traceable_tensor_subclasses,
    )
    if fake_tensors_available:
        tensor_list = tensor_list + (torch._subclasses.FakeTensor,)
    return istype(obj, tensor_list)
```

Can this share logic with that?

yanboliang: No, I don't think sharing the logic can simplify code. Can you take a look at the complete implementation of ```call_isinstance```? They are different purpose.

voznesenskym: Why do we have two different valid sets of subclasses for torch tensor? Is that full list not valid? Which elements of that list do we not want to support?

yanboliang: Can you take a look at the complete implementation of ```call_isinstance```? It's using ```dtype``` to decide the tensor type. I think here the only problem is we don't handle ```torch.nn.Parameter``` well. For other tensor type listed at ```istensor```, we already handle them well. Or if you can suggest a better and simplified way to reuse them?

voznesenskym:
```
def valid_tensor_subtypes():
    tensor_list = (
        torch.Tensor,
        torch.nn.Parameter,
        *config.traceable_tensor_subclasses,
    )
    if fake_tensors_available:
        tensor_list = tensor_list + (torch._subclasses.FakeTensor,)
    return tensor_list
```

```
def istensor(obj):
    """Check of obj is a tensor"""
    return istype(obj, valid_tensor_subtypes())
```

```
def check_type(ty):
            if ty not in tensortype_to_dtype:
                if self.python_type() in valid_tensor_subtypes():
                  ...
```

voznesenskym: Meta comment - please stop resolving the conversations until the concern is addressed!

yanboliang: Can you look at [these code](https://github.com/pytorch/torchdynamo/blob/main/torchdynamo/variables/tensor.py#L161), ```self.python_type()``` already return ```torch.Tensor``` if it's a ```FakeTensor```, why we need to add redundant code?
I'd recommend you to take a closer look at ```TensorVariable```, actually for regular tensor, fake tensor and other tensors, their python type is ```torch.Tensor```, only parameter tensor returns ```torch.nn.Parameter```, that' what this PR fixed? Thank you!

voznesenskym: Because if we ever get a new tensor subtype, we would want to support it in both places, right?

yanboliang: No, if there is a new tensor subtype, we should not update where I changed, we should do some update like [this line](https://github.com/pytorch/torchdynamo/blob/main/torchdynamo/variables/tensor.py#L161) to make its ```class_type``` is ```torch.Tensor```.

anijain2305, line 314: use `isinstance(self.python_type(), ty)`

yanboliang: sync up with @anijain2305 offline and find ```issubclass``` is the most concise way.

anijain2305:
@voznesenskym Can you PTAL?

anijain2305 (approved):

PR #1017, opened 2022-08-25
FIx sizevars typechecking and add to mypy file

Chillee:
Mostly the same as this (https://github.com/pytorch/torchdynamo/pull/880) but without ghstack lol.

### DIFF
--- a/torchinductor/sizevars.py
+++ b/torchinductor/sizevars.py
@@ -2,14 +2,19 @@ import collections
 import dataclasses
 import functools
 import logging
+from typing import Callable
 from typing import Dict
 from typing import List
+from typing import Tuple
 
 import sympy
 from sympy import Expr
 from sympy import Integer
 from sympy import Symbol
 
+from . import ir
+from .codegen.common import IndentedBuffer
+from .utils import VarRanges
 from .utils import freeze_inputs
 from .utils import immutable_dict
 from .virtualized import V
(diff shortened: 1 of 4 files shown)


torchinductor/sizevars.py
```diff
@@ -314,7 +320,8 @@ def strideof(name):
             code.writeline(f"{name}_stride = {name}.stride()")
             return f"{name}_stride"
 
-        needed = set(map(str, self.var_to_val.keys())) - set(self.replacements.keys())
+        needed_vars = set(self.var_to_val.keys()) - set(self.replacements.keys())
```
Chillee, line 323: @jansel I think this was previously just a bug? `self.replacements` is a dict from `sympy.Symbol` to `sympy.Expr`. And here you're deduplicating against a set with strings.

jansel: I think you are correct

jansel (approved):

torchinductor/sizevars.py
Chillee: Seems like fixing this bug causes some issues in `test/test_torchinductor.py::GpuTests::test_triton_conv_cuda`.

I get a `NameError: name 's0' is not defined` (cc: @pyjhzwh )

Chillee: Gonna revert to the old behavior for now and add a note.

PR #1018, opened 2022-08-25
Add test suite that runs with fake_tensor_propagation=False

mlazos:
Matches the CI run in pytorch


test/test_no_fake_tensors.py
```diff
@@ -0,0 +1,47 @@
...
+import torchdynamo
+
+from . import test_functions
+from . import test_misc
+from . import test_modules
+from . import test_repros
+from . import test_unspec
+
+
+def make_no_fake_fn(fn):
```
voznesenskym, line 14: I know I mentioned making a separate file for this, and I think we need to for the sake of pytest and our sanity, but I think there's a good amount of logic here we can share with the dynamic shape test this file is based on. 

Maybe we can pull out this function and the one below it into a shared module?

mlazos: Done, way cleaner now, and supports an arbitrary number of patches 👯

voznesenskym: Very nice :) it LGTM and our recent config changes should keep that patch valid.

voznesenskym (approved):

PR #1019, opened 2022-08-25
Run batchnorm inplace without grad

mlazos:
Even without dynamo, running an inplace operation on a tensor with grad throws, not sure why it doesn't throw with Fake Tensors

### DIFF
--- a/test/test_repros.py
+++ b/test/test_repros.py
@@ -992,12 +992,13 @@ class ReproTests(torchdynamo.testing.TestCase):
         self.assertEqual(cnt.op_count, 5)
 
     def test_batch_norm_act(self):
-        a = torch.randn(5, 1, 28, 28)
-        model = BatchNormAct2d(1).eval()
-        correct = model(a)
-        cnt = torchdynamo.testing.CompileCounter()
-        with torchdynamo.optimize_assert(cnt):
-            self.assertTrue(same(model(a), correct))
+        with torch.no_grad():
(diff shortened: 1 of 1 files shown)


jansel:
What is the error?  Why did this used to work?

mlazos:
> What is the error? Why did this used to work?

The error is “RuntimeError: a leaf Variable that requires grad is being used in an in-place operation.”

IMO when we have FakeTensors on, this should have failed as well with the same error, but I think it’s a bug in FakeTensor. This batch norm is inplace so it should fail anytime a tensor has requires_grad

ezyang:
A repro would be helpful, but the most likely explanation is we aren't preserving the requires-gradness of tensor when converting to fake tensor.

mlazos:
> ~A repro would be helpful,~ Oops this is a PR; but the most likely explanation is we aren't preserving the requires-gradness of tensor when converting to fake tensor.

Yes agreed, you can just run the test with `fake_tensor_propagation` off and it fails. I'm investigating this for moving the dynamo pin in PyTorch. 

@jansel I'm changing this test to have the correct/expected behavior, the bug is that in the original version it didn't fail.

mlazos:
I looked deeper into this and I think I figured out why it fails without fake tensors, it's actually yet another bug in clone_input, it's because `grad_fn` isn't properly copied. 

Should we return to clone here? Is the reason we didn't do that because striding isn't copied?

mlazos:
Closing in favor of #1032

PR #1029, opened 2022-08-25
Don't delete MemoryPlanningLine referenced in outputs

voznesenskym:


torchinductor/codegen/wrapper.py
```diff
@@ -296,10 +296,24 @@ def generate(self):
         result = IndentedBuffer()
         result.splice(self.header)
         result.splice(self.prefix)
+
+        def extract_name(out):
```
voznesenskym, line 300: this feels gross, I wonder if it would be better served as a util on ir.py?

voznesenskym:
Benchmarking this now

voznesenskym:
Bench seems fine.

torchinductor/codegen/wrapper.py
```diff
@@ -296,10 +296,24 @@ def generate(self):
...
+            o = out
+            while hasattr(o, "data"):
+                if hasattr(o, "name"):
+                    return o.name
+                else:
+                    o = o.data
+            return ""
+
+        out_names = [extract_name(out) for out in V.graph.graph_outputs]
```
jansel, line 309: We collect names of all outputs here:
https://github.com/pytorch/torchdynamo/blob/da8bb9a875beede7b721425087d06ebae7bc76f8/torchinductor/scheduler.py#L849
and, here:
https://github.com/pytorch/torchdynamo/blob/da8bb9a875beede7b721425087d06ebae7bc76f8/torchinductor/scheduler.py#L600

I'd suggest adding a function like:
```
V.graph.get_output_names()
````

That combines the 3 places

File: torchinductor/codegen/wrapper.py [hunk 3]
```diff
@@ -296,10 +296,24 @@ def generate(self):
...
         with result.indent():
             while self.lines and isinstance(self.lines[-1], MemoryPlanningLine):
                 # these lines will be pointless
-                self.lines.pop()
+                if self.lines[-1].node.name not in out_names:
```
jansel, line 313: Just put this in the condition of the while loop above to be more concise

jansel (changes requested):

torchinductor/graph.py
```diff
@@ -309,3 +308,10 @@ def compile_to_fn(self):
             setattr(mod, name, value)
 
         return mod.call
+
+    def get_output_names(self):
```
jansel, line 312: Update the other uses to use this

jansel (approved):

PR #1032, opened 2022-08-25
Only copy grad and requires grad for leaf tensors

mlazos:
Previously we would copy grad and requires_grad for non-leaf tensors. This could result in errors when running in-place ops like batchnorm, because some non-leaf tensors would have requires_grad set by pytorch core, and after copying, the copied tensor now has `is_leaf == True`. So putting this through an in-place op would throw "leaf tensor with grad being used in inplace op"

Because other parts of the stack will generate grad_fn and set requires_grad, it's okay to not populate them for non-leaf tensors in dynamo.

With help from @anijain2305

### DIFF
--- a/torchdynamo/utils.py
+++ b/torchdynamo/utils.py
@@ -287,16 +287,18 @@ def clone_input(x):
         result.as_strided_(x.size(), x.stride(), cache_line_offset)
         try:
             result.copy_(x.clone())
-            result.requires_grad_(x.requires_grad)
-            if x.grad is not None:
+            if x.is_leaf:
+                result.requires_grad_(x.requires_grad)
+            if x.is_leaf and x.grad is not None:
                 result.grad = clone_input(x.grad)
         except RuntimeError:
             # RuntimeError: unsupported operation: more than one element of the written-to
             # tensor refers to a single memory location. Please clone() the tensor before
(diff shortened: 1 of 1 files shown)


anijain2305 (approved):
Looks good to me. But want to wait for @jansel review to be sure.

ezyang:
fwiw, meta tensor converter has a very, very rigorous version of this. Code is at https://github.com/pytorch/pytorch/blob/15b560a5c4d638c82e738f3496e2faf95fc328a5/torch/_subclasses/meta_utils.py#L148-L223

anijain2305:
> fwiw, meta tensor converter has a very, very rigorous version of this. Code is at https://github.com/pytorch/pytorch/blob/15b560a5c4d638c82e738f3496e2faf95fc328a5/torch/_subclasses/meta_utils.py#L148-L223

This looks so complicated. I suggest to merge this PR first (if the logic makes sense) as this is blocking Dynamo pin in PyTorch CI. And follow up quickly with the better fix. Is that ok?

soumith:
yea lets merge it and move towards a rigorous version after.
Basically, without using the rigorous version, we're only buying time (I looked at the version and yea all those cases need to be carefully copied over)

jansel (approved):

PR #1037, opened 2022-08-26
Allow ConcatKernel realize_into to try to produce a reinterpret view rather than fail

voznesenskym:
There's probably a more optimal impl where we don't drop into slice at all, but rather skip both slice and a lot of the concat kernel (cc @ngimel), but this gets us past some model failures for now.


ngimel (approved):

voznesenskym:
Looks fine on bench.

PR #1038, opened 2022-08-26
Add utility for timing dynamo compilation

wconstab:
usage:
```
from .utils import print_compilation_metrics
print_compilation_metrics()
```

instrumentation:
```
from .utils import dynamo_timed
@dynamo_timed
my_function()
```

sample output:
```
Function                         Runtimes (s)
-------------------------------  --------------------
convert_frame_assert             ['0.0005']
GraphLowering.run                ['0.0637', '0.0204']
Scheduler.compute_dependencies   ['0.0001', '0.0001']
Scheduler.fuse_nodes_once        ['0.0001', '0.0002']
Scheduler.__init__               ['0.0417', '0.0597']
Scheduler.codegen                ['0.0223', '0.0225']
WrapperCodeGen.generate          ['0.0055', '0.0141']
GraphLowering.compile_to_fn      ['0.1230', '0.0988']
compile_fx_inner                 ['0.1881', '0.1217']
compile_fx.<locals>.fw_compiler  ['0.1881']
compile_fx.<locals>.bw_compiler  ['0.1218']
```

### DIFF
--- a/torchdynamo/convert_frame.py
+++ b/torchdynamo/convert_frame.py
@@ -32,6 +32,7 @@ from .guards import GuardedCode
 from .symbolic_convert import InstructionTranslator
 from .utils import CleanupManager
 from .utils import counters
+from .utils import dynamo_timed
 from .utils import filter_stack
 from .utils import format_bytecode
 from .utils import guard_failures
@@ -214,6 +215,7 @@ def format_error_msg(exc, frame):
     return msg
 
 
+@dynamo_timed
 def convert_frame_assert(compiler_fn: Callable, guard_export_fn=None, one_graph=True):
     """Fully convert a frame into an FX graph"""
     init_logging()
@@ -375,7 +377,8 @@ def convert_frame(compiler_fn: typing.Callable, guard_export_fn=None):
         except BackendCompilerFailed:
             raise
         except Exception:
-            pass
+            raise
+            # pass
         return None
 
     _convert_frame._torchdynamo_orig_callable = compiler_fn
--- a/torchdynamo/convert_frame.py
+++ b/torchdynamo/convert_frame.py
@@ -215,13 +215,13 @@ def format_error_msg(exc, frame):
     return msg
 
 
-@dynamo_timed
 def convert_frame_assert(compiler_fn: Callable, guard_export_fn=None, one_graph=True):
     """Fully convert a frame into an FX graph"""
     init_logging()
 
     compiler_fn = wrap_compiler_fn(compiler_fn)
 
+    @dynamo_timed
     def _convert_frame_assert(frame: types.FrameType, cache_size: int):
         code = frame.f_code
         input_codes.add(code)
--- a/torchdynamo/convert_frame.py
+++ b/torchdynamo/convert_frame.py
@@ -377,8 +377,7 @@ def convert_frame(compiler_fn: typing.Callable, guard_export_fn=None):
         except BackendCompilerFailed:
             raise
         except Exception:
-            raise
-            # pass
+            pass
         return None
 
     _convert_frame._torchdynamo_orig_callable = compiler_fn
--- a/torchdynamo/utils.py
+++ b/torchdynamo/utils.py
@@ -37,6 +37,31 @@ troubleshooting_url = (
 
 log = logging.getLogger(__name__)
 
+# profiling compilation time
+compilation_metrics = collections.OrderedDict()
+
+
+def dynamo_timed(func):
+    def time_wrapper(*args, **kwargs):
+        t0 = time.time()
+        r = func(*args, **kwargs)
+        key = func.__qualname__
+        if key not in compilation_metrics:
+            compilation_metrics[key] = []
+        compilation_metrics[key].append(time.time() - t0)
+        return r
+
+    return time_wrapper
+
+
+def print_compile_times():
+    rows = [
+        (k, list(map(lambda f: f"{f:.4f}", compilation_metrics[k])))
+        for k in compilation_metrics
+    ]
+    print(tabulate.tabulate(rows, headers=("Function", "Runtimes (s)")))
+
+
 LOGGING_CONFIG = {
     "version": 1,
     "formatters": {
--- a/torchdynamo/utils.py
+++ b/torchdynamo/utils.py
@@ -54,14 +54,37 @@ def dynamo_timed(func):
     return time_wrapper
 
 
-def compile_times():
-    rows = [
-        (k, list(map(lambda f: f"{f:.4f}", compilation_metrics[k])))
-        for k in compilation_metrics
-    ]
-    out = "TorchDynamo compilation metrics:\n"
-    out += tabulate.tabulate(rows, headers=("Function", "Runtimes (s)"))
-    return out
+def compile_times(repr="str", aggregate=False):
+    """
+    Get metrics about torchdynamo frontend/backend compilation times.
+
+    Accumulates information from functions tagged with `@dynamo_timed`.
+
+    repr='str' returns a printable string for user interaction, and 'csv'
+    returns headers, rows which can be logged for output
+
+    aggregate causes values from multiple compilations (e.g. split graphs)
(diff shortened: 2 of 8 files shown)


torchdynamo/utils.py
```diff
@@ -37,6 +37,44 @@
 
 log = logging.getLogger(__name__)
 
+# profiling compilation time
+compilation_metrics = collections.OrderedDict()
+@contextlib.contextmanager
+def timed_compile(key):
```
voznesenskym, line 43: 1) Do we always want this enabled? This feels a bit like a debug /profiling thing.

2) The usage feels a little off:

```
with timed_compile("compile_fx_aot"):
    with timed_compile("normalize_ir"):
        with overrides.patch_functions():
            model_ = normalize_ir(model_, example_inputs_)
            model_ = overrides.replace_fx(model_)
        num_example_inputs = len(example_inputs_)
        cudagraphs = BoxedBool(config.triton.cudagraphs)
```

Why are we nesting two? one inside `compile_fx_aot` and one outside `normalize_ir`? This feels like it should be a decorator `@timed_compile` that just takes the fn name and uses that as the key.

3) The notion of `compile` here is lost on me - if we follow (1) and (2) and make this a decorator that no-ops when the appropriate debug flag is not set, can we just call it @time

Maybe something like:
```
def time(func):
    def wrap(*args, **kwargs):
        t0 = time.time()
        r = func(*args, **kwargs)
        compilation_metrics[func.__name__] = t0 - time.time()
        return r
```

torchdynamo/utils.py
```diff
@@ -37,6 +37,44 @@
...
+    global compilation_metrics
+    pop = compilation_metrics
+    assert key not in compilation_metrics, "trying to double-record"
```
voznesenskym, line 46: What if we recursively re-enter? This feels like it imposes runtime ordering asserts. Why do we need to do this pop and store dance? I'm a little too dumb to understand this logic.  Can we:

1) make t0
2) call the func
3) make t1
4) store t1 in a `dict[name, list[time]]`

torchdynamo/utils.py
```diff
@@ -37,6 +37,44 @@
...
+        t0 = time.time()
+        yield
+    except Exception:
+        raise
+    finally:
+        t1 = time.time()
+        compilation_metrics["self"] = t1 - t0
+        compilation_metrics = pop
+
+def print_compile_times(parent_time=None, level=0, rows=[], curr_dict=compilation_metrics):
```
voznesenskym, line 59: What does this output look like?

torchdynamo/utils.py
voznesenskym: Further thought - if we do an @time, we should consider @profile also that runs cprofile, perhaps?

voznesenskym:
Stolen from internet, but:

```
def profileit(name):
    def inner(func):
        def wrapper(*args, **kwargs):
            prof = cProfile.Profile()
            retval = prof.runcall(func, *args, **kwargs)
            # Note use of name from outer scope
            prof.dump_stats(name)
            return retval
        return wrapper
    return inner
```

+ usage:

```
@profileit("compile_fx_profile")
def compile_fx(model_: torch.fx.GraphModule, example_inputs_: List[torch.Tensor]):
```

Gives us a real .prof file. 

Running it with snakeviz then shows us a real local site with interactive perf and all that:

image

Maybe add a cprofile decorator alongside the @time decorator suggested above?

wconstab:
@voznesenskym before I go into the details of your responses, did you see my PR desc?  It shows the sample output.

I think the first thing I want to hash out is the motivation and usefulness of this.  It should be distinct from 'cprofile' as it offers a limited set of measurements that we could track in our nightlies, rather than a full profile that has a lot of detail but isn't directly diffable day to day.

What I wanted to achieve is a high level view of compile time spent in general regions of the stack (dynamo, aot, triton kernels).  Instead of accumulating measurements from subsequent runs under the same key, I am trying to get measurements for just one compilation but show how they relate to each other (tree structure) just like cprofile would.

voznesenskym:
> @voznesenskym before I go into the details of your responses, did you see my PR desc?  It shows the sample output.
> 
> 

Yes

> 
> I think the first thing I want to hash out is the motivation and usefulness of this.  It should be distinct from 'cprofile' as it offers a limited set of measurements that we could track in our nightlies, rather than a full profile that has a lot of detail but isn't directly diffable day to day.
> 

Yes, most of my comments are around @timed - which addresses what your PR does - the cprofile part is totally secondary and orthogonal and just came up as an idea after seeing this. 

> 
> What I wanted to achieve is a high level view of compile time spent in general regions of the stack (dynamo, aot, triton kernels).  Instead of accumulating measurements from subsequent runs under the same key, I am trying to get measurements for just one compilation but show how they relate to each other (tree structure) just like cprofile would.

Agreed. See my feedback on turning this into @timed?

wconstab:
ok I have a redo that takes most of your suggestions @voznesenskym 

1) are there any more places we should obviously decorate before i move to start running this nightly?
2) what is the right format to aggregate this data into in the benchmark runner and on disk?

3)
> Do we always want this enabled? This feels a bit like a debug /profiling thing.

I might lean towards leaving it enabled always, but that really doesn't matter to me as it would be trivial to enable it within the benchmark runs.

4) 
> Maybe add a cprofile decorator alongside the @time decorator suggested above?

sure, i can add this too. I might do it in a separate PR. I could land your exact version now too.  Do we want it to dump each f"{name}.prof" file to disk in the cwd? or is some more elaborate scheme needed to stash all the profs in some kind of dict mirroring the timing dict?

also, the prof decorator would certainly be useful as a thing people add to code, run, then delete before checkin.  do you also think it's useful to make the `dynamo_timed` decorator have a mode where `config.dynamo_profiling=True` causes `dynamo_timed` to also dump a profile for each of those funcs for zero code change profiling?

voznesenskym:
> ok I have a redo that takes most of your suggestions @voznesenskym 
> 
> 
> 
> 1) are there any more places we should obviously decorate before i move to start running this nightly?

Wanna add it to explains output, and to export (not as part of output)? If not, I can do it in a later pr. Just having this util is already huge.

> 
> 2) what is the right format to aggregate this data into in the benchmark runner and on disk?
> 

That's an @anijain2305 question, I do not know, my apologies. 

> 
> 3)
> 
> > Do we always want this enabled? This feels a bit like a debug /profiling thing.
> 
> 
> 
> I might lean towards leaving it enabled always, but that really doesn't matter to me as it would be trivial to enable it within the benchmark runs.
> 

I'm alright with either. 

> 
> 4) 
> 
> > Maybe add a cprofile decorator alongside the @time decorator suggested above?
> 
> 
> 
> sure, i can add this too. I might do it in a separate PR. I could land your exact version now too.  Do we want it to dump each f"{name}.prof" file to disk in the cwd? or is some more elaborate scheme needed to stash all the profs in some kind of dict mirroring the timing dict?
> 
> 
> 
> also, the prof decorator would certainly be useful as a thing people add to code, run, then delete before checkin.  do you also think it's useful to make the `dynamo_timed` decorator have a mode where `config.dynamo_profiling=True` causes `dynamo_timed` to also dump a profile for each of those funcs for zero code change profiling?

Yeah this does sound like a later pr. The idea of riding it as a config on timed is cool. We'd have to see what recursive and/or layered invocations of cprofile look like - you may end up having to do some magic to not end up with a ton of reports.

torchdynamo/convert_frame.py
```diff
@@ -375,7 +377,8 @@ def _convert_frame(frame: types.FrameType, cache_size: int):
         except BackendCompilerFailed:
             raise
         except Exception:
-            pass
+            raise
+            # pass
```
voznesenskym, line 381: Nit rm?

voznesenskym (approved):
Noice

voznesenskym:
> @voznesenskym i have added it to .explain(), but i think it's unfortunate that bc there isn't a 'real' backend used in there it only shows the frontend time for now.

I plan on adding backend support here.

PR #1040, opened 2022-08-26
Consolidate CI benchmark skip list

desertfire:


jansel (approved):

desertfire:
@jansel , should we retire `check_coverage` given that we have most of the models running on CI now?

jansel:
check_coverage counts how many ops dynamo captures (versus uncaptured).  Do other CI job check this?  I thought they just check for correctness, which wouldn't detect us falling back to eager.

PR #1045, opened 2022-08-26
Fix how torchbench is installed in CircleCI

desertfire:
Summary: leverage https://github.com/pytorch/benchmark/pull/1158 to use read torchbench data from S3 instead of using git lfs

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,7 +8,7 @@ commands:
           name: Generate date for cache key
           command: date +%F > .circleci-date
       - restore_cache:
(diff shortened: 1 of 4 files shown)


ezyang (approved):
nice!

jansel:
Yeah... this is a frustrating one... I tried a bunch of the same things and couldn't get `git-lfs` to work in cricleci.  Hence the horrible tarball hack.

Hopefully you can figure out.  I tried multiple versions of git-lfs.  I tried the skip smudge sutff.  Perhaps it is related to auth.  Maybe try clearing the `ssh-agent` state.

xuzhao9:
Updated the way of tarball handling in torchbench, moving from git-lfs to Amazon S3 for easier deployment.

PR #1047, opened 2022-08-26
Delete offline autotuner

ezyang:
* #1053
* #1052
* #1051
* #1049
* __->__ #1047
* #1042

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -112,18 +112,6 @@ build-deps: clone-deps
 	make setup_lint
 	python setup.py develop
 
-offline-autotune-cpu: develop
-	rm -rf subgraphs
-	python benchmarks/torchbench.py --offline-autotune -n3
-	python autotune.py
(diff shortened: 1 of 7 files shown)


jansel:
I think we can also remove this entire file: https://github.com/pytorch/torchdynamo/blob/main/benchmarks/autotune.py

ezyang:
did it in another pr in the stack

zzpmiracle:
Sorry, @ezyang @jansel why do we delete the autotuner? I think it a nice work in torchdynamo to capare backends for each subgraph. Does it not work or have some side-effects?

ezyang:
The offline autotuner is not good for general use, as it requires a lot of work on the users part; you'd rather just have one backend that works well for everything. Additionally we were not maintaining this code. You are welcome to run your own autotuning experiments if you like!

PR #1056, opened 2022-08-26
[RFC][WIP] Backend Agnostic Minifier for TorchDynamo produced Fx Graphs

anijain2305:
This work is based on @Chillee minifier infrastructure. 

The existing minifier works after AotAutograd is run, and runs on the generated forward and backward pass. Therefore, it cannot minify issues in the Aot Autograd tracing, decomps, partitioner etc.

This PR adds minifier to the TorchDynamo produced Fx graph (opposed to the Aot Autograd produced Fx Graphs). This makes it backend-agnostic. In the generated repro file, we use `torchdynamo.optimize("compiler_name")` (opposed to `make_fx` in the existing infra).

However, there are few points to notice
* Existing infra first dumps a file that we can run further to minify. It uses `make_fx` to generate the Fx Graph module for `minifier`. This makes it pretty nice and isolated.
* We can't use `make_fx` here because we want to run torchdynamo.optimize compiler here. Therefore, the minification runs in the main process. This might be problematic if the process crashes during the minification process.

There is WIP. I am trying to get the design right before cleaning up. Please comment.


voznesenskym:
> * We can't use `make_fx` here because we want to run torchdynamo.optimize compiler here. Therefore, the minification runs in the main process. This might be problematic if the process crashes during minification.
> 
> 

Can we use export? I believe it has a very similar api to make_fx, in that it takes a callable and args and produces an fx graph.

torchdynamo/config.py
```diff
@@ -106,6 +106,13 @@
 # 3: Always dumps the last graph ran out to minify_repro.py, useful for segfaults/irrecoverable errors
 repro_level = int(os.environ.get("COMPILER_REPRO_LEVEL", 0))
 
+# TorchDynamo produced Fx GraphModule compilation debug info
+# 0: Nothing printed out when compilation fails
+# 1: Dump the initial graph to repro.tar.gz
+# 2/3: Minifies and Dumps the minified repro to repro.tar.gz
+dynamo_repro_level = int(os.environ.get("TORCHDYNAMO_REPRO_LEVEL", 0))
```
Chillee, line 113: Maybe we should call this AOT_AUTOGRAD_REPRO_LEVEL? This is mainly for minifying issues that occur at the AOTAutograd level, not the dynamo level.

torchdynamo/debug_utils.py
```diff
@@ -93,7 +157,7 @@ def generate_repro_string(gm, args):
 }
 
 
-def dump_state(gm, args, compiler_name):
+def dump_post_aot_graph_state(gm, args, compiler_name):
```
Chillee, line 160: I would prefer naming like `dump_compiler_repro_state`.

torchdynamo/debug_utils.py
```diff
@@ -259,3 +331,235 @@ def debug_wrapper(gm, example_inputs, **kwargs):
...
+
+    setup_module = textwrap.dedent(
+        f"""
+        mod = Repro().cuda()
+        opt_mod = torchdynamo.optimize("{compiler_name}")(mod)
+
+        """
+    )
+
+    # TODO - Figure out the amp state
```
Chillee, line 387: todo resolved?

Chillee (commented):
Mostly LGTM, modulo naming nits.

torchdynamo/debug_utils.py
anijain2305: Sounds good.

torchdynamo/config.py
anijain2305: I earlier thought BACKEND_REPRO_LEVEL here, but then its confusing with COMPILER_REPRO_LEVEL.

We are going to use this for Aot Autograd, but others might use this for their backend. Is this any better?

* BACKEND_REPRO_LEVEL - Uses this PR
* INDUCTOR_REPRO_LEVEL 
* NVFUSER_REPRO_LEVEL

The last 2 flags set the `compiler_repro_flag` internally. Just for the user, this might be more clear?

anijain2305: Or, just have COMPILER_REPRO_LEVEL as the user-level contract

COMPILER_REPRO_LEVEL = 0 : Do nothin
COMPILER_REPRO_LEVEL = 1 : Minifies the Backend
COMPILER_REPRO_LEVEL = 2 : Minifies for the compiler (exisiting one)

And do away with current leveling. We can have a separate flag REPRO_DETAIL for development purposes and fixing issues with minifier.

torchdynamo/config.py
```diff
@@ -99,12 +99,16 @@
...
 # Compiler compilation debug info
-# 0: Nothing printed out when compilation fails
-# 1: Dump the graph out to repro.py if compilation fails
-# 2: Dumps the graph out to minify_repro.py with a minifier if compilation fails
-# 3: Always dumps the last graph ran out to minify_repro.py, useful for segfaults/irrecoverable errors
-repro_level = int(os.environ.get("COMPILER_REPRO_LEVEL", 0))
+# 1: Dumps the original graph out to repro.py/repro.tar.gz if compilation fails
+# 2: Either 1) Dumps the original graph out to minify_repro.py with a minifier if compilation fails
+#        or 2) Minifies and dumps the minified graph to repro.py/repro.tar.gz if compilation fails
+repro_level = int(os.environ.get("REPRO_LEVEL", 2))
```
jansel, line 111:
```suggestion
repro_level = int(os.environ.get("TORCHDYNAMO_REPRO_LEVEL", 2))
```
I worry about colliding with names used by other projects.

torchdynamo/config.py
```diff
@@ -99,12 +99,16 @@
     "torch._decomp",
 }
 
+# Debug Flag to try minifier at different stages. Possible values are {None, "aot", "dynamo"}
+# None - Minifier is switched off
+# dynamo - Runs minifier on the TorchDynamo produced graphs, if compilation fails
+# aot - Runs minifier on the Aot Autograd produced graphs, if compilation fails
+repro_after = os.environ.get("REPRO_AFTER", None)
```
jansel, line 106:
```suggestion
repro_after = os.environ.get("TORCHDYNAMO_REPRO_AFTER", None)
```

Chillee (approved):
LGTM

Issue #979, opened 2022-08-23
optimizer.step() keeps to recompile and hits cache limit


yanboliang:
[```optimizer.step()```](https://github.com/pytorch/pytorch/blob/master/torch/optim/sgd.py#L120) keeps to recompile for every model training iteration, because:
* Graph break in [this line](https://github.com/pytorch/pytorch/blob/master/torch/optim/sgd.py#L145) due to unimplemented ```collections.defaultdict```.
* This graph break will be guarded by ```TUPLE_ITERATOR_LEN``` and the length of tuple iterator reduced as the for loop iterates, so this guard keeps failing and hit cache limit.

After discussing with @jansel, the right fix should be avoiding graph inside of the for loop if possible, a.k.a add support for ```collections.defaultdict```. This can also benefit other parts, as this data structure is used widely inside of Pytorch.

PR #1057, opened 2022-08-26
Add DefaultDictVariable

yanboliang:
Fix https://github.com/pytorch/torchdynamo/issues/979

### DIFF
--- a/test/test_optimizers.py
+++ b/test/test_optimizers.py
@@ -71,7 +71,7 @@ class OptimizerTests(torchdynamo.testing.TestCase):
     # NB: in python versions < 3.8, we don't capture graphs when
     # breaks occur in a loop
     test_radam = make_test(
-        torch.optim.RAdam, exp_frame_cnt=(0 if sys.version_info < (3, 8) else 5)
+        torch.optim.RAdam, exp_frame_cnt=(0 if sys.version_info < (3, 8) else 6)
     )
 
     # ASGD has a small optimization that avoids averaging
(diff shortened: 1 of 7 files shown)


test/test_optimizers.py
```diff
@@ -71,15 +71,15 @@ def setUpClass(cls):
     # NB: in python versions < 3.8, we don't capture graphs when
     # breaks occur in a loop
     test_radam = make_test(
-        torch.optim.RAdam, exp_frame_cnt=(0 if sys.version_info < (3, 8) else 5)
+        torch.optim.RAdam, exp_frame_cnt=(0 if sys.version_info < (3, 8) else 6)
```
yanboliang, line 74: No graph break on ```defaultdict.__getitem__```, so we have 1 more compiled fn.

torchdynamo/variables/dicts.py
```diff
@@ -201,6 +202,55 @@ def _key_to_var(cls, tx, key, **options):
...
+                return self.getitem_const(args[0])
+            else:
+                if self.default_factory is None:
+                    raise KeyError(f"{k}")
+                else:
+                    if isinstance(k, torch.nn.Parameter):
+                        tx.store_dict_key(global_key_name(k), k)
+                    new_val = collections.OrderedDict(self.items)
+                    if self.default_factory is list:
+                        default_var = ListVariable([], mutable_local=MutableLocal())
```
yanboliang, line 236: I was thinking reuse ```VariableBuilder``` but gave up, this is not variables passed as function arguments, so no ```source``` and ```guard```. Right now other ```default_factory``` except for ```tuple/list/dict``` would cause graph break, but I do search in PyTorch codebase and found these three types can cover most cases.

jansel (approved):

torchdynamo/variables/dicts.py
```diff
@@ -201,6 +202,55 @@ def _key_to_var(cls, tx, key, **options):
...
+        args: "List[VariableTracker]",
+        kwargs: "Dict[str, VariableTracker]",
+    ) -> "VariableTracker":
+        from . import ListVariable
+        from . import TupleVariable
+
+        options = VariableTracker.propagate(self, args, kwargs.values())
+
+        if name == "__getitem__":
+            k = ConstDictVariable.get_key(args[0])
```
voznesenskym, line 224: minor nit - assert args len before here?

voznesenskym (approved):

PR #1059, opened 2022-08-26
avoid value dependent output shape slicing

eellison:
From code comment:
Avoid a .item() call in the tensor slice that would attempt to get a value out
fake tensors, and which would determine the output shape of the slice.
It is a workaround until https://github.com/pytorch/pytorch/pull/83567
is landed and there is more complete support for breaking on data dependent operators.

I'm deferring on landing 83567 because there are a bunch of spurious `.item()` calls that get invoked for 0-element tensors and im worried about regressing perf with a bunch of graph breaks. To avoid that im adding handling for constant tensor tracking but that's a bit of a ways off and this issue with slicing is blocking correctness/some other models.

### DIFF
--- a/test/test_misc.py
+++ b/test/test_misc.py
@@ -1844,28 +1844,6 @@ class MiscTests(torchdynamo.testing.TestCase):
         self.assertEqual(y, 11)
         self.assertEqual(z, 61)
 
-    def test_slicing_dynamic_shape(self):
-        def fn(y):
-            x = torch.ones(8)
-            idx = y[0]
-            out = x[idx:]
-            return (out + 3) * 5
-
-        counter = torchdynamo.testing.CompileCounter()
-        opt_fn = torchdynamo.optimize(counter)(fn)
-        out = opt_fn(torch.ones(10, dtype=torch.long))
-        # idx should be 1 -> slicing off [1:] of 8 elem tensor
-        self.assertEqual(list(out.shape), [7])
-
-        dyn = torchdynamo.config.dynamic_shapes
-        expected_ops = 5 if dyn else 4
-        expected_frame = 1 if dyn else 2
-
-        self.assertEqual(expected_ops, expected_ops)
-        self.assertEqual(expected_frame, expected_frame)
-
-        self.assertEqual(list(opt_fn(torch.tensor([4])).shape), [4])
-
     def test_cross_entropy_loss_fancy_ctor(self):
         output = None
         rand_5 = torch.randn(5)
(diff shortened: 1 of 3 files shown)


torchdynamo/variables/tensor.py
```diff
@@ -148,6 +150,23 @@ def context():
                 if use_fake_tensors:
                     example_value = fake_wrapper(example_value)
 
+        # Avoids a .item() call in the tensor slice that would attempt to get a value out
+        # fake tensors, and which would determine the output shape of the slice.
+        # It is a workaround until https://github.com/pytorch/pytorch/pull/83567
+        # is landed and there is more complete support for breaking on data dependent operators.
+        if proxy.node.target == operator.getitem and use_fake_tensors:
```
yanboliang, line 157: Is this a workaround for ```dynamic_shapes``` only or not?

torchdynamo/variables/tensor.py
```diff
@@ -148,6 +150,23 @@ def context():
...
+            if (
+                isinstance(args[0], FakeTensor)
+                and isinstance(args[1], slice)
+                and any(
+                    isinstance(e, FakeTensor)
+                    for e in (args[1].start, args[1].stop, args[1].step)
+                )
+            ):
+                raise torch._subclasses.fake_tensor.DynamicOutputShapeException(
```
jansel, line 166: Does this actually trigger a graph break in dynamo?  Do we catch it properly?  Perhaps add a test with some ops after it ensure they are captured.

If you call `unimplemented("dynamic shape slicing")` that should gracefully trigger a graph break.

We also support `config.dynamic_shapes`, which I would hope would be able to capture this (but wouldn't know the shape of the output).

torchdynamo/variables/tensor.py
eellison: What do you mean `dynamic_shapes`? you mean the config, `dynamic_shapes` ?

yanboliang: Yes

test/test_misc.py
```diff
@@ -1844,6 +1844,28 @@ def forward(self, x):
...
+            out = x[idx:]
+            return (out + 3) * 5
+
+        counter = torchdynamo.testing.CompileCounter()
+        opt_fn = torchdynamo.optimize(counter)(fn)
+        out = opt_fn(torch.ones(10, dtype=torch.long))
+        # idx should be 1 -> slicing off [1:] of 8 elem tensor
+        self.assertEqual(list(out.shape), [7])
+
+        dyn = torchdynamo.config.dynamic_shapes
```
yanboliang, line 1860: We have an util function [```ifdyn```](https://github.com/pytorch/torchdynamo/blob/d88a0ac0e4a799617e59a05fec79d329781f3ea7/test/test_repros.py#L32) can help this.

jansel (approved):

PR #1060, opened 2022-08-26
Don't decompose addmm, take 2 

ngimel:
I messed up #912 with merge gone wrong, this is take 2, conditionally decomposes if we don't codegen addmm with triton. Perf numbers in #912 still hold.


torchinductor/decomposition.py
```diff
@@ -106,16 +106,19 @@ def clamp(x, min=None, max=None):
...
-def addmm(input, mat1, mat2):
-    return torch.mm(mat1, mat2) + input
-
-
 @register_decomposition([aten.tanh])
 def tanh(x):
     return 2.0 / (1.0 + torch.exp(-2.0 * x)) - 1.0
 
 
+@register_decomposition([aten.addmm])
```
Chillee, line 114: Perhaps we should put these in a special section.

Chillee (approved):
Do we need to make a new `ExterneKernelOut` class everytime we want to use an out variant?

ngimel:
> Do we need to make a new `ExterneKernelOut` class everytime we want to use an out variant?

I don't know tbh, was copying what was done in other kernels

torchinductor/decomposition.py
ngimel: I don't expect many of conditional decomps, if there are a few we can combine them.

jansel:
> Do we need to make a new `ExterneKernelOut` class everytime we want to use an out variant?

The automatic fallbacks burn in shapes, so for stuff we want to be long lived we should define shape formulas for them.

Chillee:
@jansel Can we just re-use core's shape propagation formulas?

jansel:
> @jansel Can we just re-use core's shape propagation formulas?

Yes, that sounds like a great idea.  Is it possible to do that now?

ngimel:
Test error is oom, we should reduce batch sizes.

PR #1061, opened 2022-08-26
Skip models with unusual behavior

anijain2305:


benchmarks/huggingface.py
```diff
@@ -482,9 +487,9 @@ def refresh_model_names_and_batch_sizes():
             SwinForImageClassification,
             SwinForMaskedImageModeling,
             SwinModel,
-            ViTForImageClassification,
-            ViTForMaskedImageModeling,
-            ViTModel,
+            # ViTForImageClassification,
+            # ViTForMaskedImageModeling,
+            # ViTModel,
```
desertfire, line 492: So these are working now? Why not just removing them completely?

anijain2305: I left it by mistake. They don't lead to any issue if we build pytorch with `USE_DISTRIBUTED=1`. I need more time to look into this one.

Let's try to get the other models in till then.

desertfire (approved):

PR #1068, opened 2022-08-27
Support pyoperator, specialize support for cond

voznesenskym:


torchdynamo/mutation_guard.py
```diff
@@ -83,9 +83,13 @@ def check(cls, obj):
 
 def is_dynamic_nn_module(obj):
     """Check for nn.Modules() created dynamically or mutated"""
-    return GenerationTracker.dynamic_classes.get(type(obj)) or GenerationTracker.check(
+    if hasattr(obj, "force_dynamic"):
+        return obj.force_dynamic
+    dyn = GenerationTracker.dynamic_classes.get(type(obj)) or GenerationTracker.check(
         obj
     )
+    print(f"Is {obj} dynamic? {dyn}")
```
jansel, line 91: print

torchdynamo/mutation_guard.py
```diff
@@ -83,9 +83,13 @@ def check(cls, obj):
 
 def is_dynamic_nn_module(obj):
     """Check for nn.Modules() created dynamically or mutated"""
-    return GenerationTracker.dynamic_classes.get(type(obj)) or GenerationTracker.check(
+    if hasattr(obj, "force_dynamic"):
```
jansel, line 86: perhaps `torchdynamo_force_dynamic`

torchdynamo/variables/torch.py
```diff
@@ -398,3 +401,115 @@ def handle_ntuple(value):
...
+                return [
+                    unwrap_real(arg_inner) for arg_inner in arg.unpack_var_sequence(tx)
+                ]
+            return arg
+
+        # Get values
+        u_args = [unwrap_real(arg) for arg in args]
+
+        # Make fx to get a graph out (All PyOperator's should produce valid fx)
+        graph = make_fx(self.value)(*u_args)
```
jansel, line 438: Could we just inline the python function of the operator rather than tracing it?

torchdynamo/variables/torch.py
```diff
@@ -398,3 +401,115 @@ def handle_ntuple(value):
...
+
+        from . import ListVariable
+        from . import TensorVariable
+        from . import UserFunctionVariable
+
+        assert kwargs is None or len(kwargs) == 0, "kwargs are not supported, yet"
+
+        def unwrap_real(arg):
+            if isinstance(arg, TensorVariable):
+                return arg.as_proxy().node.meta["example_value"]
```
jansel, line 425: Using example values is unsound.

jansel (changes requested):

torchdynamo/mutation_guard.py
voznesenskym: sure :)

torchdynamo/variables/torch.py
voznesenskym: Done! good suggestion.

torchdynamo/variables/torch.py
voznesenskym: Yea. I wasn't sure what else to use here.

It felt alright because we enforce TensorVariable to have it in order to propagate:

```
def propagate_args_kwargs(node):
        def visit(n: torch.fx.Node):
            return n.meta["example_value"]
```

That being said, I'd rather not rely on it. Open to suggestions on how best to trace this!

torchdynamo/mutation_guard.py
voznesenskym: gone

torchdynamo/variables/torch.py
voznesenskym: Discussed offline - added guard safety check.

torchdynamo/convert_frame.py
```diff
@@ -161,6 +161,7 @@ def has_tensor(obj):
             and hasattr(obj, "__dict__")
             and len(getattr(obj, "__dict__"))
         ):
+            print("Considering type?", obj.__class__)
```
jansel, line 164: stray print

torchdynamo/variables/torch.py
```diff
@@ -398,3 +401,119 @@ def handle_ntuple(value):
...
+            i = 0
+            while not next_name:
+                candidate = f"name_{i}"
+                if candidate in tx.output.nn_modules:
+                    i += 1
+                else:
+                    next_name = candidate
+
+            gm.__name__ = next_name
+            src = LocalSource(gm)
```
jansel, line 458: LocalSource should take the name of a local variable in the root frame dynamo is tracing.

torchdynamo/variables/builder.py
```diff
@@ -380,6 +382,10 @@ def index_source(key):
                 for k in ("start", "stop", "step")
             ]
             return SliceVariable(items, guards=make_guards(GuardBuilder.TYPE_MATCH))
+        elif isinstance(value, PyOperator):
+            return TorchPyOperator(
+                value, guards=self.make_guards(GuardBuilder.TYPE_MATCH)
```
jansel, line 387: I wonder if TYPE_MATCH is enough.... is every PyOperator a different type?

jansel (approved):

torchdynamo/variables/builder.py
voznesenskym: Good question - same type. I wonder if I should add a GuardBuilder.NAME_MATCH?

PR #1070, opened 2022-08-27
[inductor] Support aten.fill.Scalar

jansel:
Needed by timm_efficientnet

### DIFF
--- a/benchmarks/common.py
+++ b/benchmarks/common.py
@@ -68,8 +68,10 @@ CI_SKIP_INFERENCE = [
     # Huggingface
     "AlbertForQuestionAnswering",
     "AllenaiLongformerBase",
+    "BartForCausalLM",
     "BertForQuestionAnswering",
     "BigBird",
+    "BlenderbotSmallForConditionalGeneration",
     "DebertaForQuestionAnswering",
     "DebertaV2ForQuestionAnswering",
     "DistilBertForQuestionAnswering",
--- a/test/test_torchinductor.py
+++ b/test/test_torchinductor.py
@@ -1619,6 +1619,16 @@ class CommonTemplate:
             ),
         )
 
+    def test_fill(self):
+        def fn(x):
+            tmp = torch.ones_like(x)
+            return tmp, aten.fill.Scalar(tmp, 2)
+
+        self.common(
+            fn,
+            (torch.randn([16, 16]),),
+        )
+
     def test_pow1(self):
         def fn(x):
             return [aten.pow(x, e) for e in range(-8, 9)]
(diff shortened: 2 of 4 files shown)


Chillee (approved):
LGTM, although in the long run we should just register the ref as a decomp + lower from the prims instead to handle type promotion properly.

ngimel:
Ref in this case has a more restrictive behavior than torch op:
```
In [11]: a=torch.randint(2,(4,), dtype=torch.int)

In [12]: torch.fill(a, 3.5)
Out[12]: tensor([3, 3, 3, 3], dtype=torch.int32)

In [13]: torch._refs.fill(a, 3.5)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Input In [13], in <cell line: 1>()
----> 1 torch._refs.fill(a, 3.5)

File /scratch/ngimel/work/pytorch/torch/_prims_common/wrappers.py:212, in out_wrapper.<locals>._out_wrapper.<locals>._fn(out, *args, **kwargs)
    209         if k not in kwargs:
    210             kwargs[k] = out_attr
--> 212 result = fn(*args, **kwargs)
    213 assert (
    214     isinstance(result, TensorLike)
    215     and is_tensor
    216     or isinstance(result, Tuple)  # type: ignore[arg-type]
    217     and len(result) == len(out_names)
    218 )
    219 if out is not None:
    220     # Naively you might expect this assert to be true, but
    221     # it's not:
   (...)
    234     # be a normal meta tensor, but this is perfectly
    235     # harmless.

File /scratch/ngimel/work/pytorch/torch/_prims_common/wrappers.py:119, in elementwise_type_promotion_wrapper.__call__.<locals>._fn(*args, **kwargs)
    112 promoted_args = {
    113     x: _maybe_convert_to_dtype(bound.arguments[x], compute_dtype)
    114     for x in self.type_promoting_arg_names  # type: ignore[union-attr]
    115     if x in bound.arguments.keys()
    116 }
    117 bound.arguments.update(promoted_args)
--> 119 result = fn(**bound.arguments)
    121 if isinstance(result, TensorLike):
    122     return _maybe_convert_to_dtype(result, result_dtype)

File /scratch/ngimel/work/pytorch/torch/_refs/__init__.py:504, in fill(a, value)
    500 if not utils.is_weakly_lesser_type(type(value), python_type):
    501     msg = "value argument of type {0} cannot be safely cast to type {1}!".format(
    502         type(value), python_type
    503     )
--> 504     raise ValueError(msg)
    506 return prims.fill(a, value)

ValueError: value argument of type <class 'float'> cannot be safely cast to type <class 'int'>!
```

Issue #916, opened 2022-08-19
tabulate.tabulate causes a lot of memory to be allocated


mlazos:
Calling tabulate.tabulate to render the output graph causes 1 GB of additional memory allocation in yolov3.

This is gated on the log level now, but as a result when DEBUG or INFO log levels are set, this will cause additional memory to be allocated.

https://github.com/pytorch/torchdynamo/blob/0e4d5ee9db6dfeba4d424b2e5d2ad96c79f1dffa/torchdynamo/utils.py#:~:text=tabulate(node_specs%2C%20headers%3D%5B%22opcode%22%2C%20%22name%22%2C%20%22target%22%2C%20%22args%22%2C%20%22kwargs%22%5D)

jansel:
What!  That is crazy!

Is tabulate doing something silly like cloning all the inputs?  Is the graph huge?

mlazos:
It's super weird, I even converted all of the inputs to a string in a separate function, and it *still* had the extra allocation. It doesn't even have direct access to the graph nodes themselves.

PR #1071, opened 2022-08-27
[inductor] TORCHINDUCTOR_TRACE=1 debug mode

jansel:
This is a new debugging tool designed to make it easier to debug/understand the internals of  TorchInductor.

Setting the environment variable `TORCHINDUCTOR_TRACE=1` will cause a debug trace directory to be created and printed:
```
$ env TORCHINDUCTOR_TRACE=1 python repro.py
torchinductor.debug: [WARNING] model_forward_0 debug trace: /tmp/torchinductor_jansel/rh/crhwqgmbqtchqt3v3wdeeszjb352m4vbjbvdovaaeqpzi7tdjxqr.debug
```

Here is an [example debug directory output](https://gist.github.com/jansel/f4af078791ad681a0d4094adeb844396) for the test program:
```
torch.nn.Sequential(
        torch.nn.Linear(10, 10),
        torch.nn.LayerNorm(10),
        torch.nn.ReLU(),
    )
```

Note each file in that debug trace can be enabled/disabled via `torchinductor.config.trace.*`.  The profile and the diagram are both disabled by default since they are expensive to generate.

A single node in this new debug format looks like:
```
buf1: SchedulerNode(ComputedBuffer)
buf1.writes = 
    {   MemoryDep(name='buf1', index=0, size=()),
        MemoryDep(name='buf1', index=0, size=(s0,))}
buf1.unmet_dependencies = {MemoryDep(name='buf0', index=c0, size=(s0,))}
buf1.met_dependencies = {MemoryDep(name='primals_2', index=c0, size=(s0,))}
buf1.group.device = cuda:0
buf1.group.iteration = (1, s0)
buf1.sizes = ([], [s0])
class buf1_loop_body:
    var_ranges = {z0: s0}
    index0 = z0
    index1 = 0
    def body(self, ops):
        get_index = self.get_index('index0')
        load = ops.load('buf0', get_index, False)
        get_index_1 = self.get_index('index0')
        load_1 = ops.load('primals_2', get_index_1, False)
        add = ops.add(load, load_1)
        get_index_2 = self.get_index('index1')
        reduction = ops.reduction('buf1', torch.float32, torch.float32, 'sum', get_index_2, add)
        return reduction
```
See the [example debug directory output](https://gist.github.com/jansel/f4af078791ad681a0d4094adeb844396) for more examples.

### DIFF
--- a/torchinductor/compile_fx.py
+++ b/torchinductor/compile_fx.py
@@ -1,25 +1,21 @@
 import dataclasses
 import functools
 import logging
-import operator
 from typing import List
 
 import torch.fx
 from functorch.compile import min_cut_rematerialization_partition
 
-from torchdynamo.debug_utils import wrap_debug
 from torchdynamo.optimizations.backends import aot_autograd
 from torchdynamo.optimizations.normalize import normalize_ir
-from torchdynamo.testing import same
 from torchdynamo.utils import identity
-from torchdynamo.utils import init_logging
 
 from . import config
 from . import overrides
+from .debug import DebugContext
 from .decomposition import select_decomp_table
 from .graph import GraphLowering
 from .utils import ceildiv
-from .utils import gen_gm_and_inputs
 from .virtualized import V
 
 log = logging.getLogger(__name__)
@@ -40,36 +36,17 @@ class BoxedBool:
         return False
 
 
-class CheckEachNode(torch.fx.Interpreter):
-    def call_function(self, target, args, kwargs):
-        expected = target(*args, **kwargs)
-        if target in (operator.getitem,):
-            return expected
-
-        gm, gm_inps = gen_gm_and_inputs(target, args, kwargs)
-        graph = GraphLowering(gm)
-        with V.set_graph_handler(graph):
-            graph.run(*args, **kwargs)
-            actual = graph.compile_to_fn()(*gm_inps)
-
-        if isinstance(expected, torch.Tensor):
-            actual = actual[0]
-
-        print(target, same(expected, actual))
-        assert same(expected, actual)
-
-        return expected
-
-
-@functools.partial(wrap_debug, compiler_name="inductor")
+@DebugContext.wrap
 def compile_fx_inner(
     gm: torch.fx.GraphModule,
     example_inputs: List[torch.Tensor],
     wrap=identity,
     cudagraphs=None,
     num_fixed=0,
+    is_backward=False,
 ):
-    init_logging()
+    log.info("Compiling %s graph", "BACKWARDS" if is_backward else "FORWARDS")
+    V.debug.fx_graph(gm, example_inputs)
 
     if cudagraphs is None:
         cudagraphs = config.triton.cudagraphs
@@ -182,10 +159,6 @@ def count_tangents(fx_g: torch.fx.GraphModule):
 
 def compile_fx(model_: torch.fx.GraphModule, example_inputs_: List[torch.Tensor]):
     """Main entrypoint to a compile given FX graph"""
-    logging.getLogger("torchinductor").setLevel(
-        logging.DEBUG if config.debug else logging.WARNING
-    )
-
     with overrides.patch_functions():
         model_ = normalize_ir(model_, example_inputs_)
         model_ = overrides.replace_fx(model_)
@@ -193,21 +166,19 @@ def compile_fx(model_: torch.fx.GraphModule, example_inputs_: List[torch.Tensor]
     cudagraphs = BoxedBool(config.triton.cudagraphs)
 
     def fw_compiler(model: torch.fx.GraphModule, example_inputs):
-        if config.debug:
-            print("FORWARD GRAPH:")
-            model.graph.print_tabular()
         fixed = len(example_inputs) - num_example_inputs
         return compile_fx_inner(
             model, example_inputs, num_fixed=fixed, cudagraphs=cudagraphs
         )
 
     def bw_compiler(model: torch.fx.GraphModule, example_inputs):
-        if config.debug:
-            print("BACKWARD GRAPH:")
-            model.graph.print_tabular()
         fixed = count_tangents(model)
         return compile_fx_inner(
-            model, example_inputs, num_fixed=fixed, cudagraphs=cudagraphs
+            model,
+            example_inputs,
+            num_fixed=fixed,
+            cudagraphs=cudagraphs,
+            is_backward=True,
         )
 
     with overrides.patch_functions():
--- a/torchinductor/config.py
+++ b/torchinductor/config.py
@@ -1,3 +1,5 @@
+import os
+
 # add some debug printouts
 debug = False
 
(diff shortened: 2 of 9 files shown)


torchinductor/config.py
```diff
@@ -101,3 +103,33 @@ class triton:
...
+# create a directory containing lots of debug information
+class trace:
+    # master switch for all debugging flags below
+    enabled = os.environ.get("TORCHINDUCTOR_TRACE", "0") == "1"
+
+    # Save python logger call >=logging.DEBUG
+    debug_log = True
+
+    # Save python logger call >=logging.INFO
+    info_log = False
```
voznesenskym, line 117: Maybe just `config.trace.save_log_level`?

torchinductor/debug.py
```diff
@@ -138,3 +169,148 @@ def in_output(snode):
...
+
+class DebugFormatter:
+    def __init__(self, handler):
+        self.fopen = handler.fopen
+        self.filename = handler.filename
+        self.handler = handler
+
+    def fx_graph(self, gm: torch.fx.GraphModule, inputs: List[torch.Tensor]):
+        with self.fopen("fx_graph.py") as fd:
+            save_graph_repro(fd, gm, inputs, "inductor")
```
voznesenskym, line 298: This saves it as a repro - in the `mod = make_fx(Repro()... ` style. Not as an fx graph (`graph.print_tabular` and the like) as one would expect for something named `fx_graph` - perhaps add support for just dumping the input fx graph, and a separate save_as_repro flag?

torchinductor/compile_fx.py
```diff
@@ -40,36 +36,17 @@ def disable(obj):
...
     gm: torch.fx.GraphModule,
     example_inputs: List[torch.Tensor],
     wrap=identity,
     cudagraphs=None,
     num_fixed=0,
+    is_backward=False,
 ):
-    init_logging()
+    log.info("Compiling %s graph", "BACKWARDS" if is_backward else "FORWARDS")
+    V.debug.fx_graph(gm, example_inputs)
```
voznesenskym, line 49: Check the config.fx_graph flag here?

voznesenskym: Oh we do this through `__getattr__` - maybe leave a comment on the class to that end? it wasn't readily obvious.

torchinductor/debug.py
```diff
@@ -138,3 +169,148 @@ def in_output(snode):
...
+        with self.fopen("compile.stats") as fd:
+            stats = pstats.Stats(self._prof, stream=fd)
+            stats.strip_dirs()
+            stats.sort_stats("cumtime")
+            stats.print_stats(100)
+            stats.sort_stats("tottime")
+            stats.print_stats(100)
+
+    def __getattr__(self, name):
+        if config.trace.enabled and getattr(config.trace, name):
```
voznesenskym, line 274: assert hasattr?

torchinductor/ir.py
```diff
@@ -3239,6 +3240,19 @@ def __init__(self, fn, args, var_ranges):
         self.root_block = LoopBodyBlock(self, fn, args)
         self.indexing = None
 
+    def debug_str(self):
+        lines = [f"var_ranges = {dict(self.var_ranges)}"]
```
voznesenskym, line 3244: Other IR uses `__str__` and implements it through `str_helper` - maybe keep that pattern here?

torchinductor/config.py
```diff
@@ -101,3 +103,33 @@ class triton:
     # should we stop a fusion to allow better tiling?
     tiling_prevents_pointwise_fusion = True
     tiling_prevents_reduction_fusion = True
+
+
+# create a directory containing lots of debug information
+class trace:
```
voznesenskym, line 109: Maybe move this up to dynamo + inductor? It would be very useful to have all the way end to end. The debug log capture is general, and the dynamo out fx_graph is the inductor in fx_graph - but for dynamo we also would be able to add:

`record_guards`
`modified_bytecode`

etc

torchinductor/debug.py
```diff
@@ -138,3 +169,148 @@ def in_output(snode):
...
+        def inner(*args, **kwargs):
+            with DebugContext():
+                return fn(*args, **kwargs)
+
+        return wrap_debug(inner, compiler_name="inductor")
+
+    @staticmethod
+    def create_debug_dir():
+        for n in DebugContext._counter:
+            dirname = os.path.join(cache_dir(), f"debug.{os.getpid()}.{n}")
```
voznesenskym, line 189: Maybe make this a config option and/or env var? I'd imagine this would be useful to control per run without modifying cache_dir

torchinductor/debug.py
```diff
@@ -138,3 +169,148 @@ def in_output(snode):
...
+            stats.print_stats(100)
+            stats.sort_stats("tottime")
+            stats.print_stats(100)
+
+    def __getattr__(self, name):
+        if config.trace.enabled and getattr(config.trace, name):
+            try:
+                return getattr(DebugFormatter(self), name)
+            except Exception:
+                log.warning("Ignoring exception in debug code", exc_info=True)
```
voznesenskym, line 278: This feels undesirable? I would want an exception in debug mode to stop the program, as proceeding after an exception in my debugging tools could mean something has gone wrong, or the output is not to be trusted / different from how I configured it / etc.

torchinductor/ir.py
voznesenskym: If we want __str__ short and debug_str() long, maybe we still share a debug_str() for both, but have it take a bool?

```
def __str__(self):
     return self.debug_str(long=False)

def debug_str(self, long=True):
     return "..."
```

torchinductor/ir.py
```diff
@@ -3372,8 +3386,16 @@ def set_indirect(new_var):
...
         )
-        return gm.forward(V.get_ops_handler())
+
+    def __call__(self):
+        return self.make_gm().forward(V.get_ops_handler())
+
+    def debug_str(self, name="block"):
+        code = self.make_gm().code
+        return re.sub(
+            r";[^\n]*", "", code.strip().replace("def forward(", f"def {name}(")
```
voznesenskym, line 3400: document the regex please

torchinductor/debug.py
jansel: It will throw an AttributeError if the attribute is missing, so this seems redundant.

torchinductor/ir.py
jansel: This is intentional, the string can get very long (it is an entire FX graph)

torchinductor/config.py
jansel: I actually find it useful to save both.  The debug level log can get too noisy, but it is nice to have it to reference if I want more details.

torchinductor/debug.py
jansel: We have seen graph.print_tabular() [causes 1 GB of additional memory allocation](https://github.com/pytorch/torchdynamo/issues/916), so I am trying to avoid using it.

The graph repro contains strictly more information, and I actually find a bit more readable.

torchinductor/config.py
jansel: There are some uses of inductor alone, so I want the debugging tools to work without dynamo.  There is also a one-to-many relationship between the two, so they don't fit neatly together.

torchinductor/debug.py
jansel: This directory is only a temporary one, so the name doesn't matter.  We rename this later to sit right beside the output code.  So if we write `foo123.py` this will be named `foo123.debug/`

torchinductor/debug.py
jansel: For an end user I disagree.  Imagine you are trying to debug a problem with `gdb`, then `gdb` throws an internal exception.  That isn't helpful, it is frustrating.  You are trying to debug your issue, not debug your debugging tools.

We are writing lots of unrelated files, and this failure only means you lose one of those files.  So we still generate lots of useful debug info.

voznesenskym: I'd fix my installation before I proceed, but it's a matter of opinion, and I don't feel too strongly about this.

voznesenskym (approved):
LGTM

torchinductor/debug.py
Chillee: I actually also detest the “graph tabular” format haha.

PR #1074, opened 2022-08-28
[inductor] Fix lru_cache for simplify_with_ranges

jansel:
This updates the caching for `simplify_with_ranges` so the cache is cleared when new replacements are added.

### DIFF
--- a/torchdynamo/debug_utils.py
+++ b/torchdynamo/debug_utils.py
@@ -19,6 +19,33 @@ def minifier_dir():
     return f"/tmp/minifier_{getpass.getuser()}"
 
 
+@functools.lru_cache(None)  # subprocess is expensive
+def _cuda_system_info_comment():
+    if not torch.cuda.is_available():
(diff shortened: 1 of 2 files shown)


torchdynamo/debug_utils.py
```diff
@@ -19,6 +19,33 @@ def minifier_dir():
     return f"/tmp/minifier_{getpass.getuser()}"
 
 
+@functools.lru_cache(None)  # subprocess is expensive
+def _cuda_system_info_comment():
```
voznesenskym, line 23: Meta comment - if you pull this into its own PR, I can stamp this. I do not know the rest of the code well enough to do so.

jansel: Yeah, I snuck this one in because it seemed like a tiny change.  I was noticing this function take half of some profiles.  :)

Chillee (approved):

PR #1075, opened 2022-08-28
[inductor] Don't autotune contiguous reductions

jansel:
This PR should improve compile times.  Previously we spent >50% of our time inside the Triton autotuner.

This removes the bulk of those autotuner calls by using a heuristic to schedule contiguous reductions instead of autotuning them.  I'd estimate this is a ~2x improvement in compile time based on some limited testing.

This also adds caching of autotuning results, which will speed up compile times dramatically (I measured 8.1x on TorchBench, including the other changes) when your cache is warm.

As a bonus, this also speeds up our geomean perf by a couple percent.  The prior autotuner was missing some choices that help some models.  Specifically for `numel=1024, reduction_numel=512`, the new heuristic picks a block of `[1, 512]` -- which is faster than the autotuner's chosen config of `[4, 512]`.

### DIFF
--- a/torchinductor/codegen/triton.py
+++ b/torchinductor/codegen/triton.py
@@ -14,7 +14,6 @@ import torch
 
 import torchinductor
 
-from .. import codecache
 from .. import config
 from .. import ir
 from ..utils import has_triton_libdevice
@@ -429,7 +428,7 @@ class TritonKernel(Kernel):
     overrides = TritonOverrides
     sexpr = texpr
 
-    def __init__(self, *groups, pid_cache={}):
+    def __init__(self, *groups, pid_cache={}, is_contiguous=False):
         super(TritonKernel, self).__init__()
         self.numels = [V.graph.sizevars.simplify(s) for s in groups]
         self.range_trees = []
@@ -442,6 +441,7 @@ class TritonKernel(Kernel):
         self.suffix = IndentedBuffer()
         self.outside_loop_vars = set()
         self.initialize_range_tree(pid_cache)
+        self.is_contiguous = is_contiguous
 
     def initialize_range_tree(self, pid_cache):
         names = ["xindex", "yindex", "zindex"][: len(self.numels) - 1] + ["rindex"]
@@ -889,14 +889,13 @@ class TritonKernel(Kernel):
                 f"""
                     import triton
                     import triton.language as tl
-                    from {codecache.__name__} import {heuristics}
-
+                    from torchinductor.triton_ops.autotune import {heuristics}
                 """
             )
 
         code.splice(
             f"""
-                @{heuristics}(size_hints={size_hints!r})
+                @{heuristics}(size_hints={size_hints!r}, contiguous={self.is_contiguous!r}, filename=__file__)
                 @triton.jit
             """
         )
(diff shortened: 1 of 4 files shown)


torchinductor/codegen/triton.py
```diff
@@ -1098,10 +1097,21 @@ def end_current_reduction_loop():
         log.info("schedule: %s", node_schedule)
         return self.codegen_node_schedule(node_schedule, numel, rnumel)
 
+    @staticmethod
+    def is_contiguous(node):
+        if node in (EnableReduction, DisableReduction):
+            return True
+        return all(
+            dep.is_contiguous()
+            for dep in itertools.chain(node.read_writes.reads, node.read_writes.writes)
```
ngimel, line 1106: should it depend on whether reduction is inner/outer, and not just on whether the dep is contiguous? Also, even for contiguous reductions we needed different autotuning configs (relying on just one resulted in 1.5x or so slower kernels)

torchinductor/triton_ops/autotune.py
```diff
@@ -18,23 +23,99 @@
...
+            def call_and_store_cache(self, *args, **kwargs):
+                result = triton.code_gen.Autotuner.__call__(self, *args, **kwargs)
+                with open(cache_filename, "w") as fd:
+                    fd.write(
+                        json.dumps(
+                            {**self.best_config.kwargs, "configs_hash": configs_hash}
+                        )
+                    )
+                self.configs = [self.best_config]
+                del Autotuner.__call__  # uninstall this hook
```
ngimel, line 115: lol probably we should upstream it to triton to avoid tricks like this

jansel: Yeah, agreed.  I'll take that as an action item.

torchinductor/codegen/triton.py
jansel: "contiguous" here means if the memory indexing formula is contiguous, not if the tensor is contiguous.  So I believe outer reduction memory deps will show up as `x0 + r0*stride`, and looks like transposed memory accesses with `is_contiguous()` being false.

I believe my change from 
```
triton_config_reduction(size_hints, 1, 2048)
```
to
```
triton_config_reduction(size_hints, 1, (rnumel if 256 <= rnumel < 2048 else 2048))
```
Fixed the issue where we needed to autotune different configs for contiguous kernels.  I actually saw some speedups from this change alone, and it improved the gmean perf on torchbench on my local GPU.

Is there a specific model or sizes I should check?

ngimel: I saw forward and backward kernels for layer norm requiring different configs (2x1024 vs 8 x 512 when rnumel was 1024) in hf torchbench models, but perf results show they are fine.

ngimel (approved):

soumith:
you need to start figuring out ways to invalidate disk cache -- atleast manually, where you bump up a version number in the disk cache manually -- but would be really good to have a way that this is just done automatically (more deeper schema of the disk-cache might needed)

ngimel:
We are caching on full kernel source and on autotuning configs not changing, so changes to inductor should be fine, but yeah we should have a manual way of invalidating cache.

jansel:
You can clear the cache with `rm -rf /tmp/torchinductor_$USER`

PR #1084, opened 2022-08-29
[Fix] Unless I read it wrong, it should be v, not x.

wschin:
As title.


jansel (approved):

voznesenskym:
Good catch. Can you add a test?

voznesenskym (approved):

PR #1087, opened 2022-08-29
Improvements to @dynamo_timed

wconstab:
- order keys in the order of entry not exit
- hook up csv output to benchmark runner

Example output:
image

(actual sheet:)
https://docs.google.com/spreadsheets/d/14uEHPAQDlOx1nPTac4i_3aNdrdQZoJmoxmrLfTBd6r4/edit?usp=sharing

### DIFF
--- a/benchmarks/common.py
+++ b/benchmarks/common.py
@@ -426,11 +426,14 @@ def speedup_experiment(args, model_iter_fn, model, example_inputs):
         ("dev", "name", "batch_size", "speedup"),
         [current_device, current_name, current_batch_size, float(speedup)],
     )
-    headers, data = torchdynamo.utils.compile_times(repr='csv', aggregate=True)
+    headers, data = torchdynamo.utils.compile_times(repr="csv", aggregate=True)
+    assert (
+        output_filename.find(".csv") > 0
+    ), f"expected output_filename to be a .csv, but got {output_filename}"
     output_csv(
-        "compilation_metrics.csv",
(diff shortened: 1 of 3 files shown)


wconstab:
@anijain2305 I'm wondering how you'd like the benchmark runner hooked up.  and filename, etc.

also wondering what will happen if one row has a missing field.

benchmarks/common.py
```diff
@@ -426,6 +426,12 @@ def maybe_profile(*args, **kwargs):
         ("dev", "name", "batch_size", "speedup"),
         [current_device, current_name, current_batch_size, float(speedup)],
     )
+    headers, data = torchdynamo.utils.compile_times(repr='csv', aggregate=True)
+    output_csv(
+        "compilation_metrics.csv",
```
anijain2305, line 431: Maybe use the string from output_filename and add "compilation_metrics" at the end? We generate different output files for different backends, dtypes etc. So, we could use the same pattern in the runner.

anijain2305 (approved):
LGTM. Only 1 small comment. 

IIUC, we don't need any changes in the runner script, because the existing command will also save this new csv file

We could simply import this spreadsheet manually in a CSV file.

anijain2305:
> also wondering what will happen if one row has a missing field.

This is ok, we can use pandas to replace it with zero.

voznesenskym (approved):

anijain2305 (approved):

PR #1089, opened 2022-08-30
Record/replay error tool for dynamo

mlazos:
Adds a tool which serializes vars and functions that were accessed during dynamo symbolic execution of the current frame.

- Adds `ExecutionRecorder` to record module accesses and local/global variable accesses
- On error (an exception caught at the convert frame level), the execution recorder converts each recorded module to a `DummyModule` which contains the serialized vars and functions that were accessed. The record is serialized to a file \<function name\>\_\<exeception name\>\_\<line num\>.rec
- Adds a new frontend to dynamo, `replay` which will replay the execution from a given record file

Caveats:
- This currently only records accesses that were traced by dynamo, if dynamo skips them or they're disabled, they will be missed.
- This does not serialize torch module/variable accesses, torch should be installed in the replay environment.


torchdynamo/symbolic_convert.py
```diff
@@ -490,31 +508,40 @@ def IMPORT_NAME(self, inst):
...
+            else:
+                source = self.import_source(module_name)
+
+        if config.replay_record_enabled:
+            self.exec_recorder.add_local_mod(module_name, value)
 
         if is_allowed(value):
             self.push(TorchVariable(value, source=source))
-        elif istype(value, types.ModuleType):
+        elif istype(value, (types.ModuleType, DummyModule)):
```
jansel, line 544: Can't we make `DummyModule` a subclass of `types.ModuleType` so this isn't needed everywhere?

torchdynamo/convert_frame.py
```diff
@@ -197,24 +214,41 @@ def format_error_msg(exc, frame):
...
             msg += f"\nfrom user code:\n {''.join(traceback.format_list([exc.real_stack[-1]]))}"
 
+        msg += replay_record_msg()
+
         msg += "\nSet torchdynamo.config.verbose=True for more information\n"
     msg += "=" * 10
     return msg
 
 
+def exception_handler(e, code, frame=None):
```
voznesenskym, line 243: Not super important, but writing files in the middle of an exception handling stack feels weird. Would we lose anything if this were called from finally instead?

torchdynamo/symbolic_convert.py
```diff
@@ -490,31 +508,40 @@ def IMPORT_NAME(self, inst):
...
-        # normally, the top-level package (the name up till the first dot) is
-        # returned, not the module named by module_name. However, when a
-        # non-empty fromlist argument is given, the module named by name is
-        # returned. Therefore, we set the source correctly here.
-        if not fromlist:
-            top_level_module_name = module_name.partition(".")[0]
-            source = self.import_source(top_level_module_name)
+        # Are we replaying? if so, load recorded module
+        recorded_name = ExecutionRecorder.LOCAL_MOD_PREFIX + module_name
+        if recorded_name in self.f_globals:
```
voznesenskym, line 513: What if we have multiple top level torchdynamo.optimize calls? Is it possible to get into a state where one leaves behind an artifact in globals, and then incorrectly causes the others to behave as if they are replaying?

voznesenskym:
Can we add tests for this? A good bit of logic here in core code paths that I don't think would be easy to ensure we do not rot away without tests.

torchdynamo/symbolic_convert.py
mlazos: This was my first approach, but it breaks if we do this, because `dill` tries to import it.

torchdynamo/convert_frame.py
mlazos: I can do this, 

> Can we add tests for this? A good bit of logic here in core code paths that I don't think would be easy to ensure we do not rot away without tests.

Yeah I'm definitely going to add tests for this, using eval or something.

mlazos:
> Can we add tests for this? A good bit of logic here in core code paths that I don't think would be easy to ensure we do not rot away without tests.

Yeah I'm definitely going to add tests for this, currently doing that

torchdynamo/convert_frame.py
mlazos: I take this back, a finally would run whether an error gets thrown or not, I don't think there is another place that this can be handled from. If you're worried about writing a file throwing another exception, I can wrap this in nested try/catch and print a warning if there is an issue with writing the file.

torchdynamo/symbolic_convert.py
mlazos: I don't think that's possible: when recording a DummyModule tree is built which doesn't modify f_globals.  Then when replaying this DummyModule tree is passed in to populate f_globals when calling torchdynamo.replay (this isn't the actual global environment, it's just a map). Now when import a module here, we check f_globals for it, and when we import a local module we also store it in f_globals with this prefix. The true global environment is never modified.

torchdynamo/convert_frame.py
mlazos: Handled the file operations with a wrapping try/except

jansel (approved):

PR #1095, opened 2022-08-30
Add aten export flag

voznesenskym:
```
def fn(a, b):
    x = a / (torch.abs(a) + 1)
    c, idx = torch.sort(x, stable=True)
    b = b * -1
    return x * b, c * idx


torchdynamo.config.export_aten_graph = False
print(torchdynamo.export(fn, torch.randn([1, 1]), torch.randn([1, 1]))[0])
torchdynamo.reset()
torchdynamo.config.export_aten_graph = True
print(torchdynamo.export(fn, torch.randn([1, 1]), torch.randn([1, 1]))[0])

```
Output:
```
GraphModule()



def forward(self, arg0, arg1):
    abs_1 = torch.abs(arg0)
    add = abs_1 + 1;  abs_1 = None
    truediv = arg0 / add;  arg0 = add = None
    sort = torch.sort(truediv, stable = True)
    getitem = sort[0]
    getitem_1 = sort[1];  sort = None
    mul = arg1 * -1;  arg1 = None
    mul_1 = truediv * mul;  truediv = mul = None
    mul_2 = getitem * getitem_1;  getitem = getitem_1 = None
    return (mul_1, mul_2)
    
GraphModule()



def forward(self, arg0, arg1):
    abs_default = torch.ops.aten.abs.default(arg0)
    add_tensor = torch.ops.aten.add.Tensor(abs_default, 1);  abs_default = None
    div_tensor = torch.ops.aten.div.Tensor(arg0, add_tensor);  arg0 = add_tensor = None
    sort_stable = torch.ops.aten.sort.stable(div_tensor, stable = True)
    getitem = sort_stable[0]
    getitem_1 = sort_stable[1];  sort_stable = None
    mul_tensor = torch.ops.aten.mul.Tensor(arg1, -1);  arg1 = None
    mul_tensor_1 = torch.ops.aten.mul.Tensor(div_tensor, mul_tensor);  div_tensor = mul_tensor = None
    mul_tensor_2 = torch.ops.aten.mul.Tensor(getitem, getitem_1);  getitem = getitem_1 = None
    return (mul_tensor_1, mul_tensor_2)
```

### DIFF
--- a/torchdynamo/config.py
+++ b/torchdynamo/config.py
@@ -110,9 +110,6 @@ repro_level = int(os.environ.get("COMPILER_REPRO_LEVEL", 0))
 # When this flag is set to False, we introduce a graph break instead of capturing.
 capture_scalar_outputs = False
 
-# True -> .export() exports a graph of aten ops.
-export_aten_graph = False
-
 
 class _AccessLimitingConfig(ModuleType):
     def __setattr__(self, name, value):
--- a/torchdynamo/config.py
+++ b/torchdynamo/config.py
@@ -110,6 +110,8 @@ repro_level = int(os.environ.get("COMPILER_REPRO_LEVEL", 0))
 # When this flag is set to False, we introduce a graph break instead of capturing.
 capture_scalar_outputs = False
 
+# True -> .export() exports a graph of aten ops.
+export_aten_graph = False 
 
 class _AccessLimitingConfig(ModuleType):
     def __setattr__(self, name, value):
--- a/torchdynamo/config.py
+++ b/torchdynamo/config.py
@@ -111,7 +111,8 @@ repro_level = int(os.environ.get("COMPILER_REPRO_LEVEL", 0))
 capture_scalar_outputs = False
 
 # True -> .export() exports a graph of aten ops.
-export_aten_graph = False 
+export_aten_graph = False
+
 
 class _AccessLimitingConfig(ModuleType):
     def __setattr__(self, name, value):
--- a/torchdynamo/eval_frame.py
+++ b/torchdynamo/eval_frame.py
@@ -397,7 +397,7 @@ def explain(f, *args, **kwargs):
     return explanation, out_guards, graphs, ops_per_graph
 
 
-def export(f, *args, **kwargs):
+def export(f, *args, aten_graph=False, **kwargs):
     f = innermost_fn(f)
 
     graph = None
@@ -504,7 +504,7 @@ def export(f, *args, **kwargs):
 
             return super().output(target, (new_result,), {})
 
-    if config.export_aten_graph:
+    if aten_graph:
         graph = make_fx(graph)(*graph_captured_input)
 
     new_graph = ChangeInputOutputSignature(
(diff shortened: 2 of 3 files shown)


torchdynamo/config.py
```diff
@@ -110,6 +110,9 @@
 # When this flag is set to False, we introduce a graph break instead of capturing.
 capture_scalar_outputs = False
 
+# True -> .export() exports a graph of aten ops.
+export_aten_graph = False
```
voznesenskym, line 114: False default for back compat

jansel: Should this be kwarg?

voznesenskym: done

SherlockNoMad (approved):
ShipIt!

torchdynamo/eval_frame.py
```diff
@@ -396,7 +397,7 @@ def guard_export_print(guards):
     return explanation, out_guards, graphs, ops_per_graph
 
 
-def export(f, *args, **kwargs):
+def export(f, *args, aten_graph=False, **kwargs):
```
SherlockNoMad, line 400: We might need to expose decomp_table and use_functionalization in the export API latter.

jansel (approved):

torchdynamo/eval_frame.py
voznesenskym: Agreed.

suo: seems like those functionalities are better exposed as passes?

silvasean: Functionalization actually requires running the code and retracing, so it is a bit more involved than just "running a pass" and could benefit from already having the input args available here for a seamless user experience.

This is very timely -- I was just playing with this locally for Torch-MLIR and I hit a few stumbling blocks with running functionalization and scripting the result.

voznesenskym: Glad it could be of use! There's more I plan to add here, happy to write something up.

silvasean: One feature request would be to support dynamic shapes. For example, to be able to say something like `torchdynamo.export(f, [TensorPlaceholder([-1, 12])])` -- we have an analogous API in Torch-MLIR [link](https://github.com/llvm/torch-mlir/blob/e6528f701a7cba8c0e10195bc4b838c143798fa0/python/torch_mlir/__init__.py#L74) and we would like to be able to use TorchDynamo under the hood rather than torch.jit.script/torch.jit.trace which each have their own limitations (script: obvious, trace: limited to static shapes).

voznesenskym: That's a good suggestion. We have symbolic shape support already, but not as an input placeholder. Let me mull this over and check out your API and see what we can do here.

silvasean: You can see examples of the API here: https://github.com/llvm/torch-mlir/blob/e6528f701a7cba8c0e10195bc4b838c143798fa0/python/test/compile_api/basic.py#L1

voznesenskym: Took a really rough stab at it here - needs a lot more tests and edge cases handled, but probably a good conversation starting PR to kind of see which direction we want to take it in https://github.com/pytorch/torchdynamo/pull/1123/

PR #1098, opened 2022-08-30
Graph break rng manipulation

eellison:
```
>>> with FakeTensorMode():
...     torch.set_rng_state(torch.ones([5056], dtype=torch.uint8))
... 
Segmentation fault (core dumped)
```

I think these were just getting silently traced out with aot autograd, which isn't preserving semantics.


torchdynamo/variables/tensor.py
```diff
@@ -113,6 +113,12 @@ def create(cls, tx, proxy, example_value=None, nnmodule=None, **options):
             def wrap_fake_exception(func):
                 return func()
 
+        if (
```
eellison, line 116: let me know if there's a list or something somewhere else I should be putting this (or starting)

torchdynamo/variables/tensor.py
```diff
@@ -113,6 +113,12 @@ def create(cls, tx, proxy, example_value=None, nnmodule=None, **options):
...
+            proxy.node.target is torch.set_rng_state
+            or proxy.node.target is torch.get_rng_state
+        ):
+            unimplemented("rng state manipulation")
```
voznesenskym, line 120: Logically sound, but a little odd that this is on TensorVariable.create(). Let me patch this down and see what builder is doing...

torchdynamo/variables/tensor.py
anijain2305: YOu could just add it here - https://github.com/pytorch/torchdynamo/blob/main/torchdynamo/allowed_functions.py#L66

torchdynamo/variables/tensor.py
voznesenskym: Yea - this is too far down. torch.* calls fallback to:

```
tensor_variable = TensorVariable.create(
                tx=tx,
                proxy=tx.output.create_proxy(
                    "call_function",
                    self.value,
                    *proxy_args_kwargs(args, kwargs),
                    current_tx=tx,
                ),
                **options,
            )
```

If we didn't figure out anything special for them.

Please (1) Move this unimplemented logic to `variables/torch.py` `def call_function`

and (2) Please make a dict of unimplemented_ops -> reason

then you can check something like

```
if self.value in unimplemented_ops:
     unimplemented(unimplemented_ops[self.value])
```

torchdynamo/variables/tensor.py
voznesenskym: This won't make an unimplemented though - this just elides it out - I actually dislike this list because it just removes func silently. Perhaps worth discussing at some point.

eellison: Potentially true but leaving for a separate PR

voznesenskym: Does this graph break? If it does not graph break, does this correctly go to the interpreter? Or do we just remove it silently? If it does not go to the interpreter, this will cause correctness issues.

eellison: uh, i believe so, on stats you get 
```
graph_break [('inline in skipfiles: get_rng_state /scratch/eellison/work/pytorch/torch/random.py', 1), ('inline in skipfiles: set_rng_state /scratch/eellison/work/pytorch/torch/random.py', 1)]
```
and 
the frame_count is 2

voznesenskym: Perfect, thank you :)

voznesenskym (approved):

mlazos (approved):

anijain2305 (approved):

torchdynamo/variables/tensor.py
anijain2305: > This won't make an unimplemented though - this just elides it out - I actually dislike this list because it just removes func silently. Perhaps worth discussing at some point.

@voznesenskym Oh I did not know that. Yes, lets clean it up if this needs refactoring. I will also look deeply to see if I am missing anything.

voznesenskym: I think its fine since it becomes an inline in skipfiles above, as per @eellison comment, which makes this fine!

eellison:
CycleGan was failing for me locally even without these changes.. also the changes are pretty unobjectionable so if it's failing probably there was some other bug (inducing a graph break should never induce correctness problems). disabling for now

voznesenskym:
> CycleGan was failing for me locally even without these changes.. also the changes are pretty unobjectionable so if it's failing probably there was some other bug (graph breaking should never create correctness problems). disabling for now

fine by me! Can you just make an issue to track re-enabling it?

eellison:
> Can you just make an issue to track re-enabling it?

will add it to the tracker

PR #1099, opened 2022-08-31
Add support for jagged tensor - remove from cudagraphs, better handling of multioutput 

voznesenskym:
Tested in fbcode with:

```
from typing import List

import torch

import torchdynamo

try:
    # pyre-ignore[21]
    from fbgemm_gpu import open_source  # noqa: F401

    from test_utils import gpu_available, gpu_unavailable
except Exception:
    torch.ops.load_library("//deeplearning/fbgemm/fbgemm_gpu:sparse_ops")
    torch.ops.load_library("//deeplearning/fbgemm/fbgemm_gpu:sparse_ops_cpu")


def _generate_jagged_tensor(
    num_jagged_dim: int,
    outer_dense_size: int,
    inner_dense_size: int,
    dtype: torch.dtype,
    device: torch.device,
):
    max_lengths = [3, 3, 3]
    x_offsets: List[torch.LongTensor] = []
    num_lengths = outer_dense_size
    for d in range(num_jagged_dim):
        lengths = torch.randint(
            low=0,
            high=max_lengths[d] * 2,
            size=(num_lengths,),
            device=device,
        )
        x_offsets.append(torch.ops.fbgemm.asynchronous_complete_cumsum(lengths))
        num_lengths = x_offsets[-1][-1].item()

    x_values = torch.rand(
        x_offsets[-1][-1] * inner_dense_size,
        dtype=dtype,
        device=device
    ).reshape(x_offsets[-1][-1].item(), inner_dense_size)

    return x_values, x_offsets, max_lengths

def fn(values, offsets, lengths):
    dense = torch.ops.fbgemm.jagged_to_padded_dense(values, offsets, lengths, 10)
    jagged_values, jagged_offsets = torch.ops.fbgemm.dense_to_jagged(
                dense, offsets, 2
            )
    return jagged_values, jagged_offsets

x_values, x_offsets, max_lengths = _generate_jagged_tensor(3, 3, 3, torch.float, 'cuda')


@torchdynamo.optimize("inductor", nopython=False)
def foo():
    return fn(x_values, x_offsets, max_lengths)

x = foo()
```

Couldn't figure out how to get fbgemm setup on OSS workflow.

### DIFF
--- a/torchinductor/ir.py
+++ b/torchinductor/ir.py
@@ -2271,8 +2271,8 @@ class ExternKernel(InputsKernel):
         if config.size_asserts:
             size = V.graph.sizevars.codegen_shape_tuple(self.get_size())
             stride = V.graph.sizevars.codegen_shape_tuple(self.get_stride())
-            wrapper.writeline(f"assert {self.get_name()}.size() == {size}")
-            wrapper.writeline(f"assert {self.get_name()}.stride() == {stride}")
+            wrapper.writeline(f"assert_size_match({self.get_name()}, {size})")
+            wrapper.writeline(f"assert_stride_match({self.get_name()}, {stride})")
 
     def get_group_stride(self):
         """
@@ -2748,6 +2748,9 @@ class FallbackKernel(ExternKernelAlloc):
         )
 
         if isinstance(example_output, (list, tuple)):
+            example_output_flat, example_output_spec = pytree.tree_flatten(
+                example_output
+            )
             packed = FallbackKernel(
                 MultiOutputLayout(tensor_args[0].get_device()),
                 kernel,
@@ -2759,18 +2762,18 @@ class FallbackKernel(ExternKernelAlloc):
                 (
                     MultiOutput(
                         FixedLayout(
-                            example_output[i].device,
-                            example_output[i].dtype,
-                            [sympy.Integer(s) for s in example_output[i].size()],
-                            [sympy.Integer(s) for s in example_output[i].stride()],
+                            example_output_flat[i].device,
+                            example_output_flat[i].dtype,
+                            [sympy.Integer(s) for s in example_output_flat[i].size()],
+                            [sympy.Integer(s) for s in example_output_flat[i].stride()],
                         ),
                         packed,
                         i,
                     )
-                    if example_output[i] is not None
+                    if example_output_flat[i] is not None
                     else None
                 )
-                for i in range(len(example_output))
+                for i in range(len(example_output_flat))
             ]
         else:
             return FallbackKernel(
--- a/torchinductor/utils.py
+++ b/torchinductor/utils.py
@@ -165,3 +165,39 @@ def cache_on_self(fn):
         return getattr(self, key)
 
     return wrapper
+
+
+def has_incompatible_cudagraph_ops(gm):
+    forbidden_list = set(
+        [
+            "fbgemm.dense_to_jagged.default",
+            "fbgemm.jagged_to_padded_dense.default",
+        ]
+    )
+    for node in gm.graph.nodes:
+        if str(node.target) in forbidden_list:
+            return True
+    return False
+
+
+def assert_size_match(item, size):
+    assert isinstance(size, tuple)
+    if size == ():
+        return
+
+    if isinstance(item, (list, tuple)):
+        return
+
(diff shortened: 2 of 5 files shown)


voznesenskym:
WIP don't review yet

torchinductor/utils.py
```diff
@@ -165,3 +165,39 @@ def wrapper(self):
...
+    )
+    for node in gm.graph.nodes:
+        if str(node.target) in forbidden_list:
+            return True
+    return False
+
+
+def assert_size_match(item, size):
+    assert isinstance(size, tuple)
+    if size == ():
```
ngimel, line 185: this seems more relaxed than previous version, is this intended?

ngimel (approved):

torchinductor/utils.py
voznesenskym: I was not sure what it means to check if size is () ... let me try to change this back and see what I hit.

voznesenskym: @jansel when do we want to do a size == () check?

jansel: What is item in this case?

torchinductor/ir.py
```diff
@@ -2748,6 +2748,9 @@ def unflatten_args(new_tensor_args, new_non_tensor_args):
         )
 
         if isinstance(example_output, (list, tuple)):
+            example_output_flat, example_output_spec = pytree.tree_flatten(
```
jansel, line 2751: We never actually use `example_output_spec` will this be correct? Won't it change the return value of the function?

torchinductor/utils.py
```diff
@@ -165,3 +165,39 @@ def wrapper(self):
...
+        return
+
+    if isinstance(item, (list, tuple)):
```
jansel, line 188: If we are calling this function then inductor thinks `item` is a tensor.  This is clearly wrong, so I think this isn't fixing the underlying issue (which is TorchInductor got the type wrong).

torchinductor/ir.py
voznesenskym: The return type is always a flat list - here we just flatten to make it. Before, we only iterated over the top level list. If we unflatten it back, there is code in lowering.py:

```
def fallback_handler(kernel):
    fallbacks.add(kernel)

    def handler(*args, **kwargs):
        result = ir.FallbackKernel.create(kernel, *args, **kwargs)
        if isinstance(result, (list, tuple)):
            return list(map(TensorBox.create, result))
        else:
            return TensorBox.create(result)
            
```

That would then have to be changed to flatten as well. If we never unflatten above, we can just return it as a flat list for this case. It's a little assumptive of this callsite/usecase, but it does not seem to be incorrect. 

We could always unflatten, return that, then flatten at callsites again as needed.

torchinductor/utils.py
voznesenskym: A tensor

jansel: Is the tensor 0D?  If so won't the normal assert work?

torchinductor/ir.py
jansel: Are the nested args unused in this example program?

voznesenskym: I believe they are used, but not sure how to check?

torchinductor/utils.py
voznesenskym: Yes - but perhaps instead of flattening, we need to keep the same top-level structure but then turn the nested contents into data that is a list or tuple of tensors, instead of flattening here. Let me try that.

voznesenskym: The problem here, I think, is that the contents of `example_output` is a set, but the size() per element is not uniform in our case. Let me think about this further. Jagged tensor breaks a lot of our `FixedLayout` assumptions. We may need something like JaggedLayout where we support indexing into this more complicated list of sizes in the shape.

voznesenskym: Or, we unflatten up there, and then maybe make StorageBox take a list...

Let me mull this over. I'd like to represent the data as correctly as possible here.

voznesenskym:
Changes Planned

torchinductor/utils.py
ngimel: Previously when size was `()` and tensor was not 0d, we would error out on assert. Here it looks like we just pass it.

jansel:

Issue #941, opened 2022-08-22
Dynamo is still not producing accurate Python stack traces


ezyang:
The frames don't have correct line numbers (if that's the problem? Hard to say.)

```
======================================================================                                 
FAIL: test_grid_sample_large_index_2d_cpu_float32 (__main__.TestNNDeviceTypeCPU)                       
----------------------------------------------------------------------                                 
Traceback (most recent call last):                                                                     
  File "/raid/ezyang/pytorch-scratch2/torch/testing/_internal/common_device_type.py", line 377, in instantiated_test                                                                                          
    result = test(self, **param_kwargs)                                                                
  File "/raid/ezyang/pytorch-scratch2/torch/testing/_internal/common_device_type.py", line 935, in dep_fn                                                                                                     
    return fn(self, *args, **kwargs)                                                                   
  File "test/test_nn.py", line 17339, in test_grid_sample_large_index_2d                               
    @dtypes(torch.float, torch.double)                                                                 
  File "test/test_nn.py", line 17339, in test_grid_sample_large_index_2d                               
    @dtypes(torch.float, torch.double)                                                                 
  File "test/test_nn.py", line 17339, in test_grid_sample_large_index_2d                               
    @dtypes(torch.float, torch.double)                                                                 
AssertionError: False is not true : View must use 64-bit indexing                                      
                                                                                                       
---------------------------------------------------------------------- 
```

Steps to reproduce:

1. Build PyTorch with https://github.com/ezyang/pytorch/pull/new/stack-trace-insuff-repro
2. `PYTORCH_TEST_WITH_DYNAMO=1 python test/test_nn.py -k test_grid_sample_large_index_2d_cpu_float32`

cc @mlazos

ezyang:
(dynamo shouldn't be failing this test either, but that's a separate matter...)

mlazos:
So I took a look at this, basically what's happening is there is a graph break, so dynamo introduces another function call (the __resume_x_x functions to continue after the graph break), but the name of the code object is the same as the original. This will be a recurring issue any time dynamo introduces new functions which didn't exist in the original code. I don't think it will be possible to recover the original stack trace (without the extra frames that dynamo introduces) because this is the interpreter calling dynamo generated code, without dynamo involvement at all, but what is possible is to track the original linenumbers and filenames and end up with something like the following:

<filename> line 17339, in test_grid_sample_large_index_2d 
<filename> <line number of original instructions> in torchdynamo_compiled_fn_0
[...]

PR #1105, opened 2022-09-01
fix line numbers/names of resume

mlazos:
Sets `co_name` of resume functions to \<graph break in \<fn_name\>\> and sets `co_firstlineno` to the line number of the graph break.

[before and after](https://gist.github.com/mlazos/fef55b64903c266bf0b69c9bf4bd2b2b)

Note:
This removes line numbers from code between the first instruction and the point that we jump to the code immediate after the graph break. As a result, if we are unrolling a loop with a graph break, we will jump back into this code, and will not have the correct line number. The reason for this is that line number tables don't allow the line number of the first instruction to be less than `co_firstlineno`. The line number tables only store deltas (only positive) from the previous line number, and this begins at `co_firstlineno`. 

Improves the state of #941

### DIFF
--- a/torchdynamo/resume_execution.py
+++ b/torchdynamo/resume_execution.py
@@ -99,18 +99,19 @@ class ContinueExecutionCache:
     generated_code_metadata = ExactWeakKeyDictionary()
 
     @classmethod
-    def lookup(cls, code, *key):
+    def lookup(cls, code, lineno, *key):
         if code not in cls.cache:
             cls.cache[code] = dict()
         key = tuple(key)
         if key not in cls.cache[code]:
-            cls.cache[code][key] = cls.generate(code, *key)
+            cls.cache[code][key] = cls.generate(code, lineno, *key)
         return cls.cache[code][key]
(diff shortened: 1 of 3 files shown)


torchdynamo/resume_execution.py
```diff
@@ -160,6 +163,14 @@ def update(instructions: List[Instruction], code_options: Dict[str, Any]):
             assert not hooks
 
             prefix.append(create_instruction("JUMP_ABSOLUTE", target=target))
+
+            # because the line number table monotonically increases from co_firstlineno
+            # remove starts_line for any instructions before the graph break instruction
+            # this will ensure the instructions after the break have the correct line numbers
+            target_ind = int(target.offset / 2)
+            for inst in instructions[0:target_ind]:
+                delattr(inst, "starts_line")
```
jansel, line 172: Shouldn't this be `= None`?

mlazos: Fixed, not sure why I did that.

PR #1106, opened 2022-09-01
Fix divide by zero in lowerings

voznesenskym:
Size inputs like (0, 256) will result in a divide by zero

### DIFF
--- a/torchinductor/lowering.py
+++ b/torchinductor/lowering.py
@@ -459,9 +459,14 @@ def expand(x, sizes):
     assert isinstance(sizes, (list, tuple))
     if tuple(x.get_size()) == tuple(sizes):
         return x
-    x.mark_reuse(
-        V.graph.sizevars.size_hint(sympy_product(sizes) / sympy_product(x.get_size()))
-    )
+
+    x_size_product = sympy_product(x.get_size())
+    if x_size_product > 0:
+        sizes_product = sympy_product(sizes)
+        x.mark_reuse(V.graph.sizevars.size_hint(sizes_product / x_size_product))
+    else:
+        x.mark_reuse(V.graph.sizevars.size_hint(0))
+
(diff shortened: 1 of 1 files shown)


torchinductor/lowering.py
```diff
@@ -459,9 +459,14 @@ def expand(x, sizes):
...
-    x.mark_reuse(
-        V.graph.sizevars.size_hint(sympy_product(sizes) / sympy_product(x.get_size()))
-    )
+
+    x_size_product = sympy_product(x.get_size())
+    if x_size_product > 0:
+        sizes_product = sympy_product(sizes)
+        x.mark_reuse(V.graph.sizevars.size_hint(sizes_product / x_size_product))
+    else:
+        x.mark_reuse(V.graph.sizevars.size_hint(0))
```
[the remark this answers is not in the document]
jansel: We can just skip the call the `mark_reuse()` entirely.  It won't do anything with 0 users.

jansel (approved):

torchinductor/lowering.py
```diff
@@ -459,9 +459,11 @@ def expand(x, sizes):
     assert isinstance(sizes, (list, tuple))
     if tuple(x.get_size()) == tuple(sizes):
         return x
-    x.mark_reuse(
-        V.graph.sizevars.size_hint(sympy_product(sizes) / sympy_product(x.get_size()))
-    )
+
+    x_size_product = sympy_product(x.get_size())
```
ngimel, line 463: while you are at it, can you add handling for negative ones in `sizes`?
```
In [7]: a=torch.randn(1,3,4)

In [8]: a.expand(3, -1, -1).size()
Out[8]: torch.Size([3, 3, 4])
```

voznesenskym: Separate PR, if you don't mind. I prefer to keep it to a single focus, easier to revert if necessary too.

PR #1107, opened 2022-09-01
Remove Context Manager in TorchDynamo tests

anijain2305:
This removes `with torchdynamo.optimize...` from all TorchDynamo tests.

There are only a couple of exceptions. These tests are specific to the usage of context manager. So, I have kept them for now.

Once this PR is merged, I will add a warning in the TorchDynamo context manager to point to the decorator usage.


jansel (approved):

penguinwu:
Remember to update examples in the documentation and internal/external Dynamo posts too.

PR #1108, opened 2022-09-01
Add `use_openmp` option to cxx config

malfet:
And do not use it by default when running on MacOS
Also introduce memoizable `cpp_supports_flag` and `cxx_flags` function

Above changes make inductor usable out-of-box on MacOS

### DIFF
--- a/torchinductor/codecache.py
+++ b/torchinductor/codecache.py
@@ -47,7 +47,7 @@ def write(source_code, ext, extra=""):
     return basename, path
 
 
-def cpp_compiler():
+def cpp_compiler() -> str:
     if isinstance(config.cpp.cxx, (list, tuple)):
         search = tuple(config.cpp.cxx)
     else:
@@ -56,7 +56,7 @@ def cpp_compiler():
 
 
 @functools.lru_cache(1)
-def cpp_compiler_search(search):
+def cpp_compiler_search(search) -> str:
     for cxx in search:
         try:
             if cxx is None:
@@ -68,6 +68,18 @@ def cpp_compiler_search(search):
             continue
     raise exc.InvalidCxxCompiler()
 
+@functools.lru_cache(1)
+def cpp_supports_flag(flag: str) -> bool:
+   rc = subprocess.run([cpp_compiler(), "-E", "-x", "c", flag, "-"], stdout=subprocess.PIPE, stderr=subprocess.PIPE, input="")
+   return rc.returncode == 0
+
+
+
+def cxx_flags() -> str:
+    flags = "-march=native -O3 -ffast-math -fno-finite-math-only"
+    if config.cpp.use_openmp:
+        flags += " -fopenmp"
+    return " ".join([f for f in flags.split() if cpp_supports_flag(f)])
 
 def install_gcc_via_conda():
     """On older systems, this is a quick way to get a modern compiler"""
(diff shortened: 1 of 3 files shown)


torchinductor/codecache.py
```diff
@@ -68,6 +68,18 @@ def cpp_compiler_search(search):
...
 
+@functools.lru_cache(1)
+def cpp_supports_flag(flag: str) -> bool:
+   rc = subprocess.run([cpp_compiler(), "-E", "-x", "c", flag, "-"], stdout=subprocess.PIPE, stderr=subprocess.PIPE, input="")
+   return rc.returncode == 0
+
+
+
+def cxx_flags() -> str:
+    flags = "-march=native -O3 -ffast-math -fno-finite-math-only"
```
voznesenskym, line 79: Can you make this config driven, but on by default? I feel like there is value in looking at the unoptimized code, as -O3 can inline and remove a lot of stuff.

torchinductor/codecache.py
```diff
@@ -68,6 +68,18 @@ def cpp_compiler_search(search):
...
+    if config.cpp.use_openmp:
+        flags += " -fopenmp"
+    return " ".join([f for f in flags.split() if cpp_supports_flag(f)])
```
voznesenskym, line 82: Do we really want a subprocess run per flag? seems easier to either map it per compiler, or just fail and fallback to running without said flags (as we do today). Thoughts?

malfet: Hmm, perhaps I don't know where the fallback code is, but `torchdynamo.optimize("inductor", nopython=True)(toy_example)` failed for me with `-march=native` unsupported flag

Also, please note that `cpp_supports_flag` memoizes the answers so it will call subprocess only once.

voznesenskym: Doesn't matter too much. The fallback would have to be written by you. You can also keep a mapping of compiler->flag, there's not exactly a massive set we support yet. No strong opinion.

torchinductor/codecache.py
```diff
@@ -68,6 +68,18 @@ def cpp_compiler_search(search):
             continue
     raise exc.InvalidCxxCompiler()
 
+@functools.lru_cache(1)
```
jansel, line 71:
```suggestion
@functools.lru_cache(128)
```
Larger cache size limit

torchinductor/codecache.py
jansel: For `-march=native` I feel we should try to replace that with the correct `-march=...` equivalent for clang.  

On gcc if you remove this flag the generated code won't use AVX vector instructions and be very slow.

torchinductor/codecache.py
```diff
@@ -68,6 +68,18 @@ def cpp_compiler_search(search):
...
+def cpp_supports_flag(flag: str) -> bool:
+   rc = subprocess.run([cpp_compiler(), "-E", "-x", "c", flag, "-"], stdout=subprocess.PIPE, stderr=subprocess.PIPE, input="")
+   return rc.returncode == 0
+
+
+
+def cxx_flags() -> str:
+    flags = "-march=native -ffast-math -fno-finite-math-only"
+    if config.cpp.use_openmp:
```
jansel, line 80: Should we set `config.cpp.threads=1` in this case?

jansel (changes requested):

torchinductor/codecache.py
soumith: `march=native` detects the current CPU and uses all instructions available on the detected CPU. If it detects that your CPU supports `avx` it will use them.

malfet: Well, here is the fun part - clang have `-march=native` on x86, but not on M1 (or probably not on aarch64 linux either)

jansel:

PR #1112, opened 2022-09-01
Add warning on using context manager

anijain2305:
@jansel Does this look ok? I want to tell users to avoid using context manager.

### DIFF
--- a/test/test_modules.py
+++ b/test/test_modules.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env pytest
 
 from copy import deepcopy
+from unittest.mock import patch
 
 import torch
 from torch.nn import functional as F
@@ -676,6 +677,7 @@ class NNModuleTests(torchdynamo.testing.TestCase):
         self.assertTrue(torchdynamo.testing.same(out2, out4))
         self.assertEqual(cnt.frame_count, 3)
 
+    @patch.object(torchdynamo.config, "raise_on_ctx_manager_usage", False)
     def test_generation_tag(self):
         cnt = torchdynamo.testing.CompileCounter()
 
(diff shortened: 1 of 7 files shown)


jansel:
Doesn't the decorator call the context manager?  Will this just warn all the time?

voznesenskym:
Can we just be bold and remove it now? This warning into deprecated into deleted flow doesn't hold much water when we are still in a pretty experimental and early stage, IMO. In semver terms, we are not yet 1.0.0 ;)

anijain2305:
> Doesn't the decorator call the context manager? Will this just warn all the time?

Decorator goes through `__call__` function here
https://github.com/pytorch/torchdynamo/blob/9292e4013c17f01c5b371fabcd6d5a2b1d9ba749/torchdynamo/eval_frame.py#L116

I am adding warning in the `__enter__` method which is called only with context manager. CI is failing because export is using context manager which I can quickly fix 
https://github.com/pytorch/torchdynamo/blob/9292e4013c17f01c5b371fabcd6d5a2b1d9ba749/torchdynamo/eval_frame.py#L466-L471

anijain2305:
> Can we just be bold and remove it now? This warning into deprecated into deleted flow doesn't hold much water when we are still in a pretty experimental and early stage, IMO. In semver terms, we are not yet 1.0.0 ;)

I would totally be on board with that. What do you think @jansel ?

jansel:
Perhaps to an fbgs search to check for usage in fbcode first.

anijain2305:
> Perhaps to an fbgs search to check for usage in fbcode first.

Ok, there are 4-5 places which seem like demos. But, I just realized that we Pytorch tests use this anyways. Lets try to get this one in first. And then clean up Pytorch and fb-internals and come back and remove the warning.

jansel (approved):

PR #1114, opened 2022-09-01
Using cProfiler

anijain2305:
@wconstab I used this to debug https://github.com/pytorch/torchdynamo/issues/525

This was the result. It was not directly obvious. But going through the stack, I found the culprit. Sending a separate PR for the fix. 

~~~
### Cprofile for _convert_frame_assert iter 1 ###
         286220504 function calls (233962842 primitive calls) in 83.329 seconds

   Ordered by: internal time
   List reduced from 1100 to 20 due to restriction <20>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
41821129/1714   31.516    0.000   72.515    0.042 /scratch/anijain/work/env/lib/python3.9/copy.py:128(deepcopy)
  2200500   10.596    0.000   43.838    0.000 /scratch/anijain/work/env/lib/python3.9/copy.py:226(_deepcopy_dict)
 83704980    7.091    0.000    7.093    0.000 {method 'get' of 'dict' objects}
2209659/1714    5.539    0.000   72.458    0.042 /scratch/anijain/work/env/lib/python3.9/copy.py:258(_reconstruct)
 54266296    4.164    0.000    4.164    0.000 {built-in method builtins.id}
1365676/4046    2.741    0.000    6.135    0.002 /scratch/anijain/work/torchdynamo/torchdynamo/variables/base.py:80(apply)
 35214941    2.612    0.000    2.612    0.000 /scratch/anijain/work/env/lib/python3.9/copy.py:182(_deepcopy_atomic)
  4405814    1.834    0.000    2.586    0.000 /scratch/anijain/work/env/lib/python3.9/copy.py:242(_keep_alive)
   227929    1.444    0.000    1.444    0.000 /scratch/anijain/work/torchdynamo/torchdynamo/variables/base.py:249(__init__)
  2203348    1.013    0.000    1.013    0.000 {method '__reduce_ex__' of 'object' objects}
8285240/8276442    0.995    0.000    1.008    0.000 {built-in method builtins.isinstance}
4417050/3680    0.963    0.000   69.649    0.019 /scratch/anijain/work/env/lib/python3.9/copy.py:263(<genexpr>)
     1714    0.949    0.001   69.642    0.041 /scratch/anijain/work/env/lib/python3.9/copy.py:200(_deepcopy_list)
  6865319    0.801    0.000    0.801    0.000 {built-in method builtins.getattr}
[...]

Issue #525, opened 2022-07-07
TorchDynamo sometime takes very long to symbolically interpret stackframes.


Chillee:
Looking at Bram's example

Example accessible internally: https://www.internalfb.com/phabricator/paste/view/P514325083

From @voznesenskym 

> Anyway, can confirm it super slow, can confirm it does not get past the first for loop, hanging on Y, dot = m(X_ref). Not sure if its hanging, as adding this in the NandNet forward:

> So its not hanging, but it certainly seems to be adding a ton of overhead and needing some special finagling,

anijain2305:
cc @wconstab Lets use your timers to diagnose this.

wconstab:
I've taken a look and found that this model is spending a lot of time in dynamo instruction translation.  I added some timing code to the translator and produced this chart of the time spent so far (it's streaming out data but not sure how far along it is, hasn't finished an iteration yet)

<img width="1017" alt="image" src="https://user-images.githubusercontent.com/4984825/187564982-15c00a0b-fb09-4675-9914-876bc7c28b8f.png">

will spend some more time thinking about how to instrument what's going on in call_function and see if there is anything we can do to speed it up

wconstab:
The overall structure of this model's forward() is to iterate over 21 layers, and inside each layer iterate over N sub-layers where N is related to the layer index.  So it's roughly an N^2 problem but its only about 400 inner loops, so that should hopefully be handleable.

Backward seems to be quite fast, which aligns with the issues coming from translation rather than compilation.

Optimizer is again quite slow, even slower than .forward()-
We fail the 'capturable' check in adam optimizer (due to not being on cuda) and take a [path the results in `.item()` calls](https://github.com/pytorch/pytorch/blob/master/torch/optim/adam.py#L309) for each of the model's many (hundreds) of parameters.  This causes a lot of dynamo graph breaks while trying to capture the optimizer.

voznesenskym:
> The overall structure of this model's forward() is to iterate over 21 layers, and inside each layer iterate over N sub-layers where N is related to the layer index. So it's roughly an N^2 problem but its only about 400 inner loops, so that should hopefully be handleable.
> 
> Backward seems to be quite fast, which aligns with the issues coming from translation rather than compilation.
> 
[...]

PR #1115, opened 2022-09-01
Improve Bytcode analysis time - Copying guards

anijain2305:
Related to https://github.com/pytorch/torchdynamo/pull/1114 and https://github.com/pytorch/torchdynamo/issues/525

deepcopy of guards is quite expensive because there are 1000s of guards. Also, these guards are copied multiple times to checkpoint state.

@jansel Do we need deepcopy here? iiuc, we only add new guards, but we dont mutate the guards. So, if we just use a `set()`, is that enough to checkpoint?

### DIFF
--- a/torchdynamo/output_graph.py
+++ b/torchdynamo/output_graph.py
@@ -1,5 +1,4 @@
 import collections
-import copy
 import itertools
 import logging
 import operator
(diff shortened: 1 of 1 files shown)


jansel (approved):
I think this should be fine.

PR #1120, opened 2022-09-02
Fix operator // for integer inputs

Lezcano:
* #1121
* __->__ #1120

Before this was returning floats rather than ints.


torchinductor/lowering.py
```diff
@@ -2877,6 +2877,20 @@ def fn(*args):
     )
 
 
+# TODO(lezcano) I believe the casting behaviour of prims.div is wrong
+# https://github.com/pytorch/pytorch/issues/84412
+# div prim performs truncation division on integer inputs
+#   and true division for floating and complex inputs
+@register_lowering([prims.div], broadcast=True)
+def div_prim(a, b):
+    is_integral = isinstance(a, (bool, int)) or is_integer_type(a)
```
voznesenskym, line 2886: is_integer_type already checks int. And we have an is_boolean_type

maybe
```suggestion
    is_integral = is_boolean_type(a) or is_integer_type(a)
```

If you want to be broader (numpy types, custom integral types), consider making an is_integral_type that looks like is_integer_type but calls into `isinstance(a, numbers.Integral)`. No strong opinion.

Lezcano: I copied this from primtorch and I didn't think too much about it. Fixed!

voznesenskym: LGTM - let @jansel take a look too.

jansel:
Ick... why don't we just have truediv/floordiv prims that are explicit?  

This behavior seems kind of broken.

cc @ezyang @ngimel @Chillee

Lezcano:
What I've seen in PrimTorch is that all the other `div`s can be implemented (sometimes in fairly non-trivial ways) in terms of this one. Now, the castings that are necessary to achieve this are quite a pain to follow. Took me a fair amount of time to understand what was dispatching to what in that jungle of decorators.

cc @mruberry

ngimel:
Given that we already have codegen for floordiv, I suggest we just add inductor-specific decomposition instead
```
@register_decomposition([aten.floor_divide.default])
def floordiv(a, b):
    return aten.div.Tensor_mode(a, b, rounding_mode="floor")
```
(it doesn't require signbit and fmod, and also doesn't pull in buggy `prim.div`. We could still add signbit and fmod, because why not)

Lezcano:
Cool, done. I'm leaving the fix for `prim.div` for correctness, in case it pops up in any other decomposition.

ngimel (approved):

ngimel:
Can you add a test for `//` to `test_torchinductor`?

Lezcano:
Added the tests. Feel free to merge if they look fine. I also caught an early return in one of the tests (see the last diff). I guess mypy is not run on the tests? @jansel

ngimel:
Hm, I'm not sure ghstack PRs merge correctly, can you redo it without ghstack?

Chillee:
@ngimel You have to use `ghstack land <pr>`

Lezcano:
Do I need to do anything?

desertfire:
I ran `ghstack land  https://github.com/pytorch/torchdynamo/pull/1120`, but it doesn't seem to land to master. @ezyang , anything I am missing?

ezyang:
looks like it's on master now to me

desertfire:
> https://github.com/pytorch/torchdynamo/pull/1120

Mind to land `https://github.com/pytorch/torchdynamo/pull/1121` as well?

PR #1121, opened 2022-09-02
Add necessary lowerings and decompositions to run adaptive_avg_pool2d

Lezcano:
* __->__ #1121
* #1120

As per title.


torchinductor/decomposition.py
```diff
@@ -17,6 +17,7 @@
 
 decompositions = get_decompositions(
     [
+        aten._adaptive_avg_pool2d,
```
jansel, line 20: Let's leave this using the existing lowering, since it is slightly faster due to masked load.

torchinductor/lowering.py
```diff
@@ -2846,6 +2845,11 @@ def truncdiv(a, b):
     return ops.truncdiv(a, b)
 
 
+@register_lowering(prims.signbit)
+def signbit(a):
+    return a < 0
```
jansel, line 2850: This can just be a decomp, doesn't need to be a lowering

jansel (changes requested):

Lezcano:
Done!

jansel (approved):

jansel:
It would be good to have a test for `signbit`

torchinductor/decomposition.py
```diff
@@ -32,6 +32,8 @@
         aten.embedding_dense_backward,
         aten.expand_as,
         aten.flip,
+        aten.floor_divide.default,
+        aten.fmod.Tensor,
```
ngimel, line 36: even though nominally we have fmod decomposition, attempting to use it leads to 
```
def forward(self, arg0_1, arg1_1):
    fmod = torch.ops.prims.fmod.default(arg0_1, arg1_1);  arg0_1 = arg1_1 = None
    return (fmod,)
    
torchinductor.graph: [WARNING] Creating implicit fallback for:
  target: prims.fmod.default
  args[0]: TensorBox(StorageBox(
    InputBuffer(name='arg0_1', layout=FixedLayout('cuda', torch.int32, size=[s0], stride=[1]))
  ))
  args[1]: TensorBox(StorageBox(
    InputBuffer(name='arg1_1', layout=FixedLayout('cuda', torch.int32, size=[s0], stride=[1]))
  ))
torchinductor.ir: [WARNING] Using FallbackKernel: torch.ops.prims.fmod.default

```
(because decomposition just calls directly into prim)
so again we'd either have to write a real decomposition on the pytorch side, or lowering on the inductor side to support fmod.

torchinductor/codegen/cpp.py
```diff
@@ -194,6 +194,10 @@ def sqrt(x):
     def rsqrt(x):
         return f"1 / std::sqrt({x})"
 
+    @staticmethod
+    def signbit(x):
```
ngimel, line 198: I don't see how this would ever be called given that we don't have lowering for signbit

jansel: I think you are missing a `register_pointwise(aten.signbit)` line.

Lezcano:
Done. I also added some tests. For signbit, there was no equivalent triton function, so I added an implementation. The implementation is not 100% correct, as it would return `1` for the case `-0.0` (ikr). In triton they have this implementation, probably taken from some STL:
https://github.com/openai/triton/blob/52d311f30208605add1c65695d071196319de76b/include/triton/external/half.hpp#L430-L442
which implementation should we have for CUDA?

torchinductor/codegen/triton.py
```diff
@@ -162,6 +162,14 @@ def randn(seed, offset, _):  # _ here to keep the contract identical to CPU rand
     def rsqrt(x):
         return f"tl.libdevice.rsqrt({x})"
 
+    @staticmethod
+    def signbit(x):
+        return f"{x} < 0"
```
ngimel, line 167: isn't there a `libdevice.signbit`? Oh ok there is but separate for floating and double types, we can keep this.

ngimel (approved):

torchinductor/codegen/triton.py
jansel: Perhaps
```suggestion
        return f""tl.libdevice.signbitd({x}) if {x}.dtype is tl.float64 else tl.libdevice.signbitf({x})"
```

Lezcano: Where is there  alist ofr all the functions in `libdevice`? Do I have to `dir()` it or what's the best way to inspect it?
Also, I'll use Jansel's proposed pattern + a `x < 0` fallback for signed integers.

Lezcano:
The error seems unrelated

torchinductor/codegen/triton.py
jansel: I look at [this file](https://github.com/openai/triton/blob/master/python/triton/language/libdevice.py) to see the list of functions.
```
tl.libdevice.signbitf({x}) if {x}.dtype is tl.float32 else {x} < 0
```
Might be simpler and support ints too.  float32 is the main thing we care about inside a kernel.

Lezcano: That one would be wrong for floating points of size other than 32 and the value `-0.0` but I guess that's alright. Otherwise, if we wanted to go with the definition from cppref, we should probably do
```python
return f"tl.libdevice.signbitf({x}) if {x}.dtype is tl.float32 else tl.libdevice.signbitd(to_dtype({x}, torch.float64))"
```

Lezcano:
@ngimel @jansel these two PRs should be ready. I don't know what needs doing with the first PR of the stack, but I don't seem to be able to re-open it. Should I open a new one?

Lezcano:
Ok, this PR had the same issue as described here https://github.com/pytorch/torchdynamo/pull/1120#issuecomment-1238447916. Please tell me if I need to do anything on my side to fix this.

torchinductor/codegen/triton.py
jansel: float16/bfloat32 are up cast to float32 inside the kernel, so the only cases are float32, float64, and the integers.

Issue #1119, opened 2022-09-02
regression for fp16


soumith:
HuggingFace models regressed from 1.95x -> 1.77x without change in coverage (36/43 in both cases).

2 days ago:
```
73%, 35/48  | 84%, 36/43  | 80%, 53/66  |
 1.86x    |    1.95x    |    1.65x   
```
-> 

last night:

```
inductor_cudagraphs | 77%, 37/48 | 84%, 36/43  | 97%, 64/66
inductor_cudagraphs |   1.82x    |    1.77x    |    1.59x 
```

Right now i have to manually spot these things.
It would be a great visual aid to plot a line graph over time.

voznesenskym:
> It would be a great visual aid to plot a line graph over time.

We could persist historical or last run results in the repo @anijain2305 - and produce a chart / delta column respectively. Thoughts?

jansel:
`git bisect` points to #1096

I think this disabled functionalization for inductor.  

cc @Chillee

anijain2305:
> > It would be a great visual aid to plot a line graph over time.
> 
> We could persist historical or last run results in the repo @anijain2305 - and produce a chart / delta column respectively. Thoughts?

Yes, I do have janky scripts here - https://github.com/anijain2305/torchdynamo_dashboard. I will also publish a diff from the last nightly to identify these regressions little bit easily. The logs are already persisted in shared filesystem in the AWS clusters.

jansel:
Looks fixed in https://github.com/pytorch/torchdynamo/issues/681#issuecomment-1236277236

anijain2305:
Lets keep this open as an action item for me to publish a diff.

anijain2305:
Sample here - https://github.com/pytorch/torchdynamo/issues/681#issuecomment-1238672152



![Image](https://user-images.githubusercontent.com/13822661/188745490-28f700d7-876f-45b4-b912-89aed65fa3fa.png)

Should take a couple of days to land and show up.

PR #1122, opened 2022-09-02
Re-enable functionalization for inductor

jansel:
Possible fix for regression in #1119


voznesenskym (approved):

voznesenskym:
IIRC the original PR this is fixing, #1096 was due to a circular dep.

PR #1123, opened 2022-09-03
[Prototype] [RFC] [Needs more tests] Add support for automatically enforcing guards on export, constraining or relaxing with PartiallyDynamicTensor

voznesenskym:


torchdynamo/eval_frame.py
```diff
@@ -511,6 +585,59 @@ def output(self, target, args, kwargs):
...
+                                enforce_shape = shape_and_dd[0]
+                                dynamic_dims = shape_and_dd[1]
+                            else:
+                                enforce_shape = arg_for_guard.shape
+                        else:
+                            if isinstance(arg_for_guard, PartiallyDynamicTensor):
+                                constrain = True
+                                shape_and_dd = arg_for_guard.shape_and_dynamic_dims()
+                                enforce_shape = shape_and_dd[0]
+                                dynamic_dims = shape_and_dd[1]
```
voznesenskym, line 622: Alternative -> Get this from the guard object. Still not sure which I prefer.

voznesenskym: This is how it would look plumbed into the guard infra.

Not sure that's a valuable direction yet - perhaps if we wanted to move this concept into general support, beyond `export` - https://github.com/pytorch/torchdynamo/commit/b2fa6c97f464737c3a72415ae01931df390f5f7e

torchdynamo/config.py
```diff
@@ -23,7 +23,7 @@
 # ERROR print exceptions (and what user code was being processed when it occurred)
 log_level = logging.WARNING
 # Verbose will print full stack traces on warnings and errors
-verbose = False
```
voznesenskym, line 26: will not land config changes, here for my convenience atm

torchdynamo/eval_frame.py
voznesenskym: Collapsed into this one.

torchdynamo/_guards.cpp
```diff
@@ -52,7 +61,30 @@ class TensorCheck {
       const auto &sizes = v.sizes();
       const auto &strides = v.strides();
       for (size_t i = 0; i < ndim; ++i) {
-        if (sizes_[i] != sizes[i] || strides_[i] != strides[i]) {
+        if (sizes_[i] != sizes[i]) {
+          bool is_dynamic_dim = false;
+          for (int dd = 0; dd < dynamic_dims_.size(); dd++) {
```
jansel, line 66: This is `O(n^2)`.  Since guard checking is on the hot path, we should just a `std::vector<bool> _is_dynamic(ndim, False)` so we can do checks for dynamic without a loop.

torchdynamo/_guards.cpp
```diff
@@ -52,7 +61,30 @@ class TensorCheck {
...
+            if (i == dim) {
+              is_dynamic_dim = true;
+              break;
+            }
+          }
+          if (!is_dynamic_dim) {
+            return false;
+          }
+        } else if (strides_[i] != strides[i]) {
+          return false;
```
jansel, line 77: A single dynamic inner dim will cause the stride checks to fail for the outer dims.

test/test_misc.py
```diff
@@ -2152,6 +2153,73 @@ def f3(x):
             f3(torch.ones(6))
         self.assertEqual(cnt.frame_count, 0)
 
+    @requires_static_shapes
+    def test_shape_change_recompile_permissive(self):
+        from torchdynamo.eval_frame import PartiallyDynamicTensor
```
jansel, line 2158: eval_frame.py is the wrong place for this

torchdynamo/_guards.cpp
```diff
@@ -99,11 +131,21 @@ class TensorCheck {
       const auto &strides = v.strides();
       for (size_t i = 0; i < ndim; ++i) {
         if (sizes_[i] != sizes[i]) {
-          // return fmt::format("tensor size mismatch at index {}. expected {},
-          // actual {}", i, sizes_[i], sizes[i]);
-          fail_reason << "size mismatch at index " << i << ". expected "
-                      << sizes_[i] << ", actual " << sizes[i];
-          return fail_reason.str();
+          bool is_dynamic_dim = false;
+          for (int dd = 0; dd < dynamic_dims_.size(); dd++) {
```
jansel, line 135: I am seeing a lot of duplicate code, we should refactor this to reduce duplication.

torchdynamo/config.py
```diff
@@ -62,7 +62,7 @@
 dynamic_propagation = True
 
 # Run the FX graph with FakeTensors
-fake_tensor_propagation = True
+fake_tensor_propagation = False
```
jansel, line 65: ?

torchdynamo/eval_frame.py
```diff
@@ -397,7 +398,88 @@ def guard_export_print(guards):
     return explanation, out_guards, graphs, ops_per_graph
 
 
-def export(f, *args, aten_graph=False, **kwargs):
+class PartiallyDynamicTensor(torch.Tensor):
```
jansel, line 401: I don't like the idea of using Tensor subclassing for this.  I think it makes the UX a bit awkward.

One option is we already have a similar concept in `torch.fx.tensor_type.*`.  See `test_gradual_type.py` in the pytorch repo.

That uses type annotations, like:
```
from torch.fx.tensor_type import TensorType, Dyn

def forward(self, x: TensorType((1, 2, 3, Dyn)), y: Dyn):
   ...
```

jansel (changes requested):

torchdynamo/_guards.cpp
voznesenskym: Yeah, this is just a hack to get us going, will change.

test/test_misc.py
voznesenskym: 100% - this is just a random place for the sake of the prototype/RFC - I will go to its own file.

torchdynamo/_guards.cpp
voznesenskym: Sorry, can you elaborate?

torchdynamo/_guards.cpp
voznesenskym: Thats the nature of this whole file, `check_verbose` and `check` are the same logic with marginally different return types - this is probably outside the scope of this PR.

torchdynamo/config.py
voznesenskym: Oh it lost my older comment - this is for easier prototyping atm, not going to land as is.

voznesenskym: image
:P

torchdynamo/eval_frame.py
voznesenskym: Hmm, I kind of prefer keeping them as Tensors. Why do you not like Tensor subclassing for this? What makes the UX awkward?

torchdynamo/_guards.cpp
jansel:
```
>>> torch.zeros((3,3,3,3)).stride()
(27, 9, 3, 1)
>>> torch.zeros((3,3,3,4)).stride()
(36, 12, 4, 1)
````

I only changed the size of the last dim, but the strides of all the other dims changed.

**Review by jansel — COMMENTED** [review]
Should this be an export-only thing that is ignored if we aren't doing an export?  This would simplify the PR a lot, as we wouldn't need any guards and could make this self-contained.

For an export this adds clear value.

For non-export, I'm not seeing the use case.  None of the backends will understand these annotations.

File: torchdynamo/_guards.cpp (hunk 4 above)
Comment (voznesenskym) [reply torchdynamo/_guards.cpp:77]: Ah, yea. Let me add a test and make sure I handle this.

File: torchdynamo/eval_frame.py (hunk 8 above)
Comment (jansel) [reply torchdynamo/eval_frame.py:401]: This is yet another way to declare what dims are static versus dynamic when we already have a few different other ones.

The other one that comes to mind is `torch.onnx.export(..., dynamic_axes={"x": [2,3]})`

`torch.fx.tensor_type.TensorType` brings with is shape propagation.

**Comment by voznesenskym** [comment]
> Should this be an export-only thing that is ignored if we aren't doing an export? This would simplify the PR a lot, as we wouldn't need any guards and could make this self-contained.
> 
> For an export this adds clear value.
> 
> For non-export, I'm not seeing the use case. None of the backends will understand these annotations.

I am torn on this, its a great question. I think it adds value in both, in terms of reducing or forcing recompiles - even if backends do not understand these annotations, they offer callers a way to use this internally to signal things to dynamo. That's a more controversial take, so I am totally open to setting that aside for now.

Because I expected this, I made sure the first commits in the stack are purely export-only, with another branch merged in over them to make it general.

For export only, check out this group of commits, it is the first 5 commits in this PR: https://github.com/pytorch/torchdynamo/pull/1123/files/80cc3bfc3fc43347ba2a57449dd2bf4c2fcf9f3d 

I am happy to trim this down to export only for now, and review it with that in mind, and we can shelve the general case implementation knowing its possible to bolt on later without changing any of the assumptions in the export-only part.

How does that sound?

File: torchdynamo/eval_frame.py (hunk 8 above)
Comment (voznesenskym) [reply torchdynamo/eval_frame.py:413]: The onnx way looks inferior to this (Matching names of tensors to dynamic axes in a dict feels brittle).

For the other approach, where we use`torch.fx.tensor_type.TensorType`... Changing to`TensorType(num1, num2, num3, dyn)` *in the signature* 

(A) feels less adaptable and like worse UX than taking your already existing inputs and just passing them in as composed through `PartiallyDynamicTensor.create(yourTensor, dynamic_dims=[3])`. 

(B) (If I understood it correctly) Requires changing the model, where callers of .export may often not be able to do that, or if they are, may not have a consistently uniform pattern to the dynamism they want to add (or constraints they want to introduce).

I think this actually makes for a less awkward UX. 

That being said, if you really dislike the idea of this being a tensor subclass, we can make it compositional instead.

File: torchdynamo/eval_frame.py [hunk 9]
```diff
@@ -409,7 +410,88 @@ def guard_export_print(guards):
...
+    """
+    PartiallyDynamicTensor (PDT) acts exactly as a Tensor for the sake of tracing,
+    except that the guards installed on input for it are either relaxed, or constrained,
+    depending on if torchdynamo.config.dynamic_shapes (dynamic_shapes) are enabled.
```
voznesenskym, line 417: Will move this to a proper place once discussions above are resolved.

torchdynamo/eval_frame.py
jansel: (A) The UX seems roughly the same, though this is adding yet another way to do something (annotate static/dynamic shapes) we already have solved in many other places. It is fragmentation for no real benefit.

I personally like type annotations better.  This subclass doesn't change the Tensor's behavior in eager mode, so it feels more like an annotation than a subclass.

(B) If you can't change the model, just add a wrapper function with annotations.

The huge problem both of these are going to have is propagation.  The annotations and the subclasses both only apply to the program inputs, and we don't have a good way or rules of propagating them through the program.

voznesenskym: > we already have solved in many other places.

I see that we solved it really just once, with `torch.fx.tensor_type.TensorType`. Once is enough, but perhaps not many? 

> If you can't change the model, just add a wrapper function with annotations.

what if populating the value needs to be dynamic? Does one need to come up with new annotation functions? What if which dim is dynamic is not determined at authoring time? 

Maybe this is contrived, but philosophically it feels like the constraint is on the tensor as an input, not on the definition of the model/function.

> The annotations and the subclasses both only apply to the program inputs, and we don't have a good way or rules of propagating them through the program.

That is fair, but is there value in such a propagation for our use case?

jansel:
> they offer callers a way to use this internally to signal things to dynamo. That's a more controversial take, so I am totally open to setting that aside for now.

The whole point of TorchDynamo (in non-export mode) is it should just work without users needing to hack up their models to workaround issues.  If we find ourselves needing to do stuff like this, we should try to fix the root cause causing people to need these annotations.

> I am happy to trim this down to export only for now, and review it with that in mind, and we can shelve the general case implementation knowing its possible to bolt on later without changing any of the assumptions in the export-only part.

Yes let's keep this export only.

I feel like this needs a design doc, because I don't totally understand the goals here.  I'm not seeing a larger consistent vision.

For design, there are two main approaches:

1) We could have everything start static, then users annotate the dynamic parts.

This seems to be what the PR is doing, but to do this correctly will require a lot more widespread changes.  Search the codebase for uses of `config.dynamic_shapes` and basically everywhere we check that we will need some sort of any_input_is_dynamic check.

IMO getting this correct will be difficult.

2) We could have everything start dynamic, then users annotate the static parts.

This way will be much easier to get correct, because you can flip `config.dynamic_shapes=True` then add static shapes back in `TensorVariable.specialize()`.

---

For both (1) and (2) we will have the problem of propagation.  What if you need to control the shape of something that isn't an input to the program? `y = torch.foobar(x)`: if you know `x` what is the rule for dynamic shapes on `y`?

We can look to FX gradual shape props, they have syntax like:
```
from torch.fx.annotate import annotate
...
y = annotate(y, TensorType((Dyn, 10)))
```
for annotating something in the middle of a program that isn't an input.


I personally don't like this that much, and would prefer something like:
```
assert y.shape[1] == 10
```
A lot of people write stuff like this in their models already.  I think dynamo should detect that and start shape specializing on `y.shape[1]` implicitly without any dynamo-specific annotations.  

This `assert` syntax also doesn't require the reader to learn anything new, and has meaningful behavior in eager mode.  If someone messes up the `assert` it will break both eager any dynamo, so the annotations won't become stale over time.

voznesenskym:
>
> Yes let's keep this export only.
>

SGTM - done already, pr trimmed down. 
 
> I feel like this needs a design doc, because I don't totally understand the goals here. I'm not seeing a larger consistent vision.
> 
> For design, there are two main approaches:
> 
> 1. We could have everything start static, then users annotate the dynamic parts.
> 
> This seems to be what the PR is doing, but to do this correctly will require a lot more widespread changes. Search the codebase for uses of `config.dynamic_shapes` and basically everywhere we check that we will need some sort of any_input_is_dynamic check.
> 

This PR support both - if dynamic shapes are enabled, the tensor input constraints and installs guard asserts that the dims passed in retain their size as exported. 

If dynamic shapes are not enabled, the tensor input relaxes the static shape requirement and removes asserts on those specific dims retaining their size as exported.


> For both (1) and (2) we will have the problem of propagation. What if you need to control the shape of something that isn't an input to the program? `y = torch.foobar(x)`: if you know `x` what is the rule for dynamic shapes on `y`?
> 

That is very much outside the scope of this PR and I think outside the scope of the `export` UX - the whole point is that you only get to control, through our UX, the shape of something that is an input to the program. If you want to change the output, process it again via something else, or change the original model. 

> This `assert` syntax also doesn't require the reader to learn anything new, and has meaningful behavior in eager mode. If someone messes up the `assert` it will break both eager any dynamo, so the annotations won't become stale over time.

this PR adds assert nodes to the graph if a flag is set - is that what you mean?

jansel:
```
def foo(a, b):
    c = a + b
    if c.shape[0] == 10:
       return torch.sin(c)
    return torch.cos(c)
```

If `dynamic_shapes=True`, you will get a graph break regardless of what annotations you supply.  So the user trying to get a single graph by adding annotations will be out of luck.

If `dynamic_shapes=False`, the annotations can remove the guard checks -- but we will still specialize on `c.shape[0]` and you will get a whole program graph that is incorrectly guarded.

voznesenskym:
> ```
> def foo(a, b):
>     c = a + b
>     if c.shape[0] == 10:
>        return torch.sin(c)
>     return torch.cos(c)
> ```
> 
> If `dynamic_shapes=True`, you will get a graph break regardless of what annotations you supply. So the user trying to get a single graph by adding annotations will be out of luck.
> 

This is a good example of where propagation would be useful - being able to propagate the fact that a and/or b is constrained to a shape (that then removes control flow, or asserts if the + is invalid), even though `dynamic_shapes=True` is great. This kind of propagation was outside the scope of this, but if switching to TensorType enables that, then maybe it is worth doing. 

> If `dynamic_shapes=False`, the annotations can remove the guard checks -- but we will still specialize on `c.shape[0]` and you will get a whole program graph that is incorrectly guarded.

Makes sense.

>I personally don't like this that much, and would prefer something like:
> assert y.shape[1] == 10
> A lot of people write stuff like this in their models already. I think dynamo should detect that and start shape specializing on y.shape[1] implicitly without any dynamo-specific annotations.

That is an interesting point. I do like this suggestion a lot. It doesn't really do anything for the all static->few dynamic  case, but its a great tool for the all dynamic->some static case. For this case user defined sizes on dims through asserts can be converted into shape propagations and guards. 

However, for the all static->few dynamic case, we would need annotations. Let's shelve this for a moment, for the sake of discussion.

If we focus on the single case, all dynamic->some static, and we do analysis on asserts, would it be feasible to dynamically convert/rewrite the containing funcs signature into ones that utilize the `torch.fx.tensor_type.TensorType`?

Alternatively, if we wanted to drive it with inputs how difficult would it be to get type propagation working from user provided input relaxed/constrained dims?

jansel:
> If we focus on the single case, all dynamic->some static, and we do analysis on asserts, would it be feasible to dynamically convert/rewrite the containing funcs signature into ones that utilize the `torch.fx.tensor_type.TensorType`?

Why do we need this?  If we go with this model, everything is dynamic that the user hasn't asserted it static.  The user should just look at the asserts and make sure they make sense (and run test inputs to see if they pass).

> Alternatively, if we wanted to drive it with inputs how difficult would it be to get type propagation working from user provided input relaxed/constrained dims?

Type propagation is an `O(num_operators_in_pytorch)` (1200+) problem, so super hard.  We should try to leverage other ongoing efforts in this area.

torchdynamo/eval_frame.py
silvasean: > Maybe this is contrived, but philosophically it feels like the constraint is on the tensor as an input, not on the definition of the model/function.

That is our use case in Torch-MLIR, fwiw. Although it does work well to define a wrapper function/class with the right annotations and then script/trace/td.export it. Introducing a new "tensor with partially dynamic shape data attached" does seem like an extra concept here.

> That is fair, but is there value in such a propagation for our use case?

This propagation is exactly what Torch-MLIR would like to get here. Currently we maintain our own shape propagation infrastructure, thankfully sharing @eellison's python definitions. The goal of this feature request is for torchdynamo.export to just give us the maximally precise shapes propagated through the entire program without us having to maintain our own shape propagation infrastructure. Is there a plan for this?

Although technically Torch-MLIR takes JIT IR as input, so I think that in principle even if torchdynamo gave the partially propagated shapes, the going down to JIT IR could lose them. Any thoughts on how best to get from TD FX graphs down to JIT IR while retaining partially specified shapes?

cc @Chillee

voznesenskym: > Is there a plan for this?

Plan? No.

Desire? Yes.

I spoke to @Chillee about this today - he mentioned FakeTensor is the closest thing we have here w/r/t the propagation behavior we want. 

I plan to start poking around this week at symbolic shapes and some of the work that @Chillee has already done there to see how far it can take us, and where the gaps are

jansel: @ezyang has a [symbolic shapes branch](https://github.com/pytorch/pytorch/pull/84246) that would allow symbolic shapes to propagate through the dispatcher using FakeTensor.

It is not ready for use yet, but at some point I'd hope it would give us shape propagation.

Issue #1126, opened 2022-09-05
NameError - torch is not defined - speech_transformer


anijain2305:
Repro -  `TORCHDYNAMO_DEBUG_FUNCTION=forward python benchmarks/torchbench.py --training --float32 --backend=eager --use-eval-mode --only=speech_transformer`

And a diff

~~~
diff --git a/torchdynamo/convert_frame.py b/torchdynamo/convert_frame.py
index 7fa490d..c3af426 100644
--- a/torchdynamo/convert_frame.py
+++ b/torchdynamo/convert_frame.py
@@ -232,6 +232,9 @@ def convert_frame_assert(compiler_fn: Callable, guard_export_fn=None, one_graph=
             and os.environ.get("TORCHDYNAMO_DEBUG_FUNCTION") != code.co_name
         ):
             return None
+
+        if "encoder" not in code.co_filename:
+            return None
         if code.co_name == "<genexpr>" and code.co_filename.endswith(
             ("transformers/file_utils.py", "transformers/utils/generic.py")
         ):
~~~

This one leads to error like 
~~~
NameError: name 'torch' is not defined
ERROR
name 'torch' is not defined
~~~

I do not fully understand how globals are handled during inlining. The file containing the parent function has import `import torch.nn as nn` (no `import torch`), but the child function file has `import torch`. It seems that children imports are not passed on the parent function while inlining.

cc @voznesenskym @jansel

voznesenskym:
Repro confirmed. I have a theory here, looking into this.

voznesenskym:
Yeah this needs instruction translation for imports, I think.

Just (hilariously) jamming in 
```
hack_output.append(create_instruction("LOAD_CONST", 2, 0))
hack_output.append(create_instruction("LOAD_CONST", 0, None))
hack_output.append(create_instruction("IMPORT_NAME",  0, "torch"))
hack_output.append(create_instruction("STORE_FAST", 1, "torch"))
instructions = hack_output + instructions
```

Gets me past this import error, onto another one:

```
ModuleNotFoundError: No module named 'torchbenchmark.models.speech_transformer.speech_transformer.transformer.get_non_pad_mask'
```

Obviously not a real fix - but demonstrative of the issue.

There's also another problem here - when we inline code the stack trace is wrong.

anijain2305:
I somehow jammed it in using https://github.com/pytorch/torchdynamo/pull/1128. But, idk.

PR #1128, opened 2022-09-06
Install missing global

anijain2305:
Related to https://github.com/pytorch/torchdynamo/issues/1126

Not really sure if this is the right fix. My understanding is the `output.root_globals` is the final `f_globals` passed on to the modified frame. During inlining, we need to ensure that `globals` from the inlined functions/files are also reflected in the `output.root_globals`. This fixes the speech_transformer issue.

@voznesenskym @jansel

### DIFF
--- a/torchdynamo/symbolic_convert.py
+++ b/torchdynamo/symbolic_convert.py
@@ -399,6 +399,8 @@ class InstructionTranslatorBase(object):
                 source = AttrSource(
                     self.import_source(self.f_globals["__name__"]), name
                 )
+                if name not in self.output.root_globals and name in self.f_globals:
+                    self.output.install_global(name, self.f_globals[name])
             else:
                 mangled_name = f"___unnamed_scope_{id(self.f_globals)}"
                 if mangled_name not in self.output.root_globals:


torchdynamo/symbolic_convert.py
```diff
@@ -399,6 +399,8 @@ def get_global_source(self, name):
                 source = AttrSource(
                     self.import_source(self.f_globals["__name__"]), name
                 )
+                if name not in self.output.root_globals and name in self.f_globals:
+                    self.output.install_global(name, self.f_globals[name])
```
voznesenskym, line 403: is this safe for all globals? Should we check if the value is a module?

voznesenskym (approved):

jansel:
Hrm... this seems like a workaround not fixing the real issue.  I think we could have incorrect results if `root_globals[name]` and `f_globals[name]` point to different objects.  Which is possible...

Where is the `NameError` coming from?

anijain2305:
@jansel The repro is described here - https://github.com/pytorch/torchdynamo/issues/1126


The forward is in this file - https://github.com/pytorch/benchmark/blob/main/torchbenchmark/models/speech_transformer/speech_transformer/transformer/encoder.py

If you look at the imports at the top, it does not have `import torch`.

The analysis of `forward` inlines functions in these file - https://github.com/pytorch/benchmark/blob/main/torchbenchmark/models/speech_transformer/speech_transformer/transformer/attention.py

The NameError is coming after the inlining is done at runtime.

jansel (changes requested):
I think I found the root cause for this, let me confirm.

jansel:
Root cause fix is here: #1129

torchdynamo/symbolic_convert.py
```diff
@@ -399,6 +399,8 @@ def get_global_source(self, name):
                 source = AttrSource(
                     self.import_source(self.f_globals["__name__"]), name
                 )
+                if name not in self.output.root_globals and name in self.f_globals:
```
jansel, line 402: Funny enough, I think this fix only worked because some completely unrelated code we inline used the global `torch`.  The reference to `torch` was coming from unspecialized integer/float graph inputs and had nothing to do with user code or inlining.

If you modify the inlined function to not use `torch.*` this fix will stop working.

See #1129

anijain2305:
Closed in favor of #1129

PR #1129, opened 2022-09-06
Fix root cause of `NameError - torch is not defined`

### LINKED ISSUE #1126: NameError - torch is not defined - speech_transformer

jansel:
Fixes #1126 
Replaces #1128 

I found this by looking at the output bytecode and seeing:
```
              6 LOAD_GLOBAL             12 (torch)
              8 LOAD_ATTR               13 (tensor)
```

Then running
```
find -name '*.py' | xargs grep -i global | grep '"torch"'
```

### DIFF
--- a/torchdynamo/codegen.py
+++ b/torchdynamo/codegen.py
@@ -315,6 +315,19 @@ class PyCodegen(object):
         output.extend(self.rot_n(num_on_stack + 1))
         self.clear_tos()
 
+    def create_load_python_module(self, mod):
+        """
+        Generate a LOAD_GLOBAL instruction to fetch a given python module.
+        """
+        root_globals = self.tx.output.root_globals
+        name = re.sub(r"^.*[.]", "", mod.__name__)
+        if root_globals.get(name, None) is mod:
+            return self.create_load_global(name, add=True)
+        mangled_name = f"___module_{name}_{id(mod)}"
+        if mangled_name not in root_globals:
+            self.tx.output.install_global(mangled_name, mod)
+        return self.create_load_global(mangled_name, add=True)
+
     def make_call_generated_code(self, fn_name: str) -> List[Instruction]:
         """Call the generated code function stored in fn_name"""
         self.extend_output(self.load_function_name(fn_name))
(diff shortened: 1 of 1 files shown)


torchdynamo/codegen.py
```diff
@@ -324,7 +337,7 @@ def make_call_generated_code(self, fn_name: str) -> List[Instruction]:
             if arg.is_unspecialized:
                 self.extend_output(
                     [
-                        self.create_load_global("torch", add=True),
+                        self.create_load_python_module(torch),
```
voznesenskym, line 340: Since we are here, do we need a `create_load_python_module` for every arg in graphargs? Or can we do it once before we iterate the graphargs?

voznesenskym (approved):

torchdynamo/codegen.py
jansel: For user code, we have a `Source()` that knows how to load the original copy (no need to add a new variable).

The issue here is the `LOAD_GLOBAL torch` was coming from code torchdynamo generated, not the user's code -- and we didn't check to make sure it is existed.

anijain2305 (approved):
Nice! This was a subtle one.

PR #1140, opened 2022-09-06
Use only reduction deps to determine if reduction is inner

ngimel:
This doesn't change the benchmarking numbers, but adds a few more kernels to "no autotune" list
Also, preserves unique kernel names to make it easier to inspect cuda profiles (before all kernels were called "kernel")

### DIFF
--- a/torchinductor/codegen/triton.py
+++ b/torchinductor/codegen/triton.py
@@ -1116,7 +1116,6 @@ class TritonScheduling:
     def is_contiguous(node):
         if node in (EnableReduction, DisableReduction):
             return True
-        # print("Node readwrites", node.read_writes.reads, node.read_writes.writes, node.read_writes.index_exprs, node.read_writes.range_vars)
         if node.is_reduction():
             return all(
                 dep.is_contiguous()
--- a/torchinductor/codegen/triton.py
+++ b/torchinductor/codegen/triton.py
@@ -919,7 +919,7 @@ class TritonKernel(Kernel):
             if tree.prefix != "r" or self.inside_reduction:
                 argdefs.append(f"{tree.prefix.upper()}BLOCK : tl.constexpr")
 
-        code.writeline(f"def {name or 'kernel'}({', '.join(argdefs)}):")
+        code.writeline(f"def {name or '{kernel_name}'}({', '.join(argdefs)}):")
         self.codegen_body()
         with code.indent():
             for old, new in self.args.aliases():
@@ -932,7 +932,7 @@ class TritonKernel(Kernel):
         wrapper = IndentedBuffer()
         wrapper.writeline("TritonCodeCache.load('''")
         wrapper.splice(code.getvalue(), strip=True)
-        wrapper.writeline("''').kernel")
+        wrapper.writeline("''').{kernel_name}")
         return wrapper.getvalue()
 
     def reshape_size_str(self, i=None, x=None):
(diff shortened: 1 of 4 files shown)


torchinductor/codegen/triton.py
```diff
@@ -1105,14 +1116,18 @@ def end_current_reduction_loop():
...
-        )
+        if node.is_reduction():
+            return all(
+                dep.is_contiguous()
+                for dep in itertools.chain(
+                    node.read_writes.reads, node.read_writes.writes
+                )
+            )
+        else:
+            return True
```
jansel, line 1127: Won't this give the wrong hint if we start passing hints to tiled pointwise?

torchinductor/codegen/triton.py
```diff
@@ -919,7 +930,7 @@ def codegen_kernel(self, name=None):
             if tree.prefix != "r" or self.inside_reduction:
                 argdefs.append(f"{tree.prefix.upper()}BLOCK : tl.constexpr")
 
-        code.writeline(f"def {name or 'kernel'}({', '.join(argdefs)}):")
+        code.writeline(f"def {name or '{kernel_name}'}({', '.join(argdefs)}):")
```
jansel, line 933: The reason I named this just `kernel` is so that the same kernel generated by different graphs (or different models entirely) would share the same cache on disk. 
 
Perhaps we could hash the source code to make naming deterministic?

jansel (changes requested):

torchinductor/codegen/triton.py
ngimel: It would, but currently pointwise heuristics doesn't accept any arguments at all. We can name this function differently.

torchinductor/codegen/triton.py
ngimel: Yeah that would work too. I wouldn't expect too many hits from different models or different graphs though

Chillee: @jansel Oh is that why we're generating a bunch of code strings for different kernels now?

ngimel: Bunch of code strings is for caching autotuning results per kernel

ngimel: Kernel numbers still make navigating profiles easier (like, we easily know how many kernels are generated, seeing that kernel10 is called many times is easier than comparing 6-digit hashes etc)

jansel: Yes, every kernel gets its own file based on the hash of its source code.  So if two models generate the same kernel, it will be shared.

A lot of models have repeating patterns, especially when they come from the same benchmark suite they might literally have different instances of the same `nn.Module` as layers inside them.

I haven't tested how many hits we are getting, though hopefully dynamic shapes will lead to more hits here.

jansel: > Kernel numbers still make navigating profiles easier (like, we easily know how many kernels are generated, seeing that kernel10 is called many times is easier than comparing 6-digit hashes etc)

Perhaps we could put this behind a debug config?  I worry it could hurt compile times. 

Is there a way to label a kernel after it is generated?

ngimel: I could check the number of kernels generated in torchbench with and without kernel numbers, and yes, we could put this behind debug config. Even in this PR we are labeling a kernel after its generated, it's generated just with `{kernel_name}` in the string, we could replace it either by `kernel` or by `kernelXX`.

jansel: Perhpas

jansel: HuggingFace might have more kernel sharing. We have stuff like: `DebertaForMaskedLM`, `DebertaForQuestionAnswering`, `DebertaV2ForMaskedLM`, `DebertaV2ForQuestionAnswering`

There are also cases where we graph break in a loop and have multiple graphs from the same code.

ngimel: Changed to be controlled by config.

jansel (approved):

PR #1141, opened 2022-09-07
Graph break on functions in torch.utils._pytree.*

yanboliang:
Fix #371

### DIFF
--- a/test/test_misc.py
+++ b/test/test_misc.py
@@ -2076,6 +2076,19 @@ class MiscTests(torchdynamo.testing.TestCase):
         self.assertTrue(same(b_grad, torch.tensor([0.0, 0.0])))
         self.assertEqual(cnts.frame_count, 2)
 
+    def test_torch_utils_pytree(self):
+        def fn(x):
+            torch.utils._pytree.tree_flatten((x, x))
+            return x + 1
+
+        x = torch.randn(3)
+        ref = fn(x)
+        cnts = torchdynamo.testing.CompileCounter()
+        opt_fn = torchdynamo.optimize(cnts)(fn)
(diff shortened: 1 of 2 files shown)


voznesenskym:
Would it be possible to support this instead?

Chillee:
I agree, I'd prefer to not graph-break on these operations.

yanboliang:
I didn't see this is a very common use case, a graph break is just to avoid confusing error. If we find this fall back breaks too many large graphs in the future(I doubt), we can add support then.

yanboliang:
> I agree, I'd prefer to not graph-break on these operations.

Oh, is there a lot of places using these functions? I don't have too many context.

voznesenskym:
> I didn't see this is a very common use case, a graph break is just to avoid confusing error. If we find this fall back breaks too many large graphs in the future(I doubt), we can add support then.

I'd rather we error and fix it - I have an old branch somewhere locally I can polish off to give this a pass, do you mind if I do? or would you like to?

yanboliang:
> I'd rather we error and fix it - I have an old branch somewhere locally I can polish off to give this a pass, do you mind if I do? or would you like to?

I can do this, it's not a big issue. Thanks.

voznesenskym:
Discussed offline - leaving here for posterity - we need to probably support (A) A TreeSpecVariable being produced from our builder (B) the list in and out of pytree can be anything, so we need to make sure that when we access it, we are correctly tracking the objects inside with their appropriate VariableTracker objects.

yanboliang:
I'll close this one now and send a new one to implement this feature soon. Thanks for all your suggestion!

PR #1142, opened 2022-09-07
Tracking compilation latency in nightly

anijain2305:
Tracking the mean of worst 50 compilation latency in nightly. Sample output here - https://github.com/anijain2305/ghissue/issues/1#issuecomment-1238928755

cc @wconstab 

Will mentioned that we can use `dynamo_timed` results and can track more detailed breakdown of different components. 

This PR, however, puts timers at the highest level and just measures the total time for first 3 iterations. Pro is that this measurement reflects the compilation time observed by the user. However, con is that it lacks the details that `dynamo_timed` provides.

Bonus is that - we can extend this for measuring user-observed peak memory footprint.

### DIFF
--- a/benchmarks/runner.py
+++ b/benchmarks/runner.py
@@ -28,6 +28,7 @@ import importlib
 import io
 import itertools
 import os
+from collections import defaultdict
 from os.path import abspath
 from os.path import exists
 
@@ -35,6 +36,7 @@ import matplotlib.pyplot as plt
 import pandas as pd
 import torch
 from matplotlib import rcParams
+from numpy.core.fromnumeric import mean
 from scipy.stats import gmean
 from tabulate import tabulate
 
@@ -101,6 +103,11 @@ DEFAULTS = {
     "devices": [
         "cuda",
     ],
+    "quick": {
+        "torchbench": ["resnet18", "resnet50"],
+        "huggingface": ["AlbertForMaskedLM", "BertForMaskedLM"],
+        "timm_models": ["resnest101e", "mobilenetv2_100"],
+    },
 }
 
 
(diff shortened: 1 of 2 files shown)


anijain2305:
> well, tracking the coarse-grain time is more useful than nothing. Idk what your tradeoffs were, is this just easier to implement or is it more that the metrics are easier to look at on a dashboard?

It's both. But, it's mostly the second one. The metrics are easier to look and cover other backends as well. 

I realized that my repo is private, these are the snapshots

image

image

wconstab:
compiler=eager means dynamo+eager right? so it is at least meaningful in terms of graph acquisition time?

the first table (mean compilation time of worst 50) looks good.  The second table (compilation latency) just has resnet- is this a WIP or did you think showing the times for just a couple of specifically chosen models makes sense?  I think it could be useful to choose some models that have particularly bad compilation times and then highlight those week to week, but i'm not sure which models.

anijain2305:
> compiler=eager means dynamo+eager right? so it is at least meaningful in terms of graph acquisition time?
Thats correct. eager means dynamo+eager, and `aot_eager` is dynamo + aot + eager. So, it should give us good enough idea.

 
> the first table (mean compilation time of worst 50) looks good. The second table (compilation latency) just has resnet- is this a WIP or did you think showing the times for just a couple of specifically chosen models makes sense?

Oh I am just showing for a couple of models. Running full suite takes around 16 hours, so I use a `quick` option to flush any issues. The models here are chosen arbitrarily.

> I think it could be useful to choose some models that have particularly bad compilation times and then highlight those week to week, but i'm not sure which models.

So, the resulting table will be sorted based on the inductor_cudagraphs compilation time and cut off at 50. So, we will automatically choose the worst models.

benchmarks/runner.py
```diff
@@ -101,6 +103,11 @@
     "devices": [
         "cuda",
     ],
+    "quick": {
```
wconstab, line 106: i think it makes sense to offer this 'quick' thing.  but it doesn't necessarily make sense to pick 2 resnets here.  For example, why not pick the handful of models known to have either high peak mem or compile times?

wconstab (approved):
thanks @anijain2305 looks good!

benchmarks/runner.py
anijain2305: I actually never intended for actual run. This was just to speedup development process, so that we can quickly run the runner and look at graphs etc. I dont want to add more models here.

I think we can add another flag, that does what you say. There, we could have handful of models (and not just 2), which represents the broader set, and can finish in an hour or so.

PR #1143, opened 2022-09-07
Install aten asserts on inputs in export

voznesenskym:
Based on https://github.com/pytorch/pytorch/pull/84617


torchdynamo/eval_frame.py
```diff
@@ -524,6 +555,66 @@ def output(self, target, args, kwargs):
...
+                            )
+                            dtype = (
+                                None
+                                if torchdynamo.config.dynamic_shapes
+                                else arg_for_guard.dtype
+                            )
+
+                            equals = new_graph.graph.create_node(
+                                "call_function",
+                                torch._tensor_equal,
```
jansel, line 595: Should we just make this one function call rather than two?  This seems too verbose.

torchdynamo/eval_frame.py
```diff
@@ -524,6 +555,66 @@ def output(self, target, args, kwargs):
...
+
+                        with new_graph.graph.inserting_after(placeholders[-1]):
+                            shape = (
+                                None
+                                if torchdynamo.config.dynamic_shapes
+                                else arg_for_guard.shape
+                            )
+                            strides = (
+                                None
+                                if torchdynamo.config.dynamic_shapes
```
jansel, line 584: If `dynamic_shapes` is true doesn't `guard_args` just do nothing?

torchdynamo/eval_frame.py
```diff
@@ -524,6 +555,66 @@ def output(self, target, args, kwargs):
...
     ).transform()
 
+    if guard_args:
+        id_to_arg_mapping = dict()
+        for i in range(len(args)):
+            a = args[i]
+            id_to_arg_mapping[id(a)] = (i, a)
+        placeholders = [n for n in new_graph.graph.nodes if n.op == "placeholder"]
+        if len(placeholders) > 0:
+            for guard in out_guards:
```
jansel, line 565: Why not just guard the shapes of all inputs (one for each placeholder)?  When would this matching logic result in a different outcome?

torchdynamo/eval_frame.py
voznesenskym: Yeah, probably, although I like the idea of being able to later use _tensor_equal in something like cond, which is definitively not an assert. I'll make this _assert_tensor_metadata, and we can always refactor if we need finer grained parts later.

torchdynamo/eval_frame.py
voznesenskym: Not all inputs are tensors, and this is actually a bit of a setup for symbolic shapes, in that we use the guard. Right now we just use it for id - but in the future, I was hoping to have the guard be the source of truth for the sym shape info.

torchdynamo/eval_frame.py
voznesenskym: The pydoc is correct, my impl is wrong ;)

> If torchdynamo is in dynamic_shapes mode, we assert that the dtypes of (tensor) input values match subsequent invocations.

dtype should always be checked. Good catch.

voznesenskym:
Need to revisit this, sorry it fell into a backlog

jansel:

PR #1144, opened 2022-09-07
Cudagraph always on and cudagraphs fallback flags

voznesenskym:


torchinductor/config.py
```diff
@@ -49,6 +49,12 @@
 # replace small reductions with pointwise, disable with `= 1`
 unroll_reductions_threshold = 8
 
+# Cudagraphify will run even if we have graph mutations
+always_attempt_cudagraphify = True
```
anijain2305, line 53: Just for my education? Why have this flag? Is this to prefetch other problems?

voznesenskym: There are cases in prod models where we over-enthusiastically decide that a graph was mutated and then abort attempting cudagraphify. This just lets it always have a go.

voznesenskym:
This may not work - if it causes a segfault - still iterating, but up for feedback

torchinductor/config.py
```diff
@@ -49,6 +49,13 @@
 # replace small reductions with pointwise, disable with `= 1`
 unroll_reductions_threshold = 8
 
+# Cudagraphify will run even if we have graph mutations, may segfault
+always_attempt_cudagraphify = False
+
+# Cudagraphify failures will not stop compilation, but will not produce cudagraphified code
+# Will not proceed past segfaults
+fallback_cudagraphify = True
```
voznesenskym, line 57: Maybe default False for debugging?

torchinductor/config.py
```diff
@@ -49,6 +49,13 @@
 # replace small reductions with pointwise, disable with `= 1`
 unroll_reductions_threshold = 8
 
+# Cudagraphify will run even if we have graph mutations, may segfault
```
jansel, line 52: I believe this will be silently incorrect, not segfault.

When would the user want this behavior?

voznesenskym: @anijain2305 asked the same thing above.

There are cases in prod models where we think we are over-enthusiastically deciding that a graph was mutated and then abort attempting cudagraphify. This just lets it always have a go. 

However, we are using fail/no fail as the indicator of success - not any correctness testing. Is there a way to check that what was produced is correct?

jansel: The graph will return the correct answer, but the mutations to the inputs will be lost.

I might call this `force_cudagraphs` or something similar -- then update the comment to say it discards input mutations.

voznesenskym: Sounds good, it should feel relatively unsafe to use, and force fits the idea.

torchinductor/compile_fx.py
```diff
@@ -60,15 +60,28 @@ def compile_fx_inner(
...
-    ):
-        compiled_fn = cudagraphify(
-            compiled_fn, example_inputs, static_input_idxs=range(num_fixed)
-        )
+    )
+
+    # TODO(voz): Add a config driven option to install a segfault handler for the sake of reporting.
+    run_cudagraphs = safe_cudagraphs or config.force_cudagraphs
+    if config.force_cudagraphs and not safe_cudagraphs:
+        log.warning("Cudagraph failure heuristics bypassed, may segfault.")
```
jansel, line 72:
```suggestion
        log.warning("Cudagraph failure heuristics bypassed, may segfault or not perform input mutation")
```

torchinductor/compile_fx.py
```diff
@@ -60,15 +60,28 @@ def compile_fx_inner(
...
+    if cudagraphs and run_cudagraphs:
+        try:
+            cudagraph_fn = cudagraphify(
+                compiled_fn, example_inputs, static_input_idxs=range(num_fixed)
+            )
+        except Exception as e:
+            if config.fallback_cudagraphify:
+                log.warning("Cudagraph failed, but running with fallback enabled.")
```
jansel, line 80: If we continue we need to disable cudagraphs in the backwards with `BoxedBool.disable(cudagraphs)`

torchinductor/compile_fx.py
```diff
@@ -60,15 +60,28 @@ def compile_fx_inner(
...
+            else:
+                raise e
+
+        compiled_fn = cudagraph_fn
```
jansel, line 84: I don't think this is needed, we can just set it directly on L75

jansel:

PR #1151, opened 2022-09-07
Export aten graph with stack trace

SherlockNoMad:
torchdynamo.export with aten_graph=True now would also have stack_trace populated.

### DIFF
--- a/test/test_export.py
+++ b/test/test_export.py
@@ -865,3 +865,30 @@ class ExportTests(torchdynamo.testing.TestCase):
         dynamo_result = out_graph(*flat_input)
 
         self.assertTrue(torchdynamo.utils.same(real_result, dynamo_result))
+
+    def test_export_with_stack_trace(self):
+        inp = torch.tensor([0.1, 0.1])
+        linear = torch.nn.Linear(2, 2)
+
+        def func(x):
+            x = x + 1
+            y = x.t()
+            y = y.relu()
+            y = linear(y)
+            return y
+
+        exported = torchdynamo.export(func, inp, aten_graph=False)
+        out_graph = exported[0]
+
+        print(out_graph)
+        for node in out_graph.graph.nodes:
+            if node.op not in {"placeholder", "output"}:
+                self.assertTrue(node.stack_trace is not None)
+
+        torchdynamo.reset()
+
+        exported = torchdynamo.export(func, inp, aten_graph=True)
+        out_graph = exported[0]
+        for node in out_graph.graph.nodes:
+            if node.op == "call_function":
+                self.assertTrue(node.stack_trace is not None)
--- a/test/test_export.py
+++ b/test/test_export.py
@@ -2,8 +2,8 @@
 from unittest.mock import patch
 
 import torch
-from torch.fx.experimental.proxy_tensor import make_fx
 import torch.utils._pytree as pytree
+from torch.fx.experimental.proxy_tensor import make_fx
 
 import torchdynamo.testing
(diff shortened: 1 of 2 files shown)


test/test_export.py
```diff
@@ -864,3 +864,30 @@ def func(a, b, c):
...
+            x = x + 1
+            y = x.t()
+            y = y.relu()
+            y = linear(y)
+            return y
+
+        exported = torchdynamo.export(func, inp, aten_graph=False)
+        out_graph = exported[0]
+
+        print(out_graph)
```
qihqi, line 882: remove?

qihqi (approved):
thanks!

voznesenskym (changes requested):
Can you add a test to ensure this does not change the graph, and only change the metadata?

torchdynamo/eval_frame.py
```diff
@@ -505,7 +505,12 @@ def output(self, target, args, kwargs):
             return super().output(target, (new_result,), {})
 
     if aten_graph:
-        graph = make_fx(graph)(*graph_captured_input)
+        # Running graph with interpreter is needed for propagating the stack_trace
+        def graph_with_interpreter(*args):
+            with torch.fx.traceback.override_stack_trace():
+                return torch.fx.Interpreter(graph).run(*args)
+
+        graph = make_fx(graph_with_interpreter)(*graph_captured_input)
```
voznesenskym, line 513: A critical contract of .export() is that the graph is identical to @torchdynamo.optimize(), with fx enabled, it needs to be identical to taking the @torchdynamo.optimize() + make_fx. If `graph_with_interpreter` changes the graph in any way other than setting metadata, it will no longer be equivalent.

SherlockNoMad: Added a test for this.

test/test_export.py
```diff
@@ -864,3 +865,59 @@ def func(a, b, c):
...
+
+        torchdynamo.reset()
+
+        def compiler(gm, sample_inputs):
+            aten_gm = make_fx(gm)(*sample_inputs)
+
+            self.assertEqual(len(aten_gm.graph.nodes), len(out_graph.graph.nodes))
+            for node1, node2 in zip(aten_gm.graph.nodes, out_graph.graph.nodes):
+                self.assertEqual(node1.op, node2.op)
+                if node1.op == "call_function":
```
voznesenskym, line 917: Hmm what about the other nodes? Entire graph?

SherlockNoMad: Other nodes might have difference in names. 
For example, one graph would have input named arg0, another graph could have arg1_0... 
I think checking the target for call_function nodes should be sufficient...

voznesenskym: Hmm, can we at least ensure that the arg signature (number and type of args) and the output spec match across the graphs?

qihqi: Wouldn't evaluating the 2 graphs and asserting of getting same result be good enough?

torchdynamo/eval_frame.py
qihqi: Just curious, any callsites that depend on this `export == @torchdynamo.optimize() + make_fx. `? In my mental model export should return equivalent to `exir.trace` as that is what it tries to replace right now.

test/test_export.py
voznesenskym: Awesome, thank you, and apologies for the insistence.

voznesenskym (approved):

PR #1152, opened 2022-09-07
Remove enable_torch_dispatch for mode rewrite

samdow:
One big change for after the mode stack PR is done (being redone as one PR, will link when that's done)

### DIFF
--- a/torchdynamo/optimizations/analysis.py
+++ b/torchdynamo/optimizations/analysis.py
@@ -121,7 +121,7 @@ def has_mutation(gm, example_inputs, inputs_only=False):
         fake_wrapper = functools.partial(wrap_to_fake_tensor, fake_mode=fake_mode)
         example_inputs = tree_map(fake_wrapper, example_inputs)
         new_gm = deepcopy_to_fake_tensor(gm, fake_mode)
-        with fake_mode.restore():
+        with fake_mode:
             ShapeAliasingAndMutationProp(new_gm).run(*example_inputs)
     else:
         new_gm = copy.deepcopy(gm)
--- a/torchdynamo/optimizations/analysis.py
+++ b/torchdynamo/optimizations/analysis.py
@@ -121,7 +121,7 @@ def has_mutation(gm, example_inputs, inputs_only=False):
         fake_wrapper = functools.partial(wrap_to_fake_tensor, fake_mode=fake_mode)
         example_inputs = tree_map(fake_wrapper, example_inputs)
         new_gm = deepcopy_to_fake_tensor(gm, fake_mode)
-        with fake_mode:
+        with fake_mode.restore() if hasattr(fake_mode, "restore") else fake_mode:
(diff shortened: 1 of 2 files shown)


ezyang (approved):
But will this run into the "you haven't setup ancestors" problem yet? You might have a better time landing everything in one go. Happy to try doing it incrementally first though

ezyang:
> Sorry I'm misunderstanding this--is the idea that we squash all the pytorch/pytorch mode PRs together, merge that, then break dynamo for a short minute while we update this code?

It's technically not breaking, since there are pins. But yeah, we would push the PRs together and update the pins simultaneously.

> @zou3519 and I were also talking that as an alternate, we could (1) change the mode stack and leave enable_torch_dispatch_mode for a moment (though it wouldn't compose correctly) and then (2) change this to use the new mode stack and (3) remove enable torch dispatch. Downside here is that enable_torch_dispatch_mode would be left for a moment even though it doesn't make sense with the new mode stack, but we might be able to avoid breaking dynamo

I don't really mind how you skin the cat, so if you want to do this instead, you can; I'm just suggesting the "one big land" because I think it is the least work in the end.

samdow:
> It's technically not breaking, since there are pins. But yeah, we would push the PRs together and update the pins simultaneously.

Ah sorry I explained this poorly. So the current dynamo pin has `enable_torch_dispatch_mode` in it. With the change on the pytorch/pytorch side, we remove `enable_torch_dispatch` so the dynamo test on pytorch fail and it can't land. However, we also can't update dynamo without that change because we're either doing restore (which will go away after that PR and break, meaning we need to update the pytorch pins twice and dynamo pins twice) or going to get an error trying to do `with fake_mode` since we haven't updated how the mode stack is done yet

I'm not sure I get updating the pins simultaneously. Is there a way to update both dynamo and pytorch in the same PR? Otherwise won't there be a land race between the repos?

ezyang:
so the way I do it (based on xla), is I push a branch to dynamo repo that is not master, and then pin to that. So temporarily, the pin doesn't match master. But then once a sufficiently new nightly is built, then you can merge the branch to main on dynamo + nightly pin update. Then you're done.

ezyang:
@samdow I'm waiting on changes to the stack representation, tag me when that's done

torchdynamo/optimizations/analysis.py
```diff
@@ -121,7 +121,7 @@ def has_mutation(gm, example_inputs, inputs_only=False):
         fake_wrapper = functools.partial(wrap_to_fake_tensor, fake_mode=fake_mode)
         example_inputs = tree_map(fake_wrapper, example_inputs)
         new_gm = deepcopy_to_fake_tensor(gm, fake_mode)
-        with fake_mode.restore():
+        with fake_mode:
```
eellison, line 124: you might consider something like `if hasattr(fake_mode, "restore")` ... and similarly below with `torch_dispatch` to coordinate with landing the other stack

Issue #1138, opened 2022-09-06
Refactor Capture Scalar Outputs to use SymInts/SymFloats


eellison:
Currently, in order to support tracking`.item()` calls, `capture_scalar_outputs ` will wrap the output of `.item()` with a 0-element tensor. We should refactor this to work with SymInts/SymFloats for a few reasons:

- fake tensor / proxy tensors will be doing their own internal constant tracking which can distinguish between calls of .item() which are statically known and those that are not, and the torchdynamo `.item()` handling interferes with this 

- this aligns with the other work ongoing to trace out integers/floats operations in the dynamic work stream 

- the current wrapping of scalars to tensors only defers the `.item()` call into the dispatcher. For instance, in adagrad, `param.addcdiv_(grad, std, value=-clr)`, `-clr` in user code is a float but because of `capture_scalar_outputs` in dynamo tracing is a 0-element tensor. when `addcdiv` is invoked in the dispatcher, the `-clr` will get converted to a scalar just as `.item()` call does. See [this issue](https://github.com/pytorch/pytorch/issues/84597) for more context. With [this pr](https://github.com/pytorch/pytorch/pull/83567), this will trigger a graph break, so the current `.item()` handling is insufficient. 
- There might be subtle semantics mismatches in the current handling, e.g., right now we are truncating python floats from double precision to floats.

thiagocrepaldi:
@eellison where can I learn more about "this aligns with the other work ongoing to trace out integers/floats operations in the dynamic work stream? I am working with PyTorch's ONNX converter and properly capturing scalars is a challenge for us too. I've seen fake tensor, proxy tensor and dynamo doing a slightly better work, but if there is a stream of work looking into a definite solution, I am much interested on it :)

voznesenskym:
@eellison do you mind if I do this work on the symbolic-shapes branch?

voznesenskym:
> @eellison where can I learn more about "this aligns with the other work ongoing to trace out integers/floats operations in the dynamic work stream? I am working with PyTorch's ONNX converter and properly capturing scalars is a challenge for us too. I've seen fake tensor, proxy tensor and dynamo doing a slightly better work, but if there is a stream of work looking into a definite solution, I am much interested on it :)

Great question. Our very own @ezyang maintains a really high quality blog here: https://dev-discuss.pytorch.org/t/state-of-symbolic-shapes-branch/777/9 (Linked to latest post)

ezyang:
[...]

PR #1163, opened 2022-09-08
Throw on data dependent outputs

eellison:
We need to flip on https://github.com/pytorch/pytorch/pull/83567 for correctness. This patch allows us to continue to trace through `.item()` by catching the exception.

It also switches the optimizers to use `fake_tensors=False` because as explained in https://github.com/pytorch/pytorch/issues/84597 and https://github.com/pytorch/torchdynamo/issues/1138, the current wrapping of `.item()` for those cases only defers the `_local_scaler_dense` call until torch dispatch layer, where fakes tensors have to throw. Those issues lay out a path forward for not graph breaking, but we should get this patch in even with the graph breaks because it's needed for correctness. 

This is ready for review but it's going to take another day before the changes in pytorch are available in the nightly and the tests will pass.


eellison (commented):
This is being tested here https://github.com/pytorch/torchdynamo/blob/main/test/test_repros.py#L887

torchdynamo/variables/tensor.py
```diff
@@ -145,32 +148,26 @@ def context():
...
+                    if use_fake_tensors and isinstance(e, DataDependentOutputException):
+                        if (
+                            config.capture_scalar_outputs
+                            and proxy.node.target == "item"
+                        ):
+                            example_value = torch.zeros(
+                                size=(), dtype=args[0].dtype
+                            ).item()
+                        else:
+                            unimplemented("data dependent operator")
```
voznesenskym, line 160: nit - can you add that it's .item here? either hardcoded or just via `f" ... {proxy.node.target}"` - makes it easier to find and debug, also makes the torchdynamo.explain graph breaks easier to grok.

torchdynamo/variables/tensor.py
```diff
@@ -145,32 +148,26 @@ def context():
...
+                    elif use_fake_tensors and isinstance(
+                        e, DynamicOutputShapeException
+                    ):
+                        unimplemented("dynamic shape operator")
```
voznesenskym, line 164: Is there any data from the exception we can plumb here for a nicer message?

eellison: we can print the dynamic shape operator, yea

torchdynamo/variables/tensor.py
```diff
@@ -145,32 +148,26 @@ def context():
...
                         )
                 except RuntimeError as e:
-                    raise TorchRuntimeError() from e
+                    if use_fake_tensors and isinstance(e, DataDependentOutputException):
+                        if (
+                            config.capture_scalar_outputs
+                            and proxy.node.target == "item"
+                        ):
+                            example_value = torch.zeros(
+                                size=(), dtype=args[0].dtype
```
voznesenskym, line 157: the args here could be malformed, can you assert that:

args len is 1
args[0] is a tensor
args[0] is a numel() of 1

eellison: This is post-dispatch, so it should already have thrown in that case

voznesenskym: Ah yea sg

test/test_optimizers.py
```diff
@@ -71,16 +71,19 @@ def setUpClass(cls):
...
 
     # ASGD has a small optimization that avoids averaging
     # This will fully capture the graph once that optimization is removed
     # NB: in python versions < 3.8, we don't capture graphs when breaks
     # occur in a loop
-    test_asgd = make_test(
-        torch.optim.ASGD, exp_frame_cnt=(0 if sys.version_info < (3, 8) else 6)
-    )
+
+    # Fails without fake tensor:
```
voznesenskym, line 82: Interesting. Can you file an issue for this? Cc @mlazos

voznesenskym (approved):
Seems reasonable

test/test_unspec.py
```diff
@@ -133,6 +133,7 @@ def fn(x):
         res2 = opt_fn(x)
         self.assertTrue(same(res1, res2))
 
+    @patch.object(torchdynamo.config, "fake_tensor_propagation", False)
```
yanboliang, line 136: I'm curious why this failed? This is one of the test case copied from Pytorch CI, where fake tensor is enabled IIRC.

eellison:
```
>>> x = torch.tensor(0)
>>> torch.randint(x, x, x)
  File "<stdin>", line 1, in <module>
TypeError: randint() received an invalid combination of arguments - got (Tensor, Tensor, Tensor), but expected one of:
 * (int high, tuple of ints size, *, torch.Generator generator, Tensor out, torch.dtype dtype, torch.layout layout, torch.device device, bool pin_memory, bool requires_grad)
 * (int high, tuple of ints size, *, Tensor out, torch.dtype dtype, torch.layout layout, torch.device device, bool pin_memory, bool requires_grad)
 * (int low, int high, tuple of ints size, *, torch.Generator generator, Tensor out, torch.dtype dtype, torch.layout layout, torch.device device, bool pin_memory, bool requires_grad)
 * (int low, int high, tuple of ints size, *, Tensor out, torch.dtype dtype, torch.layout layout, torch.device device, bool pin_memory, bool requires_grad)
 ```
 
 the int->torch.tensor(0) conversion causes an error

PR #1165, opened 2022-09-08
Downgrade gym to fix CI

jansel:
Our CI is failing with an error from the gym package. A [new version of it](https://pypi.org/project/gym/0.26.0/#history) was pushed today, so I am hoping using an older one will fix things.


voznesenskym (approved):

desertfire:
Merge as it passes CI now.

Issue #872, opened 2022-08-17
Incorrect stride for tensor shape in model


Chillee:
https://www.internalfb.com/phabricator/paste/view/P522662643

Running minifier...

```
RuntimeError('setStorage: sizes [256, 256], strides [4096, 1], storage offset 0, and itemsize 4 requiring a storage size of 4178944 are out of bounds for storage of size 262144')
```

Chillee:
Minifier result

```
import torch
from torch import tensor, device
import torch.fx as fx
from torchdynamo.testing import rand_strided
from math import inf
from torchinductor.compile_fx import compile_fx_inner
from torch.fx.experimental.proxy_tensor import make_fx

class Repro(torch.nn.Module):



    def forward(self, arg181_1, arg182_1, arg183_1, arg194_1, arg210_1, convert_element_type_default_10, add_tensor_54, mul_tensor_53, mm_default_62, mm_default_64, mm_default_66, mm_default_68, mm_default_70, mm_default_72, mm_default_74, mm_default_76, mm_default_78, mm_default_80, mm_default_82, mm_default_84, mm_default_86, mm_default_88, mm_default_90, add_tensor_65):
        mm_default_36 = torch.ops.aten.mm.default(mul_tensor_53, arg181_1);  mul_tensor_53 = arg181_1 = None
        mm_default_38 = torch.ops.aten.mm.default(mm_default_36, arg182_1);  mm_default_36 = arg182_1 = None
        add_tensor_56 = torch.ops.aten.add.Tensor(add_tensor_54, mm_default_38);  add_tensor_54 = mm_default_38 = None
        mul_tensor_55 = torch.ops.aten.mul.Tensor(add_tensor_56, convert_element_type_default_10);  add_tensor_56 = convert_element_type_default_10 = None
        mm_default_40 = torch.ops.aten.mm.default(mul_tensor_55, arg183_1);  mul_tensor_55 = arg183_1 = None
        mm_default_60 = torch.ops.aten.mm.default(add_tensor_65, arg194_1);  add_tensor_65 = arg194_1 = None
        slice_tensor_16 = torch.ops.aten.slice.Tensor(mm_default_60, 1, 0, 100);  mm_default_60 = None
        mm_default_92 = torch.ops.aten.mm.default(slice_tensor_16, arg210_1);  slice_tensor_16 = arg210_1 = None
[...]

PR #1168, opened 2022-09-08
[inductor] Include last element hint in buffer_reuse_key

jansel:
Should fix root cause of #872

### DIFF
--- a/torchinductor/codegen/wrapper.py
+++ b/torchinductor/codegen/wrapper.py
@@ -12,6 +12,7 @@ from .. import codecache
 from .. import config
 from .. import ir
 from ..utils import has_triton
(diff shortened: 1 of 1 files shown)


voznesenskym (approved):

torchinductor/codegen/wrapper.py
```diff
@@ -24,10 +25,15 @@
 
 
 def buffer_reuse_key(node: ir.Buffer):
```
voznesenskym, line 27: Should we consider is_contiguous in the key as well?

jansel: This is for reusing memory, so layout of the freed buffer doesn't need to match the layout of what we reuse it for.

PR #1173, opened 2022-09-08
Debug guide + restructure in TROUBLESHOOTING.md

mlazos:
First draft ready for review, any suggestions on structure/content are welcome.

Long term TODOs:
* Add section on detailed usage record replay tool once PR is completed
* Add section on debugging in fbcode once best practices there have been established (if this should be public?)
* Fill in section on debugging numerical issues


TROUBLESHOOTING.md
```diff
@@ -6,27 +6,13 @@ In the mean time, you may need to diagnose a particular issue and determine if i
...
-- the expected behavior
-- your OS/python/pytorch version
-
 ## Graph Breaks
 Given a program like this,
 
 ```
-@torchdynamo.optimize(...):
-def some_fun(x):
+with torchdynamo.optimize(...):
```
jansel, line 13: Don't we want to stick with the decorator syntax?

jansel (commented):
Should we move this to the documentation folder?

TROUBLESHOOTING.md
mlazos: Yeah good catch, I left this here from the previous version.

documentation/TROUBLESHOOTING.md
```diff
@@ -0,0 +1,343 @@
...
+
+Some graph break reasons are insurmountable to TorchDynamo, and can't be easily fixed.
+- calling into a C extension other than torch is invisible to torchdynamo, and could do arbitrary things without TorchDynamo being able to introduce necessary guards to ensure that the compiled program would be safe to reuse
+
+### Identifying the cause of a graph break
+
+To identify all graph breaks in a program and the associated reasons for the breaks, `torchdynamo.explain` can be used. This tool runs TorchDynamo on the supplied function and aggregates the graph breaks that are encountered. Here is an example usage:
+
+```py
+from typing import List
```
soumith, line 31: this import can be removed, as it is unused

documentation/TROUBLESHOOTING.md
```diff
@@ -0,0 +1,343 @@
...
+
+def toy_example(a, b):
+    x = a / (torch.abs(a) + 1)
+    print("woo")
+    if b.sum() < 0:
+        b = b * -1
+    return x * b
+
+explained = torchdynamo.explain(toy_example, torch.randn(10), torch.randn(10))
+print(explained[0])
```
soumith, line 43: what would `explained[1]` and the rest of `explained` contain?
worth throwing in a one-liner comment

documentation/TROUBLESHOOTING.md
```diff
@@ -0,0 +1,343 @@
...
+
+At a high level, the TorchDynamo stack consists of a graph capture from Python code (TorchDynamo) and a backend compiler. In this example the backend compiler consists of backward graph tracing (AOTAutograd) and graph lowering (TorchInductor)*. Errors can occur in any component of the stack and will provide full stack traces, but there are backends which enable users to run each component of the stack without the others enabled to narrow down the exact component which is causing the error.
+
+There are some backend options which can enable you to determine which component is causing the error if you're unable to understand the error message that is generated. These are the following:
+
+- `"eager"`: only runs torchdynamo forward graph capture and then runs the captured graph with PyTorch. This provides an indication as to whether TorchDynamo is raising the error.
+
+- `"aot_nop"`: runs torchdynamo to capture a forward graph, and then AOTAutograd to trace the backward graph without any additional backend compiler steps. PyTorch eager will then be used to run the forward and backward graphs. This is useful to narrow down the issue to AOTAutograd.
```
soumith, line 78: maybe this should be renamed to just `aot` and not `aot_nop`?

documentation/TROUBLESHOOTING.md
voznesenskym: We can also unpack explain's return tuple into 4 args, and keep the same names as in the function...

documentation/TROUBLESHOOTING.md
jansel: In the performance dashboard we call this `aot_eager`, we should pick one and stick with it.

penguinwu:
Please do not expect readers to read the doc from beginning to end. The doc will grow longer and longer as we add more content and most people just want to quickly find the tools that help their problem and just read that part. 

Structure-wise:
- Please create a TOC so that people can jump directly to the sections/tools that they are interested in.
- How about adding a table of useful commands (flags) and tools up front w/ links to how to use the tools (e.g., how to use the minifier, how to enable debugging output, etc)?
- Definitely need a subsection on the minifier with an obvious subtitle so that people can jump directly to it
- We start the doc w/ graph breaks. For most users, they may be confused of what is graph breaks and why we can about it. Graph breaks seem an advanced topic that should be moved to later section, perhaps as part of performance tuning
- Please add the missing sections if the tools are still under development (and just say that tools are under development) because we are using the doc as a blueprint for our tools development too. This way we know whether there is any gap in our tools.

documentation/TROUBLESHOOTING.md
anijain2305: Here - https://github.com/pytorch/torchdynamo/pull/1247

mlazos: Yes I went with `aot_eager` as well

documentation/TROUBLESHOOTING.md
```diff
@@ -0,0 +1,383 @@
...
+| Minifier for TorchInductor          | If the error is known to occur after AOTAutograd find smallest subgraph wich reproduces errors during TorchInductor lowering | set environment variable TORCHDYNAMO_REPRO_AFTER="aot"                       |
+| `torchdynamo.explain`               | Find graph breaks and display reasoning for them                                                                             | `torchdynamo.explain(fn, *inputs)`                                           |
+| Record/Replay (in progress)         | Record and replay frames which to reproduce errors during graph capture                                                      | `torchdynamo.config.replay_record_enabled = True`                            |
+| TorchDynamo function name filtering | only compile functions with the given name to reduce noise when debugging an issue                                           | set environment variable TORCHDYNAMO_DEBUG_FUNCTION=\<name\>                 |
```
jansel, line 31: We also have `TORCHINDUCTOR_TRACE=1` https://github.com/pytorch/torchdynamo/pull/1071

documentation/TROUBLESHOOTING.md
```diff
@@ -0,0 +1,383 @@
...
+
+# Guide to Diagnosing Runtime Errors
+Below is the TorchDynamo compiler stack. 
+
+<img src="./images/td_stack.png" width=800>
```
jansel, line 36: Perhaps move TorchDynamo.png in the root to this images directory too.

documentation/TROUBLESHOOTING.md
```diff
@@ -0,0 +1,383 @@
...
+
+
+## TorchInductor Errors
+If the error doesn't occur with the `"eager"` backend, then the backend compiler is the source of the error ([example error](https://gist.github.com/mlazos/2f13681e3cc6c43b3911f336327032de])). There are [different choices](https://github.com/pytorch/torchdynamo/blob/0b8aaf340dad4777a080ef24bf09623f1aa6f3dd/README.md#existing-backends) for backend compilers for torchdynamo, with TorchInductor or nvfuser fitting the needs of most users. This section focuses on TorchInductor as the motivating example, but some tools will be usable with other backend compilers. 
+
+Below is the portion of the stack which we are focusing on:
+
+<img src="./images/torchinductor_backend.png" width=600>
```
jansel, line 106: This image isn't exactly right.  The forwards graph also goes through AOT autograd, not just the backwards.

documentation/TROUBLESHOOTING.md
mlazos: Done. Added section with your explanation and examples from your PR as well.

jansel (approved):

PR #1187, opened 2022-09-09
[OSS Benchmarks] Separate out accuracy and perf testing

anijain2305:
This PR
* separates out accuracy and perf testing. We now need to specify `--accuracy` or `--performance`.
* For accuracy, we use **model.eval()**. For performance, **we use model.train()**.
* For accuracy, we use a batch size of 2 to ensure that fp64 model can also fit in the GPU memory. bs = 2 allows to test batch_norm etc
* Keeps`skip_accuracy_check` for inductor for now.
* Propagates the changes to runner and nightly infra.
* Pretrained for TIMM models to have deterministic weights.

Problems encountered
* Inductor is showing accuracy issues. For Fp32, tracked here - https://github.com/pytorch/torchdynamo/issues/1190
* Had to skip many models in the CI, because they dont fit in memory for fp64.
* TIMM models might take more time in CI now, as CI has to fetch pretrained weights for each model.

cc @jansel

### DIFF
--- a/.circleci/check_csv.py
+++ b/.circleci/check_csv.py
@@ -7,17 +7,16 @@ import pandas as pd
 
 def check_csv(filename):
     """
-    Basic accuracy checking. If a model fails, the speedup will be 0.
+    Basic accuracy checking.
     """
 
-    actual = pd.read_csv(filename)
+    df = pd.read_csv(filename)
 
     failed = []
-    for model_name in actual["name"]:
-        speedup = float(actual.loc[actual["name"] == model_name]["speedup"])
-        status = "PASS"
-        if speedup == 0:
-            status = "FAIL"
+    for _, row in df.iterrows():
+        model_name = row["name"]
+        status = row["accuracy"]
+        if "pass" not in status:
             failed.append(model_name)
 
         print(f"{model_name:34} {status}")
--- a/benchmarks/common.py
+++ b/benchmarks/common.py
@@ -5,7 +5,6 @@ import copy
 import csv
 import functools
 import io
-import itertools
 import logging
 import os
 import random
@@ -58,18 +57,23 @@ CI_SKIP_INFERENCE = [
     # TorchBench
     "DALLE2_pytorch",
     "detectron2",
+    "dlrm",
+    "DALLE2_pytorch",
+    "fambench_dlrm",
     "fastNLP_Bert",
     "hf_Reformer",
     "moco",
     "pyhpc_",
     "Super_SloMo",
     "tacotron2",
+    "pytorch_unet",
     "yolov3",
     # Huggingface
     "AlbertForQuestionAnswering",
     "AllenaiLongformerBase",
     "BartForCausalLM",
     "BertForQuestionAnswering",
+    "BartForConditionalGeneration",  # OOM
     "BigBird",
     "BlenderbotSmallForConditionalGeneration",
     "DebertaForQuestionAnswering",
@@ -137,6 +141,7 @@ CI_SKIP_TRAINING = [
     "PegasusForCausalLM",
     # OOM
     "BigBird",
+    "TrOCRForCausalLM",
     # TIMM
     "dpn107",
     "convit_base",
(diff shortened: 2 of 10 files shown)


benchmarks/common.py
```diff
@@ -1567,16 +1631,34 @@ def main(runner, original_dir=None):
...
-        args.repeat = 2
         args.exclude += CI_SKIP_INFERENCE
         if args.training:
             args.exclude += CI_SKIP_TRAINING
 
+    if args.accuracy:
+        # Use small batch size. We use >1 batch size to ensure we test
+        # batch_norm type of operators that work on batch dims.
+        # TODO - Go through the failures for batch size = 2
+        args.batch_size = 2
```
anijain2305, line 1642: Check if args.batch_size is None

benchmarks/common.py
```diff
@@ -1856,8 +1920,12 @@ def main(runner, original_dir=None):
                     model_name,
                     batch_size=batch_size,
                 )
-            except NotImplementedError:
-                log.warning(f"{args.only} failed to load")
+            except NotImplementedError as e:
+                print(e)
+                import traceback
+
+                print(traceback.format_exc())
```
anijain2305, line 1927: Clean up

benchmarks/timm_models.py
```diff
@@ -92,6 +92,8 @@ def pip_install(package):
 SKIP = {
     # Unusual training setup
     "levit_128",
+    # pretrained True does not work
+    "hardcorenas_a",
```
anijain2305, line 96: Add pretrained models from the CI here as well.

torchdynamo/utils.py
```diff
@@ -721,18 +721,19 @@ def same(
...
+                multiplier = 2
 
-                if fp64_ref.numel() < 500:
-                    # In the presence of noise, noise might dominate our error
-                    # metric for smaller tensors.
-                    multiplier = 2
+                # if fp64_ref.numel() < 500:
+                #     # In the presence of noise, noise might dominate our error
+                #     # metric for smaller tensors.
+                #     multiplier = 2.5
```
anijain2305, line 729: Clean up and always set multiplier as 2x

benchmarks/torchbench.py
```diff
@@ -59,7 +59,6 @@
     "detectron2_fasterrcnn_r_50_fpn",
     "detectron2_maskrcnn_r_101_c4",
     "detectron2_maskrcnn_r_101_fpn",
-    "detectron2_maskrcnn_r_50_c4",
```
desertfire, line 62: Why dropping this one?

.circleci/check_csv.py
```diff
@@ -7,17 +7,16 @@
...
     failed = []
-    for model_name in actual["name"]:
-        speedup = float(actual.loc[actual["name"] == model_name]["speedup"])
-        status = "PASS"
-        if speedup == 0:
-            status = "FAIL"
+    for _, row in df.iterrows():
+        model_name = row["name"]
+        status = row["accuracy"]
+        if status != "pass":
```
desertfire, line 19: With this change, our CI coverage will drop significantly if we skip all the failed-accuracy-check benchmarks. With the uncertainty on when we can fix those errors, this worries me. Can we distinguish accuracy check failures and other failures, and only check for other failures here?

anijain2305: This worries me as well. One way is to just fallback to skip_accuracy_check for now, and get this PR in. This will ensure that non-accuracy failures will get tested through CI.

And then prioritize the accuracy failures. What do you think?

benchmarks/torchbench.py
anijain2305: It's the other way around. I am actually removing this from the skip list, so that it shows up in nightly.

torchdynamo/utils.py
```diff
@@ -772,18 +772,19 @@ def same(
...
 
                 passes_test = res_error <= (multiplier * ref_error + 1e-5)
                 if not passes_test:
-                    log.warning(
+                    log.error(
                         f"RMSE (res-fp64): {res_error:.5f}, (ref-fp64): {ref_error:.5f}"
                     )
+                    # import pdb; pdb.set_trace()
```
desertfire, line 787: nit: clean up

desertfire (approved):
`Removes skip_accuracy_check for inductor. This enables better accuracy testing for TorchInductor.` in the description needs to be updated.

If the added TIMM pretrained model downloading causes CI to slow down too much, I can add more circleci caching later.

Issue #1185, opened 2022-09-09
[inductor][ddp] File not found when trying to rename tmp_path to path in code cache


davidberard98:
Repro:
Run with:
* inductor & dynamo, main branch
* pytorch, master branch
* torchbench, https://github.com/pytorch/benchmark/blob/wconstab/ddp_experiments/ddp_experiments.py
* 8 gpus per node, 2 nodes (I can repro more reliably with n>=2 nodes), resnet50 model (cannot repro on other models)

```
/data/home/dberard/miniconda/envs/bench-fast/lib/python3.8/site-packages/torchvision/models/_utils.py:208: UserWarning: The parameter 'pretrained' is deprecated since 0.13 and will be removed in 0.15, please use 'weights' instead.
  warnings.warn(
/data/home/dberard/miniconda/envs/bench-fast/lib/python3.8/site-packages/torchvision/models/_utils.py:223: UserWarning: Arguments other than a weight enum or `None` for 'weights' are deprecated since 0.13 and will be removed in 0.15. The current behavior is equivalent to passing `weights=ResNet50_Weights.IMAGENET1K_V1`. You can also use `weights=ResNet50_Weights.DEFAULT` to get the most up-to-date weights.
  warnings.warn(msg)
WARNING:root:Using TorchDynamo with a context manager will be deprecated soon.Please read https://github.com/pytorch/torchdynamo#usage-example to use TorchDynamo using an annotation.
ERROR:root:Error while processing frame
Traceback (most recent call last):
  File "/data/home/dberard/miniconda/envs/bench-fast/lib/python3.8/site-packages/torchdynamo/eval_frame.py", line 225, in catch_errors
    log.debug(f"skipping {frame.f_code.co_name} {frame.f_code.co_filename}")
  File "/fsx/users/dberard/scratch-local/bench-fast/benchmark/torchbenchmark/util/model.py", line 188, in invoke
    self.train()
  File "/fsx/users/dberard/scratch-local/bench-fast/benchmark/torchbenchmark/util/framework/vision/model_factory.py", line 65, in train
    pred = self.model(data)
  File "/data/home/dberard/miniconda/envs/bench-fast/lib/python3.8/site-packages/torch/nn/modules/module.py", line 1190, in _call_impl
    return forward_call(*input, **kwargs)
  File "/data/home/dberard/miniconda/envs/bench-fast/lib/python3.8/site-packages/torch/nn/parallel/distributed.py", line 1040, in forward
[...]

PR #1191, opened 2022-09-10
Use mkstemp for codecache file creation

jansel:
Possible fix for #1185

### DIFF
--- a/torchinductor/codecache.py
+++ b/torchinductor/codecache.py
@@ -8,6 +8,7 @@ import random
 import re
 import subprocess
 import sysconfig
+import tempfile
 import types
 from ctypes import cdll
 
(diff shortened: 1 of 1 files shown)


torchinductor/codecache.py
```diff
@@ -39,10 +39,10 @@ def write(source_code, ext, extra=""):
         os.makedirs(subdir, exist_ok=True)
     path = os.path.join(subdir, f"{basename}.{ext}")
     if not os.path.exists(path):
-        # use a random temp file for thread safety
-        tmp_path = f"{path}.{random.randint(0, 2**31)}"
-        with open(tmp_path, "w") as fd:
-            fd.write(source_code)
+        # use a temp file for thread safety
+        fd, tmp_path = tempfile.mkstemp(dir=subdir)
```
voznesenskym, line 43: Not sure if its a concern, but mkstemp adds new restrictions here:

> The file is readable and writable only by the creating user ID. If the platform uses permission bits to indicate whether a file is executable, the file is executable by no one. The file descriptor is not inherited by child processes.

voznesenskym (approved):

torchinductor/codecache.py
voznesenskym: More detail, I say this because in the ddp context of #1185 - we may run into permission errors, if say, the file is copied onto another machine and cannot be executed there. We may want a os.chmod here.

jansel: We are writing this file to `/tmp/torchinductor_$USER` so other users should have their own directory.

PR #1192, opened 2022-09-10
[inductor] Misc compile time improvements

jansel:
This is a smaller improvement in compile time.  Roughly 5%.

### DIFF
--- a/torchinductor/config.py
+++ b/torchinductor/config.py
@@ -25,7 +25,7 @@ inplace_buffers = False
 benchmark_harness = True
 
 # control store vs recompute heuristic
-realize_reads_threshold = 4
+realize_reads_threshold = 5
 realize_bytes_threshold = 2000
 
 # fallback to eager for random/dropout, this is slow but useful for debugging
@@ -72,9 +72,6 @@ class triton:
 
     # Use cudagraphs on output code
     cudagraphs = True
-
-    # Monkey patching to lower overheads
-    hackery = False
 
     # choose conv backend, "aten" or "triton" or "autotune"
     convolution = "aten"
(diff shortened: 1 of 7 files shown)


torchinductor/config.py
```diff
@@ -25,7 +25,7 @@
 benchmark_harness = True
 
 # control store vs recompute heuristic
-realize_reads_threshold = 4
+realize_reads_threshold = 5
```
jansel, line 28: I measured this being slightly faster on torchbench.

torchinductor/scheduler.py
```diff
@@ -701,7 +701,7 @@ def topological_sort_schedule(self):
         def visit(n):
             if n not in seen:
                 seen.add(n)
-                for dep in n.unmet_dependencies:
+                for dep in sorted(n.unmet_dependencies, key=lambda d: d.name):
```
jansel, line 704: This makes our output wrapper code deterministic again.

torchinductor/config.py
jansel: I reverted this because it caused an illegal memory access on one timm model.

torchinductor/config.py
```diff
@@ -73,9 +73,6 @@ class triton:
     # Use cudagraphs on output code
     cudagraphs = True
 
-    # Monkey patching to lower overheads
-    hackery = False
```
Chillee, line 77: Nice :)

torchinductor/ir.py
```diff
@@ -3180,22 +3182,25 @@ def realize_hint(self):
         """
         Called on buffers we expect to be forced to realize later.
         """
-        if self.num_reads() > 1:
+        if isinstance(self.data, (Pointwise, Reduction)) and self.num_reads() > 1:
```
Chillee, line 3185: When does this ever matter?

Chillee (approved):

torchinductor/ir.py
jansel: It just avoids a call to `num_reads`, which is expensive.

PR #1195, opened 2022-09-11
Autotune

ngimel:
Add more reduction autotuning cases, this mostly doesn't change benchmarking numbers (marginal decreases on a couple models), but improves compile times 30-75% (1.30 -> 1 minute on hf_Bert, 12 min -> 4 min on densenet). Densenet now doesn't have autotuned kernels at all, but there are still some autotuned kernels for other nets.

### DIFF
--- a/torchinductor/triton_ops/autotune.py
+++ b/torchinductor/triton_ops/autotune.py
@@ -327,7 +327,7 @@ def reduction_heuristics(size_hints, reduction_hint=False, filename=None):
         elif reduction_hint == ReductionHint.OUTER:
             return apply_triton_config(outer_config)
         elif reduction_hint == ReductionHint.OUTER_TINY:
-            return apply_triton_config(size_hints, 2 * (256//rnumel), rnumel)
+            return apply_triton_config(size_hints, 2 * (256 // rnumel), rnumel)
         if not config.triton.autotune:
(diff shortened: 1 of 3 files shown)


torchinductor/triton_ops/autotune.py
```diff
@@ -320,8 +321,13 @@ def reduction_heuristics(size_hints, reduction_hint=False, filename=None):
         contiguous_config = triton_config_reduction(
             size_hints, 1, (rnumel if 256 <= rnumel < 2048 else 2048), num_stages=1
         )
+        outer_config = triton_config_reduction(size_hints, 128, 8)
         if reduction_hint == ReductionHint.INNER:
             return apply_triton_config(contiguous_config)
+        elif reduction_hint == ReductionHint.OUTER:
+            return apply_triton_config(outer_config)
+        elif reduction_hint == ReductionHint.OUTER_TINY:
+            return apply_triton_config(size_hints, 2 * (256 // rnumel), rnumel)
```
jansel, line 330: Won't `256 // rnumel` be `0` if rnumel==512?  Perhaps we should guard this based on `rnumel<=256`?

ngimel: Yeah I initially used tiny hint only for rnumel less than 256, but then tweaked it some more, I'll make sure the numbers agree.

jansel (approved):

PR #1201, opened 2022-09-12
Flip fake tensors on

eellison:
The cudagraph changes were already approved  [here](https://github.com/pytorch/torchdynamo/pull/1160) and landed incorrectly with ghstack lol.

Other changes
- only decompose batch_norm2d on cuda, was running into a fake tensor issue with cpu 
- additional model skips 

I'll work on fixing these [issues](https://github.com/pytorch/torchdynamo/issues/1215) but I still think it's worth landing in its current state, especially since some of these will require fixes in core which can take a couple days to make it to dynamo ci.

### DIFF
--- a/benchmarks/common.py
+++ b/benchmarks/common.py
@@ -115,6 +115,8 @@ CI_SKIP_TRAINING = [
     "speech_transformer",  # from functionalization
     "vision_maskrcnn",  # from functionalization
     "timm_efficientnet",  # from functionalization (only fails for inductor)
+    "hf_Bert",
+    "soft_actor_critic",
     # OOM
     "resnet50_quantized_qat",
     # Huggingface
--- a/benchmarks/common.py
+++ b/benchmarks/common.py
@@ -109,6 +109,7 @@ CI_SKIP_TRAINING = [
     "pytorch_struct",
     "timm_regnet",
     "vgg16",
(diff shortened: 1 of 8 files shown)


Chillee (approved):

benchmarks/common.py
```diff
@@ -110,10 +110,13 @@ def set_model_name(name):
     "pytorch_struct",
     "timm_regnet",
     "vgg16",
+    "drq",
     "Background_Matting",  # from functionalization
     "speech_transformer",  # from functionalization
     "vision_maskrcnn",  # from functionalization
     "timm_efficientnet",  # from functionalization (only fails for inductor)
+    "hf_Bert",
+    "soft_actor_critic",
```
anijain2305, line 119: Can you open bugs for tracking this?

torchdynamo/__init__.py
```diff
@@ -16,6 +17,9 @@
 from .utils import guard_failures
 from .utils import orig_code_map
 
+functorch.compile.config.use_fake_tensor = True
```
anijain2305, line 20: I remember it was causing some cyclic dependencies - https://github.com/pytorch/torchdynamo/pull/1096

anijain2305 (commented):
LGTM

PR #1204, opened 2022-09-12
Reland #1195 heuristics PR

ngimel:
Per title

### DIFF
--- a/test/test_torchinductor.py
+++ b/test/test_torchinductor.py
@@ -37,7 +37,8 @@ try:
     assert get_decompositions([torch.ops.aten.trace])
     # Requires functorch
     from torchinductor.compile_fx import compile_fx_inner
-except (ImportError, ModuleNotFoundError, AssertionError):
+except (ImportError, ModuleNotFoundError, AssertionError) as e:
+    print(e)
     raise unittest.SkipTest("requires sympy/functorch")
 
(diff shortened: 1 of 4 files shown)


desertfire:
Repost the chat msg here,
"
I was able to narrow down this to https://github.com/pytorch/torchdynamo/blob/395bbc8d221f6e41192c8a7a8345d8970e6b730c/torchinductor/ir.py#L33. The CI test was on CPU, and this line introduces a import triton which eventually triggers an exception at https://github.com/pytorch/torchdynamo/blob/136b7ea308f38bfca3094b130b15cbd0ce6b89b2/test/test_torchinductor.py#L41
"

test/test_torchinductor.py
```diff
@@ -37,7 +37,8 @@
     assert get_decompositions([torch.ops.aten.trace])
     # Requires functorch
     from torchinductor.compile_fx import compile_fx_inner
-except (ImportError, ModuleNotFoundError, AssertionError):
+except (ImportError, ModuleNotFoundError, AssertionError) as e:
+    print(e)
```
ngimel, line 41:
```suggestion
except (ImportError, ModuleNotFoundError, AssertionError):
```

desertfire (approved):

Issue #1200, opened 2022-09-12
TorchBench training error: AttributeError: 'NoneType' object has no attribute 'type


desertfire:
Repro:
```
 benchmarks/torchbench.py -d cuda --inductor --training --float32 --use-eval-mode --no-skip -k  vision_maskrcnn
```
Error:
```
Traceback (most recent call last):
  File "/scratch/binbao/work/triton/python/triton/code_gen.py", line 1309, in _compile
    generator.visit(self.parse())
  File "/scratch/binbao/work/triton/python/triton/code_gen.py", line 754, in visit
    return super().visit(node)
  File "/scratch/binbao/work/env/lib/python3.9/ast.py", line 407, in visit
    return visitor(node)
  File "/scratch/binbao/work/triton/python/triton/code_gen.py", line 245, in visit_Module
    ast.NodeVisitor.generic_visit(self, node)
  File "/scratch/binbao/work/env/lib/python3.9/ast.py", line 415, in generic_visit
    self.visit(item)
  File "/scratch/binbao/work/triton/python/triton/code_gen.py", line 754, in visit
    return super().visit(node)
  File "/scratch/binbao/work/env/lib/python3.9/ast.py", line 407, in visit
    return visitor(node)
  File "/scratch/binbao/work/triton/python/triton/code_gen.py", line 306, in visit_FunctionDef
    has_ret = self.visit_compound_statement(node.body)
  File "/scratch/binbao/work/triton/python/triton/code_gen.py", line 239, in visit_compound_statement
    self.last_ret = self.visit(stmt)
  File "/scratch/binbao/work/triton/python/triton/code_gen.py", line 754, in visit
    return super().visit(node)
  File "/scratch/binbao/work/env/lib/python3.9/ast.py", line 407, in visit
    return visitor(node)
  File "/scratch/binbao/work/triton/python/triton/code_gen.py", line 350, in visit_Assign
    values = self.visit(node.value)
  File "/scratch/binbao/work/triton/python/triton/code_gen.py", line 754, in visit
    return super().visit(node)
  File "/scratch/binbao/work/env/lib/python3.9/ast.py", line 407, in visit
    return visitor(node)
  File "/scratch/binbao/work/triton/python/triton/code_gen.py", line 434, in visit_BinOp
    return getattr(lhs, fn)(rhs, _builder=self.builder)
  File "/scratch/binbao/work/triton/python/triton/language/core.py", line 46, in wrapper
    return fn(*args, **kwargs)
[...]

PR #1211, opened 2022-09-13
Fix a gradient is None issue in max_pool2d_with_indices_backward

desertfire:
Summary: For https://github.com/pytorch/torchdynamo/issues/1200

### DIFF
--- a/torchinductor/lowering.py
+++ b/torchinductor/lowering.py
@@ -2173,12 +2173,14 @@ def max_pool2d_with_indices_backward(
             h // stride[0] - max(0, (h - kernel_size[0]) // stride[0])
             for h in range(kernel_size[0] * 2)
         ]
+        + [1]
     )
     w_window_size = max(
         [
             w // stride[1] - max(0, (w - kernel_size[1]) // stride[1])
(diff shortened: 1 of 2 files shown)


torchinductor/lowering.py
```diff
@@ -2173,12 +2173,14 @@ def max_pool2d_with_indices_backward(
             h // stride[0] - max(0, (h - kernel_size[0]) // stride[0])
             for h in range(kernel_size[0] * 2)
         ]
+        + [1]
```
voznesenskym, line 2176: Why does adding [1] here fix it?

desertfire: It's to make h_window_size/w_window_size at least has value 1. I can rewrite this to use another `max` if that makes things more clear.

eellison, line 2176: nit, i think it would be a little more intuitive to fold the one into the expression than as `+ [1]`

eellison (approved):
LGTM - factor out / correct the window calculation fix for `avg_pool2d_backward` as well ?

jansel (approved):

PR #1218, opened 2022-09-13
Improve torchdynamo.explain() explanation

suo:
Improve the stack trace to point to the actual location of the graph
break.

Before, we were lazily computing the stack trace at the time we were
compiling the subgraph. However, if were graph breaking inside an
inlined function, the sequence of events looked like:
1. Catch the graph break exception
2. Restore the old graph state <-- thus overriding the user stack location
3. Compute the stack trace and add it as a message to the compiled graph

So the graph break explanation pointed to the caller, not the actual
location of the graph break. By computing the stack trace at the same
time we're processing the graph break exception, we can fix that.

This also starts printing the whole stack trace that we have instead of
just the most recent line, which I think is more useful (but if people
find it noisy I can change it back).

Looking for advice on testing.

### DIFF
--- a/torchdynamo/symbolic_convert.py
+++ b/torchdynamo/symbolic_convert.py
@@ -117,7 +117,10 @@ def generic_jump(truth_fn: typing.Callable, push: bool):
         elif isinstance(value, TensorVariable) and self.should_compile_partial_graph():
             # compile a partial subgraph prefix then jump into user code
             self.push(value)
-            self.output.compile_subgraph(self, msg="generic_jump")
+            user_stack = self.format_frame_summary()
+            self.output.compile_subgraph(
+                self, msg=f"generic_jump {typestr(value)}\n{user_stack}"
+            )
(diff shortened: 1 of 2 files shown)


torchdynamo/symbolic_convert.py
```diff
@@ -117,7 +117,10 @@ def inner(self: "InstructionTranslatorBase", inst: Instruction):
         elif isinstance(value, TensorVariable) and self.should_compile_partial_graph():
             # compile a partial subgraph prefix then jump into user code
             self.push(value)
-            self.output.compile_subgraph(self, msg="generic_jump")
+            user_stack = self.format_frame_summary()
+            self.output.compile_subgraph(
+                self, msg=f"generic_jump {typestr(value)}\n{user_stack}"
+            )
```
voznesenskym, line 123: Nice.

voznesenskym (approved):

Issue #965, opened 2022-08-23
[TIMM: nasnetalarge, pnasnet5large] AssertionError: None


desertfire:
Repro:
```
benchmarks/timm_models.py -d cuda --inductor --training --float32 --use-eval-mode --quiet -k nasnetalarge
benchmarks/timm_models.py -d cuda --inductor --training --float32 --use-eval-mode --quiet -k pnasnet5large
```
Error:
```
  File "/fsx/users/binbao/torchdynamo-tip/torchinductor/codegen/common.py", line 488, in generate
    assert isinstance(expr, str), expr
AssertionError: None
ERROR
```

desertfire:
@Chillee , minifier hits OOM on this one.

Chillee:
@desertfire  can you repro it with a smaller batch size?

Chillee:
I'm actually getting segfaults/`std::bad_alloc` errors with `nasnetalarge`.

Minimal repro

```

import torch
from torch import tensor, device
import torch.fx as fx
from torchdynamo.testing import rand_strided
from math import inf
from torch.fx.experimental.proxy_tensor import make_fx

class Repro(torch.nn.Module):

    def forward(self, sub_tensor_1, unsqueeze_default):
        gather_default = torch.ops.aten.gather.default(sub_tensor_1, 1, unsqueeze_default);  sub_tensor_1 = unsqueeze_default = None
        return (gather_default,)

args = [((1, 1000), (1000, 1), torch.float32, 'cuda'), ((1, 1), (1, 1), torch.int64, 'cuda')]
args = [rand_strided(shape, stride, dtype, device) for shape, stride, dtype, device in args]
mod = make_fx(Repro())(*args)

from torchinductor.compile_fx import compile_fx_inner

compiled = compile_fx_inner(mod, args)
compiled(*args)
```

desertfire:
Ok, `--batch_size=1` worked, although it took over a hour for minifer to run.

Minified version:
```

import torch
from torch import tensor, device
import torch.fx as fx
from torchdynamo.testing import rand_strided
from math import inf
from torch.fx.experimental.proxy_tensor import make_fx

class Repro(torch.nn.Module):



    def forward(self, arg1851_1, getitem_393):
        avg_pool2d_backward_default_12 = torch.ops.aten.avg_pool2d_backward.default(getitem_393, arg1851_1, [1, 1], [2, 2], [0, 0], False, False, None);  getitem_393 = arg1851_1 = None
[...]

PR #1222, opened 2022-09-14
Fix a gradient is None issue in avg_pool2d_backward

desertfire:
Summary: For https://github.com/pytorch/torchdynamo/issues/965

### DIFF
--- a/torchinductor/lowering.py
+++ b/torchinductor/lowering.py
@@ -2531,13 +2531,13 @@ def avg_pool2d_backward(
 
     h_window_size = max(
         [
-            h // stride[0] - max(0, (h - kernel_size[0]) // stride[0])
+            max(h // stride[0] - max(0, (h - kernel_size[0]) // stride[0]), 1)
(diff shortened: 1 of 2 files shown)


torchinductor/lowering.py
```diff
@@ -2618,15 +2618,15 @@ def fn(idx):
                     scale,
                 )
 
+                mask = ops.and_(
+                    ops.lt(ph, phend),
+                    ops.lt(pw, pwend),
+                )
                 if gradient is None:
-                    # don't need mask for 0, 0
```
desertfire, line 2622: @jansel , I had to change this to pass the added test. I wonder if `max_pool2d_with_indices_backward ` has the same issue, https://github.com/pytorch/torchdynamo/blob/138eb96c36c6ba5a14e65d28fad3cbd5a6e6dcfc/torchinductor/lowering.py#L2224, although I haven't tried to construct a test case for that.

ngimel (approved):

Issue #1178, opened 2022-09-08
Bert model training fails with inductor


yanboliang:
Repro:
```
import os

import torch
from datasets import load_dataset
from torch.utils.data import DataLoader
from transformers import AutoModelForSequenceClassification
from transformers import AutoTokenizer

import torchdynamo

torchdynamo.config.fake_tensor_propagation = False
torchdynamo.config.capture_scalar_outputs = True

os.environ["TOKENIZERS_PARALLELISM"] = "false"
device = torch.device("cuda") if torch.cuda.is_available() else torch.device("cpu")
num_samples = 100
num_epochs = 3


def data_processing(num_samples):
    dataset = load_dataset("yelp_review_full")
    tokenizer = AutoTokenizer.from_pretrained("bert-base-cased")

    def tokenize_function(examples):
        return tokenizer(examples["text"], padding="max_length", truncation=True)

    tokenized_datasets = dataset.map(tokenize_function, batched=True)

    tokenized_datasets = tokenized_datasets.remove_columns(["text"])
    tokenized_datasets = tokenized_datasets.rename_column("label", "labels")
    tokenized_datasets.set_format("torch")

    small_train_dataset = (
        tokenized_datasets["train"].shuffle(seed=42).select(range(num_samples))
    )
    small_eval_dataset = (
        tokenized_datasets["test"].shuffle(seed=42).select(range(num_samples))
    )

    train_dataloader = DataLoader(small_train_dataset, shuffle=True, batch_size=8)
    eval_dataloader = DataLoader(small_eval_dataset, batch_size=8)

    return train_dataloader, eval_dataloader

if __name__ == "__main__":
    train_dataloader, eval_dataloader = data_processing(num_samples)
    model = AutoModelForSequenceClassification.from_pretrained(
        "bert-base-cased", num_labels=5
    )
    model.to(device)
    model.train()
    opt_model = torchdynamo.optimize("inductor")(model)
    for i, batch in enumerate(train_dataloader, 0):
        batch = {k: v.to(device) for k, v in batch.items()}
        outputs = opt_model(**batch)
```
Error:
```
Traceback (most recent call last):
  File "/fsx/users/ybliang/conda/envs/dynamo-gpu/lib/python3.9/site-packages/functorch/_src/aot_autograd.py", line 56, in preserve_rng_state
    yield
[...]

PR #1226, opened 2022-09-14
Fix caching of simplify_loops results

### LINKED ISSUE #926: Inductor Error: aten.native_layer_norm_backward.default  Illegal Memory Access
`python /scratch/eellison/work/torchdynamo/benchmarks/microbenchmarks/operatorbench.py --op=aten.native_layer_norm_backward.default --dtype=float32 --suite=huggingface`

>Running aten.native_layer_norm_backward.default
Iter 0
Iter 1
Iter 2
Iter 3
Iter 4
Iter 5
Iter 6
Iter 7
Iter 8
Iter 9
Iter 10
error aten.native_layer_norm_backward.default
CUDA error: an illegal memory access was encountered
CUDA kernel errors might be asynchronously reported at some other API call,so the stacktrace below might be incorrect.
For debugging consider passing CUDA_LAUNCH_BLOCKING=1.
Iter 11```

Same error happens with `native_layer_norm.default`

> CUDA_VISIBLE_DEVICES=6 python /scratch/eellison/work/torchdynamo/benchmarks/microbenchmarks/operatorbench.py --op=aten.native_layer_norm.default --dtype=float32 --suite=huggingface

ngimel:
Fixes #926, #1178
`simplify_loops` uses `sizevars.simplify` a lot, so its cache should be invalidated on any changes to `sizevars.replacements`. To make it easier, I moved `simplify_loops` to `sizevars`, and added invalidation on changes to `replacements`, similar to `simplify_with_ranges`.


jansel (approved):
Awesome!  Good catch finding this.

PR #1229, opened 2022-09-14
Enhance operatorbench.py for op tests

desertfire:
Summary:
1. Skip nvfuser measurement by default
2. Add an argument for controlling repeat times, and set it to 1 when running in the accuracy-checking mode

### DIFF
--- a/benchmarks/microbenchmarks/operator_inp_utils.py
+++ b/benchmarks/microbenchmarks/operator_inp_utils.py
@@ -15,6 +15,7 @@ import torch
 from torch.utils._python_dispatch import TorchDispatchMode
 from torch.utils._pytree import tree_flatten
 from torch.utils._pytree import tree_map
+from torch.testing import make_tensor
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


eellison (approved):
Nice! for accuracy testing I think we should also fix the input generation - I think default `empty` initialized tensors will be a lot of zeros. Here was patch for it:
```
diff --git a/benchmarks/microbenchmarks/operator_inp_utils.py b/benchmarks/microbenchmarks/operator_inp_utils.py
index b75eab23..dec0e70d 100644
--- a/benchmarks/microbenchmarks/operator_inp_utils.py
+++ b/benchmarks/microbenchmarks/operator_inp_utils.py
@@ -15,6 +15,7 @@ import torch
 from torch.utils._python_dispatch import TorchDispatchMode
 from torch.utils._pytree import tree_flatten
 from torch.utils._pytree import tree_map
+from torch.testing import make_tensor
 
@@ -89,9 +90,14 @@ def deserialize_sparse_tensor(size, dtype, layout, is_coalesced, nnz=None):
 
 def deserialize_tensor(size, dtype, stride=None):
     if stride is not None:
-        return torch.empty_strided(size, stride, dtype=dtype)
+        out = torch.empty_strided(size, stride, dtype=dtype)
     else:
-        return torch.empty(size, dtype=dtype)
+        out = torch.empty(size, dtype=dtype)
+    try:
+        out.copy_(make_tensor(size, dtype=dtype, device="cpu"))
+    except:
+        return out
+    return out
 
 
 def serialize_tensor(e):

```

PR #1238, opened 2022-09-15
Move Functorch Config Change

eellison:
Following https://github.com/pytorch/torchdynamo/pull/1096 - according to @anijain2305 previous import may caused circular import

### DIFF
--- a/torchdynamo/__init__.py
+++ b/torchdynamo/__init__.py
@@ -1,6 +1,3 @@
-# TODO: remove this config entirely
-import functorch.compile
-
 from . import allowed_functions
 from . import convert_frame
 from . import eval_frame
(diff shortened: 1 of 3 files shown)


anijain2305 (approved):

anijain2305 (dismissed):
Wait this might only enable it for Aot Autograd. We need it for inductor as well

https://github.com/pytorch/torchdynamo/blob/main/torchinductor/compile_fx.py#L234

desertfire (approved):
LGTM

Chillee (commented):
Since fake tensor/functionalize are enabled in core, can we just delete these from here lol.

eellison:
> Since fake tensor/functionalize are enabled in core, can we just delete these from here lol. 
I still need to land that haha

PR #1240, opened 2022-09-15
add oneDNN graph backend for CPU

wenzhe-nrv:
This PR adds oneDNN graph backend to Dynamo. 
It is based on the oneDNN graph context manager API `fuser3` added here https://github.com/pytorch/pytorch/pull/82491

### DIFF
--- a/torchdynamo/optimizations/backends.py
+++ b/torchdynamo/optimizations/backends.py
@@ -103,7 +103,7 @@ def nvfuser_ofi(subgraph):
 
 
 @create_backend
-def onednn_graph(subgraph):
+def onednn(subgraph):
     with torch.jit.fuser("fuser3"):
         return reload_jit_model(subgraph)
 
--- a/torchdynamo/optimizations/backends.py
+++ b/torchdynamo/optimizations/backends.py
@@ -103,6 +103,12 @@ def nvfuser_ofi(subgraph):
 
 
 @create_backend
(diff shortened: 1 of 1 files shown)


jansel (approved):

torchdynamo/optimizations/backends.py
```diff
@@ -102,6 +102,12 @@ def nvfuser_ofi(subgraph):
         return reload_jit_model_ofi(subgraph)
 
 
+@create_backend
+def onednn_graph(subgraph):
```
jansel, line 106:
```suggestion
def onednn(subgraph):
```
Using the string `"onednn"` seems nicer than `"onednn_graph"`

wenzhe-nrv: Agree, that's nice without `_`. But `onednn_graph` is a bit different than onednn, thus using that might be miss leading. Let me think if there's a better naming.

wenzhe-nrv: good suggest. I'll confirm the naming with our team. Thanks.

jansel: The final call is up to you, happy to merge this name too.

wenzhe-nrv: @jansel I updated the name. Thanks for suggesting.

PR #1248, opened 2022-09-15
Update InductorTritonTemplateKernel.md

jspark1105:
torchinductor.config.triton.mm = "triton" to use the template?


Chillee (approved):
Makes sense.

By the way, I think @jansel had some thoughts on how the template API could be improved.

jansel:
Yeah, I want to remove the need for so much glue code to map args/strides.

jansel (approved):

PR #1250, opened 2022-09-16
Fail on using the context manager

anijain2305:
I have removed the context manager usage in PyTorch CI - https://github.com/pytorch/pytorch/pull/85124

Internally, there are 3 files where context manager is used. And I am talking to migrate to use annotation.

With this, we can remove the context manager.

### DIFF
--- a/torchdynamo/eval_frame.py
+++ b/torchdynamo/eval_frame.py
@@ -105,14 +105,10 @@ class _TorchDynamoContext:
 
     def __enter__(self):
         if config.raise_on_ctx_manager_usage:
-            raise RuntimeError("TorchDynamo is used with a context manager")
-        else:
-            log.warning(
-                (
-                    "Using TorchDynamo with a context manager will be deprecated soon."
-                    "Please read https://github.com/pytorch/torchdynamo#usage-example "
-                    "to use TorchDynamo using an annotation."
-                )
+            raise RuntimeError(
+                "torchdynamo.optimize(...) is used with a context manager. "
+                "Please refer to https://github.com/pytorch/torchdynamo#usage-example "
+                "to use torchdynamo.optimize(...) as an annotation/decorator. "
             )
         self.on_enter()
         self.prior = set_eval_frame(self.callback)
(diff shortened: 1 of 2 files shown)


voznesenskym (approved):

anijain2305:
@jansel Wondering if you are ok with this. We should be done with internal migration soon.

torchdynamo/eval_frame.py
```diff
@@ -104,15 +104,16 @@ def __init__(
...
-        log.warning(
-            (
-                "Using TorchDynamo with a context manager will be deprecated soon."
-                "Please read https://github.com/pytorch/torchdynamo#usage-example "
-                "to use TorchDynamo using an annotation."
-            )
-        )
         if config.raise_on_ctx_manager_usage:
             raise RuntimeError("TorchDynamo is used with a context manager")
+        else:
```
jansel, line 109: Should we just remove the warning now that raising is the default behavior?  If someone sets the config they should know what they are doing and the warning will just be annoying

torchdynamo/eval_frame.py
```diff
@@ -104,15 +104,16 @@ def __init__(
...
     def __enter__(self):
-        log.warning(
-            (
-                "Using TorchDynamo with a context manager will be deprecated soon."
-                "Please read https://github.com/pytorch/torchdynamo#usage-example "
-                "to use TorchDynamo using an annotation."
-            )
-        )
         if config.raise_on_ctx_manager_usage:
             raise RuntimeError("TorchDynamo is used with a context manager")
```
jansel, line 108: I'd suggest some more context in this error string that perhaps mentions the correct usage.

jansel (commented):
I am fine with this, a few comments.

torchdynamo/eval_frame.py
anijain2305: Will do.

torchdynamo/eval_frame.py
anijain2305: Yeah, thats correct. Will remove it.

PR #1252, opened 2022-09-16
Update CI skip list for HF and TIMM

desertfire:


anijain2305:
Is it ok if we just test models with batch_size = 2? Will that drop the coverage a lot?

jansel (approved):

desertfire:
> Is it ok if we just test models with batch_size = 2? Will that drop the coverage a lot?

That will be ok. I saw your PR has that change, so I will leave it to you.

PR #1259, opened 2022-09-16
Small changes so torchbench.py works in fbcode

jansel:


benchmarks/torchbench.py
```diff
@@ -7,18 +7,23 @@
...
 
 import torch
-from common import BenchmarkRunner
-from common import main
 
 from torchdynamo.testing import collect_results
 from torchdynamo.testing import reduce_to_scalar_loss
 from torchdynamo.utils import clone_inputs
 
+with patch.object(sys, "path", [abspath(dirname(__file__))] + sys.path):
```
voznesenskym, line 20: For fbcode, can we just fix the buck targets for this to get the right deps in place, instead of patching the sys path?

jansel: I don't know buck well enough to know how to do this.  You want to give it a try?

changing this to `.common` works in fbcode, by not in oss.

bertmaher (approved):
While I feel like it should be possible to fix TARGETS so this works without the conditional logic, I'm not sure it's worth fiddling with it now :-p

jansel:
> While I feel like it should be possible to fix TARGETS so this works without the conditional logic, I'm not sure it's worth fiddling with it now :-p

Yeah, if this weren't just testing code I would have invested more into it.

Issue #1189, opened 2022-09-10
torchbench models hf_Bert, hf_GPT2_large failing with `A view was created in no_grad mode...`


davidberard98:
hf_Bert (and hf_GPT2_large) fail with this error:

```
Traceback (most recent call last):
  File "run.py", line 271, in <module>
    run_one_step(test, model_flops=model_flops, model=m, export_dcgm_metrics_file=export_dcgm_metrics_file, stress=args.stress)
  File "run.py", line 61, in run_one_step
    func()
  File "/fsx/users/dberard/scratch-local/bench-fast/benchmark/torchbenchmark/util/model.py", line 188, in invoke
    self.train()
  File "/data/home/dberard/miniconda/envs/bench-fast/lib/python3.8/site-packages/torchdynamo/eval_frame.py", line 166, in _fn
    return fn(*args, **kwargs)
  File "/fsx/users/dberard/scratch-local/bench-fast/benchmark/torchbenchmark/util/framework/huggingface/model_factory.py", line 111, in train
    def train(self):
  File "/data/home/dberard/miniconda/envs/bench-fast/lib/python3.8/site-packages/torchdynamo/eval_frame.py", line 166, in _fn
    return fn(*args, **kwargs)
  File "/data/home/dberard/miniconda/envs/bench-fast/lib/python3.8/site-packages/functorch/_src/aot_autograd.py", line 889, in forward
    return compiled_f(
  File "/data/home/dberard/miniconda/envs/bench-fast/lib/python3.8/site-packages/functorch/_src/aot_autograd.py", line 880, in new_func
    return compiled_fn(args)
  File "/data/home/dberard/miniconda/envs/bench-fast/lib/python3.8/site-packages/functorch/_src/aot_autograd.py", line 228, in g
    return f(*args)
RuntimeError: A view was created in no_grad mode and its base or another view of its base has been modified inplace with grad mode enabled. Given that this use case is ambiguous and error-prone, it is forbidden. You can clarify your code by moving both the view and the inplace either both inside the no_grad block (if you don't want the inplace to be tracked) or both outside (if you want the inplace to be tracked).
```

This issue is probably in the optimizer, because commenting out `optimizer.step()` causes the error to go away. https://github.com/pytorch/benchmark/blob/main/torchbenchmark/util/framework/huggingface/model_factory.py#L115

Repro: run torchbench with https://github.com/pytorch/benchmark/pull/1173:
```
[...]

PR #1260, opened 2022-09-17
[Bytecode Transformation] Add missing torch.set_grad_enable calls in Fx Graphs

anijain2305:
On the main branch, we put set_grad_enabled calls into a ContextWrappingVariable

https://github.com/pytorch/torchdynamo/blob/41c44bc1d080d6cf063419a4166732b983b84eef/torchdynamo/variables/torch.py#L161-L166

However, `set_grad_enabled` can be called w/o context manager as well. And, the test added in this PR fails.

To do this, we explicitly call `enter` of GradModeVariable on seeing a `POP_TOP`

### DIFF
--- a/test/test_misc.py
+++ b/test/test_misc.py
@@ -516,15 +516,15 @@ class MiscTests(torchdynamo.testing.TestCase):
             return x
 
         with torch.no_grad():
-            torchdynamo.testing.standard_test(self, fn=fn1, nargs=2, expected_ops=3)
-            torchdynamo.testing.standard_test(self, fn=fn2, nargs=2, expected_ops=3)
+            torchdynamo.testing.standard_test(self, fn=fn1, nargs=2, expected_ops=5)
+            torchdynamo.testing.standard_test(self, fn=fn2, nargs=2, expected_ops=5)
             torchdynamo.testing.standard_test(self, fn=fn3, nargs=2, expected_ops=5)
             torchdynamo.testing.standard_test(self, fn=fn4, nargs=2, expected_ops=5)
         with torch.enable_grad():
             torchdynamo.testing.standard_test(self, fn=fn1, nargs=2, expected_ops=5)
             torchdynamo.testing.standard_test(self, fn=fn2, nargs=2, expected_ops=5)
-            torchdynamo.testing.standard_test(self, fn=fn3, nargs=2, expected_ops=3)
-            torchdynamo.testing.standard_test(self, fn=fn4, nargs=2, expected_ops=3)
+            torchdynamo.testing.standard_test(self, fn=fn3, nargs=2, expected_ops=5)
+            torchdynamo.testing.standard_test(self, fn=fn4, nargs=2, expected_ops=5)
 
     def test_build_tuple_unpack(self):
         def fn1(a, b, c):
--- a/test/test_misc.py
+++ b/test/test_misc.py
@@ -2378,6 +2378,27 @@ class MiscTests(torchdynamo.testing.TestCase):
             g1(torch.randn(10), torch.randn(10))
             self.assertEqual(count_graph_break_msgs(log.output), 1)
 
+    def test_inplace_param_update(self):
+        def fn(param, y):
+            prev_grad = torch.is_grad_enabled()
+            try:
+                torch.set_grad_enabled(False)
+                torch.set_grad_enabled(True)
+                torch.set_grad_enabled(False)
+                param.add_(y)
+            finally:
+                torch.set_grad_enabled(prev_grad)
+
+        y = torch.randn(4)
+        x = torch.nn.Parameter(torch.randn(4))
+        fn(x, y)
+
+        cnts = torchdynamo.testing.CompileCounter()
+        opt_fn = torchdynamo.optimize(cnts, nopython=True)(fn)
+        opt_fn(x, y)
+        self.assertEqual(cnts.frame_count, 1)
+        self.assertEqual(cnts.op_count, 5)
+
 
 class TestTracer(JitTestCase):
     def test_jit_save(self):
--- a/torchdynamo/symbolic_convert.py
+++ b/torchdynamo/symbolic_convert.py
@@ -1243,7 +1243,6 @@ class InstructionTranslatorBase(object):
         self.output.guards.add(
             GlobalWeakRefSource(name).create_guard(GuardBuilder.WEAKREF_ALIVE)
         )
-        # self.f_globals[name] = weakref.ref(value)
         if name not in self.output.root_globals:
             self.output.install_global(name, weakref.ref(value))
 
--- a/torchdynamo/symbolic_convert.py
+++ b/torchdynamo/symbolic_convert.py
@@ -604,7 +604,14 @@ class InstructionTranslatorBase(object):
         self.push(None)
 
     def END_FINALLY(self, inst):
-        assert self.pop() is None
+        tos = self.pop()
+        if sys.version_info < (3, 8):
+            # python3.7 and 3.8 can have END_FINALLY without BEGIN_FINALLY
+            assert tos is None or (
+                tos.is_python_constant() and tos.as_python_constant() is None
+            )
+        else:
+            assert tos is None
 
     def FOR_ITER(self, inst):
         it = self.pop()
@@ -1236,7 +1243,9 @@ class InstructionTranslatorBase(object):
         self.output.guards.add(
             GlobalWeakRefSource(name).create_guard(GuardBuilder.WEAKREF_ALIVE)
         )
-        self.f_globals[name] = weakref.ref(value)
+        # self.f_globals[name] = weakref.ref(value)
+        if name not in self.output.root_globals:
+            self.output.install_global(name, weakref.ref(value))
 
     @property
     def fake_mode(self):
--- a/torchdynamo/variables/misc.py
+++ b/torchdynamo/variables/misc.py
@@ -277,8 +277,6 @@ class GradModeVariable(ContextWrappingVariable):
         return variables.ConstantVariable(None, **VariableTracker.propagate(self))
 
     def _call_func(self, tx, value):
-        if self.target_value == self.initial_value:
-            return
         tx.output.graph.create_node(
             "call_function", torch._C._set_grad_enabled, (value,), {}
         ),
--- a/torchdynamo/variables/misc.py
+++ b/torchdynamo/variables/misc.py
@@ -262,14 +262,19 @@ class ContextWrappingVariable(ContextManagerVariable):
 
 
 class GradModeVariable(ContextWrappingVariable):
+    @staticmethod
+    def create(tx, target_value, initial_value=None, **kwargs):
+        var = GradModeVariable(target_value, initial_value, **kwargs)
+        var._call_func(tx, target_value)
+        return var
+
     def __init__(self, target_value, initial_value=None, **kwargs):
         super(GradModeVariable, self).__init__(
             target_value=target_value, initial_value=initial_value, **kwargs
         )
 
     def enter(self, tx):
-        assert self.initial_value == torch.is_grad_enabled()
-        return super(GradModeVariable, self).enter(tx)
+        return variables.ConstantVariable(None, **VariableTracker.propagate(self))
 
     def _call_func(self, tx, value):
         if self.target_value == self.initial_value:
--- a/test/test_subgraphs.py
+++ b/test/test_subgraphs.py
@@ -428,8 +428,7 @@ class SubGraphTests(torchdynamo.testing.TestCase):
         self._common(fn, 2, 9)
         torchdynamo.reset()
         with torch.no_grad():
-            # fewer ops (no disable grad) if already disabled
-            self._common(fn, 2, 5)
+            self._common(fn, 2, 9)
 
     def test_resume_with_no_grad2(self):
         def fn(a, b):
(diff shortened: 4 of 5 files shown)


voznesenskym:
If grad is changed in the middle of the module, does this PR capture a call to torch to enable/disable in the graph as an fx node with a call_function? Or does it just set the global state for tracing until it changes again/ exits?

anijain2305:
> If grad is changed in the middle of the module, does this PR capture a call to torch to enable/disable in the graph as an fx node with a call_function? Or does it just set the global state for tracing until it changes again/ exits?

This PR is more about the former. The resulting graphs will have enable/disable in the Fx graph. TorchDynamo already puts them in the Fx graph for context managers (`with torch.no_grad`, `with torch.enabled_grad` etc), but is not putting it when used standalone like `torch.set_grad_enabled(True)`.

voznesenskym:
> > If grad is changed in the middle of the module, does this PR capture a call to torch to enable/disable in the graph as an fx node with a call_function? Or does it just set the global state for tracing until it changes again/ exits?
> 
> This PR is more about the former. The resulting graphs will have enable/disable in the Fx graph. TorchDynamo already puts them in the Fx graph for context managers (`with torch.no_grad`, `with torch.enabled_grad` etc), but is not putting it when used standalone like `torch.set_grad_enabled(True)`.

Yes, that is what I thought. Can we put them in the graph? We should be able to identify this as a `call_function` on a TorchVariable as its a torch.* call, right? We can then:

1) Create a `call_function` proxy to `torch.set_grad_enabled(val)`
2) Set the global state for the tracing as you do in this PR. 

Thoughts?

voznesenskym:
> > If grad is changed in the middle of the module, does this PR capture a call to torch to enable/disable in the graph as an fx node with a call_function? Or does it just set the global state for tracing until it changes again/ exits?
> 
> 
> 
> This PR is more about the former. The resulting graphs will have enable/disable in the Fx graph. TorchDynamo already puts them in the Fx graph for context managers (`with torch.no_grad`, `with torch.enabled_grad` etc), but is not putting it when used standalone like `torch.set_grad_enabled(True)`.

I think I misunderstood your comment here :) if this PR already writes them into the fx graph, then that's perfect.

test/test_misc.py
```diff
@@ -2378,6 +2378,25 @@ def count_graph_break_msgs(msgs):
             g1(torch.randn(10), torch.randn(10))
             self.assertEqual(count_graph_break_msgs(log.output), 1)
 
+    def test_inplace_param_update(self):
+        def fn(param, y):
+            prev_grad = torch.is_grad_enabled()
+            try:
+                torch.set_grad_enabled(False)
```
voznesenskym, line 2385: Can we test setting the same value twice?

test/test_misc.py
```diff
@@ -2378,6 +2378,25 @@ def count_graph_break_msgs(msgs):
             g1(torch.randn(10), torch.randn(10))
             self.assertEqual(count_graph_break_msgs(log.output), 1)
 
+    def test_inplace_param_update(self):
```
voznesenskym, line 2381: Would you open to adding a test for export with this?

voznesenskym (approved):

anijain2305:
Oh, this is exactly what this PR does. This PR calls the `enter` method of `GradModeVariable` which is implemented here 

https://github.com/pytorch/torchdynamo/blob/d2c4fccbfd55d252e84838f466f7054ba04f272e/torchdynamo/variables/misc.py#L274-L280

So, this PR does not do anything fancy. We were already calling this function for SETUP_WITH here 

https://github.com/pytorch/torchdynamo/blob/d2c4fccbfd55d252e84838f466f7054ba04f272e/torchdynamo/symbolic_convert.py#L584

But, we were not calling it when `torch.set_grad_enabled` is called without any context manager. So, this PR calls it when POP_TOP is called and the TOS is a GradModeVariable.

However, after writing this, I think we should unite the two and implement this in CALL_FUNCTION instead of `SETUP_WITH` and `POP_TOP`. Let me check.

torchdynamo/symbolic_convert.py
```diff
@@ -994,7 +1001,11 @@ def NOP(self, inst):
         pass
 
     def POP_TOP(self, inst):
-        self.pop()
+        val = self.pop()
+        if isinstance(val, GradModeVariable):
```
jansel, line 1005: I'd expect this to fail on:
```
tmp = torch.set_grad_enabled(True)
```

POP_TOP only happens if the user doesn't store the result.

anijain2305: Should we then then effectively call `GradModeVariable` `enter` method on `CALL_METHOD` or `CALL_FUNCTION`?

voznesenskym: Why does it have to be from the bytecode of CALL_METHOD or CALL_FUNCTION? I haven't run it, but this should already be a `call_function` on a `TorchVariable` right? The problem is that it goes down into GradModeVariable's ctor, which is a no-op, and GradModeVariable itself is a ContextWrappingVariable, whereas this is not a context.

The right fix here would be to make a helper util like the one I described above, where we create:

```
 node = tx.output.graph.create_node(
            "call_function", torch._C._set_grad_enabled, (value,), {}
        )
```

And wrap it in a VariableTracker.

voznesenskym (changes requested):
RC after @jansel comments. See my comment above, I think we can do this via just `tx.output.graph.create_node`

torchdynamo/symbolic_convert.py
anijain2305: Yes, that sounds good. I meant something along this lines as well. Basically put the grad state op in Fx Graph during the symbolic convesion of call_function (which is TorchVariable in this case), instead of delaying it to `SETUP_WITH` or `POP_TOP`.

test/test_misc.py
anijain2305: Added

anijain2305:
cc @jansel @voznesenskym Can you please review again?

anijain2305:
cc @jansel @voznesenskym Can you please review again?

jansel (approved):

torchdynamo/symbolic_convert.py
```diff
@@ -1236,7 +1243,9 @@ def store_dict_key(self, name, value):
         self.output.guards.add(
             GlobalWeakRefSource(name).create_guard(GuardBuilder.WEAKREF_ALIVE)
         )
-        self.f_globals[name] = weakref.ref(value)
+        # self.f_globals[name] = weakref.ref(value)
```
voznesenskym, line 1246: rm comment

torchdynamo/variables/misc.py
```diff
@@ -262,14 +262,19 @@ def call_function(
 
 
 class GradModeVariable(ContextWrappingVariable):
+    @staticmethod
+    def create(tx, target_value, initial_value=None, **kwargs):
+        var = GradModeVariable(target_value, initial_value, **kwargs)
```
voznesenskym, line 267: initial value is really weird, because its None here, and then used to early return out of _call_func, but `_ initial_value()` just gets the current state of `torch.is_grad_enabled()` Which gets called in the super through:

```
if initial_value is None:
            initial_value = self._initial_value()
        self.initial_value = initial_value
```

And we never actually set it.

Can we just:

1) remove it from the ctor of this class
2) Fix the `ContextWrappingVariable` ctor to not call self._initial_value()
3) Instead have `ContextWrappingVariable` just take a value

It also look like its why we have a ContextManagerVariable and a ContextWrappingVariable - two confusing names, one of which is the base (ContextManagerVariable) which acts as a past-through for the of `FakeContextWrappingVariable` (as it cannot call super, due to not having an initial value).

Other nits:

- FakeContextWrappingVariable is too tightly coupled to `"autograd.profiler.profile"` for its name 

- ContextWrappingVariable being used for something that is not a context (set grad enabled in this case)

torchdynamo/variables/misc.py
```diff
@@ -262,14 +262,19 @@ def call_function(
...
             target_value=target_value, initial_value=initial_value, **kwargs
         )
 
     def enter(self, tx):
-        assert self.initial_value == torch.is_grad_enabled()
-        return super(GradModeVariable, self).enter(tx)
+        return variables.ConstantVariable(None, **VariableTracker.propagate(self))
 
     def _call_func(self, tx, value):
         if self.target_value == self.initial_value:
```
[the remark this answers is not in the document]
anijain2305: I am seeing the calls at the right place. For this code

~~~

def fn(param, y):
    prev_grad = torch.is_grad_enabled()
    try:
        torch.set_grad_enabled(False)
        param.add_(y)
    finally:
        torch.set_grad_enabled(prev_grad)

~~~

I am seeing this Fx graph

~~~
 __compiled_fn_0 <eval_with_key>.1 opcode         name                 target                                 args        kwargs
-------------  -------------------  -------------------------------------  ----------  --------
placeholder    param                param                                  ()          {}
placeholder    y                    y                                      ()          {}
call_function  _set_grad_enabled    <built-in function _set_grad_enabled>  (False,)    {}
call_method    add_                 add_                                   (param, y)  {}
call_function  _set_grad_enabled_1  <built-in function _set_grad_enabled>  (True,)     {}
output         output               output                                 ((),)       {}
~~~


These calls are inserted in the static helper function added in this PR which calls `self._call_func` which is implemented here - 


https://github.com/pytorch/torchdynamo/blob/f8a46decc66d99cc1f3104214f4099d6493d4b4a/torchdynamo/variables/misc.py#L274-L280

voznesenskym:
```
import torch
import torchdynamo

class MyModule(torch.nn.Module):
    def forward(self, x):
        torch.set_grad_enabled(True)
        y = torch.sin(x) + torch.tensor([0.5], requires_grad=True)
        return y * y


module = MyModule()
graph, guards = torchdynamo.export(module, torch.tensor([[0.0, 0], [0, 0]]))
print(graph)
torch.set_grad_enabled(False)
res = graph(torch.tensor([0.5]))
print("Grad required?", res.requires_grad)
```

If we do not capture that `torch.set_grad_enabled(True)` due to the *current* state of grad in torch, the exported graph will not contain `torch._C._set_grad_enabled(True)` and produce a different result for `requires_grad`

torchdynamo/variables/misc.py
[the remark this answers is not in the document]
voznesenskym: Because you went from True to False ;) See my comment below.

anijain2305:
> ```
> import torch
> import torchdynamo
> 
> class MyModule(torch.nn.Module):
>     def forward(self, x):
>         torch.set_grad_enabled(True)
>         y = torch.sin(x) + torch.tensor([0.5], requires_grad=True)
>         return y * y
> 
> 
> module = MyModule()
> graph, guards = torchdynamo.export(module, torch.tensor([[0.0, 0], [0, 0]]))
> print(graph)
> torch.set_grad_enabled(False)
> res = graph(torch.tensor([0.5]))
> print("Grad required?", res.requires_grad)
> ```
> 
> If we do not capture that `torch.set_grad_enabled(True)` due to the _current_ state of grad in torch, the exported graph will not contain `torch._C._set_grad_enabled(True)` and produce a different result for `requires_grad`

I see. I did not think about the export case. This could be fixed by cancelling the early out, and always setting the target_value.

voznesenskym:
> > ```
> > import torch
> > import torchdynamo
> > 
> > class MyModule(torch.nn.Module):
> >     def forward(self, x):
> >         torch.set_grad_enabled(True)
> >         y = torch.sin(x) + torch.tensor([0.5], requires_grad=True)
> >         return y * y
> > 
> > 
> > module = MyModule()
> > graph, guards = torchdynamo.export(module, torch.tensor([[0.0, 0], [0, 0]]))
> > print(graph)
> > torch.set_grad_enabled(False)
> > res = graph(torch.tensor([0.5]))
> > print("Grad required?", res.requires_grad)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > If we do not capture that `torch.set_grad_enabled(True)` due to the _current_ state of grad in torch, the exported graph will not contain `torch._C._set_grad_enabled(True)` and produce a different result for `requires_grad`
> 
> I see. I did not think about the export case. This could be fixed by cancelling the early out, and always setting the target_value.

yessir :) just remove those 2 lines.

torchdynamo/variables/misc.py
anijain2305: Ah, its a mess. We have incrementally added fixes for grad support, and code is messy now. We need a refactor.

Regarding the initial value, whats the key concern? I want to point out that we do need `initial_value` for the `exit` of the context manager. Initial value just keeps track of the initial torch grad state, and uses it during the `exit` call.

We do not really set the `initial_state` during the `enter` or `init` function, because we don't have to. For the export usecase, I would assume that there would be a guard that checks that the torch initial state is not different from the exported graph state.

torchdynamo/variables/misc.py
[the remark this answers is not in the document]
anijain2305: Yup, I got your point now. Thanks for the example, that made it clear. I have removed the two lines.

anijain2305:
cc @davidberard98 is it possible to try this PR if this fixes your error in https://github.com/pytorch/torchdynamo/issues/1189?fbclid=IwAR2MGOPKeRpTWCAtVsLuagSlWFGEfd_OBSkRU0a9p0uCX1hKTuaGhHFKMR4

voznesenskym (approved):
Thank you :)

anijain2305:
> Thank you :)

No, thank you! Thanks for catching an important corner case, I would have completely missed it.

PR #1261, opened 2022-09-17
[RFC] [WIP] Add export specializing escape hatch

voznesenskym:
This is a way to allow users to explicitly specialize their graphs where they know that the baked in value suffices.  We probably need https://github.com/pytorch/torchdynamo/pull/1143/files before this. CC: @suo as the POC for the request.

This is an early stage PR, put up to get feedback.

```
class MyModule2(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.param = torch.nn.Parameter(torch.rand(4, 2))
        self.linear = torch.nn.Linear(2, 2)

    def forward(self, x):
        res = {}
        y = torch.sin(x)
        with torchdynamo.specialize.specialize():
            y = torch.nonzero(self.linear(x))
        return y
```
`graph, guards = torchdynamo.export(module, torch.tensor([[0.0, 0], [0, 0]]))` Captures a whole graph where before we would graph break.

### DIFF
--- a/torchdynamo/config.py
+++ b/torchdynamo/config.py
@@ -7,6 +7,8 @@ from types import ModuleType
 
 import torch
 
+import torchdynamo.utils
+
 try:
     import torch._prims
     import torch._refs
@@ -21,12 +23,14 @@ except ImportError:
 # INFO print compiled functions + graphs
 # WARN print warnings (including graph breaks)
 # ERROR print exceptions (and what user code was being processed when it occurred)
+# NOTE: changing log_level will automatically update the levels of all torchdynamo loggers
 log_level = logging.WARNING
-# Verbose will print full stack traces on warnings and errors
-verbose = False
 
 # the name of a file to write the logs to
 log_file_name = None
+
+# Verbose will print full stack traces on warnings and errors
+verbose = False
 
 # verify the correctness of optimized backend
 verify_correctness = False
@@ -130,6 +134,9 @@ class _AccessLimitingConfig(ModuleType):
     def __setattr__(self, name, value):
         if name not in _allowed_config_names:
             raise AttributeError(f"{__name__}.{name} does not exist")
+        # automatically set logger level whenever config.log_level is modified
+        if name == "log_level":
+            torchdynamo.utils.set_loggers_level(value)
         return object.__setattr__(self, name, value)
 
 
--- a/torchdynamo/config.py
+++ b/torchdynamo/config.py
@@ -69,7 +69,7 @@ guard_nn_modules = False
 dynamic_propagation = True
 
 # Run the FX graph with FakeTensors
-fake_tensor_propagation = True
+fake_tensor_propagation = False
 
 # run FX normalization passes in optimizer
 normalize_ir = False
--- a/torchdynamo/eval_frame.py
+++ b/torchdynamo/eval_frame.py
@@ -91,6 +91,7 @@ class _TorchDynamoContext:
         self,
         callback,
         on_enter=nothing,
+        on_exit=nothing,
         backend_ctx_ctor=null_context,
         patch_fn=nothing,
     ):
@@ -99,6 +100,7 @@ class _TorchDynamoContext:
         self.callback = callback
         self.prior = unset
         self.on_enter = on_enter
+        self.on_exit = on_exit
         self.extra_ctx_ctor = backend_ctx_ctor
         patch_fn()
 
@@ -121,6 +123,7 @@ class _TorchDynamoContext:
         set_eval_frame(self.prior)
         self.prior = unset
         self.backend_ctx.__exit__(exc_type, exc_val, exc_tb)
+        self.on_exit()
 
     def __call__(self, fn):
         fn = innermost_fn(fn)
@@ -187,8 +190,9 @@ class _TorchDynamoContext:
 
 
 class OptimizeContext(_TorchDynamoContext):
-    def __init__(self, callback, backend_ctx_ctor):
+    def __init__(self, callback, backend_ctx_ctor, pre_on_enter=nothing, post_on_exit=nothing):
         def on_enter():
+            pre_on_enter()
             global most_recent_backend
             if (
                 most_recent_backend is not None
@@ -197,11 +201,17 @@ class OptimizeContext(_TorchDynamoContext):
                 raise ResetRequired()
             most_recent_backend = compiler_fn
             install_generation_tagging_init()
+        
+        def on_exit():
+            # Could just pass post_on_exit through, but this reads much clearer
+            post_on_exit()
 
         compiler_fn = innermost_fn(callback)
+        print("CALLBACK IS", callback)
         super().__init__(
             callback=callback,
             on_enter=on_enter,
+            on_exit=on_exit,
             backend_ctx_ctor=backend_ctx_ctor,
             patch_fn=TorchPatcher.patch,
         )
@@ -215,6 +225,29 @@ class RunOnlyContext(_TorchDynamoContext):
 class DisableContext(_TorchDynamoContext):
     def __init__(self):
         super().__init__(callback=None)
+
+
+specializing = False
+
+
+class SpecializeContext(OptimizeContext):
+# class SpecializeContext:
+    def __init__(self):
+        def on_enter():
+            global specializing
+            assert not specializing, "Torchdynamo specialization cannot be nested yet."
+            specializing = True
+
+        def on_exit():
+            global specializing
+            specializing = False
+            
+        super().__init__(
+            callback=callback,
+            backend_ctx_ctor=backend_ctx_ctor,
+            pre_on_enter=on_enter,
+        )
+        # specializing = prior
 
 
 def catch_errors_wrapper(callback):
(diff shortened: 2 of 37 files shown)


torchdynamo/config.py
```diff
@@ -69,7 +69,7 @@
 dynamic_propagation = True
 
 # Run the FX graph with FakeTensors
-fake_tensor_propagation = True
+fake_tensor_propagation = False
```
voznesenskym, line 72: will not land

soumith:
i think it's poorly named. you probably want to think of a more intuitive one.
`with torchdynamo.export_raw()` or something

voznesenskym:
> i think it's poorly named. you probably want to think of a more intuitive one. `with torchdynamo.export_raw()` or something

Agreed, but I am not sure export_raw is it either... perhaps something like `.remove_dynamism` or `.inline_result`? @suo thoughts?

suo:
Ummm…elsewhere we use "concrete" to describe this kind of behavior (e.g. `concrete_args`) so maybe `torchdynamo.concrete_graph()`?

voznesenskym:
> Ummm…elsewhere we use "concrete" to describe this kind of behavior (e.g. `concrete_args`) so maybe `torchdynamo.concrete_graph()`?

Sure, that's fine by me.

soumith:
i would go with `.inline_graph` or `.remove_dynamism` in that order.
`concrete_graph` is alien to non-compiler folks

suo:
in my defense, `concrete_args` is a named argument in both jax and FX APIs!

but you are probably right lol

suo:
I lied, it's called `static_argnums` in jax :P

torchdynamo/eval_frame.py
```diff
@@ -188,21 +192,31 @@ def _fn(*args, **kwargs):
...
                 and most_recent_backend is not compiler_fn
             ):
                 raise ResetRequired()
             most_recent_backend = compiler_fn
+            global most_recent_context
+            most_recent_context = backend_ctx_ctor
             install_generation_tagging_init()
 
+        def on_exit():
+            # Could just pass post_on_exit through, but this reads much clearer
```
jansel, line 212: Let's remove this extra wrapper.  It will just add overhead.

torchdynamo/skipfiles.py
```diff
@@ -154,6 +162,8 @@ def check(filename, allow_torch=False):
         return True
     if filename in FILENAME_ALLOWLIST:
         return False
+    if filename in lazy_allowlist():
```
jansel, line 165: Perhaps refactor the existing list to be lazy + cached rather than adding a second list.

jansel (changes requested):
I don't fully understand the semantics of this.  Some examples in docs and tests would help.  What gets specialized?  What invariants do we maintain?  How do we check for correctness?

Is this just for handling nonzero? If so I feel like just adding a new op like:
```
@torchdynamo.allow_in_graph
def nonzero_fixed_size(..., output_shape=...):
   rv = nonzero(...)
   assert rv.shape == output_shape
   return rv
```

torchdynamo/skipfiles.py
voznesenskym: sg

torchdynamo/eval_frame.py
voznesenskym: sg

voznesenskym:
> I don't fully understand the semantics of this. Some examples in docs and tests would help. What gets specialized? What invariants do we maintain? How do we check for correctness?

Yeah, agreed. I can add that.

> Is this just for handling nonzero? If so I feel like just adding a new op like:
> 
> ```
> @torchdynamo.allow_in_graph
> def nonzero_fixed_size(..., output_shape=...):
>    rv = nonzero(...)
>    assert rv.shape == output_shape
>    return rv
> ```

It's for any part of the graph the user sees fit to specialize on, not just nonzero - any torch call is valid. @suo and I spoke about just having a dynamo nonzero op, but two things came up:

1) Theres a lot of other ops we may want to specialize on in this way
2) If we do go with doing it op by op, we probably want to do it in pytorch itself, with those ops explicitly acting as a specialized version of the original op. (perhaps something like nonzero.specialized vs nonzero.default?) 

Agreed this needs a lot more docs and tests and explanations.

suo:
To chime in on the use case: it's not just for nonzero, there are many cases in export where the user "knows more" than dynamo does about the code and would benefit from a escape hatch.

As another example, quantized modules store whether fake quantization is enabled [on a torch.tensor](https://github.com/pytorch/pytorch/blob/09be44de7b56495bcb5ad1d47376200cbb853097/torch/ao/quantization/fake_quantize.py#L49-L52). It's frequently used in control flow (and thus is a graph break as we are branch on value), but is static in practice.

soumith:
> As another example, quantized modules store whether fake quantization is enabled [on a torch.tensor](https://github.com/pytorch/pytorch/blob/09be44de7b56495bcb5ad1d47376200cbb853097/torch/ao/quantization/fake_quantize.py#L49-L52). It's frequently used in control flow (and thus is a graph break as we are branch on value), but is static in practice.

Wouldn't the right thing to do be to first export, and then run a pass to erase the guards around the value of `fake_quant_enabled`?

suo:
Today, we don't guard on tensor values, only metadata. So something that does, e.g. `if self.fake_quant_enabled[0] == 1` is a graph break.

Conceivably this could be a guard, but there are complications to doing guards on value in general (guarding on value can be expensive, it can can lead to large number of recompilations, floating point comparisons ruin things, etc.).

jansel:
I'm not convinced a context manager is the right UX for this.  I also don't like the name specialized, it is not clear to me (even after reading the PR) exactly what is getting specialized and what isn't.  I think it will be confusing to reason about the correctness.

I'd also like to try to make whatever we do have meaningful semantics in eager mode.  If we are adding some constraint (like nonzero returns a static shape), I'd prefer that constraint become an assert in eager mode.  This will make there be no divergence between eager/dynamo, if you use it incorrectly you get an error in both.

I think `torchdynamo.allow_in_graph()` already gives a good enough escape hatch for stuff like `nonzero()`.  Just define a custom wrapper, like the example above, which should already work.

For things like specializing on the value of a tensor (e.g. `if self.fake_quant_enabled[0] == 1`), perhaps we need a new decorator like `@torchdynamo.assume_constant_result`.  Any function with that decorator would get evaluated once at compile time and whatever it returns would get burned into the graph as a constant.  I don't love this, but I feel like it is easier to reason about what it does.

suo:
We should separate the discussion about `dynamic_output_shape` ops (like `nonzero`) from the discussion of whether we want a specializing escape hatch.

For the `dynamic_output_shape` ops: I think it makes sense to add equivalent static-output-shape ops for each of these. There are only 15 in `native_functions.yaml`, so it's not that wild. And there are unique benefits to doing this (memory planning becomes easier because you don't have data-dependent shapes running around).

The ops thing doesn't cover all use cases (e.g. the stupid fake quant thing); so export will also need a more general-purpose escape hatch as well. For this I'm fine with the `@torchdynamo.assume_constant_result` idea. It's a little less user-friendly but does provider nicer boundaries for us and keeps the rules easier to explain.

voznesenskym:
> We should separate the discussion about `dynamic_output_shape` ops (like `nonzero`) from the discussion of whether we want a specializing escape hatch.
> 
> For the `dynamic_output_shape` ops: I think it makes sense to add equivalent static-output-shape ops for each of these. There are only 15 in `native_functions.yaml`, so it's not that wild. And there are unique benefits to doing this (memory planning becomes easier because you don't have data-dependent shapes running around).
> 
> The ops thing doesn't cover all use cases though (e.g. the stupid fake quant thing); so export will also need a more general-purpose escape hatch as well. For this I'm fine with the `@torchdynamo.assume_constant_result` idea. It's a little less user-friendly but does provider nicer boundaries for us and keeps the rules easier to explain.

I'm fine with adding static-output-shape ops at the torch level, and then a `assume_constant_result`, but what does it provide over this? The whole point of this is to (A) only use it during export (B) provide a graph that trades off on correctness and enforces that via guards. 

> I don't love this, but I feel like it is easier to reason about what it does.

I don't love it either - open to suggestions though, other than this escape hatch and `assume_constant_result`, what else can we offer?

jansel:
For the case of `fake_quant_enabled`, can we just use a normal python bool?  (Which should already work.)  Is using a tensor here some artifact of a prior TorchScript hack?

This doesn't seem like a very natural way to write a program.

suo:
I'm not sure. The quantization code references making it a buffer to work with NCCL replication during DDP. It wouldn't be for TorchScript, which supports bools. Changing it would be backwards incompatible for any model using eager quantization though (as the buffer is in a PyTorch core module). And the "buffer-as-bool" pattern is fairly common because it ensures the boolean will be represented in the `state_dict`, so there's a user tradeoff to changing it to a Python bool.

Big picture—this is only one of many different kinds of issues we will see on models we want to export. I don't think it's practical to try to restructure the code in all instances vs. providing a pragmatic escape hatch.

jansel:

PR #1262, opened 2022-09-17
[TIMM] Use pretrained models and deterministic inputs to help CI

anijain2305:
As title

Removed the models for which we TIMM does not have pretrained models. This will reduce flakiness in CI as we move to accuracy testing.


desertfire (commented):
Do we want to keep these models for the perf testing part?

anijain2305:
> Do we want to keep these models for the perf testing part?

@desertfire I think it's ok to skip them. There are in total 69 models in TIMM (which is already pretty large number), and there is always another model in the category for the above skipped models. So, I think its ok to skip.

desertfire (approved):

PR #1265, opened 2022-09-18
Expand CI to include aot-eager for HF and TIMM

desertfire:


desertfire:
@anijain2305 , there are many accuracy failures for aot-eager. In our performance dashboard, I know we are ignoring accuracy checking for inductor, but are we doing the same for aot-eager? Because the pass rate for aot-eager in the dashboard is definitely larger than what I am seeing here.

anijain2305:
@desertfire This is unexpected. It could be from fake tensors. Could you try once with fake_tensors OFF by setting `functorch.compile.config.use_fake_tensor = False` in the repo?

So many variations in eager is also not expected (while OOMs are expected). Lets try to get other issues resolved before we merge this PR.

desertfire:
> IMO it is better to stem the bleeding (land extra CI first with xfails) and then fix them after, since without the CI you're blind to further regressions on the non-CI components.

Agree with that. The reason I am giving it another round of run is some of the failures are not reproducible locally, but looks true after SSH to CI machine and rerun. I am trying to see if Animesh's recent pretrain changes to TIMM would help to reduce some of those inconsistency.

desertfire:
@anijain2305 , should we get this in?

.circleci/config.yml
```diff
@@ -84,27 +84,131 @@ jobs:
...
+    resource_class: gpu.nvidia.large
+    steps:
+      - checkout
+      - install_deps
+      - run:
+          name: Huggingface AotAutograd Eager training run
+          command: |
+            source .circleci/setup_env.sh
+            make develop
+            python benchmarks/huggingface.py --accuracy --ci --training --use-eval-mode --amp --backend=aot_eager -d cuda --output=aot_eager_hf_training.csv
```
anijain2305, line 183: You can remove `--use-eval-mode`. `--accuracy` sets it up.

.circleci/config.yml
```diff
@@ -114,11 +218,11 @@ jobs:
       - checkout
       - install_deps
       - run:
-          name: TorchBench inference run
+          name: TorchBench TorchInductor inference run
           command: |
             source .circleci/setup_env.sh
             make develop
-            python benchmarks/torchbench.py --accuracy --ci --skip-accuracy-check -d cuda --inductor --float32 --output=inductor_torchbench_inference.csv
+            python benchmarks/torchbench.py --accuracy --ci -d cuda --inductor --float32 --output=inductor_torchbench_inference.csv
```
anijain2305, line 225: Oh nice!

anijain2305 (approved):

.circleci/config.yml
desertfire: Good to know. Let me do that next time when I update this file.

PR #1271, opened 2022-09-19
Two small fixes found onboarding RWKV

voznesenskym:


torchdynamo/variables/nn_module.py
```diff
@@ -115,11 +115,15 @@ def var_getattr(self, tx, name):
 
         if name in base_dict:
             subobj = base_dict[name]
-        elif name in base_dict["_modules"] and name not in all_class_attribute_names:
```
voznesenskym, line 118: JIT-ed NN Module do not have these attr, so instead of a keyerror we shuffle on to let inspect. getattr_static handle it (or fail with a better message).

torchinductor/lowering.py
```diff
@@ -1321,7 +1321,9 @@ def full_like(x, fill_value, **kwargs):
 
 def tensor_constructor(fill_value):
     # torch.zeros, torch.ones, etc
-    def inner(*size, dtype=None, device=None, layout=0, pin_memory=False):
+    def inner(
+        *size, dtype=None, device=None, layout=0, pin_memory=False, memory_format=None
```
voznesenskym, line 1325: Unused arg, but allows construction of tensors like `y = torch.empty((B, T, C), device='cuda', memory_format=torch.contiguous_format)`

jansel (approved):

torchinductor/lowering.py
ngimel: We should probably check that it's `contiguous`, otherwise we'd silently ignore things like `channels_last`?

voznesenskym: We don't use the value, do you want me to assert that it is contiguous_format?

ngimel: Yes, assert that it's either None or contiguous_format (similarly to how we do with pin_memory and layout)

PR #584, opened 2022-07-15
Introduce prims+nvfuser backend

SherlockNoMad:
As title.

To run with prims+nvfuser backend: 

$ python torchbench.py --training --isolate **--prims-nvfuser** --devices=cuda

PR #1281, opened 2022-09-20
PrimTorch + nvFuser backend

IvanYashchuk:
This PR adds a new AOT Autograd-based backend with the "--nvprims-nvfuser" cli argument for benchmarks.

### DIFF
--- a/benchmarks/common.py
+++ b/benchmarks/common.py
@@ -1402,6 +1402,12 @@ def parse_args():
         "--prims-nvfuser", action="store_true", help="user prims + nvfuser backend"
     )
     parser.add_argument(
+        "--nvprims-nvfuser", action="store_true", help="nvprims + nvfuser backend"
+    )
+    parser.add_argument(
+        "--nvprims-aten", action="store_true", help="nvprims + aten backend"
+    )
+    parser.add_argument(
         "--dump-raw-metrics",
         action="store_true",
         help="dump raw timing metrics from speedup experiment",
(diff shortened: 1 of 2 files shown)


benchmarks/common.py
```diff
@@ -1401,6 +1401,12 @@ def parse_args():
     parser.add_argument(
         "--prims-nvfuser", action="store_true", help="user prims + nvfuser backend"
     )
+    parser.add_argument(
```
anijain2305, line 1404: Nit - We can directly use `--backend=nvprims_nvfuser`, if you didn't know about this option.

IvanYashchuk: I didn't know! I will try and remove this code then.

torchdynamo/optimizations/training.py
```diff
@@ -431,6 +471,12 @@ def create_aot_backends():
     # directly lowers to NVFuser without relying no Torchscript.
     BACKENDS["prims_nvfuser"] = aot_prims_nvfuser
```
jansel, line 472: Is this backend still needed?  What is the difference with `nvprims_nvfuser`?

IvanYashchuk: It's not needed for our team and there are no plans to use it or recommend anyone to use it. It wasn't me who added this, so I'm not going to touch it.

"prims_nvfuser" was added in https://github.com/pytorch/torchdynamo/pull/584 and it makes a decision of translating aten->prims based on a static outdated and not maintained table from https://github.com/pytorch/pytorch/blob/09965957cd8ecc696852e73022892b3ad4475783/torch/fx/passes/backends/nvfuser.py#L70-L71
The performance of this backend is terrible, on the order of 5% of eager mode's performance.

"nvprims_nvfuser" decomposes aten ops into _nvprims_, a subset of prims that are guaranteed to be executable by nvFuser, and the decomposition is used only if given aten op is fully decomposable into nvprims.

jansel: Let's just remove it if you don't recommend anyone use it.

cc @SherlockNoMad

IvanYashchuk:
@anijain2305 I reverted changes to `benchmarks/common.py` to prefer `--backend=nvprims_nvfuser` argument. Other than removing "prims_nvfuser" that I'm still hesitant to do (https://github.com/pytorch/torchdynamo/pull/1281#discussion_r977379888) this PR is ready for another round of review.

IvanYashchuk:
@anijain2305 @jansel ^

jansel (approved):

torchdynamo/optimizations/training.py
SherlockNoMad: I created this backend for your team's use. 
Now that you have your own implementation, please feel free to remove this.

IvanYashchuk:
Can someone hit the merge button?

desertfire:
You can run `make format` to fix the lint error.

IvanYashchuk:
@desertfire, thanks, I ran it. Could you please approve to run the workflows?

anijain2305 (approved):

anijain2305:
Thanks @IvanYashchuk for the contribution. Your PR is merged.

PR #1288, opened 2022-09-21
Skip cuda memory reset when testing on cpu

desertfire:


benchmarks/common.py
```diff
@@ -1212,8 +1212,9 @@ def run_performance_test(
             optimized_model_iter_fn = optimize_ctx(self.model_iter_fn)
 
             try:
-                torch.cuda.reset_peak_memory_stats()
-                torch.cuda.empty_cache()
+                if self.args.devices == ["cuda"]:
```
anijain2305, line 1215: Maybe use `current_device`?

desertfire: It makes sense. The way how `args.devices` is currently used in other places may need clean up as well, but it's low priority as we don't really mix multiple device measurement.

PR #1289, opened 2022-09-21
Refactor context management (less hierarchy, fix initial value, remove grad assumptions), add autocast

voznesenskym:


jansel (approved):

torchdynamo/variables/misc.py
```diff
@@ -306,7 +363,7 @@ def exit(self, tx, *args):
         return variables.ConstantVariable(None, **VariableTracker.propagate(self))
 
     def fn_name(self):
-        return "autograd.profiler.profile"
+        return "torch.amp.autocast_mode.autocast"
```
yanboliang, line 366: Is this typo or on purpose?

voznesenskym: No, good catch.

Issue #885, opened 2022-08-18
[inductor] Lower aten.index_add_/aten.index_add


desertfire:
Should be fairly similar to how `aten.index_put_/aten.index_put` is currently handled, see
 * https://github.com/pytorch/torchdynamo/blob/7cc850ca54745a995b792947f49871cf93ef0081/torchinductor/lowering.py#L1445
 * https://github.com/pytorch/torchdynamo/blob/7cc850ca54745a995b792947f49871cf93ef0081/torchinductor/decomposition.py#L357

eellison:
I looked into this a bit, I'm not sure it has that much in common with `index_put`, since `index_add` can accumulate arbitrarily many indices of the same value.  Also, the usage seems to be pretty different - 
for `fastNLP_Bert` you have inputs: `Operator: aten.index_put_.default
cnt: 1, ((T([6, 476], i64), [T([6], i64), T([6], i64)], T([], i64)), {})` (aka indices of length 6). 

For e.g. `hf_Longformer` you have the following: `((T([2359296], f16), 0, T([4718592], i64), T([4718592], f16)), {})` (aka, indices of length 4718592).

I can't think of a lowering of `index_add` within inductor that makes a ton of sense at least within the current IR. Maybe a custom template op (similar to matmul/conv) might make sense ?  Or maybe there's a way to generalize this pattern with embedding_bag. Both are pretty common in huggingface. 

cc @ngimel @jansel

jansel:
You should be able to map this to `ir.Scatter(..., scatter_mode="atomic_add")`.

`ir.Scatter` is very similar to `ir.Pointwise`, except:

`ir.Pointwise` computes
```
out[x] = ...
```

`ir.Scatter` computes
```
out[output_indexer(x)] = ...
```

If you set `scatter_mode="atomic_add"` it instead does:
```
out[output_indexer(x)] += ...
```
In a thread-safe way.


The iteration ranges of scatter then become the iteration space of the scatter loops (which will be different than the size of the output).

The lowering should be something like this (warning, this is untested):
```
def index_add_(self, dim, index, source, *, alpha=1):
    index_loader = index.make_loader()
    source_loader = source.make_loader()

    def output_indexer(index):
        index = list(index)
        index[dim] = index_loader([index[dim]])
        return index

    def fn(index):
        return ops.mul(
            ops.constant(alpha, source.get_dtype()), 
            source_loader(index)
        )
    
    return ir.Scatter(
[...]

PR #1292, opened 2022-09-21
Remove lowering for index_select and add decomp for index_{add,add_}

lezcano:
The decompositons and refs for these functions were added in core in https://github.com/pytorch/pytorch/pull/85002. They generate the same code as the hand-written lowering.

Fixes https://github.com/pytorch/torchdynamo/issues/885


jansel (approved):

lezcano:
The error seems unrelated, but is it?

desertfire:
> The error seems unrelated, but is it?

I would suggest you to verify it locally. You can find the command used by CI at the top of the `Huggingface inference run` section.

lezcano:
The failures seem unrelated.

PR #1298, opened 2022-09-21
Fix test_distributed config for CI

wconstab:
avoid tests that depend on nccl for now

switch from cuda to cpu as this CI runner has no cuda, and afaik the coverage/failure mode would have been the same

### DIFF
--- a/test/test_distributed.py
+++ b/test/test_distributed.py
@@ -55,7 +55,8 @@ class TestDistributed(torchdynamo.testing.TestCase):
             )
         )
         cls.rank = 0
-        cls.device = f"cuda:{cls.rank}"
+        cls.device = f"cpu:{cls.rank}"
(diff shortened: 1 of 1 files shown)


wconstab:
cc @anijain2305 @davidberard98  looks like something fixed DDP+dynamo on tip, these tests started passing again for me :) not sure if you fixed it or if this just happened by chance.

voznesenskym (approved):

mlazos (approved):

anijain2305:
> cc @anijain2305 @davidberard98 looks like something fixed DDP+dynamo on tip, these tests started passing again for me :) not sure if you fixed it or if this just happened by chance.

I didn't do anything. I hope these things happen more often :D

anijain2305 (approved):

PR #1299, opened 2022-09-21
Fix problems in debug_utils when working on internal environment

desertfire:


torchdynamo/debug_utils.py
```diff
@@ -101,12 +101,16 @@ def _cuda_system_info_comment():
...
-    model_str += f"{cuda_version_out}\n"
+    try:
+        cuda_version_out = subprocess.run(["nvcc", "--version"], stdout=subprocess.PIPE)
+        cuda_version_lines = cuda_version_out.stdout.decode().split("\n")
+        cuda_version_out = "".join(
+            [f"# {s} \n" for s in cuda_version_lines if s not in [""]]
+        )
+        model_str += f"{cuda_version_out}\n"
+    except FileNotFoundError:
+        print("nvcc does not exist")
```
voznesenskym, line 112: Instead of printing here, can you write out "nvcc not found in path" to the model_str?

voznesenskym (approved):

desertfire:
Tests all passed.  Not showing as success because `GitHub rate limited us while adding a commit status of success: API rate limit exceeded for user ID 6422482.`

Issue #1172, opened 2022-09-08
Convert simple usage of copy.copy/copy.deepcopy to pytorch ops


jansel:
This code will currently trigger a graph break:
```
def foo(x):
   y = torch.randn_like(x) 
   z = copy.deepcopy(y)
   return z+1
```

We should be able to handle this by mapping it to `torch.clone()` and not graph breaking.

williamwen42:
Considering reverting the change since `clone` and `deepcopy` semantics are not exactly the same, e.g.
```
x = torch.rand(10)
x.attr = 1
y = copy.deepcopy(x)
print(getattr(y, 'attr', None))  # prints 1
z = x.clone()
print(getattr(z, 'attr', None))  # prints None
```

williamwen42:
For reference, `copy.copy` and assignment are not semantically equivalent either, e.g.
```
x = torch.rand(10)
x.attr = 1
y = copy.copy(x)
print(getattr(y, 'attr', None))  # prints None
z = x
print(getattr(z, 'attr', None))  # prints 1
```

PR #1300, opened 2022-09-21
Convert copy.copy/deepcopy calls on tensors to pytorch ops

williamwen42:
Address #1172.

Test: https://gist.github.com/williamwen42/45ddec1a973388800d0890a1bf33f489

### DIFF
--- a/torchdynamo/utils.py
+++ b/torchdynamo/utils.py
@@ -784,7 +784,6 @@ def same(
                     log.error(
                         f"RMSE (res-fp64): {res_error:.5f}, (ref-fp64): {ref_error:.5f}"
                     )
-                    # import pdb; pdb.set_trace()
                 return passes_test
 
             return False
--- a/torchdynamo/utils.py
+++ b/torchdynamo/utils.py
@@ -784,6 +784,7 @@ def same(
                     log.error(
                         f"RMSE (res-fp64): {res_error:.5f}, (ref-fp64): {ref_error:.5f}"
                     )
+                    # import pdb; pdb.set_trace()
                 return passes_test
 
             return False
--- a/torchdynamo/variables/functions.py
+++ b/torchdynamo/variables/functions.py
@@ -182,7 +182,9 @@ class UserFunctionVariable(BaseUserFunctionVariable):
     ) -> "VariableTracker":
         # handle copy.copy and copy.deepcopy on tensors
         if self.fn is copy.copy or self.fn is copy.deepcopy:
-            if len(args) >= 1 and isinstance(args[0], variables.TensorVariable):
+            if len(args) != 1:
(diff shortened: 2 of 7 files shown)


williamwen42:
@jansel to confirm, the semantics for converting `copy.copy` should just be tensor assignment, i.e. `y = x`?

torchdynamo/variables/functions.py
```diff
@@ -174,6 +177,22 @@ def bind_args(self, parent, args, kwargs):
     def export_freevars(self, parent, child):
         pass
 
+    def call_function(
+        self, tx, args: "List[VariableTracker]", kwargs: "Dict[str, VariableTracker]"
+    ) -> "VariableTracker":
+        # handle copy.copy and copy.deepcopy on tensors
+        if self.fn is copy.copy or self.fn is copy.deepcopy:
+            if len(args) >= 1 and isinstance(args[0], variables.TensorVariable):
```
jansel, line 185: What will happen if len(args) == 2?

jansel (changes requested):
Add some test cases for this

torchdynamo/variables/functions.py
williamwen42: It should generally be == 1, but it's possible for `copy.deepcopy` to receive 2 args. Should we graph break in that case, or just assume the user still wants a torch.clone on the first argument?

torchdynamo/utils.py
```diff
@@ -784,7 +784,6 @@ def same(
                     log.error(
                         f"RMSE (res-fp64): {res_error:.5f}, (ref-fp64): {ref_error:.5f}"
                     )
-                    # import pdb; pdb.set_trace()
```
voznesenskym, line 787: Was this left behind as a hint to folks to debug accuracy errors? this might be worth keeping...

torchdynamo/variables/functions.py
```diff
@@ -174,6 +177,22 @@ def bind_args(self, parent, args, kwargs):
...
+                if self.fn is copy.copy:
+                    return args[0]
+                elif self.fn is copy.deepcopy:
+                    return variables.TorchVariable(torch.clone).call_function(
+                        tx, args, kwargs
+                    )
+            unimplemented("copy.copy/copy.deepcopy called on non-tensor")
```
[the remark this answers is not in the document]
williamwen42: If you copy a non-tensor, then dynamo would attempt to inline the copy function.

torchdynamo/variables/functions.py
williamwen42: Changed so that graph break will occur if args is not length 1.

torchdynamo/utils.py
williamwen42: Will keep.

jansel (approved):

mlazos (approved):

PR #1302, opened 2022-09-21
Remove UnspecializedNumpyVariable - stop special casing numpy

voznesenskym:


anijain2305:
What is the context behind it? Is specializing on numpy integers causing too many problems?

torchdynamo/variables/tensor.py
```diff
@@ -646,42 +646,16 @@ def inline_torch_function_unwrapped(
...
-
 class UnspecializedPythonVariable(TensorVariable):
     """
     This is a 1-element tensor represents unspecialized python float/int.
     """
 
     def __init__(self, proxy: torch.fx.Proxy, **kwargs):
         raw_value = kwargs.pop("raw_value", None)
+        if isinstance(raw_value, np.number):
+            raw_value = raw_value.item()
```
anijain2305, line 657: Ah, got the point. Nice!

voznesenskym:
> What is the context behind it? Is specializing on numpy integers causing too many problems?

No, just cleaning up code we don't need.

anijain2305 (approved):
+100

PR #1294, opened 2022-09-21
Update pytorch nightly version

desertfire:
Summary: Patch the CPU nightly version to work around a nightly build issue in core

PR #1306, opened 2022-09-21
Turn test_symbolic into an xfail

ezyang:


desertfire:
Err, already done this in https://github.com/pytorch/torchdynamo/pull/1294

ezyang:
You merged a skip. This is an xfail. It will fail when it starts passing so we know to delete it.

ezyang:
blurgh this passes on cpu lol. ok too much trouble

PR #1310, opened 2022-09-22
GradModeVariable should guard on the initial grad state

yanboliang:
This bug was found when I run training loss benchmark, where ```GRAD_MODE``` guard keeps failing and hits cache limit.

```GradModeVariable``` has a singleton guard on the grad state, but the state was decided when ```GuardBuilder.GRAD_MODE``` was called. It was after the code object transformation, where the grad state may be changed. 

The fix is to use the grad state at the entry of the frame execution.

### DIFF
--- a/torchdynamo/convert_frame.py
+++ b/torchdynamo/convert_frame.py
@@ -72,6 +72,9 @@ input_codes = Tracker()
 output_codes = Tracker()
 
 
+initial_grad_state = None
+
+
 @functools.wraps(original_forward_from_src)
 def fx_forward_from_src_skip_result(*args, **kwargs):
     # we monkey patch FX to prevent infinite loop of trying to convert
@@ -315,6 +318,9 @@ def convert_frame_assert(
 
         if not has_tensor_in_frame(frame):
             return None
+
+        global initial_grad_state
+        initial_grad_state = torch.is_grad_enabled()
 
         return _compile(
             frame.f_code,
--- a/torchdynamo/eval_frame.py
+++ b/torchdynamo/eval_frame.py
@@ -53,6 +53,7 @@ null_context = contextlib.nullcontext
 unset = object()
 compile_lock = threading.RLock()
 most_recent_backend = None
+initial_grad_state = None
 
 
 def remove_from_cache(f):
@@ -190,6 +191,8 @@ class OptimizeContext(_TorchDynamoContext):
     def __init__(self, callback, backend_ctx_ctor):
         def on_enter():
             global most_recent_backend
+            global initial_grad_state
+            initial_grad_state = torch.is_grad_enabled()
             if (
                 most_recent_backend is not None
                 and most_recent_backend is not compiler_fn
--- a/torchdynamo/eval_frame.py
+++ b/torchdynamo/eval_frame.py
@@ -53,7 +53,6 @@ null_context = contextlib.nullcontext
 unset = object()
 compile_lock = threading.RLock()
 most_recent_backend = None
-initial_grad_state = None
 
 
 def remove_from_cache(f):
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: torchdynamo/variables/misc.py [hunk 1]
```diff
@@ -262,6 +257,10 @@ def call_function(
 
 
 class GradModeVariable(ContextWrappingVariable):
+    """represents torch.{no_grad,enable_grad,set_grad_mode}()"""
+
+    _guards_singleton = {Guard("", GuardSource.GLOBAL, GuardBuilder.GRAD_MODE)}
+
```
yanboliang, line 263: code refactor, they are relevant to grad mode only.

torchdynamo/eval_frame.py
```diff
@@ -190,6 +191,8 @@ class OptimizeContext(_TorchDynamoContext):
     def __init__(self, callback, backend_ctx_ctor):
         def on_enter():
             global most_recent_backend
+            global initial_grad_state
+            initial_grad_state = torch.is_grad_enabled()
```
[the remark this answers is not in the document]
yanboliang: Oh, yes, you are right!

anijain2305 (approved):
LGTM. But, lets wait for @jansel review to be sure.

torchdynamo/convert_frame.py
```diff
@@ -72,6 +72,9 @@ def clear(self):
 output_codes = Tracker()
 
 
+initial_grad_state = None
```
jansel, line 75: Ideally we wouldn't use a global for this.

jansel (approved):

torchdynamo/convert_frame.py
yanboliang: I'll address this as follow-up, thanks!

torchdynamo/convert_frame.py
```diff
@@ -316,6 +319,9 @@ def format_guard_failures(code):
         if not has_tensor_in_frame(frame):
             return None
 
+        global initial_grad_state
+        initial_grad_state = torch.is_grad_enabled()
```
voznesenskym, line 323: It's odd to me that we special case this, and to @jansel's point, not great that it's global - what we should have instead is a state snapshot that is passed around. Grad is just one such state.

torchdynamo/guards.py
```diff
@@ -404,11 +406,11 @@ def OBJECT_MUTATION(self, guard: Guard):
         mutation_guard.watch(self.get(guard.name), self.guarded_code)
 
     def GRAD_MODE(self, guard: Guard):
-        """Guard on the current value of torch.is_grad_enabled()"""
+        """Guard on the initial grad state"""
         assert guard.name == ""
         assert guard.source is GuardSource.GLOBAL
         code = None
-        if torch.is_grad_enabled():
```
voznesenskym, line 411: Can you explain this to me a little more? When would torch.is_grad_enabled() not line up with the state stored in `convert_frame.initial_grad_state` - is it because someone swapped grad state, but it happened in a skip / graph break? If so, shouldn't we actually be checking the current torch.is_grad_enabled()? Why have they deviated?

yanboliang: Take a look at the unit test I added, if grad_enabled is True before the ```torch.set_grad_enabled``` and user's code sets it to ```False``` before graph break, then when this guard creating, the state is False which is not line up with the state stored in ```convert_frame.initial_grad_state``` which is True.

torchdynamo/convert_frame.py
yanboliang: I'll address the global variable issue in a follow up PR. We are planning to publish the E2E training performance result, which is blocked by this PR.

torchdynamo/guards.py
voznesenskym: But if user code says grad is false, why are we using true?

yanboliang: We are guarding on the enter state, not the exit state, right?

voznesenskym: Yes, enter state. 

If the user code says grad is set to false, we should reflect that in the guard.

What I don't understand is how torch.is_grad_enabled() can give us the wrong value for the state of grad... Why do we want the initial value, instead of the user set one?

PR #1311, opened 2022-09-22
Dashboard code and fixes in runner infra

anijain2305:
Tried putting most of the dashboard code in here, so that others can use and modify it as well.

Pending - Testing

### DIFF
--- a/benchmarks/common.py
+++ b/benchmarks/common.py
@@ -29,7 +29,6 @@ from torchdynamo.optimizations import backends
 from torchdynamo.optimizations.log_args import conv_args_analysis
 from torchdynamo.profiler import Profiler
 from torchdynamo.profiler import fx_insert_profiling
-from torchdynamo.testing import CompileCounterWithBackend
 from torchdynamo.testing import dummy_fx_compile
 from torchdynamo.testing import format_speedup
 from torchdynamo.testing import same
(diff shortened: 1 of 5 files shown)


anijain2305:
@desertfire Can I get a review on this? It 

* Cleans up the peak memory compression ratio as we discussed earlier (it was showing GBs, so I made it a ratio such that numbers like 1.2x make sense)
* Moves all of the dashboard code at 1 place. This will help other ppl use the same infra and make changes to dashboard whenever necessary.

The last 3 entries in Perforamnce Dashboard are from this PR.

benchmarks/torchbench.py
```diff
@@ -188,6 +188,7 @@
 }
 
 DONT_CHANGE_BATCH_SIZE = {
+    "demucs",
```
desertfire, line 191: This seems to fail on CI.

benchmarks/common.py
```diff
@@ -285,6 +284,10 @@ def print_summary(filename):
                 print(col.ljust(width), f"{data[col].mean():.1%}")
             elif col in ("graphs", "graph_calls", "captured_ops", "total_ops"):
                 print(col.ljust(width), f"{data[col].mean():.1f}")
+            elif col in ("compilation_latency"):
+                print(col.ljust(width), f"mean={data[col].mean():.1f} seconds")
+            elif col in ("compression_ratio"):
```
desertfire, line 289: I find the naming is a little confusing here. First, having `memory` in the naming can make the meaning more explicit. Second, I guess the reason you picked compression_ratio `eager_peak_mem / dynamo_peak_mem`(instead of the inverse) to measure the memory reduction effect is probably following how we compute the speedup, but I feel for memory reduction the metric is uncommon? But that could be a personal preference.

desertfire (approved):
Approve with optional-to-be-addressed comment

benchmarks/common.py
anijain2305: I have typically used memory compression ratio in the research before. Higher the compression ratio, better the technique. It kind of makes graphs/conclusions easy to read. 

Another popular one is percentage reduction in memory. I think it's compression ratio is ok for now. I can add `memory` to be more precise.

benchmarks/torchbench.py
anijain2305: We will have to skip this one. The smallest batch size demucs support is 4, and even that is unable to fit in CI's GPU memory.

PR #1319, opened 2022-09-22
Support .numel with config.capture_scalar_outputs

voznesenskym:


torchdynamo/variables/tensor.py
```diff
@@ -419,7 +420,16 @@ def call_method(
             sizes = [variables.ConstantVariable(x) for x in self.size]
             constant_result = SizeVariable(sizes, **options)
         elif name == "numel" and self.size is not None:
-            constant_result = ConstantVariable(product(self.size), **options)
+            if config.capture_scalar_outputs:
```
jansel, line 423: This path is only hit in static shape mode, so the number of elements is a constant and guarded on.

jansel (changes requested):
Add tests.

jansel:

PR #1320, opened 2022-09-22
Assert memory format

voznesenskym:


jansel (approved):
make sure this passes CI

PR #1328, opened 2022-09-23
[WIP] Let Dynamo step through DistributedDataParallel forward

anijain2305:
cc @wconstab @davidberard98

### DIFF
--- a/torchdynamo/variables/builder.py
+++ b/torchdynamo/variables/builder.py
@@ -255,6 +255,12 @@ class VariableBuilder:
                 return self.tx.output.side_effects.track_object_existing(
                     self.source, value, result
                 )
+            elif issubclass(
+                value.__class__, torch.nn.parallel.distributed.DistributedDataParallel
+            ):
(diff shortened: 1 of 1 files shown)


jansel (approved):

davidberard98 (approved):
This fixes our resnet50 issue.

anijain2305:
@davidberard98 I am hesitating to merge this because I think we can add a category of ops for which we want Dynamo to let in. This could be FSDP very soon.

Do you want to take a stab at this? Basically add a category in allowed_functions.py and if nn_module falls in that category, we let Dynamo step in. Let me know if thats ok.

wconstab:
> I am hesitating to merge this because I think we can add a category of ops for which we want Dynamo to let in. This could be FSDP very soon.

anijain2305:
Ok, we can do a refactor later. We can land this for now.

I don;t expect this to be many modules, but I do expect a few related to distributed, like FSDP and DDP, maybe the param_flat thing that you are seeing as well.

Issue #1148, opened 2022-09-07
TorchBench training error: RuntimeError: Default process group has not been initialized, please make sure to call init_process_group.


desertfire:
Repro:
```
benchmarks/torchbench.py -d cuda --inductor --training --float32 --no-skip -k moco
benchmarks/torchbench.py -d cuda --inductor --training --float32 --no-skip -k mobilenet_v3_large
```
Error:
```
Traceback (most recent call last):
  File "benchmarks/torchbench.py", line 354, in <module>
    main(TorchBenchmarkRunner(), original_dir)
  File "/fsx/users/binbao/torchdynamo-tip/benchmarks/common.py", line 1854, in main
    device, name, model, example_inputs, batch_size = runner.load_model(
  File "benchmarks/torchbench.py", line 260, in load_model
    benchmark = benchmark_cls(
  File "/fsx/users/binbao/torchbenchmark/torchbenchmark/util/model.py", line 16, in __call__
    obj = type.__call__(cls, *args, **kwargs)
  File "/fsx/users/binbao/torchbenchmark/torchbenchmark/models/moco/__init__.py", line 68, in __init__
    self.model = torch.nn.parallel.DistributedDataParallel(
  File "/fsx/users/binbao/pytorch-release/torch/nn/parallel/distributed.py", line 601, in __init__
    self.process_group = _get_default_group()
  File "/fsx/users/binbao/pytorch-release/torch/distributed/distributed_c10d.py", line 493, in _get_default_group
    raise RuntimeError(
RuntimeError: Default process group has not been initialized, please make sure to call init_process_group.
```

desertfire:
Assign to @wconstab as it is DDP related.

wconstab:
this looks like a benchmark config issue to me.  can you shed any light on why it's popping up now? (did we just start running this model for the first time, or something else changed?)

if we're trying to benchmark a model wrapped in DDP we should init the process group first.  Since we don't support  DDP benchmarking in the `torchdynamo/benchmarks/*` infra, there isn't any infra in there that inits the process group.

it looks like this model got wrapped in DDP either intentionally or perhaps unwittingly?

desertfire:
For context, the issue was hidden behind some other bugs. We can skip this model for now (cc @anijain2305), but thought it would be an interesting use case for your study.

wconstab:
[...]

PR #1336, opened 2022-09-24
[Inductor] Corner case empty graph

anijain2305:
Fixes the latest errors seen in https://github.com/pytorch/torchdynamo/issues/1148

### DIFF
--- a/torchinductor/compile_fx.py
+++ b/torchinductor/compile_fx.py
@@ -9,6 +9,7 @@ from functorch.compile import make_boxed_compiler
 from functorch.compile import min_cut_rematerialization_partition
 from torch._subclasses.fake_tensor import FakeTensor
 from torch.utils._mode_utils import no_dispatch
+from torchdynamo.utils import count_calls
 
 from . import config
 from . import overrides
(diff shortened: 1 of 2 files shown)


torchinductor/compile_fx.py
```diff
@@ -72,6 +73,9 @@ def compile_fx_inner(
     num_fixed=0,
     is_backward=False,
 ):
+    if count_calls(gm.graph) == 0:
```
voznesenskym, line 76: surprised we get here - iirc dynamo raises `raise exc.SkipFrame()` if `count_calls` returns 0. Did you see this downstream of dynamo?

anijain2305: Yeah, I am surprised too. This is DDP magic. Need to look into more details whats going on.

voznesenskym: Can you dump the graph? Perhaps its a graph of constants? Or something funky like an empty gm?

jansel: Could be from a backwards graph...

jansel (approved):

voznesenskym:
> @voznesenskym I added the offending graph as a test. The original graph is not empty, but the AOT Autograd generated fwd graph is empty.

lovely, ty

voznesenskym (approved):

anijain2305:
CI failure is unrelated.

PR #1337, opened 2022-09-24
Skip test_distributed

mlazos:
This test suite hangs in PyTorch core. cc @wconstab 

disabling to move the pin


voznesenskym (approved):
Sad times

PR #1338, opened 2022-09-24
Use new Triton runtime

jansel:
@ptillet recently rewrote the Triton runtime in https://github.com/openai/triton/pull/644

This should dramatically reduce CPU overheads when cudagraphs is disabled.

This updates TorchInductor to use that new runtime.  We now call `triton.compile()`, and no longer use `triton.jit()`.  

There is also some early support for parallel compiles, but still need optimize that part.

Note this PR breaks support for Triton versions prior to 998fd5f9afe166247f441999c605dfe624ca9331.


test/test_torchinductor.py
```diff
@@ -37,7 +37,8 @@
     assert get_decompositions([torch.ops.aten.trace])
     # Requires functorch
     from torchinductor.compile_fx import compile_fx_inner
-except (ImportError, ModuleNotFoundError, AssertionError):
+except (ImportError, ModuleNotFoundError, AssertionError) as e:
+    print(f"{type(e)}: {e}")
```
voznesenskym, line 41: Nit: Can you print to stderr?

torchinductor/codecache.py
```diff
@@ -161,8 +165,9 @@ def load(cls, source_code):
                 mod = types.ModuleType(f"{__name__}.{key}")
                 mod.__file__ = path
                 exec(code, mod.__dict__, mod.__dict__)
-                cls.cache[key] = mod
-                cls.cache[key].key = key
+                if key not in cls.cache:
```
voznesenskym, line 168: Is this not already always the case because of `if key not in cls.cache:` not on L162?

torchinductor/codecache.py
```diff
@@ -174,7 +179,54 @@ def patch_triton_dir():
...
+    @functools.lru_cache(1)
+    def pool():
+        assert config.compile_threads > 1
+        return ThreadPoolExecutor(config.compile_threads)
+
+    @classmethod
+    def submit(cls, task):
+        if config.compile_threads <= 1:
+            return task()
+        return cls.pool().submit(task)
```
voznesenskym, line 205: If you make a class that composes `ThreadPoolExecutor` you can take all the `.compile_threads <= 1` logic out of this class, and instead just have that class define a `submit` that either delegates to the pool or returns it as is.

torchinductor/codecache.py
```diff
@@ -174,7 +179,54 @@ def patch_triton_dir():
...
+
+        def task():
+            kernel.precompile()
+            return kernel
+
+        return self.submit(task)
+
+    def cpp(self, source_code):
+        def task():
+            return CppCodeCache.load(source_code).kernel
```
voznesenskym, line 224: The cache load happens at subtly different times between these two. Triton's loads in the call to `triton()`, and the cpp one loads on the thread pool as it gets dispatched on a task. The incongruent behavior may lead to surprises later, or force CppCodeCache.load to be thread safe.

torchinductor/codecache.py
```diff
@@ -174,7 +179,54 @@ def patch_triton_dir():
...
+
+        return self.submit(task)
+
+    def wait(self, scope: Dict[str, Any]):
+        if config.compile_threads > 1:
+            for key, result in list(scope.items()):
+                if isinstance(result, Future):
```
voznesenskym, line 231: should we assert?

torchinductor/codegen/wrapper.py
```diff
@@ -245,6 +250,13 @@ def __init__(self):
 
         self.allocated = set()
         self.freed = set()
+        self.get_cuda_stream = functools.lru_cache(None)(self.get_cuda_stream)
+
+    def get_cuda_stream(self, index):
+        name = f"stream{index}"
+        # self.writeline("torch.cuda.set_device(torch.cuda.current_device())")
```
voznesenskym, line 257: rm comment?

torchinductor/codegen/wrapper.py
```diff
@@ -245,6 +250,13 @@ def __init__(self):
 
         self.allocated = set()
         self.freed = set()
+        self.get_cuda_stream = functools.lru_cache(None)(self.get_cuda_stream)
+
+    def get_cuda_stream(self, index):
```
voznesenskym, line 255: called get, but does a write_line. Maybe `write_get_cuda_stream`?

torchinductor/codegen/triton.py
```diff
@@ -1175,7 +1214,7 @@ def codegen_node_schedule(self, node_schedule, numel, reduction_numel):
             kernel_name = wrapper.next_kernel_name()
             wrapper.kernels[src_code] = kernel_name
             subs_name = kernel_name if config.triton.ordered_kernel_names else "kernel"
-            src_code = src_code.format(kernel_name=subs_name)
+            src_code = src_code.replace("KERNEL_NAME", subs_name)
```
voznesenskym, line 1217: Yes, join the dark side.

torchinductor/codecache.py
jansel: This PR adds multithreading.  I've been experimenting with loading these in parallel in a racy way.  If another thread won the race, we want to reuse its result to have fewer compiles.

torchinductor/codegen/triton.py
```diff
@@ -993,7 +1029,10 @@ def call_kernel(self, code, name: str):
             if tree.prefix != "r":
                 grid.append(expr)
         call_args = ", ".join(call_args)
-        code.writeline(f"{name}[grid({', '.join(grid)})]({call_args})")
+        stream = code.get_cuda_stream(V.graph.scheduler.current_device.index)
```
voznesenskym, line 1032: nit: `stream_name =`

torchinductor/codecache.py
jansel: No, there are other types of objects in the module.

torchinductor/codecache.py
jansel: The C++ cache load calls gcc, which is expensive (and also inherently thread safe).

torchinductor/compile_fx.py
```diff
@@ -102,12 +100,43 @@ def compile_fx_inner(
...
+        and inputs[i].device.type == "cuda"
+    ]
+
+    if len(check_inputs) == 0:
+        return model
+
+    def run(*new_inputs):
+        for i in check_inputs:
+            if new_inputs[i].data_ptr() % ALIGNMENT:
+                if isinstance(new_inputs, tuple):
```
voznesenskym, line 134: Feels a little odd to have this check before accessing new_inputs[i] above

torchinductor/codegen/triton.py
jansel: format was false alarming on dicts

torchinductor/codecache.py
jansel: There are only 2 instances of this related ThreadPoolExecutor.  The others are skipping different things.

torchinductor/compile_fx.py
jansel: This is a performance optimization to avoid a copy in the (very) common case where everything is aligned.

torchinductor/triton_ops/autotune.py
```diff
@@ -1,70 +1,163 @@
...
 
-class CachingAutotuner(triton.code_gen.Autotuner):
+class CachingAutotuner(KernelInterface):
     """
-    Simplified version of Triton autotuner that has no invalidation key
-    and caches the best config to disk.
+    Simplified version of Triton autotuner that has no invalidation
+    key and caches the best config to disk to improve cold start times.
+    Unlike the main triton Autotuner, this version can precompile all
+    configs, and does not rely on the Triton JIT.
```
voznesenskym, line 40: Awesome

torchinductor/triton_ops/autotune.py
```diff
@@ -1,70 +1,163 @@
...
+        torch.cuda.set_device(torch.cuda.current_device())
+        compile_meta = copy.deepcopy(self.meta)
+        for k, v in cfg.kwargs.items():
+            compile_meta["constants"][self.fn.arg_names.index(k)] = v
+        compile_meta["num_warps"] = cfg.num_warps
+        compile_meta["num_stages"] = cfg.num_stages
+
+        if config.compile_threads > 1:
+            major, minor = torch.cuda.get_device_capability(compile_meta["device"])
+            compile_meta["cc"] = major * 10 + minor
```
voznesenskym, line 70: nit - if minor crosses double digits, which is valid semver, this will become wrong (1.11 becomes 21, so does 2.1) Consider just keeping semver (major.minor) or * 100 to reduce collision.

torchinductor/triton_ops/autotune.py
```diff
@@ -1,70 +1,163 @@
...
+            try:
+                p = multiprocessing.Process(
+                    target=triton.compile,
+                    args=(self.fn,),
+                    kwargs={**compile_meta, "warm_cache_only": True},
+                )
+                p.start()
+                p.join()
```
voznesenskym, line 78: Why are we doing this in a separate process?

torchinductor/triton_ops/autotune.py
```diff
@@ -1,70 +1,163 @@
...
+            except Exception:
+                log.exception("Error in async Triton compile")
+                # continue on to hopefully get a better error message below
+
+        binary = triton.compile(
+            self.fn,
+            **compile_meta,
+        )
```
voznesenskym, line 86: if the multiprocess call to compile succeed above, we still run this, right? Is that intended?

torchinductor/codecache.py
voznesenskym: Oh nice, that makes sense. Should we add a comment to that effect?

torchinductor/compile_fx.py
voznesenskym: Understood!

torchinductor/triton_ops/autotune.py
jansel: This behavior is to match Triton's API.

torchinductor/triton_ops/autotune.py
jansel: Because of the python GIL.

torchinductor/triton_ops/autotune.py
jansel: Then this one will be a cache hit and run very fast.

torchinductor/triton_ops/autotune.py
voznesenskym: Can you elaborate? We are not releasing the GIL here, and we are just waiting for the process to finish anyway. Yes, that new process has its own Python/ it's own GIL, but this parent process is just waiting, I think. But I could totally be missing something. What is this doing for the GIL?

test/test_torchinductor_opinfo.py
```diff
@@ -91,7 +91,7 @@ def process(device_type):
     "mvlgamma.mvlgamma_p_5": {f32, f64, i32, i64},  # flaky
     "cumprod": {f32, f64},  # flaky
     "_masked.prod": {f32, f64},  # flaky
-    "empty_like": {b8, f16, f32, f64},  # flaky
+    "empty_like": {b8, f16, f32, f64, i32, i64},  # flaky
```
ngimel, line 94: lol we shouldn't be checking results of empty_like, they are undefined, but that's for another PR

torchinductor/codegen/triton.py
```diff
@@ -28,11 +29,35 @@
...
+    raise NotImplementedError(f"unhandled {type(arg)}: {arg}")
+
+
+def config_of(args):
+    from triton.runtime.jit import JITFunction
+
+    divisible_by_16 = [
+        i
+        for i, arg in enumerate(args)
+        if isinstance(arg, TensorArg)
```
ngimel, line 55: I'm a bit concerned by saying all tensor args are divisible by 16, e.g. if there are some slicing op inside the compiled graph, and we end up calling `codegen_reference` with an odd offset https://github.com/pytorch/torchdynamo/blob/dedca39600c76a967c8881a5f1ae3aecf9fcc7be/torchinductor/ir.py#L1336

torchinductor/compile_fx.py
```diff
@@ -102,12 +100,43 @@ def compile_fx_inner(
...
+            if graph.mutated_inputs:
+                log.warning("skipping cudagraphs due to input mutation")
+            elif complex_memory_overlap_inputs:
+                log.warning("skipping cudagraphs due to complex input striding")
 
-    return compiled_fn
+    return align_inputs(compiled_fn, example_inputs, range(num_fixed))
+
+
+def clone_preserve_strides(x):
```
ngimel, line 112: if inputs are slices of the same tensor, this blows up the memory, and also removes aliasing but that's pre-existing

torchinductor/triton_ops/autotune.py
ngimel: has never happened so far, and compute capabilities don't even follow semvers

voznesenskym: Point taken :P

torchinductor/triton_ops/autotune.py
jansel: We call this function from many python threads.  `p.join()` releases the GIL, so other threads can start their own processes and compile other kernels in parallel.  This multiprocessing parallelism is pretty standard in Python code.

test/test_torchinductor_opinfo.py
jansel: This test is flakey on master.

torchinductor/codegen/triton.py
jansel: `codegen_reference()` is only used for calling extern kernels (and return values) I believe.  For Triton kernels we apply the offset inside the indexing formula.

torchinductor/compile_fx.py
jansel: You think we should switch to recompiling if alignments change? I did this based on your suggestion when we talked in person.

torchinductor/triton_ops/autotune.py
voznesenskym: I missed the part where this was getting called from many threads, that makes sense, thanks.

voznesenskym:
LGTM from first principles.

torchinductor/compile_fx.py
ngimel: Idk, maybe we should handle arg aliasing some other way? It's problematic now anyways (and we are already doing this for cuda graphs) so it's not a regression and definitely not blocking for this PR.

torchinductor/codecache.py
voznesenskym: Interesting issue with potential relevance: https://github.com/pytorch/torchdynamo/issues/1347

jansel:
This PR now has a single test failure:
```
CUDA_LAUNCH_BLOCKING=1 python benchmarks/timm_models.py --accuracy -d cuda --inductor --float32 --training --only swin_base_patch4_window7_224
cuda train swin_base_patch4_window7_224
...
  File "/tmp/torchinductor_jansel/nd/cndxz5mmbqajkiakvbzxdczaralpd7t57ajewnph4cqbiabmemjf.py", line 6011, in call
    kernel76.run(buf1055, _unsafe_view_12, getitem_19, reciprocal_8, buf1061, buf1063, 3328, 121, grid=grid(3328), stream=stream0)
  File "/home/jansel/torchdynamo/torchinductor/triton_ops/autotune.py", line 157, in run
    return launcher(
  File "<string>", line 4, in launcher
RuntimeError: Triton Error [CUDA]: an illegal memory access was encountered
```

At first I thought this was related to the alignment code in the new Triton runtime, however, if I disable that with:
```patch
diff --git a/torchinductor/codegen/triton.py b/torchinductor/codegen/triton.py
index 51123ce..439dfc4 100644
--- a/torchinductor/codegen/triton.py
+++ b/torchinductor/codegen/triton.py
@@ -50,10 +50,6 @@ def config_of(args):
     from triton.runtime.jit import JITFunction
 
     divisible_by_16 = [
-        i
-        for i, arg in enumerate(args)
-        if isinstance(arg, TensorArg)
-        or V.graph.sizevars.maybe_guard_multiple_of(arg.expr, JITFunction.divisibility)
     ]
     return instance_descriptor(tuple(divisible_by_16), ())
 
diff --git a/torchinductor/compile_fx.py b/torchinductor/compile_fx.py
index 56c7d5c..efb6463 100644
--- a/torchinductor/compile_fx.py
+++ b/torchinductor/compile_fx.py
@@ -106,7 +106,7 @@ def compile_fx_inner(
             elif complex_memory_overlap_inputs:
                 log.warning("skipping cudagraphs due to complex input striding")
 
-    return align_inputs(compiled_fn, example_inputs, range(num_fixed))
+    return compiled_fn
 
 
 def clone_preserve_strides(x):
diff --git a/torchinductor/config.py b/torchinductor/config.py
index 7d9c38e..fd336b3 100644
--- a/torchinductor/config.py
+++ b/torchinductor/config.py
@@ -73,7 +73,7 @@ class cpp:
 class triton:
 
-    cudagraphs = True
+    cudagraphs = False
 
     # choose conv backend, "aten" or "triton" or "autotune"
     convolution = "aten"
```
It still gets an illegal memory access.

I tried setting `TORCHDYNAMO_REPRO_AFTER` = `aot`/`dynamo`, but both of them fall with IMA's inside the repro code:
```
  File "/home/jansel/torchdynamo/torchdynamo/debug_utils.py", line 501, in dump_backend_repro_as_tarfile
    gm.to_folder(gm_dir, "Repro")
  File "/home/jansel/conda/envs/torchdynamo/lib/python3.8/site-packages/torch/fx/graph_module.py", line 423, in to_folder
    torch.save(self.state_dict(), folder / 'state_dict.pt')
  File "/home/jansel/conda/envs/torchdynamo/lib/python3.8/site-packages/torch/serialization.py", line 422, in save
    _save(obj, opened_zipfile, pickle_module, pickle_protocol)
  File "/home/jansel/conda/envs/torchdynamo/lib/python3.8/site-packages/torch/serialization.py", line 646, in _save
    storage = storage.cpu()
  File "/home/jansel/conda/envs/torchdynamo/lib/python3.8/site-packages/torch/storage.py", line 120, in cpu
    return torch.UntypedStorage(self.size()).copy_(self, False)
RuntimeError: CUDA error: an illegal memory access was encountered
```
and
```
  File "/home/jansel/conda/envs/torchdynamo/lib/python3.8/site-packages/torch/_tensor_str.py", line 115, in __init__
    tensor_view, torch.isfinite(tensor_view) & tensor_view.ne(0)
RuntimeError: CUDA error: an illegal memory access was encountered
```
and
```
  File "/home/jansel/conda/envs/torchdynamo/lib/python3.8/site-packages/torch/cuda/random.py", line 62, in cb
    default_generator.set_state(new_state_copy)
RuntimeError: CUDA error: an illegal memory access was encountered
```

It feels almost like there is some CUDA corruption causing random failures everywhere.  I am stumped...

@ngimel any ideas here?  I think you mentioned seeing similar stuff elsewhere.

I'm not sure if this is related to this PR.  Perhaps we should just skip this test and turn it into an issue.

soumith:
@voznesenskym was also mentioning today that (without this PR), running OpInfo tests on `main`, but reordering the tests is causing them to fail in strange ways including IMEs. Voz said he'll start looking into that under `compute-sanitizer`

jansel:
We see this same error in https://github.com/pytorch/torchdynamo/pull/1362 so I am going to skip that test and spin this off into a new issue:  #1365

ngimel (approved):

voznesenskym (approved):

PR #1340, opened 2022-09-25
Get past inline in skipfiles for op testing, either pass tests or uncover real issues. (Enable ~600 more tests)

voznesenskym:


voznesenskym:
This gets past hundreds of current skipfile errors, which in turn progresses downstream to segfaults due to memory corruption.

We should consider running each op in its own process...

torchdynamo/config.py
```diff
@@ -86,7 +86,11 @@
 
 # If a PyTorch module is in this allowlist, torchdynamo will be allowed
 # to inline objects from it or its children.
-skipfiles_inline_module_allowlist = {torch.nn, torch.distributions}
+skipfiles_inline_module_allowlist = {
+    torch.nn,
+    torch.distributions,
+    torch.testing,
```
voznesenskym, line 92: we can also patch this in during tests, if we don't want it here, but I do not think it matters much.

ngimel (approved):

test/test_torchinductor_opinfo.py
```diff
@@ -91,9 +91,15 @@ def process(device_type):
     "mvlgamma.mvlgamma_p_5": {f32, f64, i32, i64},  # flaky
     "cumprod": {f32, f64},  # flaky
     "_masked.prod": {f32, f64},  # flaky
+    "_masked.std": {b8, f16, f32, f64, i32, i64},  # segfault
+    "histic": {b8, f16, f32, f64, i32, i64},  # segfault
```
ngimel, line 95: histc?

voznesenskym: histc

voznesenskym: This whole list is garbage now since we fixed the indentation bug you pointed out, I have a local branch with a ton more exclusions. I'll push that up once it passes locally.

voznesenskym:
No time to see this through, unfortunately, @SherlockNoMad  feel free to take it.

PR #1341, opened 2022-09-25
Timm models - Keep loss value small

anijain2305:
For the TIMM models that are failing accuracy with TorchInductor, I found the loss value is quite high (spasnet - 8, pnasnet - 7). Few months ago, we observed a behavior that for large loss values, small changes in the order of accumulation of gradients (fan-out in the forward pass means accumulation in the backward pass) can lead to accuracy deviations.

In our benchmarking infra, we use random input values, while the models are trained with pre-processed data, which could explain these high loss values.

This PR tries to lower the loss value. I tried locally and it passes accuracy checks for many Inductor models.

### DIFF
--- a/benchmarks/timm_models.py
+++ b/benchmarks/timm_models.py
@@ -10,7 +10,6 @@ import warnings
 import torch
 from common import BenchmarkRunner
 from common import main
-from torch._subclasses import FakeTensor
 
 from torchdynamo.testing import collect_results
 from torchdynamo.utils import clone_inputs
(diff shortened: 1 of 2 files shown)


benchmarks/timm_models.py
```diff
@@ -297,10 +296,9 @@ def _gen_target(self, batch_size, device):
         )
 
     def compute_loss(self, pred):
-        if isinstance(pred, FakeTensor) and not isinstance(self.target, FakeTensor):
```
anijain2305, line 300: @eellison The earlier code seemed out of place w/r/t FakeTensor. Please review.

jansel (approved):

PR #1344, opened 2022-09-25
Test CI

desertfire:


desertfire:
@SherlockNoMad , looks like test_torchinductor_opinfo still running on both cpu and cuda in the `unit_test` task.

SherlockNoMad:
> @SherlockNoMad , looks like test_torchinductor_opinfo still running on both cpu and cuda in the `unit_test` task.

Yes, I also noticed this in local machine... looks like a pytest bug 

As a workaround, we can use -k 'TestInductorOpInfoCUDA'

Issue #1174, opened 2022-09-08
Add tiling to TorchInductor C++ backend 


jansel:
The [C++ backend](https://github.com/pytorch/torchdynamo/blob/main/torchinductor/codegen/cpp.py) currently does not do any tiling like the [Triton backend does](https://github.com/pytorch/torchdynamo/blob/8be5842ed74e7e16bc734814befa5365305d5ca5/torchinductor/codegen/triton.py#L1249).

We could likely map this to `#pragama omp tile(...)` somewhat easily.

desertfire:
@EikanWang mentioned he is not able to assign this one to himself. @jansel , is there any registration steps that new contributors need to follow? I believe @EikanWang has already assigned CLA when working on pytorch core.

jansel:
I sent an invite.

EikanWang:
Seems like I still could not assign this task to me. But anyway, we are working on this issue.

EikanWang:
@jansel , I clicked the invitation and tried to accept it. But it always directs [https://github.com/pytorch/torchdynamo/invitations](https://github.com/pytorch/torchdynamo/invitations) to [https://github.com/pytorch/torchdynamo](https://github.com/pytorch/torchdynamo). May I know if it is as expected since I still cannot assign an issue?

jansel:
Weird, I can try again.

zhuhaozhe:
Hi, @jansel.
I start to find  `#pragma omp tile(...)` on [openmp 5.1 doc](https://www.openmp.org/wp-content/uploads/OpenMPRefCard-5.1-web.pdf) and GCC12 has the initial support of OpenMP 5.1 https://www.openmp.org/resources/openmp-compilers-tools/.
Although compiler support is limited for `#pragma omp tile(...)`, we can do this loop transformation without openmp naive support.
![image](https://user-images.githubusercontent.com/54701539/197709044-bed542f0-2e1e-4955-bb4a-f1acfd406053.png)

We can decide how to do tiling while creating [CppKernel](https://github.com/pytorch/torchdynamo/blob/77f2e194dfbc887c68115ca5c52bfc0d59f7df4a/torchinductor/codegen/cpp.py#L311) and attach tiling info on the CppKernel instance.
Then we can 
[...]

PR #1348, opened 2022-09-26
Add linear eltwise fusion for cpu in inference mode

chunyuan-w:
An FX transformation is added to fuse linear with eltwise OPs in torchinductor for CPU in inference mode.

The fusion OP is implemented in https://github.com/pytorch/pytorch/pull/85622 and will be treated as an extern kernel call in torchinductor.

The fusion of linear with the below OPs is supported:

- relu
- sigmoid
- tanh
- hardswish
- leaky_relu
- hardtanh
- gelu


torchinductor/ir.py
```diff
@@ -2840,6 +2840,49 @@ def __init__(self, layout, input, index):
     def should_allocate(self):
         return False
 
+class LinearReLU(ExternKernelAlloc):
```
EikanWang, line 2843: `LinearReLU` is specific for `Linear` + `ReLU`. Should this class be `LinearEltwise`?

torchinductor/overrides.py
```diff
@@ -163,3 +230,15 @@ def rand_like(x, **kwargs):
 
 
 replacements = {torch.nn.functional.dropout: lowmem_dropout, torch.rand_like: rand_like}
+
+computation_op = nn.Linear
```
EikanWang, line 234: Currently, the computation operator only covers `Linear`. In the future, we also need to support `Conv` and `Matmul`. Should the `computation_op` be a list?

EikanWang:
@jansel , @desertfire , this PR demonstrates how to integrate oneDNN post-op fusions in TorhcInductor. Please feel free to comment on the design.

test/test_linear_fusion_cpu.py
```diff
@@ -0,0 +1,68 @@
...
+        nn.Tanh(),
+        nn.Hardswish(),
+        nn.LeakyReLU(0.1, inplace=False),
+        nn.Hardtanh(min_val=-0.5, max_val=4, inplace=False),
+        nn.GELU(approximate="none"),
+        nn.GELU(approximate="tanh"),
+    ]
+    return eltwise_list
+
+class TestFuseFx(QuantizationTestCase):
```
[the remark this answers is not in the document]
chunyuan-w: I want leverage this [checkGraphModuleNodes](https://github.com/pytorch/pytorch/blob/ea81138bd6b688554ad307e370bebeffd264f1b7/torch/testing/_internal/common_quantization.py#L583) function to check whether the fusion is successful or not and this function is defined in the [QuantizationTestCase](https://github.com/pytorch/pytorch/blob/ea81138bd6b688554ad307e370bebeffd264f1b7/torch/testing/_internal/common_quantization.py#L332) class.

torchinductor/ir.py
chunyuan-w: Fixed the name.

test/test_linear_fusion_cpu.py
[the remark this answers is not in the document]
jgong5: How about we add note to explain why we have to inherit `QuantizationTestCase` here? We may consider to move `checkGraphModuleNodes` util function to a common place later to avoid confusion.

torchinductor/ir.py
```diff
@@ -2840,6 +2840,49 @@ def __init__(self, layout, input, index):
     def should_allocate(self):
         return False
 
+class LinearEltwise(ExternKernelAlloc):
+    kernel = "torch.ops.mkldnn_prepacked.linear_eltwise"
```
jgong5, line 2844: About the naming: Is the post-op unary only? Also, to be consistent with existing IR names, would "pointwise" be better than "eltwise"?

torchinductor/overrides.py
```diff
@@ -36,6 +39,70 @@ def replace_fx(gm: torch.fx.GraphModule):
...
+                    eltwise,
+                    op_name,
+                    op_info,
+                    linear.in_features,
+                    linear.out_features,
+                    linear.bias is not None,
+                    linear.weight.device,
+                    linear.weight.dtype)
+
+def fuse_fx(gm: torch.fx.GraphModule, example_inputs):
```
jgong5, line 83: Not sure if it is proper to add backend-specific FX passes here. Would it better to add a FX pass registration mechanism here (e.g., a registration decorator similar to how lowerings are registered) and move the backend-specific passes to a separate backend file?

test/test_linear_fusion_cpu.py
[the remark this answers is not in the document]
chunyuan-w: Comments are added.

torchinductor/overrides.py
chunyuan-w: Refactored the code to be more extensible to add support for Conv and Matmul in the future.

soumith:
ideally, ideep could have a python API + package, or directly be a C++ dependency in Inductor.
I don't like the separation of bindings, where for every fusion you want to implement, you will send a PR to pytorch/pytorch and then hook up that pattern-match into inductor.
Inductor's strength isn't in pattern-matching to fast kernels, but is to generate code dynamically.
For example, it's a purely artificial restriction that the epilogue only allows the non-linearities specified above.

Also, the blackbox nature here is a bit uncomfortable, we would ideally have inductor over time have C++ generated code for CPU. A lowered IR can be passed to a powerful "CPU compiler" for example -- like LLVM. It's not ideal that we're wrapping around another high-level compiler.

jansel:
`make format` should fix the lint errors\

soumith:
> @soumith are you ok with the Intel folks exposing a C++-level API from PyTorch core? This will avoid the need to install two copies of MKL because PyTorch hides the symbols. When TorchDynamo is available in core this will become a bit simpler...

This part seems fine if that's what we want.

But more concretely, @jansel what is your long-term vision for a CPU backend for PyTorch? I'd like to see a proper, runtime codegen. But it's not clear to me what a statically compiled C++-level API gives us other than things like hard-coded epilogue fusions.

jansel:
> But more concretely, @jansel what is your long-term vision for a CPU backend for PyTorch? I'd like to see a proper, runtime codegen. But it's not clear to me what a statically compiled C++-level API gives us other than things like hard-coded epilogue fusions.

I agree, the focus should be on more general purpose codegen.

Calling MKL in this way wasn't my idea. I would push stuff like https://github.com/pytorch/torchdynamo/issues/1174 and https://github.com/pytorch/torchdynamo/issues/1175 which I believe the Intel folks are also looking at.

I think having a FX pass above inductor that does pattern-matching style fusions does fit into the larger plan though.  That is how we plan to integrate nvFuser, AIT, cuDNN, etc.  I don't see MKL as very different from those.

soumith:
> I think having a FX pass above inductor that does pattern-matching style fusions does fit into the larger plan though. That is how we plan to integrate nvFuser, AIT, cuDNN, etc. I don't see MKL as very different from those.

I'm pretty supportive of that. I think what threw me off was that this PR isn't integrating above inductor, but within inductor.

jgong5:
> Since we have codepaths to leverage these already (to_mkldnn, etc.), lets focus on long-term for the Inductor backend. It doesn't seem worth it to do the same kind of work again just for short-term inference wins that people can get via other codepaths.
> I'd like to see a proper, runtime codegen. But it's not clear to me what a statically compiled C++-level API gives us other than things like hard-coded epilogue fusions.
> I'm pretty supportive of that. I think what threw me off was that this PR isn't integrating above inductor, but within inductor.

@soumith My two cents. Integrating oneDNN post-ops into the FX pass inside inductor is an intermediate step towards a full-fledged codegen for conv and gemm. There is existing path that leverages oneDNN (a.k.a. mkldnn) but additional fusions with memory-bound operations, like those in BERT (multi-head-attention, add+layernorm), are still missing in the stock PyTorch. We implemented some rigid pattern-matching based TS graph rewrites in IPEX (https://github.com/intel-innersource/frameworks.ai.pytorch.ipex-cpu/blob/cpu-device/csrc/jit/passes/graph_rewrite_mha.cpp#L226) while these things can be more generally supported by inductor. Adding oneDNN post-ops to the inductor path combines the goodness from both oneDNN fusion and inductor's current codegen. In the long run, the C++-level API can expose micro-kernels as the building blocks of conv and gemm. Then, the conv and gemm would be codegen as a loop-nest over the micro-kernels. Does this plan make sense to you?

jansel:
> I'm pretty supportive of that. I think what threw me off was that this PR isn't integrating above inductor, but within inductor.

`fuse_fx()` is effectively "above inductor", since aot_autograd/normalization is called from inductor -- and most of these pattern matchers will to run after that -- inductor will need to also call the pattern matchers.

test/inductor/test_torchinductor.py
```diff
@@ -199,9 +216,15 @@ def upcast_fn(x):
             else:
                 return x
 
+        def get_original_lowp_dtype(example_inputs):
+            dtypes = [x.dtype for x in example_inputs if isinstance(x, torch.Tensor)]
+            dtype_set = set(dtypes)
+            return dtype_set.pop() if len(dtype_set) == 1 else torch.half
```
jgong5, line 222: This is used to support bfloat16 type? What if some of the inputs are bfloat16 (or half) and some are float?

test/inductor/test_torchinductor.py
```diff
@@ -1274,6 +1297,40 @@ def fn(a, b):
             check_lowp=False,
         )
 
+    @unittest.skipIf(HAS_CUDA, "only support bf16 linear unary fusion on cpu")
```
jgong5, line 1300: If it is CPU only, how about put it inside `if HAS_CPU:` check?

jgong5 (changes requested):

jansel:

torchinductor/ir.py
```diff
@@ -2943,6 +2943,56 @@ def should_allocate(self):
         return False
 
 
+class LinearEltwise(ExternKernelAlloc):
+    kernel = "torch.ops.mkldnn._linear_pointwise"
```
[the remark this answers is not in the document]
chunyuan-w: Got it. I'll rename it in the new PR submitted to pytorch repo.

PR #1351, opened 2022-09-26
Update PyTorch pin

desertfire:


jansel:
@desertfire I think we can actually just remove the failing torchaudio dep, it is not used.

desertfire:
> @desertfire I think we can actually just remove the failing torchaudio dep, it is not used.

I was thinking the same. Let me do that.

jansel (approved):

desertfire:
@voznesenskym , for the unit test failure here, `TestInductorOpInfoCUDA::test_comprehensive_masked_log_softmax_cuda_float64`, I think you already skipped the test in https://github.com/pytorch/torchdynamo/pull/1340. Do you happen to know what the problem is here?

ngimel:
#1358 was failing when running just the generated code, so it's unlikely aot or fake tensor problem (I fixed it by removing constexpr annotations, which is a triton bug), but the failure is very subtle and easily disappears when trying to minimize/change something

PR #1357, opened 2022-09-26
Remap copy.copy/copy.deepcopy only if argument really is a Tensor

williamwen42:
Fix case where `copy.copy`/`copy.deepcopy` is called on a `TensorVariable` but the wrapped value is not exactly a `torch.Tensor`, e.g. a `nn.Parameter`.

Fixes this test failure (`PYTORCH_TEST_WITH_DYNAMO=1 pytest test/test_subclass.py -k test_deepcopy `): https://github.com/pytorch/pytorch/blob/207a5a8fa9bfd9361038d46636c0440290c171bb/test/test_subclass.py#L48-L59

### DIFF
--- a/test/test_functions.py
+++ b/test/test_functions.py
@@ -5,6 +5,7 @@ import inspect
 import itertools
 import operator
 from typing import Any
+from unittest.mock import patch
 
 import torch
 from torch import sub
@@ -733,6 +734,9 @@ class FunctionTests(torchdynamo.testing.TestCase):
         with self.assertRaises(torchdynamo.exc.Unsupported):
             h(torch.rand(10))
 
+    # test does not work with fake_tensor_propagation set to False since
+    # the Parameter gets silently converted to a tensor before the deepcopy call
+    @patch.object(torchdynamo.config, "fake_tensor_propagation", True)
     def test_copy_parameter(self):
         import copy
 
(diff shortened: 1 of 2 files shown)


mlazos:
Rather than avoiding this case, is it possible to implement copying for the Parameter?

test/test_functions.py
```diff
@@ -732,3 +733,18 @@ def h(x):
 
         with self.assertRaises(torchdynamo.exc.Unsupported):
             h(torch.rand(10))
+
+    # test does not work with fake_tensor_propagation set to False since
+    # the Parameter gets silently converted to a tensor before the deepcopy call
```
anijain2305, line 738: Hmm, this seems like we should fix w/o `fake_tensor` as well.

torchdynamo/variables/functions.py
```diff
@@ -188,9 +188,48 @@ def call_function(
...
-                        tx, args, kwargs
-                    )
+                    if args[0].class_type is torch.Tensor:
+                        return args[0].call_method(tx, "clone", [], {})
+                    elif args[0].class_type is torch.nn.Parameter:
+                        # equivalent to
+                        # torch.nn.Parameter(
+                        #     args[0].clone(memory_format=torch.preserve_format)), args[0].requires_grad)
+                        # )
+                        return variables.TorchVariable(
```
anijain2305, line 198: Can we do a refactor such that code is readable?

mlazos: Agreed, I went over this offline with Will, do you think this a good approach? I was looking for second opinions. We felt this was hacky but I was like it does do what we want in essence.

jansel (approved):

mlazos (approved):

williamwen42:
Considering closing the PR since deepcopy may not always match with clones (see #1172).

Issue #1358, opened 2022-09-26
mixer_b16_224- Illegal CUDA memory access


anijain2305:
`benchmarks/timm_models.py --training -dcuda --accuracy --training --inductor --only=mixer_b16_224`

Error
~~~
  File "/tmp/torchinductor_anijain/ki/ckiub6r23dvmx2jzso3olsygu457mor7owjvwr2fei6we5roqbwh.py", line 1837, in call
    kernel0[grid(392)](convolution, addmm, addmm_2, addmm_3, addmm_5, addmm_6, addmm_8, addmm_9, addmm_11, addmm_12, addmm_14, addmm_15, addmm_17, addmm_18, addmm_20, addmm_21, addmm_23, addmm_24, addmm_26, addmm_27, addmm_29, addmm_30, addmm_32, addmm_33, addmm_35, getitem_49, reciprocal_48, getitem_47, reciprocal_46, getitem_45, reciprocal_44, getitem_39, reciprocal_38, getitem_37, reciprocal_36, getitem_31, reciprocal_30, getitem_29, reciprocal_28, getitem_23, reciprocal_22, getitem_21, reciprocal_20, getitem_15, reciprocal_14, getitem_13, reciprocal_12, getitem_7, reciprocal_6, getitem_5, reciprocal_4, buf5, primals_147, buf0, buf1, buf2, buf3, buf4, buf8, buf24, buf43, buf94, buf113, buf164, buf183, buf233, buf252, buf304, buf323, buf375, buf394, buf11, 392, 768)
  File "/scratch/anijain/work/triton/python/triton/code_gen.py", line 999, in __call__
    return self.kernel(*wargs, **kwargs, grid=self.grid)
  File "/scratch/anijain/work/torchdynamo/torchinductor/triton_ops/autotune.py", line 50, in __call__
    timings = {
  File "/scratch/anijain/work/torchdynamo/torchinductor/triton_ops/autotune.py", line 51, in <dictcomp>
    config: self._bench(*args, config=config, **kwargs)
  File "/scratch/anijain/work/torchdynamo/torchinductor/triton_ops/autotune.py", line 42, in _bench
    return super()._bench(*args, config=config, **kwargs)
  File "/scratch/anijain/work/triton/python/triton/code_gen.py", line 1054, in _bench
    return triton.testing.do_bench(kernel_call)
  File "/scratch/anijain/work/triton/python/triton/testing.py", line 138, in do_bench
    torch.cuda.synchronize()
[...]

PR #1362, opened 2022-09-26
don't codegen constexpr for xnumel and rnumel

ngimel:
Tentative fix for #1358, that looks like triton bug. The same kernel either works or doesn't, depending on whether xnumel and rnumel are annotated as tl.constexpr. Note, this regresses perf a bit for static shapes (e.g. hf_Bert goes from 1.131 to 1.125) but correctness is more important, and we'd move to dynamic shapes soon anyway. 
Triton issue https://github.com/openai/triton/issues/714

### DIFF
--- a/torchinductor/codegen/triton.py
+++ b/torchinductor/codegen/triton.py
@@ -934,11 +934,7 @@ class TritonKernel(Kernel):
 
         argdefs, _ = self.args.python_argdefs()
 
-        if config.dynamic_shapes:
-            maybe_const = ""
-        else:
-            maybe_const = ": tl.constexpr"
-
+        maybe_const = ""
         for tree in self.range_trees:
             if tree.prefix != "r" or self.inside_reduction:
                 argdefs.append(f"{tree.prefix}numel{maybe_const}")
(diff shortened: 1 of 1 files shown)


torchinductor/codegen/triton.py
```diff
@@ -934,11 +934,7 @@ def codegen_kernel(self, name=None):
 
         argdefs, _ = self.args.python_argdefs()
 
-        if config.dynamic_shapes:
-            maybe_const = ""
-        else:
-            maybe_const = ": tl.constexpr"
-
+        maybe_const = ""
```
voznesenskym, line 937: its only used on L940, you can probably just rm the whole thing. 

Or leave it commented out with an explanation?

ngimel: Yeah I didn't remove it because we might want to go back to it, I'll leave a comment.

voznesenskym: We also have `argdefs.append(f"{tree.prefix.upper()}BLOCK : tl.constexpr")` - is that one alright?

ngimel: Yes those need to be constexprs and powers of 2 - triton codegen wouldn't work otherwise.

ngimel:
hm, test failure is really surprising - so sometimes `constexpr` is actually making things correct? (Not really correct, that model was failing accuracy check, so who knows what was happening, but at least no IMA). I'll look into this.

jansel:
This conflicts with the new runtime, but I also included this fix into the new runtime already -- so I think we can close this.

Issue #1203, opened 2022-09-12
Autograd function causes unnecessary graph break


suo:
Found on a quantized pyspeech model internally. The following code raises an exception:
```
import torch
import torchdynamo

class MyAutogradFunc(torch.autograd.Function):
    @staticmethod
    def forward(ctx, foo):
        return foo + foo

    @staticmethod
    def backward(ctx, grad_output):
        return grad_output

class Foo(torch.nn.Module):
    def __init__(self):
        super().__init__()

    def forward(self, foo):
        return MyAutogradFunc().apply(foo)

f = Foo()
@torchdynamo.optimize(nopython=True)
def fn():
    f(torch.ones(2, 3))

fn()
```

The exception is `torchdynamo.exc.Unsupported: call_function AutogradFunctionVariable() [] {}`.

Looking at the code in `misc.py`, changing the forward like this will work:
```
# don't instantiate the autograd func
return MyAutoGradFunc.apply(foo)
```

But both should work; seems like we're just missing a definition of `call_function` on `AutogradFunctionVariable`.

suo:
cc @voznesenskym heads up that I ran into this enabling models for the executorch september milestone (work-aroundable by patching upstream but would be nice to fix).

suo:
Another should-be-equivalent variation of `Foo` that doesn't work:
```

class Foo(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.fn = MyAutogradFunc.apply

    def forward(self, foo):
        return self.fn(foo)
```
produces:
```
torchdynamo.exc.Unsupported: call_function UserDefinedObjectVariable(apply) [TensorVariable()] {}
```

PR #1375, opened 2022-09-27
Fix autograd and class/static method related graph breaks

williamwen42:
Fix autograd graph breaks reported in #1203.

Also fix classmethod/staticmethod graph breaks that I found by playing around with the above autograd graph breaks.

### DIFF
--- a/test/test_misc.py
+++ b/test/test_misc.py
@@ -2536,6 +2536,30 @@ class MiscTests(torchdynamo.testing.TestCase):
 
         f2()  
 
+    def test_object_classmethod(self):
+        class C():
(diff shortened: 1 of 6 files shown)


test/test_misc.py
```diff
@@ -2520,6 +2520,74 @@ def fn():
         res = opt_fn()
         self.assertTrue(same(ref, res))
 
+    def test_autograd_function_equivalence(self):
```
mlazos, line 2523: Also check the result of these calls just to make sure it's working as we expect.

anijain2305 (approved):

PR #1381, opened 2022-09-27
Add support for torch/_tensor.py @_handle_torch_function_and_wrap_type_error_to_not_implemented marked methods

voznesenskym:
Tested with `compute-sanitizer --tool memcheck pytest test/test_torchinductor_opinfo.py -k rmatmul___cuda_float16`


mlazos:
Why can't these functions be built as TorchVariables? `TorchDunderFunctions` are exactly this [else clause](https://github.com/pytorch/torchdynamo/blob/ddec1bd1ae511ff6fe11e0ecf4d24e1656bcaf79/torchdynamo/variables/torch.py#L223)

torchdynamo/variables/builder.py
```diff
@@ -352,6 +353,11 @@ def index_source(key):
                 value, guards=make_guards(GuardBuilder.FUNCTION_MATCH)
             )
         elif istype(value, types.FunctionType):
+            if value in tensor_dunder_fns:
```
mlazos, line 356: Can we add these to the `is_allowed` list, and have it handled by the existing clause?

mlazos: Resolved offline, these are builtins which bridge to a torch op, I'm okay with special casing it here.

voznesenskym: discussed offline, will keep as is.

torchdynamo/variables/functions.py
```diff
@@ -16,6 +16,7 @@
 from ..source import AttrSource
 from ..source import GetItemSource
 from ..utils import make_cell
+from ..utils import proxy_args_kwargs
```
mlazos, line 19: Unused import?

mlazos (approved):

voznesenskym:
> This job appears to have stopped responding, try re-running it on the circleci job. Boo.

PR #1384, opened 2022-09-27
change mul lowering to allow triton to handle bool args

ngimel:
Per title

### DIFF
--- a/torchinductor/lowering.py
+++ b/torchinductor/lowering.py
@@ -2916,6 +2916,16 @@ def div(a, b):
     )
 
 
+@register_lowering([aten.mul], broadcast=True)
+def mul(a, b):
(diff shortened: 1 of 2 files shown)


torchinductor/lowering.py
```diff
@@ -3006,9 +3015,10 @@ def sum_(x, axis=None, keepdims=False, *, dtype=None):
 register_pointwise(aten.gt, type_promote=False, override_dtype=torch.bool)
 register_pointwise(aten.eq, type_promote=False, override_dtype=torch.bool)
 register_pointwise(aten.ne, type_promote=False, override_dtype=torch.bool)
-register_lowering(aten.__and__, type_promote=False)(
```
jansel, line 3009: Why remove `__and__`?

ngimel: It's not removed, it's moved one line down, to use `logical_and` definition.

voznesenskym (approved):

PR #1386, opened 2022-09-27
Use libdevice version for triton ops when possible

desertfire:


jansel (changes requested):
I think we can remove `has_triton_libdevice()`, we no longer support triton versions where it is false.

I believe the `tl.*` versions are faster, so we may want to benchmark this change.  If they are, we could put them behind a `config.approximations` flag.

desertfire:
> I think we can remove `has_triton_libdevice()`, we no longer support triton versions where it is false.
> 
> I believe the `tl.*` versions are faster, so we may want to benchmark this change. If they are, we could put them behind a `config.approximations` flag.

I have restructured this PR to only switch to libdevice ops for fp64 to unblock a few opinfo tests. For the performance vs approximation concern, let me create an issue for tracking and someone can work on it later.

desertfire:
https://github.com/pytorch/torchdynamo/issues/1518

jansel (approved):

PR #1390, opened 2022-09-27
Delaying reinplacing-decomps to lowering

anijain2305:
Currently few of the reinplacing-like decomps introduce mutation in the graph. This upsets dce and cse optimizations in Aot Autograd. So, moving such transformations from decomps to lowering.

Pending - tests. Not sure if op-level testing is done here.

### DIFF
--- a/torchinductor/lowering.py
+++ b/torchinductor/lowering.py
@@ -1560,6 +1560,11 @@ def index(x, indices):
     )
 
 
+@register_lowering([aten.index_put])
+def index_put(x, indices, values, accumulate=False):
+    return index_put_(clone(x), indices, values, accumulate)
+
+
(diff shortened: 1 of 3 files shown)


torchinductor/lowering.py
```diff
@@ -1560,6 +1560,11 @@ def fn(idx):
     )
 
 
+@register_lowering([aten.index_put])
```
ngimel, line 1563: please add a comment why this needs to be a lowering and not decomp

anijain2305:
@jansel @ngimel This is ready for review. I looked into `test_torchinductor.py` and we have tests for these ops there.

jansel (approved):

Issue #1353, opened 2022-09-26
Does not support channels-last Conv for CppCodeCache


EikanWang:
Repro

```python
import unittest
import contextlib
from unittest.mock import patch

import torch
import torchdynamo
from torchdynamo.testing import same
from torchdynamo.optimizations.log_args import conv_args_analysis

from torch.fx.passes.shape_prop import ShapeProp
import torch.fx as fx
from torchinductor.compile_fx import compile_fx
import torchinductor.config
from torchinductor import config

config.normalize_ir = True

class MockModule(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.conv = torch.nn.Conv2d(3, 3, 1, 1)
        self.relu = torch.nn.Sigmoid()
    def forward(self, x):
        return self.relu(self.conv(x))

class TestCase(unittest.TestCase):
    @classmethod
    def setUpClass(cls):
        cls._stack = contextlib.ExitStack()
        cls._stack.enter_context(patch.object(config, "debug", True))
        cls._stack.enter_context(patch.object(config.cpp, "min_chunk_size", 1))

    @classmethod
    def tearDownClass(cls):
        cls._stack.close()

class CppTests(TestCase):
    def test_channels_last(self):
        torchdynamo.reset()

        mod = MockModule()
        mod = mod.to(memory_format=torch.channels_last).eval()

        torchinductor.metrics.reset()
        @torchdynamo.optimize_assert(compile_fx)
        def fn(x):
            return mod(x)

        x = torch.randn(2, 3, 50, 60).to(memory_format=torch.channels_last)
        result = fn(x)
        assert same(result, mod(x))
```
For `CppCodeCache`, it would convert the input to channels-first anyway as follows.
```C++
extern "C" void kernel(const float* __restrict__ in_ptr0,
                       float* __restrict__ out_ptr0,
                       const long ks0)
{
    #pragma omp parallel
    {
        #pragma omp for collapse(3)
        for(long i0=0; i0<ks0; ++i0)
        {
            for(long i1=0; i1<3; ++i1)
            {
                for(long i2=0; i2<3000; ++i2)
                {
                    {
                        {
                            auto tmp0 = in_ptr0[i1 + (3*i2) + (9000*i0)];
                            out_ptr0[i2 + (3000*i1) + (9000*i0)] = tmp0;
                        }
                    }
                }
            }
        }
   ```

EikanWang:
[...]

PR #1394, opened 2022-09-28
torchinductor:fix conv2d channels_last issue for CPU backend

XiaobingSuper:
For torchinductor CPU path, there set a wrong layout format for conv2d channels_last path, which will assert stride mismatch error:
```
    def call(primals_1, primals_2, primals_3):
        primals_1_size = primals_1.size()
        s0 = primals_1_size[0]
        primals_3_size = primals_3.size()
        s1 = primals_3_size[0]
        s2 = primals_3_size[2]
        buf0 = aten.convolution(primals_3, primals_1, None, (1, 1), (0, 0), (1, 1), False, (0, 0), 1)
        assert buf0.size() == (s1, 3, 16, 16)
>       assert buf0.stride() == (768, 256, 16, 1)
E       AssertionError
```
This PR will fix https://github.com/pytorch/torchdynamo/issues/1353

### DIFF
--- a/torchinductor/ir.py
+++ b/torchinductor/ir.py
@@ -1518,6 +1518,12 @@ class Layout(IRNode):
                 return False
         return True
 
+    def is_channels_last_stride_ordered(self):
+        # create channels_last order(NCHW, NCDHW, the C is the first order).
+        order = [0] + list(reversed(range(1, len(self.stride) - 1)))
+        order = [len(order)] + order
+        return self.is_stride_ordered(order)
+
     def as_fixed(self):
         return FixedLayout(
             self.device,
@@ -3103,7 +3109,13 @@ class Convolution(ExternKernelAlloc):
                 x.get_dtype(),
             )
         else:
-            output_layout_str = "torch.contiguous_format"
+            # if x or weight have one channels_last format, it will call channels_last path,
+            # which align with aten.convolutuion path(only support 2d case).
+            if len(x.get_size()) == 4 and (x.get_layout().is_channels_last_stride_ordered() or \
+                weight.get_layout().is_channels_last_stride_ordered()):
+                output_layout_str = "torch.channels_last"
+            else:
+                output_layout_str = "torch.contiguous_format"
 
         if output_layout_str == "torch.channels_last":
             stride_order = [0] + list(reversed(range(1, len(kernel_size) + 1)))
--- a/torchinductor/ir.py
+++ b/torchinductor/ir.py
@@ -3109,13 +3109,21 @@ class Convolution(ExternKernelAlloc):
                 x.get_dtype(),
             )
         else:
-            # if x or weight have one channels_last format, it will call channels_last path,
-            # which align with aten.convolutuion path(only support 2d case).
-            if len(x.get_size()) == 4 and (x.get_layout().is_channels_last_stride_ordered() or \
-                weight.get_layout().is_channels_last_stride_ordered()):
+            output_layout_str = "torch.contiguous_format"
+            # If x or weight have one channels_last(2d or 3d) format, it will call channels_last path,
+            # which align with aten.convolutuion path(cpu only support 2d case now).
+            # TODO: after cpu 3d convolution support channels_last path, the size check can be removed.
+            # TODO: the gpu channels_last path depend on cudnn version, see
+            # https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/native/ConvUtils.h.
+            if (
+                x.get_device().type == "cpu"
+                and len(x.get_size()) == 4
+                and (
+                    x.get_layout().is_channels_last_stride_ordered()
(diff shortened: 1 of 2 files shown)


torchinductor/ir.py
```diff
@@ -3007,7 +3013,13 @@ def create(
                 x.get_dtype(),
             )
         else:
-            output_layout_str = "torch.contiguous_format"
+            # if x or weight have one channels_last format, it will call channels_last path,
+            # which align with aten.convolutuion path(only support 2d case).
+            if len(x.get_size()) == 4 and (x.get_layout().is_channels_last_stride_ordered() or \
```
EikanWang, line 3018: Please add TODO here as it does not support conv3d now.

test/test_torchinductor.py
```diff
@@ -1255,6 +1255,30 @@ def fn(x, w, b):
             check_lowp=False,
         )
 
+    def test_conv2d_channels_last(self):
+        m = torch.nn.Sequential(
+            torch.nn.Conv2d(3, 3, 1, 1),
+            ToTuple(),
+        )
+        m = m.to(memory_format=torch.channels_last)
+        x = torch.randn([2, 3, 16, 16]).to(memory_format=torch.channels_last)
```
EikanWang, line 1264: Since the `conv` runs into the channel-last path if the activation or weight is channel-last, please add more test cases to check only the activation or weight is the channels-lat.

test/test_torchinductor.py
```diff
@@ -1255,6 +1255,30 @@ def fn(x, w, b):
...
+            (torch.randn([2, 3, 16, 16]),),
+        )
+
+    def test_conv3d_channels_last(self):
+        m = torch.nn.Sequential(
+            torch.nn.Conv3d(3, 3, 1, 1),
+            ToTuple(),
+        )
+        m = m.to(memory_format=torch.channels_last_3d)
+        x = torch.randn([2, 3, 16, 16, 16]).to(memory_format=torch.channels_last_3d)
```
EikanWang, line 1276: Same as above

torchinductor/ir.py
```diff
@@ -1481,6 +1481,12 @@ def is_stride_ordered(self, order):
                 return False
         return True
 
+    def is_channels_last_stride_ordered(self):
+        # create channels_last order(NCHW, NCDHW, the C is the first order).
```
EikanWang, line 1485: What's the behavior of `conv1d`?

torchinductor/ir.py
```diff
@@ -1481,6 +1481,12 @@ def is_stride_ordered(self, order):
...
+        order = [0] + list(reversed(range(1, len(self.stride) - 1)))
+        order = [len(order)] + order
```
EikanWang, line 1487: Is the order correct? Regarding a nhwc tensor, the order will be [3, 0, 2, 1]. Is it as expected?

EikanWang (changes requested):

EikanWang:
Please fix the lint errors - `make format`

test/test_torchinductor.py
XiaobingSuper: added.

test/test_torchinductor.py
XiaobingSuper: added.

torchinductor/ir.py
XiaobingSuper: For conv1d case, aten.convolution always returns a continuous tensor even the input is NHC format.

torchinductor/ir.py
XiaobingSuper: Yes, it is expected. For NCHW, the stride order is [3, 2, 1, 0], but for the NHWC, the stride order is [3, 0, 2, 1].

torchinductor/ir.py
XiaobingSuper: Yes, added.

XiaobingSuper:
@Chillee , please help review this PR.

test/test_torchinductor.py
```diff
@@ -1255,6 +1256,50 @@ def fn(x, w, b):
             check_lowp=False,
         )
 
+    @unittest.skipIf(HAS_CUDA, "only support cpu channels_last")
```
jansel, line 1259: Is there a reason this needs to be CPU only?  Does it error if you enable it for GPU?

XiaobingSuper: This PR only enables CPU channels last, for GPU, it will be failed, for GPU path, I find that the channels last depend on cudnn version, see https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/native/ConvUtils.h, I need to find a method to check the version at python side, and then create another PR to fix GPU path.

Chillee (approved):
LGTM.

XiaobingSuper:
code rebased.

EikanWang (approved):

PR #1395, opened 2022-09-28
Create `assume_constant_result` decorator for specializing escape hatch

voznesenskym:
A far less intrusive approach to https://github.com/pytorch/torchdynamo/pull/1261, as discussed on the same PR


torchdynamo/eval_frame.py
```diff
@@ -556,6 +557,16 @@ def graph_with_interpreter(*args):
     return (new_graph, out_guards)
 
 
+def assume_constant_result(fn):
+    setattr(fn, "DYNAMO_MARKED_CONSTANT", True)
+    # @functools.wraps(fn)
+    # def inner(*args, **kwargs):
+    #     return fn(*args, **kwargs)
+    if config.fake_tensor_propagation:
+        unimplemented("Constant result capture is not supported with fake tensors.")
```
jansel, line 566: `unimplemented` it meant to be raised while symbolically evaluating, not from user facing APIS.

Let's raise an exception directly here.

torchdynamo/variables/functions.py
```diff
@@ -263,6 +282,23 @@ def call_function(
...
+def invoke_and_store_as_constant(tx, fn, name, source, options, args, kwargs):
+    def convert(x):
+        if isinstance(x, variables.TensorVariable):
+            return x.proxy.node.meta["example_value"]
+        return x.as_python_constant()
+
+    args = [convert(x) for x in args]
+    kwargs = {k: convert(x) for k, v in kwargs.items()}
+    res = fn(*args, **kwargs)
+    return tx.output.add_submodule(
```
jansel, line 294: This will only work for tensors I believe.  Add a test of a function that returns `.size(1)`.  Might need some logic similar to VariableBuilder.

torchdynamo/variables/functions.py
```diff
@@ -263,6 +282,23 @@ def call_function(
...
+        res,
+        name,
+        source=NNModuleSource(GetItemSource(source, name)),
```
jansel, line 297: This source isn't right.

`source` is how we can retrieve an input value from user data structures.

jansel (changes requested):

torchdynamo/eval_frame.py
voznesenskym: Yeah, that's a good point.

torchdynamo/variables/functions.py
voznesenskym: Correct. I was thinking that we should only support tensors in this API. That feels like it works better with the notions around when people would use this, and keeps the whole thing simpler w/r/t to just baking in a single constant result. Adding support for data structures and scalars can be added later if sorely needed. What do you think?

torchdynamo/variables/functions.py
voznesenskym: Good catch, ty.

torchdynamo/variables/functions.py
```diff
@@ -78,8 +79,17 @@ def closure_vars(self, tx):
 class UserFunctionVariable(BaseUserFunctionVariable):
```
mlazos, line 79: You're making some significant changes to this object, could we make a separate `ConstantFunctionVariable` type that we create in the builder if the right attributes are present?

mlazos:
I really want this for in effect ignoring a method that only changes some state that I don't care about,  can you elaborate on why this doesn't work with FakeTensors enabled? Also I would need to return `None` from the said method, is that supported? (I saw a comment that you only support tensors).

voznesenskym:
> I really want this for in effect ignoring a method that only changes some state that I don't care about, can you elaborate on why this doesn't work with FakeTensors enabled? Also I would need to return `None` from the said method, is that supported? (I saw a comment that you only support tensors).

We use example_value to invoke the function. With FakeTensors, we do not have the real values around to pass into said function. 

I am going to work on supporting other types.

torchdynamo/variables/functions.py
voznesenskym: We now support non-tensor types properly.

voznesenskym:
> I really want this for in effect ignoring a method that only changes some state that I don't care about, can you elaborate on why this doesn't work with FakeTensors enabled? Also I would need to return `None` from the said method, is that supported? (I saw a comment that you only support tensors).

I added an example of returning None from control flow in a test, give that a whirl :)

torchdynamo/variables/functions.py
voznesenskym: Well, `UserFunctionVariable` can be const, but so can `UserMethodVariable` and it felt a little gross to fork it into ConstantUserFunctionVariable and ConstantUserMethodVariable, but I am not strongly opposed.

torchdynamo/variables/functions.py
```diff
@@ -263,6 +281,22 @@ def call_function(
...
+def invoke_and_store_as_constant(tx, fn, name, options, args, kwargs):
+    def convert(x):
+        if isinstance(x, variables.TensorVariable):
+            return x.proxy.node.meta["example_value"]
+        return x.as_python_constant()
+
+    args = [convert(x) for x in args]
+    kwargs = {k: convert(v) for k, v in kwargs.items()}
+    res = fn(*args, **kwargs)
+    return tx.output.register_attr_or_module(
```
jansel, line 293: Does this handle constants correctly?

voznesenskym: Yes, let me add a test for that.

test/test_export.py
```diff
@@ -928,3 +928,420 @@ def compiler(gm, sample_inputs):
...
+        module = MyModule()
+        graph, _ = torchdynamo.export(module, torch.tensor([[0.0, 0], [0, 0]]))
+        result = graph(torch.tensor([[1.0, 0.0], [0, 0]]))
+        self.assertTrue(torchdynamo.utils.same(result, real_result))
+        result = graph(torch.tensor([[1, 0], [0.25, 0.25]]))
+        self.assertTrue(torchdynamo.utils.same(result, real_result))
+
+    @patch.object(torchdynamo.config, "fake_tensor_propagation", False)
+    def test_export_with_constant_free_function_and_class_method(self):
+        @torchdynamo.eval_frame.assume_constant_result
```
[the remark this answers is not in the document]
voznesenskym: SG wasn't sure if we want it as a top level API, or somewhere else like torchdynamo.specializing

torchdynamo/eval_frame.py
```diff
@@ -556,6 +556,14 @@ def graph_with_interpreter(*args):
     return (new_graph, out_guards)
 
 
+def assume_constant_result(fn):
+    setattr(fn, "DYNAMO_MARKED_CONSTANT", True)
```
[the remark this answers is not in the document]
voznesenskym: We can't dunder it to avoid name mangling rule. I'll do single.

documentation/GuardsOverviewPt1.md
```diff
@@ -152,7 +152,7 @@ A `ConstantVariable` is an instance of`VariableTracker`. `VariableTracker` repre
 When it comes to representing an object inside TorchDynamo, a VariableTracker does exactly what it says - it tracks a given variable. Its an extremely flexible class, but there are a few points to keep in mind:
 
 - It manages the `guard` relationship around the underlying object through:
-	- create_guard
+	- make_guard
```
mlazos, line 155: This rename is fine, but I'm curious why?

torchdynamo/source.py
```diff
@@ -227,3 +238,20 @@ def name(self):
...
+        return [codegen.create_load_global(self.source_name, add=False)]
+
+    def guard_source(self):
+        return GuardSource.CONSTANT
+
+    def name(self):
+        return self.source_name
+
+    def make_guard(self, fn, is_volatile=False):
+        raise NotImplementedError()
```
mlazos, line 257: Can you help me understand how ConstantSource works? In my head the impl for assume constant doesn't need any extra guards, but I feel like I'm missing something. Especially since I saw add_submodule in the call_function code.

documentation/GuardsOverviewPt1.md
voznesenskym: We had 2, create_guard and make_guard. Both in source.py. I picked the one we had less of and renamed it.

torchdynamo/source.py
voznesenskym: Exactly, constant does not need guards. We use the ConstantSource as a signal in source-handling code to avoid setting up guards, creating spurious input args, etc. See where we call `is_constant_source`

jansel (approved):

voznesenskym:
Thanks @jansel.

It's a deeper change than expected, so I am going to give it a pass on the bench to make sure we don't have any odd surprises. I do not expect any.

voznesenskym:
Looks fine on local bench.

Issue #1347, opened 2022-09-26
Contention in CppCodeCache on execution with multiple processes


jgong5:
Repro
```python
import torch
import torchdynamo

@torchdynamo.optimize()
def f(a):
    return a.softmax(dim=1) + a.sum()

a = torch.rand([100,100])
f(a)
```
Run the python script with the following bash script
```bash
python test_multi.py &
python test_multi.py &
python test_multi.py &
python test_multi.py &
python test_multi.py &
python test_multi.py &
python test_multi.py &
python test_multi.py &
python test_multi.py &
python test_multi.py &
python test_multi.py &
python test_multi.py &
python test_multi.py &
python test_multi.py &
python test_multi.py &
python test_multi.py &
wait
```
Error message dumped:
OSError: /tmp/torchinductor_jgong5/2u/c2uuffcbelf4a6jb5mb6dlxpunim7xtiox6artovnkozwiop5x3v.so: file too short

`CppCodeCache` should be implemented with a multi-process safe way.

voznesenskym:
That's fair, and timely because the question of this objects thread safety actually came up in this review: https://github.com/pytorch/torchdynamo/pull/1338#discussion_r979356341

Your example is somewhat contrived, however, and it's not readily obvious from it exactly where the thread safety issue arises. Would you be open to trying to make a smaller repro? If not, I can take a look.

jgong5:
@voznesenskym Good to know it was observed earlier. I won't say the case is contrived since running multiple instances at the same time is common in inference. I tried to simplify the repro a bit earlier but it cannot be reproduced sometimes due to the randomness. I believe making the compiled function bigger would make it more stable to repro (maybe add more aten calls to `f`). We are working on a fix.

voznesenskym:
> @voznesenskym Good to know it was observed earlier. I won't say the case is contrived since running multiple instances at the same time is common in inference. I tried to simplify the repro a bit earlier but it cannot be reproduced sometimes due to the randomness. I believe making the compiled function bigger would make it more stable to repro (maybe add more aten calls to `f`). We are working on a fix.

Sounds good, lmk if you need a hand. https://github.com/pytorch/torchdynamo/pull/1335 Is still under review, but you want to test with warmup + multiple inputs in a loop (closer to the inference case), this simple tool can help. 

[...]

PR #1400, opened 2022-09-28
Fix cache contention for multiple processes

Valentine233:
Fix cache contention for multiple processes

#1347


torchinductor/codecache.py
```diff
@@ -34,16 +35,20 @@ def code_hash(code):
...
-    if not os.path.exists(subdir):
-        os.makedirs(subdir, exist_ok=True)
-    path = os.path.join(subdir, f"{basename}.{ext}")
-    if not os.path.exists(path):
-        # use a temp file for thread safety
-        fd, tmp_path = tempfile.mkstemp(dir=subdir)
-        with os.fdopen(fd, "w") as f:
-            f.write(source_code)
-        os.rename(tmp_path, path)
+    if not os.path.exists(os.path.join(cache_dir(), "locks")):
```
voznesenskym, line 38: Is this process safe too? Could this possibly race?

Valentine233: I didn't encounter races when testing, but you are right. Is it better to just not create the directory "locks/"?

voznesenskym: I think any file manipulation you do needs to be locked, so I would just start with acquiring a file lock and trust that the lib does the right thing. If avoiding the dir helps us do that, by all means .

jgong5: Yep, I don't think you have to explicitly create the lock file before instantiating a `FileLock`. Instead, you can just instantiate the `FileLock` with the lock file name.

Valentine233:
I am wondering how to run workflows again. Could sb help, pls?

Chillee:
First time contributor needs somebody to manually approve.

voznesenskym:
Edited summary, hope you don't mind.

torchinductor/codecache.py
```diff
@@ -10,6 +10,7 @@
 import tempfile
 import types
 from ctypes import cdll
+from filelock import Timeout, FileLock
```
voznesenskym, line 13: Timeout unused?

torchinductor/codecache.py
```diff
@@ -34,16 +35,18 @@ def code_hash(code):
...
-    if not os.path.exists(subdir):
-        os.makedirs(subdir, exist_ok=True)
-    path = os.path.join(subdir, f"{basename}.{ext}")
-    if not os.path.exists(path):
-        # use a temp file for thread safety
-        fd, tmp_path = tempfile.mkstemp(dir=subdir)
-        with os.fdopen(fd, "w") as f:
-            f.write(source_code)
-        os.rename(tmp_path, path)
+    lock = FileLock(os.path.join(cache_dir(), basename + ".lock"))
```
voznesenskym, line 38: Consider Timeout here.

torchinductor/codecache.py
voznesenskym, line 13: From the docs:

```
All log messages by this library are made using the DEBUG_ level, under the filelock name. On how to control displaying/hiding that please consult the [logging documentation of the standard library](https://docs.python.org/3/howto/logging.html). E.g. to hide these messages you can use:

logging.getLogger("filelock").setLevel(logging.INFO)
```

Not super important, but consider modifying the log level to match whatever is stipulated in config.

voznesenskym:
LGTM, will stamp when CI passes.

torchinductor/codecache.py
```diff
@@ -155,14 +163,16 @@ class PyCodeCache:
...
-        if key not in cls.cache:
-            with open(path) as f:
-                code = compile(f.read(), path, "exec")
-                mod = types.ModuleType(f"{__name__}.{key}")
-                mod.__file__ = path
-                exec(code, mod.__dict__, mod.__dict__)
-                cls.cache[key] = mod
-                cls.cache[key].key = key
+        lock = FileLock(os.path.join(cache_dir(), key + ".lock"), timeout=600)
+        with lock:
```
jgong5, line 167: We don't have to do the lock here since the shared data are read only, right?

torchinductor/codecache.py
```diff
@@ -34,16 +36,18 @@ def code_hash(code):
...
-    if not os.path.exists(subdir):
-        os.makedirs(subdir, exist_ok=True)
-    path = os.path.join(subdir, f"{basename}.{ext}")
-    if not os.path.exists(path):
-        # use a temp file for thread safety
-        fd, tmp_path = tempfile.mkstemp(dir=subdir)
-        with os.fdopen(fd, "w") as f:
-            f.write(source_code)
-        os.rename(tmp_path, path)
+    lock = FileLock(os.path.join(cache_dir(), basename + ".lock"), timeout=600)
```
jgong5, line 39: Suggest to avoid using magic number `600` here. Use a constant var instead.

jgong5 (changes requested):

jgong5:
@Valentine233 Can you add UT to cover the fix?

test/test_multi.sh
```diff
@@ -0,0 +1,19 @@
...
+python test_multi.py &
+python test_multi.py &
+python test_multi.py &
+python test_multi.py &
+python test_multi.py &
+python test_multi.py &
+python test_multi.py &
+python test_multi.py &
+python test_multi.py &
+wait
```
jgong5, line 19: @voznesenskym Do you know how to incorporate such a test into the CI?

voznesenskym: Multi process testing is tricky. You'd have to do a lot of mocking, and ensure that it reliably fails 100% of the time without this fix, or the flaky nature of the test won't give us the protection we want from a test.

jgong5: @voznesenskym I agree with your point that it might not provide 100% guarantee of protection but if the UT is created from the real usages, like multi-process inference, it would provide a good level of confidence that we are good to support such usages. On the other hand, if we already have such multi-process tests in the integration tests, we probably don't have to add a similar UT here. I'm open for better suggestions.

jgong5: @Valentine233 If we decide to add such a test, please make sure the test has good result checkers, e.g. the compiled function of each process should return expected result, the cache is cleared before tests etc.

torchinductor/codecache.py
```diff
@@ -38,16 +42,18 @@ def code_hash(code):
...
-    if not os.path.exists(subdir):
-        os.makedirs(subdir, exist_ok=True)
-    path = os.path.join(subdir, f"{basename}.{ext}")
-    if not os.path.exists(path):
-        # use a temp file for thread safety
-        fd, tmp_path = tempfile.mkstemp(dir=subdir)
-        with os.fdopen(fd, "w") as f:
-            f.write(source_code)
-        os.rename(tmp_path, path)
+    lock = FileLock(os.path.join(cache_dir(), basename + ".lock"), timeout=LOCK_TIMEOUT)
```
jansel, line 45: I don't think this lock is needed.

`os.rename(tmp_path, path)` is atomic, so even if two threads same write the file it will be correct.

jansel (changes requested):

torchinductor/codecache.py
Valentine233: > I don't think this lock is needed.
> 
> `os.rename(tmp_path, path)` is atomic, so even if two threads write the same file it will be correct.

@jansel Thanks for your advice. Will `if not os.path.exists(subdir):  os.makedirs(subdir, exist_ok=True)` encounter races here?

Valentine233: Maybe it would be better to use this?
`if not os.path.exists(subdir):
     try:
         os.makedirs(subdir, exist_ok=True)
     except:
         pass`

jgong5: > Will if not os.path.exists(subdir): os.makedirs(subdir, exist_ok=True) encounter races here?

I don't think it is a problem since `exist_ok=True` is provided.

test/test_multi.sh
jgong5: @voznesenskym Any comments on the test? How about we add multi-instance inference tests into the integration test later and remove such UTs here? We can add an issue to track this.

voznesenskym: I don't have a problem with the idea of testing - I do not know how we can make a test that fails 100% of the time without this fix. If we have a test that can give false negatives (regression, but test still passes) - that is not a test worth having.

torchinductor/codecache.py
```diff
@@ -139,19 +148,24 @@ class CppCodeCache:
...
-                try:
-                    subprocess.check_output(cmd, stderr=subprocess.STDOUT)
-                except subprocess.CalledProcessError as e:
-                    raise exc.CppCompileError(cmd, e.output)
-
-            cls.cache[key] = cdll.LoadLibrary(output_path)
-            cls.cache[key].key = key
+        lock_dir = os.path.join(cache_dir(), "locks")
+        if not os.path.exists(lock_dir):
+            os.makedirs(lock_dir, exist_ok=True)
```
jgong5, line 153: Duplicate logic of creating the lock_dir. How about abstracting a function for this?

jgong5 (approved):
Please file an issue for the test to cover multiple instance inference scenario.

torchinductor/codecache.py
```diff
@@ -63,15 +74,18 @@ def cpp_compiler():
...
-            if cxx is None:
-                cxx = install_gcc_via_conda()
-            subprocess.check_output([cxx, "--version"])
-            return cxx
-        except (subprocess.SubprocessError, FileNotFoundError):
-            continue
-    raise exc.InvalidCxxCompiler()
+    lock_dir = get_lock_dir()
+    lock = FileLock(os.path.join(lock_dir, "g++.lock"), timeout=LOCK_TIMEOUT)
+    with lock:
```
jansel, line 79: I think the lock should only be needed around `install_gcc_via_conda()`.  If we are using an existing compiler, it should be fine to test it in parallel.

torchinductor/codecache.py
```diff
@@ -139,19 +153,22 @@ class CppCodeCache:
...
-                try:
-                    subprocess.check_output(cmd, stderr=subprocess.STDOUT)
-                except subprocess.CalledProcessError as e:
-                    raise exc.CppCompileError(cmd, e.output)
-
-            cls.cache[key] = cdll.LoadLibrary(output_path)
-            cls.cache[key].key = key
+        lock_dir = get_lock_dir()
+        lock = FileLock(os.path.join(lock_dir, key + ".lock"), timeout=LOCK_TIMEOUT)
+        with lock:
```
jansel, line 158: Should we have a fast path that doesn't acquire the lock on a cache hit?

torchinductor/codecache.py
Valentine233: Yes, make sense. I will move the lock before `install_gcc_via_conda()`.

jansel (approved):

PR #1405, opened 2022-09-28
Generate TARGETS file for inductor kernels in fbcode

### LINKED ISSUE #1097: Generate buck targets for inductor, inductor generated kernels are not runnable in fbcode atm
This needs to:

1) Redirect the compile cache somewhere TARGETS works
2) Produce a valid TARGETS file so we can buck run the generated py file
3) Probably get an e2e test in fbcode

bertmaher:
Fixes #1097

### DIFF
--- a/torchinductor/debug.py
+++ b/torchinductor/debug.py
@@ -327,3 +327,23 @@ class DebugFormatter:
 
     def output_code(self, filename):
         shutil.copy(filename, self.filename("output_code.py"))
+        V.debug.buck_targets()
+
+    def buck_targets(self):
+        with self.fopen("TARGETS") as fd:
+            fd.write("""
+load("@fbcode_macros//build_defs:python_binary.bzl", "python_binary")
+
+python_binary(
+    name = "output_code",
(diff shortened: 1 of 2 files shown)


torchinductor/debug.py
```diff
@@ -327,3 +327,23 @@ def graph_diagram(self, nodes: SchedulerNodeList):
 
     def output_code(self, filename):
         shutil.copy(filename, self.filename("output_code.py"))
+        V.debug.buck_targets()
+
+    def buck_targets(self):
+        with self.fopen("TARGETS") as fd:
+            fd.write("""
```
jansel, line 334: Suggest `textwrap.dedent()` so you can indent this a bit nicer.

jansel (commented):
`make format` should fix lint issues

voznesenskym (changes requested):
We are missing the C++ side of things here (Headers, .so). I am also not sure we generate a single file. a `glob` rule for the out dir may cover things better.

Where is it writing TARGETS to atm?

jansel:

PR #1258, opened 2022-09-16
Fix inplace ops type promotion and graph outputs

ngimel:
This fixes #1245 and causes inplace ops to return `self` rather than a copy of self, but I'm not sure if that's the correct fix

PR #1406, opened 2022-09-28
[inductor] Fix input mutation issue

### LINKED ISSUE #1374: Mutation bug
Modifying an output after use creates a cyclical reference. E.g in the code below:
```

def forward(self, arg0_1, arg1_1, arg2_1):
    relu = torch.ops.aten.relu.default(arg2_1);  arg2_1 = None
    permute = torch.ops.aten.permute.default(arg0_1, [1, 0])
    addmm = torch.ops.aten.addmm.default(arg1_1, relu, permute);  permute = None
   ....
    copy__1 = torch.ops.aten.copy_.default(arg1_1, add_1);  arg1_1 = add_1 = None
    return ()

```
This triggers a bug in the scheduler while computing dependencies.

Chatted with @jansel about this and he suggested I remove this code block: 
https://github.com/pytorch/torchdynamo/blob/64980a6c55e39f4a727f5b16b781478167300c59/torchinductor/lowering.py#L2853

Removing the code above fixed the issue.

jansel:
Fixes #1374

### DIFF
--- a/test/test_torchinductor.py
+++ b/test/test_torchinductor.py
@@ -2443,7 +2443,8 @@ class CommonTemplate:
 
     def test_input_mutation4(self):
         def fn(a):
-            return torch.relu_(a)
+            torch.relu_(a)
+            return a
 
         arg1 = torch.randn([1, 64], device=self.device)
         arg2 = arg1.clone()
--- a/test/test_torchinductor.py
+++ b/test/test_torchinductor.py
@@ -2441,6 +2441,19 @@ class CommonTemplate:
         self.assertTrue(same(actual1, correct1))
         self.assertTrue(same(arg1, arg2))
 
+    def test_input_mutation4(self):
+        def fn(a):
+            return torch.relu_(a)
+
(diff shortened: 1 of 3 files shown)


test/test_torchinductor.py
```diff
@@ -2441,6 +2441,19 @@ def fn(a):
         self.assertTrue(same(actual1, correct1))
         self.assertTrue(same(arg1, arg2))
 
+    def test_input_mutation4(self):
+        def fn(a):
+            return torch.relu_(a)
```
ngimel, line 2446: but this is already working? We can return the input and it is propely mutated

jansel: I'll change it to do that.

ngimel: I meant, it already works today, https://github.com/pytorch/torchdynamo/pull/1258/files#diff-7ed3e49a7cfe8882d728601f53c7fa4cc2252b2ac160d140cb99167dd69ff551R2238 is pretty much the same test

ngimel (approved):

PR #1407, opened 2022-09-28
[inductor] Add decomp for bernoulli

jansel:
Something in upstream PyTorch both broke the low memory dropout stuff in `overrides.py` and changed dropout to map to `bernoulli` rather than `bernoulli_`.


Chillee:
Possibly functionalization being turned on?

Chillee (approved):

PR #1408, opened 2022-09-28
update the install instructions on the readme and install_requires

albanD:
Since the "binary" install is based on a git install, only the packages listed as install_requires are actually installed. So we need them there.


README.md
```diff
@@ -37,20 +37,27 @@ bugs and rough edges.
...
 Python 3.8 is recommended.
 Python 3.7 through 3.10 are supported and tested.
+Make sure to have a development version of python installed locally as well.
 
-TorchDynamo requires the latest development/nightly build of PyTorch or
-building [PyTorch from source]. TorchDynamo also requires [functorch]
-from the PyTorch repository.  The Makefile target `make setup_nightly_gpu`
-contain the commands used by our CI to setup dependencies.
+To use GPU back ends (and in particular Triton), please make sure to have cuda
+11.7 installed locally.
```
jansel, line 43: 11.6 also should work.  Not sure about other versions.

albanD: Since I listed the binaries for 11.7 and don't want people to get into trouble with the cuda toolchain not matching the install I left 11.7 here.
Do you want me to move this into the "binary" install and mention in the from source that the user should have 11.6+ locally?

albanD: Updated the PR to reflect this.

jansel: It might be good to drop a link for other versions.

jansel (approved):

albanD:
FYI @desertfire after this PR, you will need to update the Readme as well when you update the CI pin of PyTorch.

desertfire:
> FYI @desertfire after this PR, you will need to update the Readme as well when you update the CI pin of PyTorch.

Are you referring to the shell example in the `Install nightly binaries` section? I thought that's just an example, so you can put something general in the description, like TODAY_DATE with an example date?

albanD:
@desertfire I think there are two things:
- we're not always following today's date. And the triton is pinned to a specific hash. so there isn't really a generic command we can give
- the install instructions are best when user can just copy/paste them into their terminal and they "just work" I think.

PR #1409, opened 2022-09-29
Find conda from environment

bertmaher:
conda sometimes installs itself as a shell function instead of a legit binary, and puts the binary in the CONDA_EXE.  This diff lets us actually find conda in this case.

### DIFF
--- a/torchinductor/codecache.py
+++ b/torchinductor/codecache.py
@@ -5,6 +5,7 @@ import hashlib
 import logging
 import os
 import re
+import shutil
 import subprocess
 import sysconfig
 import tempfile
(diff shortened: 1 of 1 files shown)


voznesenskym (approved):

torchinductor/codecache.py
```diff
@@ -81,7 +81,7 @@ def install_gcc_via_conda():
         log.info("Downloading GCC via conda")
         subprocess.check_call(
             [
-                "conda",
+                os.environ.get("CONDA_EXE", "conda"),
```
voznesenskym, line 84: consider checking the path if this returns None.

```
conda =  os.environ.get("CONDA_EXE", "conda")
if conda is None:
    conda = shutil.which('conda')
```

voznesenskym: Also consider checking that its there after both said checks, before dropping into subprocess.check_call.

bertmaher: Done -- although I had to make slightly more extensive changes to the surrounding logic (the code was sort of assuming that the subprocess call would fail, and if I avoided the subprocess call, I didn't have a way to signal failure!)

jansel (approved):

PR #1410, opened 2022-09-29
Fix CUDA Error- context is destroyed

jansel:
Running our standalone launchers at the end of generated code gave a `CUDA: Error- context is destroyed` error.

This fixes that, though I am not sure why the error happened.


torchinductor/codecache.py
```diff
@@ -192,6 +193,9 @@ def load(cls, source_code):
 
 
 class AsyncCompile:
+    def __init__(self):
+        self._context_keepalive = None
```
voznesenskym, line 197: In C++, I would recommend RAII and making a clear object whose job it is to maintain a 1:1 lifecycle with the cuda context.

However, looking at the docs, it seems like this api is undocumented and also not called by us directly. So I guess this is the best we can do.

voznesenskym (approved):

voznesenskym:
@jansel can you run `compute-sanitizer --tool memcheck $whatever_command_you_ran_to_see_this_error`? I wonder if have a teardown/destruction order race where a context can sometimes be accessed after/during teardown.

jansel:
I don't think this is a race.  I think it is that Triton's compiler wants an active context to work.

PR #1412, opened 2022-09-29
[inductor] Fix alignment issue with new runtime

jansel:
This fixes a minor issue in #1338 that only shows up in one internal model.

### DIFF
--- a/torchinductor/codegen/triton.py
+++ b/torchinductor/codegen/triton.py
@@ -47,14 +47,15 @@ def signature_of(arg):
 
 
 def config_of(args):
-    from triton.runtime.jit import JITFunction
+    from ..compile_fx import ALIGNMENT
 
-    divisible_by_16 = [
-        i
-        for i, arg in enumerate(args)
-        if isinstance(arg, TensorArg)
-        or V.graph.sizevars.maybe_guard_multiple_of(arg.expr, JITFunction.divisibility)
-    ]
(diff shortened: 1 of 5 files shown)


soumith:
add a test?

jansel:
> add a test?

I was trying to make a more minimal repro for this but having a hard time.  Passing the wrong alignment hints to Triton don't always produce invalid code.

torchinductor/codegen/triton.py
```diff
@@ -47,14 +47,15 @@ def signature_of(arg):
...
+        assert isinstance(x, SizeArg)
+        return V.graph.sizevars.maybe_guard_multiple_of(x.expr, ALIGNMENT)
 
-    divisible_by_16 = [
-        i
-        for i, arg in enumerate(args)
-        if isinstance(arg, TensorArg)
-        or V.graph.sizevars.maybe_guard_multiple_of(arg.expr, JITFunction.divisibility)
-    ]
+    divisible_by_16 = [i for i, arg in enumerate(args) if is_aligned(arg)]
```
ngimel, line 58: lol the place I worried about turned out to be problematic

ngimel (approved):

torchinductor/codegen/triton.py
jansel: Yeah it did :)

Issue #1404, opened 2022-09-28
Inductor cuda codegen generating wrong argument name 


SherlockNoMad:
Hi folks, I found that the inductor codegen for the ’tmp’ variable name is stateful accross unit tests

For example, when I run all the cuda opinfo tests all together, neg_cuda_float32 is generating a tmp42, which is not defined in the context. 
```
FAILED OP neg on cuda with torch.float32, at 7:59:
def kernel(in_ptr0, out_ptr0, xnumel, XBLOCK : tl.constexpr):
    xoffset = tl.program_id(0) * XBLOCK
    xindex = xoffset + tl.reshape(tl.arange(0, XBLOCK), [XBLOCK])
    xmask = xindex < xnumel
    x0 = xindex
    tmp0 = tl.load(in_ptr0 + x0, xmask)
    tl.store(out_ptr0 + x0 + tl.zeros([XBLOCK], tl.int32), tmp42, xmask)
                                                           ^
```

However, when I try to run this single test in isolation, I am getting a correct kernel, 
```
@pointwise_heuristics(size_hints=[512], filename=__file__)
@triton.jit
def kernel(in_ptr0, out_ptr0, xnumel, XBLOCK : tl.constexpr):
    xoffset = tl.program_id(0) * XBLOCK
    xindex = xoffset + tl.reshape(tl.arange(0, XBLOCK), [XBLOCK])
    xmask = xindex < xnumel
    x0 = xindex
    tmp0 = tl.load(in_ptr0 + x0, xmask)
    tmp1 = -tmp0
    [tl.store](http://tl.store/)(out_ptr0 + x0 + tl.zeros([XBLOCK], tl.int32), tmp1, xmask)
).kernel
```
Notice that the wrong kernel missing the compute line ‘tmp1 = -tmp0’ 

I am wondering if there anything stateful thing about the variable name codegen in Inductor?

desertfire:
https://github.com/pytorch/torchdynamo/blob/3eb350f6870f01e42918df245318a78c8b5d016b/torchinductor/codegen/common.py#L435 is how a new tmp variable is named. I don't think `CSE` should be carried between tests. Can you give `torchdynamo.config.raise_on_assertion_error=True` a try to see if the problem still exists?

voznesenskym:
[...]

PR #1417, opened 2022-09-29
Fix leaked inductor internal information seen in the opinfo test

desertfire:
Summary:
1) the context manager fix in virtualized.py solves https://github.com/pytorch/torchdynamo/issues/1404
2) set `torchdynamo.config.raise_on_assertion_error = True`
3) Update the opinfo CPU skip and expect list, and now the failures look more real. CUDA skip list is coming in another PR.


test/test_torchinductor_opinfo.py
```diff
@@ -847,4 +709,7 @@ def fn(*args, **kwargs):
 instantiate_device_type_tests(TestInductorOpInfo, globals())
 
 if __name__ == "__main__":
+    raise_on_assertion_error = torchdynamo.config.raise_on_assertion_error
+    torchdynamo.config.raise_on_assertion_error = True
     run_tests()
+    torchdynamo.config.raise_on_assertion_error = raise_on_assertion_error
```
jansel, line 715: No need to restore this.  If we are in `__main__` the process is about to exit.

jansel (approved):

Issue #1267, opened 2022-09-19
pytorch core test failure: AssertionError: bool indices are not supported yet


desertfire:
Repro:
```
PYTORCH_TEST_WITH_INDUCTOR=1   pytest test/test_indexing.py -k test_empty_index_cuda
```
Error:
```
Traceback (most recent call last):
  File "/fsx/users/binbao/pytorch/test/test_indexing.py", line 962, in test_empty_index
    def test_empty_index(self, device):
  File "/fsx/users/binbao/pytorch/test/test_indexing.py", line 965, in <graph break in test_empty_index>
    self.assertEqual(x[idx].numel(), 0)
  File "/fsx/users/binbao/pytorch/test/test_indexing.py", line 970, in <graph break in test_empty_index>
    self.assertEqual(x, y)
  File "/fsx/users/binbao/torchdynamo-tip/torchdynamo/eval_frame.py", line 167, in _fn
    return fn(*args, **kwargs)
  File "/fsx/users/binbao/pytorch/functorch/_src/aot_autograd.py", line 904, in forward
    return compiled_f(
  File "/fsx/users/binbao/pytorch/functorch/_src/aot_autograd.py", line 890, in new_func
    compiled_fn = create_aot_dispatcher_function(
  File "/fsx/users/binbao/torchdynamo-tip/torchdynamo/utils.py", line 76, in time_wrapper
    r = func(*args, **kwargs)
  File "/fsx/users/binbao/pytorch/functorch/_src/aot_autograd.py", line 468, in create_aot_dispatcher_function
    return aot_dispatch_base(flat_fn, fake_flat_tensor_args, aot_config)
  File "/fsx/users/binbao/pytorch/functorch/_src/aot_autograd.py", line 278, in aot_dispatch_base
    compiled_fw = aot_config.fw_compiler(fw_module, flat_args)
  File "/fsx/users/binbao/pytorch/functorch/_src/aot_autograd.py", line 237, in f
    out_f = compiler(fx_g, inps)
  File "/fsx/users/binbao/torchdynamo-tip/torchdynamo/utils.py", line 76, in time_wrapper
    r = func(*args, **kwargs)
  File "/fsx/users/binbao/torchdynamo-tip/torchinductor/compile_fx.py", line 246, in fw_compiler
    return compile_fx_inner(
  File "/fsx/users/binbao/torchdynamo-tip/torchdynamo/debug_utils.py", line 332, in debug_wrapper
[...]

PR #1419, opened 2022-09-29
[Inductor] Support aten.index_put_ with bool index

sangongs:
When the index list is just a single bool index and the value to assign has only one element, we dispatch `aten.index_put_` as masked fill. Otherwise, we fallback.

See the discussion in issue #1267.

Test plan:
- Unit tests
- `PYTORCH_TEST_WITH_INDUCTOR=1   pytest test/test_indexing.py -k test_empty_index_cuda`

### DIFF
--- a/test/test_torchinductor.py
+++ b/test/test_torchinductor.py
@@ -2679,12 +2679,22 @@ class CommonTemplate:
 
         self.common(
             fn,
-            [
+            (
+                torch.randn([1024, 4, 2]),
+                torch.randn([1024, 4, 2]) > 0,
+                torch.randn([]),
+                False,
+            ),
+        )
+
+        self.common(
+            fn,
+            (
                 torch.randn([1024, 4, 2]),
                 torch.randn([1024, 4, 2]) > 0,
                 torch.randn([]),
                 True,
-            ],
+            ),
         )
 
     def test_index_put_fallback1(self):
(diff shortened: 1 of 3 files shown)


torchinductor/lowering.py
```diff
@@ -1578,8 +1578,34 @@ def index_put(x, indices, values, accumulate=False):
     return index_put_(clone(x), indices, values, accumulate)
 
 
+def index_put_as_masked_fill(self, indices, value, accumulate):
+    if value.get_device() != self.get_device():
+        value = to_device(value, self.get_device())
+    if accumulate:
+        value = add(self, value)
```
ngimel, line 1585: You should add a test for `accumulate` = True

test/test_torchinductor.py
```diff
@@ -2671,6 +2671,55 @@ def fn(a, b, c):
...
+
+    def test_index_put_fallback1(self):
+        def fn(a, b, c, d):
+            a = a.clone()
+            torch.ops.aten.index_put_(a, [b], c, d)
+            return a
+
+        self.common(
+            fn,
+            [
```
ngimel, line 2698: nit: use tuples, not lists, for args

test/test_torchinductor.py
```diff
@@ -2671,6 +2671,55 @@ def fn(a, b, c):
...
+                torch.randn([3]),
+                torch.as_tensor([True, True, False]),
+                torch.randn([2]),
+                random.choice([True, False]),
```
ngimel, line 2702: just test both cases

test/test_torchinductor.py
```diff
@@ -2671,6 +2671,55 @@ def fn(a, b, c):
...
+            return a
+
+        self.common(
+            fn,
+            [
+                torch.randn([1, 2, 3]),
+                torch.as_tensor([0, 1]),
+                torch.as_tensor([True, True, False]),
+                torch.randn([]),
+                random.choice([True, False]),
```
ngimel, line 2719: again, just test both cases

PR #1425, opened 2022-09-30
try removing some skips

ngimel:
Locally they are passing

### DIFF
--- a/test/test_torchinductor_opinfo.py
+++ b/test/test_torchinductor_opinfo.py
@@ -136,10 +136,6 @@ inductor_skips["cuda"] = {
     "log2": {b8, i32, i64},
     "logsumexp": {f64},
     "lu_unpack": {f32, f64},  # RuntimeError: CUDA error
-    "nan_to_num": {
-        i32,
-        i64,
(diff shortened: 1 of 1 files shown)


Chillee (approved):

test/test_torchinductor_opinfo.py
```diff
@@ -175,11 +170,9 @@ def process(device_type):
     "scatter_reduce.mean": {f16, f32, f64, i32, i64},  # segfault
     "scatter_reduce.prod": {f16, f32, f64, i32, i64},  # segfault
     "scatter_reduce.sum": {b8, f16, f32, f64, i32, i64},  # segfault
-    "softmax": {b8, f16, f32, f64, i32, i64},  # segfault
+    "softmax": {b8, f64, i32, i64},  # segfault
```
soumith, line 173: what's happening for softmax and f16, f32?
can you track that issue somewhere if it isn't already

ngimel: fp16 and fp32 are passing, I removed them from skips. For integer inputs something wrong is going on with type promotion I think, for fp64 triton frequently has trouble generating code.

Issue #1359, opened 2022-09-26
TorchInductor OpInfo accuracy failure Tracker


SherlockNoMad:
Following OpInfo tests are failing due to accuracy issue:

To reproduce the error, run 
```
pytest -sv  test/test_torchinductor_opinfo.py -k test_comprehensive_<opname>_<device_type>_<dtype>
```

cc @anijain2305 

- [ ] _masked.argmax	cuda	torch.float16	
- [ ] _masked.argmax	cuda	torch.float32	
- [ ] _masked.argmax	cuda	torch.float64	
- [ ] _masked.argmax	cuda	torch.int32	
- [ ] _masked.argmax	cuda	torch.int64	
- [ ] _masked.argmin	cuda	torch.float16	
- [ ] _masked.argmin	cuda	torch.float32	
- [ ] _masked.argmin	cuda	torch.float64	
- [ ] _masked.argmin	cuda	torch.int32	
- [ ] _masked.argmin	cuda	torch.int64	
- [ ] _masked.cumprod	cuda	torch.float16	
- [ ] _masked.cumsum	cuda	torch.float16	
- [ ] add	cuda	torch.bool	
- [ ] addbmm	cuda	torch.float16	
- [ ] addmv	cuda	torch.float16	
- [ ] addr	cuda	torch.float16	
- [ ] argmax	cuda	torch.float16	
- [ ] argmax	cuda	torch.float32	
- [ ] argmax	cuda	torch.float64	
- [ ] argmax	cuda	torch.int32	
- [ ] argmax	cuda	torch.int64	
- [ ] argmin	cuda	torch.float16	
- [ ] argmin	cuda	torch.float32	
- [ ] argmin	cuda	torch.float64	
- [ ] argmin	cuda	torch.int32	
- [ ] argmin	cuda	torch.int64	
- [ ] baddbmm	cuda	torch.float16	 
- [ ] cross	cuda	torch.float16	
- [ ] cumsum	cuda	torch.float16	
- [ ] diff	cuda	torch.float16	
- [ ] dist	cuda	torch.float16	
- [ ] linalg.cholesky	cuda	torch.float32	
- [ ] linalg.cholesky	cuda	torch.float64	
- [ ] linalg.cholesky_ex	cuda	torch.float32	
- [ ] linalg.cholesky_ex	cuda	torch.float64	
- [ ] linalg.cross	cuda	torch.float16	
- [ ] linalg.lu_factor	cuda	torch.float32	
[...]

PR #1430, opened 2022-09-30
Fix type conversion issues for logical ops

desertfire:
Summary: https://github.com/pytorch/torchdynamo/issues/1359

Test Plan:
PYTORCH_ALL_SAMPLES=1 pytest test/test_torchinductor_opinfo.py -k test_comprehensive_logical


desertfire:
I also renamed some flags in `lowering.py` to make them more meaningful.

desertfire:
> Looks good, if CI agrees.

Hmm, somehow CI gives a different result on fft.ifft and fft.ihfft. Reverting anyway.

eellison (commented):
any thought of using prims here ? that's their intended purpose right

jansel (approved):

ngimel:
It is, but prims bundle explicit broadcasting and type promotion and we need only the latter.

PR #1431, opened 2022-09-30
Re enable batch norm decomp on cpu

eellison:
Previously I had to disable the decomp because of a FakeTensor bug. That has since been fixed (I think - double checking with this PR, works locally).


desertfire:
Nice, so what actually fixes this? Something in fake tensor?

desertfire (approved):

PR #1435, opened 2022-09-30
Fix lowering of ceil/floor/round/trunc for int dtypes

SherlockNoMad:
When input is integer type, lowering for ceil and floor is a nop. 

tl.libdevice.ceil doesn't support integer type.


torchinductor/lowering.py
```diff
@@ -490,6 +490,22 @@ def squeeze_(x, dim=None):
     return x
 
 
+@register_lowering(aten.ceil)
+def ceil(x):
+    if is_integer_type(x):
+        return x
+    fn = ops_wrapper(aten.ceil.__name__)
```
voznesenskym, line 497: we already have ops.floor and ops.ciel, would it suffice to use those instead?

```
@make_pointwise
def ceil(x):
    return ops.ceil(x)
```

SherlockNoMad: tl.libdevice.ceil doesn't support int types 

we have this dtype specific lower to avoid calling op.ceil for int type.

voznesenskym: ah ok, that lgtm then, let @ngimel take a pass at it too

torchinductor/lowering.py
```diff
@@ -490,6 +490,22 @@ def squeeze_(x, dim=None):
...
+    return make_pointwise(fn)(x)
+
+
+@register_lowering(aten.floor)
+def floor(x):
+    if is_integer_type(x):
+        return x
+    fn = ops_wrapper(aten.floor.__name__)
```
jansel, line 505: Let's just use `"floor"` directly.  I don't think the `__name__` is making the code more generic, since if it ever changed it would break things.

SherlockNoMad: fixed.

jansel (approved):

ngimel (approved):

Issue #1243, opened 2022-09-15
Need easier way to tell which step of the optimized path fails (dynamo + inductor)


yf225:
During internal benchmark enablement, there are multiple times where Dynamo silently starts to not generate any graph and not calling Inductor, but the workflow doesn't throw any error because Dynamo is (rightfully) "optional" with graceful eager fallback.

I learned from Michael Lazos that he plans to add operator stats to show which ops fall back to eager vs. using Inductor. But I believe more importantly we need easier way to tell which step of the compilation workflow has issue that causes the fallback, so that on the performance engineer side we can go in and figure out why the fallback happens.

Specifically it would be great to have:
1. A mode where we always error on fallback (maybe some ops only have fallback? in that case can have whitelist for those ops)
2. We want a way to know which step failed. The logs is currently a bit long and I also don’t know exactly what they mean in terms of what fails. If we can have “Step 1: XYZ, Step 2: XYZ, Step 3: XYZ, failed!” printed in the log, it’s easier to know that it's Step 3 that failed.

jansel:
Perhaps we could adapt torchdynamo.explain to this?  cc @voznesenskym

voznesenskym:
> Perhaps we could adapt torchdynamo.explain to this?  cc @voznesenskym 

Yea seems to be right up it's ally. I can take this one, if you don't mind.

voznesenskym:
@yf225 have you tried explain yet? It gives you graph breaks + graphs, and ostensibly the graphs are what would get lowered to inductor. Atm, we explain graph breaks, but not fallbacks. 

For what you call fallbacks: 

> "with graceful eager fallback." 

I think is what we refer to as graph breaks. It is when we go to eager mode instead of producing a graph. 

> A mode where we always error on fallback

If we take the definition of fallback you gave, we have that, in two forms - `torchdynamo.optimize(backend, nopython=True)` (for sending a whole graph down to a backend) and torchdynamo.export() for exporting a graph out of dynamo. 

However, we also have other definitions of fallback, which is when inductor generates a fallback via `ExternKernel`/`FallbackKernel` - you can see the list of such ops in lowering.py, but it can also happen through other paths. Are you interested in capturing these as well? 

Feel free to ping me on internal chat, I am happy to go over this in more detail as well.

yf225:
Thanks @voznesenskym - some raw notes from offline discussion:

- Dynamo layers of invocation
  - User defined torch model
[...]

PR #1436, opened 2022-09-30
Log beginning and end of steps, update logging levels

williamwen42:
Implement part of #1243.

Log the beginning and the end of:
- the initial call to a function optimized by torchdynamo
- the call to the backend compiler function
- ~the AOT autograd call in torchinductor~ (UPDATE: this shouldn't be added until dynamo is merged with the rest of pytorch)
- the forward and backward compiler functions in torchinductor

Create new logging level (in between DEBUG and INFO) for dynamo generated graphs, bytecode, and guards.

Sample outputs: https://gist.github.com/williamwen42/b81842def05f500f617d8a78ad2caf7b


torchdynamo/config.py
```diff
@@ -18,11 +18,14 @@
     HAS_REFS_PRIMS = False
 
 
+INFO_CODE = 15
```
mlazos, line 21: Unless we want this to be configurable, we should hide this elsewhere in TorchDynamo, maybe utils.py?

mlazos: Or even it's own module, we've said this a few times for other things, so we could create a logging_utils module or something.

torchdynamo/eval_frame.py
```diff
@@ -157,6 +159,9 @@ def __call__(self, *args, **kwargs):
 
         @functools.wraps(fn)
         def _fn(*args, **kwargs):
+            if self.first_ctx:
+                log.info("torchdynamo begin tracing")
```
yf225, line 163: It would be great to prefix these milestones with "Step X:", e.g. "Step 1: torchdynamo begin tracing", "Step 2: torchdynamo done tracing", so that user can just search for "Step X+1" to find if/where the next step is run, instead of needing to know the name of next step before they can search for it in the log.

torchinductor/compile_fx.py
```diff
@@ -71,7 +71,7 @@ def compile_fx_inner(
     num_fixed=0,
     is_backward=False,
 ):
-    log.info("Compiling %s graph", "BACKWARDS" if is_backward else "FORWARDS")
+    log.info("Inductor compiling %s graph", "BACKWARDS" if is_backward else "FORWARDS")
```
yf225, line 74: For each graph we are trying to compile, let's give each of them an ID and print the ID as well. For example:
- "Inductor compiling graph 1 FORWARDS"
- "Inductor compiling graph 1 BACKWARDS"
- "Inductor compiling graph 2 FORWARDS"
- "Inductor compiling graph 2 BACKWARDS"
etc.

torchdynamo/eval_frame.py
```diff
@@ -157,6 +159,9 @@ def __call__(self, *args, **kwargs):
...
+
```
yf225, line 164: Also since TorchDynamo + TorchInductor is a deep stack, user will benefit from a blueprint of the steps (e.g. what Steps we will go through and what each step does) in a short summary. Maybe we can put it as part of README.md file?

torchdynamo/logging.py
```diff
@@ -0,0 +1,61 @@
...
+    "version": 1,
+    "formatters": {
+        "torchdynamo_format": {"format": "%(name)s: [%(levelname)s] %(message)s"},
+    },
+    "handlers": {
+        "torchdynamo_console": {
+            "class": "logging.StreamHandler",
+            "level": "DEBUG",
+            "formatter": "torchdynamo_format",
+            "stream": "ext://sys.stdout",
```
jansel, line 32: Normally this would be stderr

torchdynamo/output_graph.py
```diff
@@ -408,13 +408,15 @@ def compile_and_call_fx_graph(self, tx, rv, root):
...
+            log.info("Step 2: done compiler function")
             assert callable(compiled_fn), "compiler_fn did not return callable"
         except Exception as e:
-            sys.stderr.write("-" * 40 + "\n")
-            sys.stderr.write("TORCHDYNAMO: backend compiler failed\n")
-            traceback.print_exc()
-            sys.stderr.write("-" * 40 + "\n")
+            log.warning("-" * 40 + "\n")
+            log.warning("TORCHDYNAMO: backend compiler failed\n")
+            log.warning(traceback.format_exc())
```
jansel, line 418: Does `log.warning(..., exc_info=True)` format this nicer?

yf225 (approved):
Thanks! It looks great to me. I'd suggest asking @msaroufim for a review too.

torchdynamo/logging.py
```diff
@@ -0,0 +1,63 @@
+import logging
+import os
+
+# logging level for dynamo generated graphs/bytecode/guards
+INFO_CODE = 15
```
voznesenskym, line 5: Just call it verbose? Or something? This feels like it overloads the word info needlessly.

voznesenskym (commented):
minor quibble otherwise lgtm

jansel (approved):

torchdynamo/convert_frame.py
```diff
@@ -396,23 +397,25 @@ def transform(instructions, code_options):
...
                 code.co_name,
                 code.co_filename,
                 code.co_firstlineno,
                 code,
-            )
+            ),
         )
-        log.info(
+        log.log(
+            torchdynamo_logging.VERBOSE,
```
mlazos, line 411: Sorry, I already know you changed this once, can you call this CODE? I don't want to use verbose because we already have a config setting verbose which makes exception traces larger.

mlazos: Also consider adding all of the different levels we support under torchdynamo_logging, this will make the interface consistent for all of the log levels

anijain2305 (approved):

torchinductor/compile_fx.py
```diff
@@ -76,11 +80,18 @@ def compile_fx_inner(
...
-    if count_calls(gm.graph) == 0:
+    if dynamo_utils.count_calls(gm.graph) == 0:
         return gm
 
-    log.info("Compiling %s graph", "BACKWARDS" if is_backward else "FORWARDS")
+    _step_logger()(
+        logging.INFO,
+        "torchinductor compiling "
+        f"{'BACKWARDS' if is_backward else 'FORWARDS'} "
+        f"graph {graph_id}",
```
yf225, line 92: Sorry for late request - it would be great if we can should total # of graphs here as well (e.g. something like "graph 1 of 10"), which will be like a progress bar in the user's mental model and helps with understanding compilation progress.

williamwen42: I'm not sure this feature is possible to implement, since the problem of "determining ahead-of-time how many graphs a program, given input, will produce" seems like an undecidable problem.

yf225: Curious why is it undecidable? I thought the # graphs a program given input will produce is deterministic (or do you mean it is not deterministic?)

williamwen42: The number of graphs a program will produce will depend on the input (including possibly randomness) as well, for example, we can't decide without input if the following program will graph break:
```python
if x.sum() > 0:
  print("break!")
else:
  pass
```

My intuition is that counting the ahead of time number of graphs is as challenging as determining if a program graph breaks at all, which is undecidable due to Rice's theorem (the problem of deciding if a program contains a non-trivial semantic property is undecidable).

PR #1438, opened 2022-09-30
Replace type_promote with type_promotion_kind

SherlockNoMad:
Some unary ops, e.g. lgamma, needs to have type_promotion_kind=INT_TO_FLOAT

### DIFF
--- a/torchinductor/lowering.py
+++ b/torchinductor/lowering.py
@@ -125,7 +125,7 @@ def decode_device(device):
     return device
 
 
-def get_promoted_dtype(*args):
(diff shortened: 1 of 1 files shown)


torchinductor/lowering.py
```diff
@@ -125,7 +125,9 @@ def decode_device(device):
     return device
 
 
-def get_promoted_dtype(*args):
+def get_promoted_dtype(
+    *args, type_promotion_kind=ELEMENTWISE_TYPE_PROMOTION_KIND.DEFAULT
```
ngimel, line 129: don't default type promotion type here, let's be explicit (in the only callsite it's defined anyway).

SherlockNoMad: fixed.

ngimel (approved):

PR #1442, opened 2022-10-01
Type promotion for lgamma, exp, sigmoid, sqrt, cos, sin, log, silu

SherlockNoMad:
Apply type_promotion_kind=ELEMENTWISE_TYPE_PROMOTION_KIND.INT_TO_FLOAT for lgamma, exp, sigmoid, sqrt, cos, sin, log, silu
l.libdevice.lgamma doesn't support int types


ngimel:
`isnan`, `isinf` are better served by dtype-specific decompositions (have to return tensors full of `False` for integral dtypes).

ngimel:
Also, can you add some more functions to INT_TO_FLOAT list? E.g. exp, sin, cos and other trig functions should be fixed by this

ngimel (approved):

PR #1443, opened 2022-10-01
Fix codegen for sign 

SherlockNoMad:
aten.sign(0) produces 0. 

In this PR
- Fix code gen to handle the case for 0 
- decompose sgn as sign


jansel (approved):

torchinductor/decomposition.py
```diff
@@ -273,7 +273,7 @@ def lift(self):
 
 @register_decomposition([aten.sgn])
 def sgn(self):
```
ngimel, line 275: can we instead import `sgn` decomposition from refs? It's doing the same thing

PR #1444, opened 2022-10-01
fix argmin/max for 1-element reduction

ngimel:


desertfire:
Nice! The cpu skip list needs to be fixed.

desertfire (approved):

PR #1447, opened 2022-10-01
Assertion to check if fake tensors have leaked into python

anijain2305:
I am seeing a fake tensor in `frame.f_locals` which is confusing to me. And I think this is wrong. Fake tensor is internal, and when someone runs a graph with real inputs, none of the output should be a FakeTensor.

So, adding an assertion to catch it somewhat early to help debug and share some repros.

### DIFF
--- a/torchdynamo/source.py
+++ b/torchdynamo/source.py
@@ -57,13 +57,8 @@ class Source:
             GuardSource.GLOBAL_NN_MODULE,
         )
 
-    def __eq__(self, other):
-        self_dict = dataclasses.asdict(self)
-        other_dict = dataclasses.asdict(other)
-        return self_dict == other_dict and self.name() == other.name()
 
-
-@dataclasses.dataclass(eq=False)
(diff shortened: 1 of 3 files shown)


jansel:
Seems fine, but tests are failing

eellison (approved):
LGTM, but you have a couple of errors

anijain2305:
Yes, the goal was to run CI to uncover any such issues. Happy that CI uncovered atleast one. It seems to be Fake tensor issue as well. We should fix that one.

eellison (commented):
I'm actually not that familiar with `GraphArg` or how it is used. Also, the test is failing correctly returns non-fake tensors which makes me a little suspect of this. We could also error in aot_eager instead.

jansel (approved):
Fix tests before landing

anijain2305:
@jansel Can you please review again as I added some `__eq_` code?

jansel (approved):

PR #1449, opened 2022-10-02
Fix an issue that does not vectorize the most outer parallel loop

EikanWang:
* __->__ #1449

Current logic replaces [` for ` with ` for {simd}`](https://github.com/pytorch/torchdynamo/blob/main/torchinductor/codegen/cpp.py#L658). But the literal will miss the end [whitespace character](https://github.com/pytorch/torchdynamo/blob/main/torchinductor/codegen/cpp.py#L654) if no reduction and the parallel is not greater than [1](https://github.com/pytorch/torchdynamo/blob/main/torchinductor/codegen/cpp.py#L655).


jansel (approved):
Note we have had some issues landing ghstack PRs in the past.

Chillee:
@ezyang cough cough

PR #1451, opened 2022-10-03
fix timed() when cuda is not available

jgong5:
`timed(...)` calls `synchronize` without checking CUDA availability, which breaks CPU when CUDA is not available. This PR fixed the problem.

### DIFF
--- a/torchinductor/utils.py
+++ b/torchinductor/utils.py
@@ -94,12 +94,14 @@ def gen_gm_and_inputs(target, args, kwargs):
 
 
 def timed(model, example_inputs, times=1):
-    synchronize()
+    if torch.cuda.is_available():
+        synchronize()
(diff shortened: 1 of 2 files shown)


torchinductor/utils.py
```diff
@@ -94,12 +94,14 @@ def gen_gm_and_inputs(target, args, kwargs):
 
 
 def timed(model, example_inputs, times=1):
-    synchronize()
+    if torch.cuda.is_available():
+        synchronize()
```
voznesenskym, line 98: it would be a tiny tad cleaner to just have a `def synchronize()` that does the `torch.cuda.is_available()` and `torch.cuda. synchronize()` inside it. Less repetition, and fits nicely into this utils.py class

desertfire (approved):

PR #1452, opened 2022-10-03
Allow cpp codegen generalized for num_threads

jgong5:
Currently, CPP codegen generates code specialized on the number of threads but torchdynamo currently doesn't guard on the number of threads per runtime environment and doesn't recompile if thread runtime config changes. It also uses the `multiprocessing.cpu_count` by default (when `config.cpp.num_threads=-1`) which returns all CPUs, not following the OMP runtime configuration. This PR tries to address the problems and makes the following changes:
1. Add a configuration `config.cpp.dynamic_threads` and when it is `True`, the generated kernel is not specialized for number of threads and doesn't recompile with changes of thread number but with minor perf regression on single-threaded workload. The default configuration is set to `True` for best perf, and calling `torchdynamo.reset()` is recommended if thread number changes at runtime. This is not needed after torchdynamo supports the guard on number of threads.
2. Align the default number of threads with the torch.get_num_threads configured by OMP runtime.


torchinductor/codegen/cpp.py
```diff
@@ -98,6 +106,13 @@ def argmax_argmin_prefix(reduction_type, src_dtype, tmpvar):
     return prefix
 
 
+def float16_reduction_prefix(rtype):
+    prefix = [
+        f"#pragma omp declare reduction({RTYPE_TO_CPP[rtype]} : {DTYPE_TO_CPP[torch.float16]} : omp_out += omp_in)"
```
jansel, line 111: Will this work for non sums?  Perhaps we should just promote fp16 to fp32 for the reduction accumulator. Using a fp16 accumulator might be numerically unstable.

jgong5: Thanks for the comment. I have fixed the reduction statement. About the reduction accumulator type for fp16, agree that we'd better use fp32 instead. Is CUDA backend using fp32 now? For CPU, fp16 support is still incomplete (e.g., no dot-product acceleration with fp16, arithmetic with SIMD was introduced in SPR just recently etc.). Can we move the complete fp16 support to the future PRs? If you agree, I can report an issue to track this.

A question about CI failures, I noticed three cases failed with something like "RuntimeError: unexpected success var_mean, torch.float16, cpu". May I know why these cases are supposed to fail? Thanks.

jansel: Inside the body of a kernel, the GPU backend does everything (including reductions) in fp32 precision.

A future PR is fine.

torchinductor/codegen/cpp.py
```diff
@@ -98,6 +106,14 @@ def argmax_argmin_prefix(reduction_type, src_dtype, tmpvar):
     return prefix
 
 
+def float16_reduction_prefix(rtype):
+    prefix = [
+        f"#pragma omp declare reduction({RTYPE_TO_CPP[rtype]}:{DTYPE_TO_CPP[torch.float16]}:"
+        + f"omp_out = omp_out {RTYPE_TO_CPP[rtype]} omp_in)"
```
jansel, line 112: Since this will only work for sum/any -- perhaps just assert the reduction is one of those.

torchinductor/config.py
```diff
@@ -56,7 +56,8 @@
 
 # config specific to codegen/cpp.pp
 class cpp:
-    threads = -1  # set to cpu_count()
+    threads = -1  # set to torch.get_num_threads()
+    specialize_num_threads = False  # whether to specialize on the number of threads
```
jansel, line 60: Does this change hurt single threaded performance?

torchinductor/codegen/cpp.py
jgong5: Added the assertion.

torchinductor/config.py
jgong5: @jansel In parallel, we can also work on adding thread number to one of the guards of torch dynamo. Thread number is similar to the shapes that can be specialized to get better perf. Not only the kernels in torch inductor, oneDNN kernels are also specialized with number of threads and would do recompilation if thread number changes. The difference from the shapes is that the thread number changes less frequently and assuming it static as default makes more sense. But before the dynamo guards support thread number guards, better we don't specialize it by default. What do you think?

jansel (changes requested):
looks good now, but need to fix tests/lints (make format)

torchinductor/config.py
jansel: Perhaps we should specialize by default then.  The only use case I can think of for dynamically changing thread count is performance measurement.  For that use case people can just call `torchdynamo.reset()`

jgong5: We have the use case that dynamically identify the number of inference streams and thread number per inference stream for best inference throughput. Streams can be assigned with different number of threads. But I agree with you that specializing thread number by default should be fine. I have changed the code accordingly.

jansel (approved):

PR #1454, opened 2022-10-03
Set inductor logging level

desertfire:
Summary: Let the inductor logging level also controlled by
torchdynamo.config.log_level


torchinductor/debug.py
```diff
@@ -20,6 +20,7 @@
 from torch.fx.passes.tools_common import legalize_graph
 
 import torchinductor
+from torchdynamo.config import log_level
```
jansel, line 23: Doing the import this way means if I change the value after import you won't pick up the change.

jansel (changes requested):

jansel (approved):

PR #1471, opened 2022-10-04
DDPOptimizer: register_attr_or_module -> add_submodule

davidberard98:
self.module is a raw torch.fx.GraphModule, not an OutputModule. So register_attr_or_module doesn't exist on self.module.

### DIFF
--- a/torchdynamo/optimizations/distributed.py
+++ b/torchdynamo/optimizations/distributed.py
@@ -165,7 +165,7 @@ class DDPOptimizer:
                         compiled_submod = self.compile_submod(submod, args, kwargs)
                         n.target = "compiled_" + n.target
                         self.module.delete_submodule(n.target)
-                        self.module.register_attr_or_module(n.target, compiled_submod)
+                        self.module.add_submodule(n.target, compiled_submod)
 
                     # then we execute the modified node using the usual logic
                     return getattr(self, n.op)(n.target, args, kwargs)


torchdynamo/optimizations/distributed.py
```diff
@@ -165,7 +165,7 @@ def run_node(self, n: Node) -> Any:
                         compiled_submod = self.compile_submod(submod, args, kwargs)
                         n.target = "compiled_" + n.target
                         self.module.delete_submodule(n.target)
-                        self.module.register_attr_or_module(n.target, compiled_submod)
+                        self.module.add_submodule(n.target, compiled_submod)
```
wconstab, line 168: was this line crashing? or never being run/tested?

wconstab: ok i guess this got changed recently.  @voznesenskym do we need to pay attention to something special here or is it fine to just use add_submodule?

voznesenskym: If you check the impl of the func, it handles a whole lot more than module, so the name was changed to match. What are you trying to do?

voznesenskym: Oh I see, this got caught in a search in replace. I think this was just untested. However, for dynamo, if it was add_submodule before and you want to keep the functionality you need to use the new name. The output graph is explicitly different and limiting.

voznesenskym (approved):

torchdynamo/optimizations/distributed.py
wconstab: > I think this was just untested. 

There is actually a test for it, but i think it got disabled due to some CI issues.  We're trying to get it re-enabled.

> if it was add_submodule before and you want to keep the functionality you need to use the new name

yes, it was add_submodule before.  However, we're not using dynamo's fx graph here.  So i'm unclear if your changes (new functionality in register_attr_or_module) apply here.

What we're doing in this code is summarized as following:
def ddp_compiler_as_dynamo_backend(gm: Dynamo's traced graphmodule, ..):
  1) analyze gm to determine split points
  2) split_gm = fx.passes.split_module.split_module(gm, ...)  # use analysis above and fx infra to split into a wrapper 'split_gm' containing calls into new child gms
  3) run split_gm through 'SubmodCompiler' which is an fx interpreter that replaces calls to child_gms with calls to compiled child_gms, using the provided aot- compiiler from the user
  
  step 3 is where we call add_submodule on split_gm, and iiuc we're modifying a vanilla fx GM here, not one produced by dynamo.  Hence we don't have access to that new functionality.
  
  Do we need it for some reason, or is this ok?

PR #1480, opened 2022-10-05
Skip failed unit tests and models since moving PyTorch pin

yanboliang:


anijain2305 (approved):

anijain2305:
Please add an issue so that we fix it tomorrow. Also try to see if you could minify it using TORCHDYNAMO_REPRO_AFTER="dynamo" ?

PR #1482, opened 2022-10-05
Add torchdynamo.graph_break()

jansel:
This is a way to force a graph break rather than relying on stuff like print.


voznesenskym (approved):

torchdynamo/allowed_functions.py
```diff
@@ -82,15 +82,13 @@ def _disallowed_function_ids():
...
         torch.distributions.normal.Normal,
-        torch.get_rng_state,
         torch.inference_mode,
         torch.set_anomaly_enabled,
         torch.set_autocast_cache_enabled,
         torch.set_autocast_cpu_dtype,
         torch.set_autocast_cpu_enabled,
         torch.set_autocast_enabled,
         torch.set_autocast_gpu_dtype,
-        torch.set_rng_state,
```
voznesenskym, line 93: Why remove these?

jansel: Before this PR these were getting inlined, which seems incorrect.   The change below makes them not inline anymore.

voznesenskym: Nice :)

jansel:
#1615

Issue #1401, opened 2022-09-28
Optimize Softmax Perf for Inductor CPU Backend


jgong5:
This issue tracks the required features for optimizing Softmax for Inductor CPU backend. The features might be needed by other reduction related ops.
Below are the numbers measured on CPX with a typical Softmax shape from BERT model as an example to demonstrate the potential of various optimizations (column on the right contains all the optimizations on the left). The optimized numbers were measured with manually altered C++ code based on Inductor cpp codegen. L3 cache is flushed before each iteration to make sure the numbers are stable. ATen and Inductor baseline numbers were measured with op benchmark while others were measured with generated output_code.py benchmarks. These two have gaps as indicated in the table and need further investigation.

Softmax (1,16,384,384,dim=3) | ATen | Inductor baseline   (simdlen=8) | +buffer reuse | +inplace buffers | +manual   vectorization on reduction
-- | -- | -- | -- | -- | --
28c | 1038us | 1873us   (1356us w/ output code) | 2201us   (1795us w/ output code) | 955us | 591us
4c | 2113us | 7116us   (6767us w/ output code) | 5582us   (5332us w/ output code) | 4041us | 1541us
1c | 6447us | 17732us   (17223us w/ output code) | 17561us   (17307us w/ output code) | 13968us | 4540us

Numbers by Ablation (the opt baseline has all the optimizations mentioned above, each column on the right having part of the optimization ablated, and the speedup ratio w/ vs. w/o the corresponding optimization)

Softmax   (1,16,384,384,dim=3) | ATen | Inductor opt   baseline (simdlen=8) | -buffer reuse | -inplace buffers | -manual   vectorization on reduction | -buffer reuse | -inplace buffers | -manual   vectorization on reduction
-- | -- | -- | -- | -- | -- | -- | -- | --
28c | 1038 | 591 | 672 | 1469 | 955 | 1.13X | 2.48X | 1.61X
4c | 2113 | 1694 | 2046 | 2826 | 4041 | 1.20X | 1.66X | 2.38X
1c | 6447 | 5407 | 6406 | 7752 | 13968 | 1.18X | 1.43X | 2.58X

1. [Buffer reuse](https://github.com/jgong5/pytorch_bench/blob/master/operator_benchmark/test_softmax_dynamic_4d_reuse/output_code.py) (1.13X - 1.18X)
[...]

Issue #823, opened 2022-08-15
Revisit inplace_buffers=True for TorchInductor


jansel:
TorchInductor has support for inplacing operations, but it is currently disabled:
https://github.com/pytorch/torchdynamo/blob/8e4abe3f66fb6b8d97d6461d64f9c3be323a523e/torchinductor/config.py#L28
due to some bugs in the past.

We should turn this back on and fix any issues it causes on our benchmark suites.

This could help with #822

Chillee:
Seems like this is currently quite broken.

Turning it on with resnet18 (`time ./benchmarks/torchbench.py -dcuda --inductor --training --no-skip --float16 -k resnet18 --batch_size 32`) results in 
```
  File "/scratch/chilli/fresh/torchdynamo/torchinductor/scheduler.py", line 348, in mark_run
    self.allocate()
  File "/scratch/chilli/fresh/torchdynamo/torchinductor/scheduler.py", line 336, in allocate
    V.kernel.args.make_inplace(
AttributeError: 'NullHandler' object has no attribute 'args'
```

jansel:
Yeah, I'm not surprised.  The code might a bit rotted a bit.

PR #1486, opened 2022-10-05
support inplace buffer reuse

jgong5:
Enable `inplace_buffers = True` feature. This addresses the issue https://github.com/pytorch/torchdynamo/issues/823 and the second optimization mentioned in  https://github.com/pytorch/torchdynamo/issues/1401. It is turned off by default now due to https://github.com/pytorch/torchdynamo/issues/1670. Currently, only one CI case gmixer_24_224 training on CUDA would fail due to this issue.

This PR made following fixes:
1. Support more than two `other_names` in `make_inplace` (common.py) to support a chain of inplace reuses.
2. Do not delete buffer if it is the graph output in `make_buffer_reuse` (wrapper.py)
3. Add a new field `V.graph.inplaced_to_remove`.  For inplace buffers subject to remove, we don't actually remove them but put them in this dedicated set. This simplifies the life cycle management of inplace buffers. This set is used to 1) avoid unnecessary store in DeferredLine; 2) avoid alias var definitions in kernel (graph.py/scheduler.py).
4. Do not allocate buffer for `ExternKernelAlloc` and `MultiOutput` (wrapper.py) since these are pre-allocated.
5. Make sure reuse key matches on inplace buffer reuse in `allocate` (scheduler.py)
6. Do not inplace reuse buffers with layout `(ir.MultiOutputLayout, ir.MutationLayout, ir.AliasedLayout)`. This is a conservative choice and can revisit later if we can still inplace reuse buffers with some of these layouts.
7. Add kernel output buffers to `available_buffer_names` in `codegen` (scheduler.py) to make sure `remaining_uses` is correct in `allocate` (scheduler.py)


jansel (changes requested):
The approach seems right, but the test failures look real.

jgong5:
@jansel There are two CI failures left. The unit test one seems not relevant to this change. For inductor_timm_training_2 gmixer_24_224 accuracy issue, weird that I cannot repro the failure locally, will check further, but feel free to let me know if you have any insights/hints. Thanks!

jgong5:
> It the minifier works smoothly, you should be able to get a smaller reproducible graph which makes debugging much easier.

Thanks a lot for the tip, will have a try! BTW, seems trunk is broken? I got something like argument number not matching with all CUDA related test...

soumith:
@jgong5 I believe that is because the `main` now requires a newer triton: https://github.com/pytorch/torchdynamo/pull/1585
Please upgrade your triton to the outlined commit and you should be good

jgong5:
> It the minifier works smoothly, you should be able to get a smaller reproducible graph which makes debugging much easier

@desertfire Can this troubleshoot accuracy issues or only handle compilation/crash errors?

jansel (changes requested):
Requesting changes to take this off my review queue, re-request review when tests pass.

jgong5:
Thanks @desertfire @soumith for the hints on the minifier tool. It helped to generate a repro, even though not a small one since the issue only happened on the first clean-cache run. I managed to narrow it down manually to a smaller repro https://github.com/pytorch/torchdynamo/issues/1670. It seems an issue inside Triton compiler or the way how TorchInductor integrates Triton? @jansel

jgong5:
@jansel Are you ok to land this PR with `inplace_buffers` set to `False` first? If it is turned on with the same PR change, only gmixer_24_224 training fails the accuracy test with the reason I explained in the description. I don't think the failure is caused by the change though but I will continue to check the root cause. Having it landed first allows others to experiment `inplace_buffers` if needed. If you are ok, the PR is in a good shape for review.

jansel (approved):

jansel:
Yes, sounds good to me.

PR #1488, opened 2022-10-05
[inductor] Workaround the shortcomings of OSX's default LLVM compiler

soumith:
This makes the OSX-default compiler work with Inductor lowering / compilation.

### DIFF
--- a/torchinductor/codegen/cpp_prefix.h
+++ b/torchinductor/codegen/cpp_prefix.h
@@ -3,7 +3,9 @@
 #include <cmath>
 #include <cstdlib>
 #include <limits>
+#ifdef _OPENMP
 #include <omp.h>
+#endif
 
 #include "ATen/core/PhiloxRNGEngine.h"
 #include <c10/util/Half.h>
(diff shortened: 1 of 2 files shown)


torchinductor/codegen/cpp_prefix.h
```diff
@@ -3,7 +3,9 @@
 #include <cmath>
 #include <cstdlib>
 #include <limits>
+#ifdef _OPENMP
 #include <omp.h>
+#endif
```
voznesenskym, line 8: Not blocking at all, just worth considering. This is safe because we only use `#pragma omp` from omp. If we ever introduce a direct call into an omp.h header, that call will fail to compile. As we don't have osx CI, that could go unnoticed.

A few potential solutions, not for this PR:

A) Add OSX CI
B) have an omp_available bool in cpp.py to guard omp calls
C) Instruct OSX / provide infra for users to install libomp.

voznesenskym (approved):
LGTM, CI failure looks potentially relevant.

jansel:
`We have migrated `torchdynamo` to `torch._dynamo` and will use the `pytorch/pytorch` repo for future development.  Please resubmit this PR to https://github.com/pytorch/pytorch/

More details and instructions to port this PR over can be found in https://github.com/pytorch/torchdynamo/issues/1588`

PR #1492, opened 2022-10-05
Update CI for TIMM models

desertfire:
Summary:
1) Relax the exception check for retry when pretrain model download fails; 
2) Enable accuracy checking for inductor training as we are over 90% passing on TIMM

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -499,7 +499,7 @@ jobs:
           command: |
             source .circleci/setup_env.sh
             make develop
-            python benchmarks/timm_models.py --accuracy --ci --skip-accuracy-check -d cuda --inductor --float32 --training \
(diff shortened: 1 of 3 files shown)


anijain2305 (approved):

anijain2305:
> 1. Relax the exception check for retry when pretrain model download fails;

Maybe the right way to handle this is caching the pretrained weights in CI?

desertfire:
> > 1. Relax the exception check for retry when pretrain model download fails;
> 
> Maybe the right way to handle this is caching the pretrained weights in CI?

Agree, but I don't know how to do it.

Pretrained models are downloaded to directories like `/home/circleci/.cache/torch/hub/checkpoints/gluon_inception_v3-9f746940.pth`, which is not under the project directory, `/home/circleci/project`. Is there an easy way to config where to download TIMM models?

PR #1494, opened 2022-10-05
Refactor torchdynamo imports to be relative

jansel:
This is an early step in migrating to core / `torch.dynamo`.


voznesenskym (approved):
Stamp for good change, predicated on CI passing ofc

PR #1500, opened 2022-10-05
Fix decomp/lowering/ExternKernel for addmm

SherlockNoMad:
- addmm decomp and lowering supports beta/alpha kwargs
- ExternKernel support codegen for kwarg-only args


ngimel (approved):
Great!

PR #1501, opened 2022-10-05
Add smoke test

williamwen42:
Add tests to check that dynamo is installed correctly. Tested by running the script `python tools/smoke_test.py` on different environments.


tools/smoke_test.py
```diff
@@ -0,0 +1,174 @@
...
+        x = torch.randn(10, 10).to(device)
+        x.requires_grad = True
+        y = fn(x)
+        torch.testing.assert_close(y, x + x)
+    except Exception:
+        sys.stderr.write(traceback.format_exc() + "\n" + err_msg + "\n\n")
+        sys.exit(1)
+
+
+def check_eager_cpu():
```
yanboliang, line 110: Can we organize the following functions into a config dict and one function to iterate across these configs? I think that is a more scalable way.

tools/smoke_test.py
```diff
@@ -0,0 +1,174 @@
...
+        "CUDA inductor sanity check failed\n"
+        + "NOTE: Please check that you installed the correct hash/version of `triton`",
+    )
+
+
+def main():
+    check_python()
+    check_pip_deps()
+    check_torch()
+    check_cuda()
```
yanboliang, line 163: For all these checks above, it's better we can print out the version details even if they pass all checks. Users can leverage this script to collect env info when they reporting bugs, and help us to narrow down where is the bug from.

tools/smoke_test.py
```diff
@@ -0,0 +1,174 @@
...
+            f"torch version not supported: {torch.__version__} "
+            f"- minimum requirement: {MIN_TORCH_VERSION}"
+        )
+
+
+def check_cuda():
+    import torch
+
+    if not torch.cuda.is_available():
+        sys.stderr.write("CUDA not available for torch\n")
```
yanboliang, line 54: It's not error, if users want try torchdynamo on CPU, CUDA is not necessary. Just print out if CUDA is available and version is supported.

tools/smoke_test.py
```diff
@@ -0,0 +1,174 @@
...
+            f"CUDA not available -- skipping CUDA check on {backend} backend\n"
+        )
+        return
+    try:
+        import torchdynamo
+
+        torchdynamo.reset()
+
+        @torchdynamo.optimize(backend, nopython=True)
+        def fn(x):
```
yanboliang, line 98: It's better to add another ```nn.Module``` example, as optimize function and module are the most popular cases dynamo should support.

yanboliang (changes requested):

tools/smoke_test.py
```diff
@@ -0,0 +1,166 @@
...
+
+MIN_CUDA_VERSION = Version("11.6")
+MIN_PYTHON_VERSION = (3, 7)
+# NOTE: dev version required, e.g. Version('1.13.0.dev123') < Version('1.13.0')
+MIN_TORCH_VERSION = Version("1.13.0.dev0")
+
+
+def check_python():
+    if sys.version_info < MIN_PYTHON_VERSION:
+        raise RuntimeError(
```
voznesenskym, line 21: make your own exception for this, so you know this is an explicit failure from the test as opposed to a runtimeerror, which anyone can make downstream of calls you make.

tools/smoke_test.py
```diff
@@ -0,0 +1,166 @@
+import os
```
voznesenskym, line 1: My opinion is that if its called test, it goes in test/, otherwise, its something else. If its not running on CI, and its under tools/, call it "verify_install.py" or something like that.

tools/smoke_test.py
```diff
@@ -0,0 +1,166 @@
...
+import re
+import subprocess
+import sys
+import traceback
+
+try:
+    import packaging.version
+except ImportError:
+    from pkg_resources import packaging
+Version = packaging.version.Version
```
voznesenskym, line 11: Add a comment as to when / why this might fail please.

tools/smoke_test.py
```diff
@@ -0,0 +1,166 @@
...
+
+MIN_CUDA_VERSION = Version("11.6")
+MIN_PYTHON_VERSION = (3, 7)
+# NOTE: dev version required, e.g. Version('1.13.0.dev123') < Version('1.13.0')
+MIN_TORCH_VERSION = Version("1.13.0.dev0")
```
voznesenskym, line 16: Is that correct?

tools/smoke_test.py
```diff
@@ -0,0 +1,166 @@
...
+    # check if torch cuda version matches system cuda version
+    cuda_home = os.getenv("CUDA_HOME")
+    if not cuda_home:
+        raise RuntimeError("$CUDA_HOME not found")
+
+    nvcc = os.path.join(cuda_home, "bin", "nvcc")
+    nvcc_version_str = (
+        subprocess.check_output([nvcc, "--version"]).strip().decode("utf-8")
+    )
+    cuda_version = re.search(r"release (\d+[.]\d+)", nvcc_version_str)
```
voznesenskym, line 83: This feels assumptive, in order:

1) That nvcc is installed there (You could have a malformed $CUDA_HOME)
2) That running --version produces the string you want

tools/smoke_test.py
```diff
@@ -0,0 +1,166 @@
...
+            x = torch.randn(10, 10).to(device)
+            x.requires_grad = True
+            y = f(x)
+            torch.testing.assert_close(y, x + x)
+            z = y.sum()
+            z.backward()
+            torch.testing.assert_close(x.grad, 2 * torch.ones_like(x))
+    except Exception:
+        sys.stderr.write(traceback.format_exc() + "\n" + err_msg + "\n\n")
+        sys.exit(1)
```
voznesenskym, line 132: This just feels like a unit test snuck into this - what is the point of this? If dynamo is installed, it should be working. If its installed, running any unit test should work. This just makes things confusing. If you want to keep the spirit of a dynamo check of this complexity here, just run pytest on test_functions.py or something light.

torchinductor/utils.py
```diff
@@ -26,6 +26,8 @@
 
 @functools.lru_cache(None)
 def has_triton():
+    if not torch.cuda.is_available():
+        return False
```
voznesenskym, line 30: Put this in its own PR?

voznesenskym (changes requested):
It all looks fine, but if we don't run this as part of regular CI, it will code rot away very quickly. 

We also should make sure we are not doing anything that might make the torch integration more difficult. I would talk to @jansel and @malfet to double check what our plans there are.

tools/smoke_test.py
williamwen42: `from pkg_resources import packaging` should be sufficient (that is, no try/except needed) since it should be bundled in the python environment.

tools/smoke_test.py
williamwen42: Not sure what the concern is here?

yanboliang:
@voznesenskym I'd like to provide more context here, actually this is a feature requirement from Peng, Jason Liang and Mark. The goal of this script is not to run on CI because we already had CI cover these checks. The goal is:
* This is a user side tools, they can use this to verify if they install dynamo and the dependencies successfully.
* If they pass this test, but they still have errors when running their own code. This can help us to collect python/CUDA/triton/dynamo version info for debug purpose. We would suggest users to submit an issue with the output of this script.

tools/smoke_test.py
voznesenskym: > I'm wary about calling this code since it's not public API, so I'm thinking of copying the parts I need.

Call it, its fine, its a private API in a repo we own, within the same overall project, for a niche tool. It certainly beats out copying the code.

voznesenskym:
> @voznesenskym I'd like to provide more context here, actually this is a feature requirement from Peng, Jason Liang and Mark. The goal of this script is not to run on CI because we already had CI cover these checks. The goal is:

We should still run it on CI so it does not coderot. I am not worried about protecting what this test protects, thats redundant, so much as keeping this logic up to date. 

> * This is a user side tools, they can use this to verify if they install dynamo and the dependencies successfully.

Yup.

> * If they pass this test, but they still have errors when running their own code. This can help us to collect python/CUDA/triton/dynamo version info for debug purpose. We would suggest users to submit an issue with the output of this script.

Yes.

I think I have all this context, and we are in agreement. No one here is against this tool.

tools/smoke_test.py
yanboliang: I think this is relevant to user guide vs. developer guide. The purpose of this smoke test is for user guide, we should not expect users are dynamo expert like us and should not expose unit tests to them.

voznesenskym: This is a unit test exposed to them in all but name.

tools/smoke_test.py
voznesenskym: I was under the impression that with all the fake tensor stuff, we basically are only able to run on nightly torch. Just want to make sure this min torch version is correct.

tools/smoke_test.py
voznesenskym: why the first try then?

tools/smoke_test.py
voznesenskym: I disagree but do not care enough to argue lol.

yanboliang: No worry, we can revisit if the requirement from Peng/Jason/Mark is not reasonable.

williamwen42: I'm inclined to keep a simple test here since (1) it's difficult for incorrect triton versions to be detected otherwise, and (2) pytest and other testing libraries should not be required dependencies for users.

.circleci/config.yml
```diff
@@ -68,6 +68,7 @@ jobs:
             source .circleci/setup_env.sh
             make develop
             mkdir test-results
+            python tools/verify_install.py
```
voznesenskym, line 71: How long does this run? I was thinking you could give it its own job. Not important if its fast.

voznesenskym (approved):
LGTM, make sure @yanboliang looks at it too, since he had prior feedback.

.circleci/config.yml
williamwen42: It should be fairly fast.

tools/smoke_test.py
williamwen42: https://peps.python.org/pep-0440/#summary-of-permitted-suffixes-and-relative-ordering, `dev0` is the smallest in the version sorting order, and we aren't using a version of torch that looks like `1.13aN`.

yanboliang (approved):

PR #1503, opened 2022-10-05
Make torchinductor use relative imports

jansel:
Preparing for a move to `torch.inductor`


voznesenskym (approved):

voznesenskym:
On the topic of inductor-->torch, our cpp cmake files need a revisit.

PR #1506, opened 2022-10-05
Run reference in original dtype for opinfo tests

SherlockNoMad:
Opinfo cuda test was comparing with eager reference result ran in fp32. 
This PR fixes this by running eager kernel in test case's original dtype.


voznesenskym (approved):

test/test_torchinductor.py
```diff
@@ -147,32 +147,38 @@ def check_model(
...
-        nonlocal has_lowp_args
-        if isinstance(x, torch.Tensor) and (
-            x.dtype == torch.float16 or x.dtype == torch.bfloat16
-        ):
-            has_lowp_args = True
-            return x.float()
-        else:
-            return x
+    if reference_in_float:
+        # check_lowp is ignored here, it's kept just to be able to call `common` with extra arg
```
ngimel, line 160: should we remove check_lowp then?

ngimel (approved):

PR #1507, opened 2022-10-05
DDPOptimizer: recompile graph after replacing subgraphs

davidberard98:
DDPOptimizer does the following:
1. split up the graph into submodules
2. replace each of the submodules with a compiled submodule

After step 2, we weren't calling `graph_module.recompile()`, which caused us to call the old submodules. The old submodules didn't have the step that unwraps tuples, which caused type errors, described below.

The old submodules will wrap outputs in tuples, even if there's only a single output. So we might have submodule1 which should output a tensor that is consumed by submodule2. In this case we need to unwrap the output tuple from submodule 1; that's what WrapperModule does. But since we weren't recompiling, the graph_module was still calling the old submodules.

For some reason, this issue only occurred when using inductor backend.

I verified locally that the new test passes.

### DIFF
--- a/test/test_distributed.py
+++ b/test/test_distributed.py
@@ -137,6 +137,28 @@ class TestDistributed(torchdynamo.testing.TestCase):
         reason="requires pytorch landing in parallel",
     )
     @patch.object(config, "optimize_ddp", True)
+    def test_graph_split_inductor(self):
+        """
+        Same as above, but using inductor backend.
+        We observed issues with inductor/fx interface in the past.
+        """
+        m, inputs, correct_outputs = self.get_model()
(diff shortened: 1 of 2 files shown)


anijain2305 (approved):

test/test_distributed.py
```diff
@@ -132,6 +132,26 @@ def opt_fn(inputs):
         self.assertTrue(same(correct_outputs, opt_outputs))
         self.assertEqual(check_splits_compiler.compiler_called, 3)
 
+    @pytest.mark.skipif(
```
wconstab, line 135: btw this skipif was intended as a transient thing, but maybe we should remove it now that the ddp change on pytorch side has landed and doesn't seem likely to get reverted.  not urgent.

Issue #1368, opened 2022-09-27
aot_eager amp training failure: conversion is supported for Half type only


desertfire:
Discovered in a recent pytorch nightly pin update, https://github.com/pytorch/torchdynamo/pull/1351

Repro:
```
benchmarks/torchbench.py --accuracy --training --amp --backend=aot_eager --only pytorch_struct
```
Error:
```
TorchDynamo optimized model failed to run because of following error
conversion is supported for Half type only
FAIL
```

anijain2305:
Repro

~~~

import torch
import torchdynamo
from torch import tensor, device
import torch.fx as fx
from torchdynamo.testing import rand_strided
from math import inf
from torchdynamo.debug_utils import run_fwd_maybe_bwd


from torch.nn import *
class Repro(torch.nn.Module):
    def __init__(self):
        super().__init__()



    def forward(self, self_s_emb, add_3):
        einsum_2 = torch.functional.einsum('ah,th->t', self_s_emb, add_3);  self_s_emb = add_3 = None
        log_softmax_2 = einsum_2.log_softmax(-1);  einsum_2 = None
        return (log_softmax_2,)


mod = Repro().cuda()
opt_mod = torchdynamo.optimize("aot_eager")(mod)


args = [((1, 256), (256, 1), torch.float32, 'cuda', True), ((30, 256), (256, 1), torch.float16, 'cuda', True)]
args = [rand_strided(sh, st, dt, dev).requires_grad_(rg) for (sh, st, dt, dev, rg) in args]


with torch.cuda.amp.autocast(enabled=True):
    ref = run_fwd_maybe_bwd(mod, args)
    res = run_fwd_maybe_bwd(opt_mod, args)
~~~

anijain2305:
@mlazos Assigning in case you are interested. Please feel free to assign it to me if you are busy. 

This might be a good way to learn about Aot Autograd and AMP, and can be useful as we turn on GradScaler for optimizers.

eellison:
actually i dont.. but this error is the result of einsum returning float16 during tracing, but then float32 upon invocation. 

```
import torch
import torchdynamo
from torch import tensor, device
import torch.fx as fx
from torchdynamo.testing import rand_strided
from math import inf
from torchdynamo.debug_utils import run_fwd_maybe_bwd


from torch.nn import *
class Repro(torch.nn.Module):
    def __init__(self):
        super().__init__()

    def forward(self, self_s_emb, add_3):
[...]

PR #1510, opened 2022-10-06
Adding test for #1368

anijain2305:
Leaving a PR so that we don't forget to add a test when #1368 gets fixed in PyTorch core.

It will fail for a few days until we fix the error and also PyTorch pin gets updated.


jansel:

PR #1513, opened 2022-10-06
Add support for _get_item_by_idx

voznesenskym:


torchdynamo/variables/nn_module.py
```diff
@@ -246,6 +246,20 @@ def call_method(
         ):
             return ConstantVariable(True, **options)
 
+        if name == "_get_item_by_idx":
+            assert args[1].is_python_constant()
+            assert isinstance(args[0], TupleVariable)
+            mod_var = args[0].items[args[1].value]
+            key = mod_var.module_key
+            submod = tx.output.get_submodule(key)
+            return tx.output.register_attr_or_module(
```
anijain2305, line 255: I think this is replaced by `add_submodule`? I saw some errors on the latest pytorch with this.

anijain2305 (approved):
Pre-approving. LGTM

torchdynamo/variables/nn_module.py
voznesenskym: I renamed this function from add_submodule to this in another PR. Perhaps that's it? It passed our CI, so i think it's okay. If we get name conflicts or issues here when we bump torch versions I'll revisit 😄

PR #1514, opened 2022-10-06
Update PyTorch pin

desertfire:


desertfire:
@yanboliang is investigating the failures.

pommedeterresautee:
@desertfire import torchdynamo raises the following import error:
```
...
ImportError: cannot import name 'enable_torch_dispatch_mode' from 'torch.utils._python_dispatch' (/home/geantvert/.local/share/virtualenvs/kernl/lib/python3.9/site-packages/torch/utils/_python_dispatch.py)
```
tried with both dev20221005 and dev20221006 nightly.

PR #1519, opened 2022-10-06
[inductor] Only skip CPU tests when host cpp compiler not found

bertmaher:
FB-internal build isn't set up to invoke a c++ compiler, just triton/ptxas.  So we can still run the GPU tests even if the CPU is skipped.


voznesenskym (approved):

voznesenskym:
> Looks fine to me, but I thought Voz solved the internal c++ compiler issue?

I was working on the triton compiler issue, I don't think I did anything for the internal cpp compiler.

PR #1521, opened 2022-10-06
Skip two opinfo tests

yanboliang:


voznesenskym:
@yanboliang is the failure for these by chance `Tensor-likes are not close!`? If so, can you move them to the inductor_override_kwargs area instead?

yanboliang:
```@yanboliang is the failure for these by chance Tensor-likes are not close!? If so, can you move them to the inductor_override_kwargs dict instead?```

@voznesenskym This is other failures, @SherlockNoMad suggested me to put them in the skip list.

PR #1522, opened 2022-10-06
Get past graph breaks in tests to enable more ops

voznesenskym:


test/inductor/test_torchinductor_opinfo.py
```diff
@@ -103,6 +103,7 @@ def process(device_type):
 inductor_skips["cpu"] = {
     "linalg.ldl_solve": {b8, f16, f32, f64, i32, i64},  # segfault
     "linalg.lu_solve": {b8, f16, f32, f64, i32, i64},  # segfault
+    "reciprocal": {b8, i32, i64},  # segfault
```
SherlockNoMad, line 106: This should have been resolved. Merge error?

voznesenskym: Yes, probably :)

test/inductor/test_torchinductor_opinfo.py
```diff
@@ -421,13 +406,27 @@ def process(device_type):
...
+
 # key can be either op_name, or (op_name, deivce_type), or (op_name, device_type, dtype)
 inductor_override_kwargs = {
     # the return value of empty is undefined
     "empty": {"assert_equal": False},
     "empty_like": {"assert_equal": False},
     "new_empty": {"assert_equal": False},
     "new_empty_strided": {"assert_equal": False},
+    "randn": {"assert_equal": False},
+    "nn.functional.tanhshrink": {"assert_equal": False},
```
SherlockNoMad, line 429: tanhshrink doesn't seem to have randomness. 
Is this a real failure?

voznesenskym:
I need to figure out why every test is failing, on CI, but not locally. Thoughts @desertfire ?

test/inductor/test_torchinductor_opinfo.py
voznesenskym: Yeah maybe, i can look once I figure out this weird CI thing...

torchdynamo/variables/torch.py
```diff
@@ -40,8 +41,10 @@
     torch._C._TensorBase.__ror__,
     torch._C._TensorBase.__rxor__,
     torch._C._TensorBase.__rand__,
+    torch.testing._internal.common_utils.disable_functorch,
```
desertfire, line 44: Do you need to explicitly import this? CI failure seems to point to this line.

voznesenskym:
This is blocked on fake_tensor not working with state setting. I'll look into fixing that.

torchdynamo/variables/misc.py
```diff
@@ -249,6 +249,33 @@ def call_function(
...
+            target_values=True, initial_values=False, **kwargs
+        )
+        self.cm = cm
+
+    def enter(self, tx):
+        def enter_functional():
+            return self.cm.__enter__()
+
+        tx.output.graph.create_node("call_function", enter_functional, (), {})
+        return variables.ConstantVariable(None, **VariableTracker.propagate(self))
```
jansel, line 264: This is unsound, enter does not always return None.

torchdynamo/variables/misc.py
```diff
@@ -249,6 +249,33 @@ def call_function(
...
+        super(SimpleEnterExitContextWrappingVariable, self).__init__(
+            target_values=True, initial_values=False, **kwargs
+        )
+        self.cm = cm
+
+    def enter(self, tx):
+        def enter_functional():
+            return self.cm.__enter__()
+
+        tx.output.graph.create_node("call_function", enter_functional, (), {})
```
jansel, line 263: This is unsound:
1) It assumes it is ok to put the `enter()` in the graph (it might contain control flow)
2) It leaks the value of `cm` at compile time into runtime
3) If `enter()` has side effects (which they often do -- for example changing grad state), we won't see those side effects at trace time.

torchdynamo/variables/tensor.py
```diff
@@ -269,6 +269,20 @@ def context():
                     need_unwrap=False,
                     **options,
                 )
+        elif isinstance(example_value, contextlib._GeneratorContextManager):
```
jansel, line 272: I don't think we can safely do this for all context managers.  We could have a list of context managers where we have manually confirmed it is safe for though.

jansel (changes requested):

torchdynamo/variables/tensor.py
voznesenskym: Sg, I can narrow it down to a subset.

torchdynamo/variables/misc.py
voznesenskym: 1. Fair
2. Yes
3. Ideally for those, we would not use this object, but I agree, its too easy to misuse. I'll tighten it down to a smaller subset of context managers that do not have these behaviors.

torchdynamo/variables/misc.py
voznesenskym: Agreed. Same as below, I'll narrow it down to a list of specific managers.

voznesenskym: RM

torchdynamo/variables/tensor.py
voznesenskym: rm

jansel (approved):

PR #1524, opened 2022-10-06
Graph Break for Unsupported Fake Tensor Instead of Throwing

eellison:
Previously we would propagate an error to the user if Fake Tensor didn't support some pytorch feature - say, quantization. This caused @mlazos to disable some pytorch tests with torchdynamo.

I think we should just log a warning and graph break instead of the hard error.

### DIFF
--- a/test/dynamo/test_misc.py
+++ b/test/dynamo/test_misc.py
@@ -1535,14 +1535,12 @@ class MiscTests(torchdynamo.testing.TestCase):
     @patch.object(torchdynamo.config, "fake_tensor_propagation", True)
     def test_unsupported_fake_tensor(self):
         def f(x):
-            return torch.quantize_per_tensor(
-                torch.tensor([-1.0, 0.0, 1.0, 2.0]), 0.1, 10, torch.quint8
-            )
+            return torch.quantize_per_tensor(x, 0.1, 10, torch.quint8)
 
         x = torch.randn(2, 2)
-        with self.assertRaises(RuntimeError):
-            opt_f = torchdynamo.optimize_assert(torchdynamo.testing.CompileCounter())(f)
-            opt_f(x)
+        cnts = torchdynamo.testing.CompileCounter()
+        opt_f = torchdynamo.optimize(cnts)(f)
+        self.assertEqual(cnts.op_count, 0)
 
         torchdynamo.reset()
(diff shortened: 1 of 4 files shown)


torchdynamo/utils.py
```diff
@@ -712,12 +712,13 @@ def wrap_fake_exception(fn):
...
-            from .exc import FakeTensorError
+            from .exc import unimplemented
 
-            raise FakeTensorError(
+            log.warning(
                 f"Unsupported: {e.reason} with fake tensor propagation. "
                 "Run with config.fake_tensor_propagation=False"
-            ) from e
+            )
+            raise unimplemented(f"Unsupported: {e.reason} with fake tensor propagation")
```
eellison, line 721: this is more targeted for devs/export path, but maybe it should be more tailored to users

test/dynamo/test_misc.py
```diff
@@ -1535,14 +1535,12 @@ def fn(x, func):
...
-                torch.tensor([-1.0, 0.0, 1.0, 2.0]), 0.1, 10, torch.quint8
-            )
+            return torch.quantize_per_tensor(x, 0.1, 10, torch.quint8)
 
         x = torch.randn(2, 2)
-        with self.assertRaises(RuntimeError):
-            opt_f = torchdynamo.optimize_assert(torchdynamo.testing.CompileCounter())(f)
-            opt_f(x)
+        cnts = torchdynamo.testing.CompileCounter()
+        opt_f = torchdynamo.optimize(cnts)(f)
```
mlazos, line 1542: do you intend to not call opt_f here?

eellison: oops, no

torchdynamo/utils.py
voznesenskym: As long as that config offers a path to avoiding the break, I'd leave the comment reference to it. Both for export, and because some graphs breaks can become perf hits.

voznesenskym (approved):

mlazos (approved):
LGTM, I think this will be good for directing users who run into fake tensor issues

torchdynamo/utils.py
eellison: > because some graphs breaks can become perf hits

I'll note that for all of the unsupported fake tensor stuff we don't support in torchinductor either

PR #1530, opened 2022-10-07
[benchmarks] Skip accuracy large models and Fix GradScaler

anijain2305:
cc @desertfire

### DIFF
--- a/benchmarks/common.py
+++ b/benchmarks/common.py
@@ -922,7 +922,20 @@ class BenchmarkRunner:
     def setup_amp(self):
         if self.args.amp and self.args.training:
             assert self.args.devices == ["cuda"], "AMP is supported only for CUDA"
-            self.grad_scaler = torch.cuda.amp.GradScaler()
+            # AMP training can lead to small loss values which can undeflow
+            # gradient values returning in zero gradients. To solve this
+            # problem, PyTorch introduces GradScaler. GradScaler is a stateful
+            # structure, that scales the loss values to prevent underflow. Loss
+            # values are big at the beginning of training (therefore not
+            # requiring scaling), while loss value tends to be small as network
+            # starts getting better (requiring scaling). GradScaler manages all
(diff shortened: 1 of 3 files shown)


benchmarks/torchbench.py
```diff
@@ -194,6 +194,16 @@
 }
 
 
+SKIP_ACCURACY_CHECK_MODELS = {
+    # Models too large to have eager, dynamo and fp64_numbers simultaneosuly
```
desertfire, line 198: In this case,  can we write the results to CPU and then compare?

desertfire: Hmm, I noticed the `_large` in their names, maybe we are ok here.

benchmarks/common.py
```diff
@@ -922,7 +922,20 @@ def __init__(self):
...
+            # values are big at the beginning of training (therefore not
+            # requiring scaling), while loss value tends to be small as network
+            # starts getting better (requiring scaling). GradScaler manages all
+            # of this fine tuning, checking the gradients are turning to inf,
+            # discarding such batches.
+
+            # Since we are not running a long iteration, default value of
+            # init_scale 65536 is going to turn all gradients to inf. Therefore,
+            # we just use a init_scale of 2.0 for benchmarking purpose.
+            self.grad_scaler = torch.cuda.amp.GradScaler(init_scale=2.0)
```
desertfire, line 938: Look ok to me. I am curious how much will this change our AMP pass rate.

desertfire (approved):

benchmarks/common.py
anijain2305: Jumped to 69% from 33% for TIMM

anijain2305:
CI failing with unrelated error.

benchmarks/torchbench.py
anijain2305: I think we should use CPU for comparison in general. It might remove some OOM error for CI. Lets handle that in a separate PR.

PR #1532, opened 2022-10-07
Update PyTorch pin

desertfire:


yanboliang (approved):

yanboliang:
Sorry I messed up this PR, opened #1550 for testing. If the fix works well, we can add it to our next pin moving.

jansel:
> There is a cuda lib version issue for certain linalg ops, `RuntimeError: Error in dlopen: /home/circleci/project/env/lib/python3.8/site-packages/torch/lib/libtorch_cuda_linalg.so: undefined symbol: cudsh_slascl_`.
> 
> @jansel , I wonder when the migration to core can finish? I would like to skip those failing tests for now than fix a CircleCI build issue, if the migration will be done soon.

My hope is to land torchdynamo to core in the next few days, then source the source-of-true over this weekend.  We will need to port testing infra over.

ngimel:
It's ok to skip these tests temporarily, the build is being fixed on master with the revert of https://github.com/pytorch/pytorch/pull/82695

Issue #1545, opened 2022-10-07
[DDPOptimizer] Compiled subgraphs sometimes return lists


davidberard98:
```python
            def compile_submod(self, submod, args, kwargs):
                """
                Compile the submodule,
                using a wrapper to make sure its output is always a tuple,
                which is required by AotAutograd based compilers
                """
                assert len(kwargs) == 0, "We assume only args for these modules"

                class WrapperModule(torch.nn.Module):
                    def __init__(self, compiled_submod, unwrap_singleton_tuple):
                        super().__init__()
                        self.compiled_submod = compiled_submod
                        self.unwrap_singleton_tuple = unwrap_singleton_tuple

                    def forward(self, *args):
                        x = self.compiled_submod(*args)
                        # TODO(whc)
                        # for some reason the isinstance check is necessary if I split one node per submod
                        # - even though I supposedly wrapped the output in a tuple in those cases, the real
                        # compiled module was still returning a tensor
                        if self.unwrap_singleton_tuple and (isinstance(x, tuple) or isinstance(x, list)): # <- HERE
                            return x[0]
                        return x

                unwrap_singleton_tuple = False
                for sn in submod.graph.nodes:
                    if sn.op == "output":
                        if not isinstance(sn.args[0], tuple):
                            unwrap_singleton_tuple = True
                            sn.args = (sn.args,) # <- Make it a tuple
                submod.recompile()

                wrapper = WrapperModule(
                    self.compiler(submod, args),
                    unwrap_singleton_tuple,
                )
                return wrapper
```

See the line marked `HERE`. We need to add an option for `isinstance(x, list)` for reasons I don't understand. We can see on the line marked "Make it a tuple" that we are already wrapping the output as a tuple, so why is the output sometimes appearing as a list?

wconstab:
i'm not too concerned about this, but at the same time i kinda want to debug it.  if you have a repro case can you paste it?

PR #1546, opened 2022-10-07
DDPOptimizer: accept list output as compiled submodule output

davidberard98:
Not sure why we need to do this, filed an issue to investigate further: https://github.com/pytorch/torchdynamo/issues/1545

Also added a missing recompile, although I don't think it does anything here in the inductor case

### DIFF
--- a/torchdynamo/optimizations/distributed.py
+++ b/torchdynamo/optimizations/distributed.py
@@ -140,6 +140,8 @@ class DDPOptimizer:
                         if not isinstance(sn.args[0], tuple):
                             unwrap_singleton_tuple = True
                             sn.args = (sn.args,)
+                submod.recompile()
+
                 wrapper = WrapperModule(
                     self.compiler(submod, args),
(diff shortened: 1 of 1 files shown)


torchdynamo/optimizations/distributed.py
```diff
@@ -130,7 +130,9 @@ def forward(self, *args):
                         # for some reason the isinstance check is necessary if I split one node per submod
                         # - even though I supposedly wrapped the output in a tuple in those cases, the real
                         # compiled module was still returning a tensor
-                        if self.unwrap_singleton_tuple and isinstance(x, tuple):
+                        if self.unwrap_singleton_tuple and (
+                            isinstance(x, tuple) or isinstance(x, list)
```
anijain2305, line 134: nit - `isinstance(x, (tuple, list))`

anijain2305 (approved):

wconstab (approved):

PR #1547, opened 2022-10-08
Fallback to aten.arange for float types

SherlockNoMad:
Fallback to aten.arange for float types as Triton kernel doesn't support it yet.


ngimel (approved):

torchinductor/lowering.py
```diff
@@ -1105,6 +1105,9 @@ def clone(x, *, memory_format=0):
     register_lowering(aten.lift_fresh_copy)(clone)
 
 
+fallback_arange = fallback_handler(aten.arange)
```
ngimel, line 1108: Should you be calling make_fallback and not fallback_handler here?

PR #1554, opened 2022-10-08
Fix lowering for scatter_reduce

SherlockNoMad:
- Fallback to aten for unsupported reduction types
- Handle 0-dim tensors

### DIFF
--- a/torchinductor/lowering.py
+++ b/torchinductor/lowering.py
@@ -1789,6 +1789,7 @@ def scatter_reduce_(self, dim: int, index, src, reduce, *, include_self: bool = 
     if reduce not in {None, "sum"} or (
         reduce == "sum" and self.get_dtype() in {torch.bool, torch.int64}
     ):
+        self.realize()
         return fallback_scatter_reduce_(
             self, dim, index, src, reduce, include_self=include_self
         )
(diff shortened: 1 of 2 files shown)


ngimel (approved):
lgtm. Can you also please check if it works with negative dim? Line 1818 doesn't look like it would work

torchinductor/lowering.py
```diff
@@ -1778,12 +1777,35 @@ def scatter_reduce(x, dim: int, index, src, reduction_type, **kwargs):
...
 def scatter_reduce_(self, dim: int, index, src, reduce, *, include_self: bool = True):
+    assert reduce in {None, "sum", "prod", "mean", "amax", "amin"}
+
     # TODO: Need to support more reduction type
-    assert reduce is None or reduce in {"sum"}
+    # For reduction of "sum", tl.atomic_add doesn't support bool or int64
+    if reduce not in {None, "sum"} or (
+        reduce == "sum" and self.get_dtype() in {torch.bool, torch.int64}
+    ):
+        return fallback_scatter_reduce_(
```
ngimel, line 1792: you need to make sure that inputs are realized before calling fallback?

SherlockNoMad: updated .

SherlockNoMad:
> lgtm. Can you also please check if it works with negative dim? Line 1818 doesn't look like it would work

Checked. It worked fine with neg dim. 
`indirect_idx[dim] = ops.indirect_indexing(index_loader(idx))` is using python's handling for neg dim.

PR #1563, opened 2022-10-09
fix ipex UT

jiayisunx:


voznesenskym:
How was this passing before?

jiayisunx:
> How was this passing before?

I guess IPEX didn't install successfully in pre-CI environment(IPEX cannot work well with the latest PyTorch). so it skipped.

jiayisunx:
For low prevision(bf16), the results before and after IPEX optimization(include folding and fusion) are not exactly the same, so I remove test_ipex_bf16 in test/dynamo/test_verify_correctness.py

jansel (approved):

PR #1569, opened 2022-10-10
Vectorize kernel by aten::vec

EikanWang:
* __->__ #1569

### DIFF
--- a/test/inductor/test_torchinductor.py
+++ b/test/inductor/test_torchinductor.py
@@ -19,6 +19,7 @@ import torchdynamo
 from torchdynamo.debug_utils import same_two_models
 from torchdynamo.testing import rand_strided
 from torchdynamo.testing import same
+from torchinductor import metrics
 from torchinductor.utils import timed
 
 try:
@@ -3724,6 +3725,75 @@ if HAS_CPU:
         @patch("torch.cuda.is_available", lambda: False)
         def test_timed_cpu_only(self):
             timed(lambda: torch.randn(10), ())
+
+        def test_vec_kernel_cpu_only(self):
+            def fn(x1, x2):
+                # Current, there are some limitations as follows.
+                #   rsqrt:
+                #     assert [both a fallback and a decomp for same kernel: aten.rsqrt.default]
+                #   round:
+                #     couldn't find symbolic meta function/decomposition
+                #   fmod/logical_and/logic_or:
+                #     vec kernel has not support to_type
+                x = torch.abs(x1)
+                x = torch.sin(x)
+                x = torch.cos(x)
+                x = torch.exp(x)
+                x = torch.sqrt(x)
+                x = torch.add(x, x1)
+                x = torch.sub(x, x2)
+                x = torch.mul(x, x1)
+                x = torch.div(x, x1)
+                x = torch.pow(x, 10)
+                x = torch.log(x)
+                x = torch.floor(x)
+                x = torch.ceil(x)
+                x = torch.trunc(x)
+                x = torch.lgamma(x)
+                x = torch.fmod(x, x2)
+                res = x + x2
+                return (res,)
+
+            @contextlib.contextmanager
+            def set_simd(simd_len):
+                org_cpp_simd_len = config.cpp.simdlen
+                config.cpp.simdlen = simd_len
+                yield
+                config.cpp.simdlen = org_cpp_simd_len
+
+            x1 = torch.randn((10, 20))
+            x2 = torch.randn((10, 20))
+
+            with set_simd(8):
+                metrics.reset()
+                traced = make_fx(fn, tracing_mode="symbolic")(x1, x2)
+                compiled = compile_fx_inner(traced, [x1, x2])
+                assert same(fn(x1, x2)[0], compiled(x1, x2)[0], equal_nan=True)
+                assert metrics.generated_simd_vec_kernel_count == 1
+
+                metrics.reset()
+                x1 = x1.permute(1, 0)
+                x2 = torch.randn((20, 10))
+                traced = make_fx(fn, tracing_mode="symbolic")(x1, x2)
+                compiled = compile_fx_inner(traced, [x1, x2])
+                assert same(fn(x1, x2)[0], compiled(x1, x2)[0], equal_nan=True)
+                assert metrics.generated_simd_vec_kernel_count == 0
+
+                metrics.reset()
+                x1 = torch.randn((10, 7))
+                x2 = torch.randn((10, 7))
+                traced = make_fx(fn, tracing_mode="symbolic")(x1, x2)
+                compiled = compile_fx_inner(traced, [x1, x2])
+                assert same(fn(x1, x2)[0], compiled(x1, x2)[0], equal_nan=True)
+                assert metrics.generated_simd_vec_kernel_count == 1
+
+                metrics.reset()
+                x1 = x1[:, 2:3]
+                x2 = x2[:, 2:3]
+                traced = make_fx(fn, tracing_mode="symbolic")(x1, x2)
+                compiled = compile_fx_inner(traced, [x1, x2])
+                assert same(fn(x1, x2)[0], compiled(x1, x2)[0], equal_nan=True)
+                assert metrics.generated_simd_vec_kernel_count == 0
 
 
 if HAS_CUDA:
--- a/torchinductor/codegen/cpp.py
+++ b/torchinductor/codegen/cpp.py
@@ -256,7 +256,7 @@ class CppFastVecOverrides(OpOverrides):
         return f"{a} || {b}"
 
 
-class FastVecChecker:
+class SimdVecChecker:
     def __init__(self):
         super().__init__()
         self.passed = True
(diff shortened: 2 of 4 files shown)


torchinductor/codegen/cpp.py
```diff
@@ -706,7 +1106,23 @@ def mark_parallel(self, par_depth):
...
+        most_inner_loop = self.loops[-1]
+        main_loop_range = ir.IndexingDiv(most_inner_loop.size, sympy_factor)
+
+        main_loop = LoopLevel(most_inner_loop.var, main_loop_range)
+        main_loop.parallel = most_inner_loop.parallel
+
+        offset = main_loop_range * sympy_factor
+        tail_loop = LoopLevel(most_inner_loop.var, most_inner_loop.size)
+        tail_loop.offset = offset
+        tail_loop.parallel = most_inner_loop.parallel
```
jgong5, line 1122: Do we need to copy the `reduction_vars` too?

torchinductor/codegen/cpp.py
```diff
@@ -504,6 +596,239 @@ def write_to_suffix(self):
...
+        self.fast_vec_list = []
+        for dict_obj in CppSimdVecOverrides.__dict__:
+            if isinstance(CppSimdVecOverrides.__dict__[dict_obj], staticmethod):
+                self.fast_vec_list.append(dict_obj)
+        self.exit_stack = contextlib.ExitStack()
+
+    def is_legal_data_access(self, var: sympy.Symbol, index: sympy.Expr):
+        return self.is_var_irrevelant(var, index) or self.is_single_step_var(var, index)
+
+    def could_be_simd_vec(self, name: str, index: sympy.Expr):
```
jgong5, line 672: name it simply as `can_vec`?

torchinductor/codegen/cpp.py
```diff
@@ -504,6 +596,239 @@ def write_to_suffix(self):
...
+        loop_with_tail.tail_loop.simd_vec = False
+
+        loop_with_tail.main_loop_body = self.simd_vec_kernel
+        loop_with_tail.tail_loop_body = self.simd_omp_kernel
+        return loop_nest
+
+    def codegen_loops(self, code, worksharing):
+        threads = config.cpp.threads
+        if threads < 1:
+            threads = torch.get_num_threads()
```
jgong5, line 776: Same logic in `CppKernel`, how about abstract a common util func?

torchinductor/codegen/cpp.py
```diff
@@ -504,6 +596,239 @@ def write_to_suffix(self):
...
+
+        itervars = self.simd_vec_kernel.itervars
+        rangs = self.simd_vec_kernel.ranges
+        loops = [LoopLevel(var, size) for var, size in zip(itervars, rangs)]
+        loops_nest_non_reduc, _ = LoopNest(loops[: self.reduction_depth]), LoopNest(
+            loops[self.reduction_depth :]
+        )
+
+        # TODO: Support reductions
+        # assert reductions is None or len(reductions.loops) == 0
```
jgong5, line 793: So, do you need to guarantee no reduction here?

torchinductor/codegen/cpp.py
```diff
@@ -504,6 +596,239 @@ def write_to_suffix(self):
...
+                (loop_with_tail.main_loop, loop_with_tail.main_loop_body),
+                (loop_with_tail.tail_loop, loop_with_tail.tail_loop_body),
+            ):
+
+                code.writelines(loop.lines())
+                with contextlib.ExitStack() as stack:
+                    stack.enter_context(code.indent())
+                    code.splice(kernel.loads)
+                    code.splice(kernel.compute)
+                    code.splice(kernel.stores)
```
jgong5, line 829: ditto: suggest to abstract common logic out to avoid duplicate code as `CppKernel`.

torchinductor/codegen/cpp.py
```diff
@@ -527,18 +852,45 @@ def can_fuse_horizontal(node1, node2):
     def can_fuse_vertical(cls, node1, node2):
         return cls.can_fuse_horizontal(node1, node2) and not node1.is_reduction()
 
-    def codegen_nodes(self, nodes):
+    def check_simd_vec(self, nodes):
```
jgong5, line 855: Simpler to name it `can_vec`?

torchinductor/codegen/cpp.py
```diff
@@ -558,7 +910,35 @@ def codegen_nodes(self, nodes):
...
+        metrics.generated_kernel_count -= 1
+        # Maitain the metrics kernel count
+        if simd_vec_kernel:
+            metrics.generated_kernel_count -= 1
+
+        cpp_kernel_proxy = CppKernelProxy(
+            kernel_group.args, kernel_group.ws.num_threads
+        )
+        cpp_kernel_proxy.simd_vec_kernel = simd_vec_kernel
+        cpp_kernel_proxy.simd_omp_kernel = simd_omp_kernel
```
jgong5, line 939: Only one kind of kernel would be used at a time, right? Then, why can't we do the dispatch here directly to invoke either the original CppKernel or the vec Cppkernel? Why have to wrap another `CppKernelProxy` here?

jgong5 (changes requested):

jansel:

torchinductor/codegen/cpp.py
EikanWang: Nope. Actually, we only use the code of `CppKernel` and `CppVecKernel` but do not use its loops. The pseudo code is as follows.
```python
for cpp_kernel_proxy_loop_1:
  for cpp_kernel_proxy_loop_2:
    for main loop of cpp_kernel_proxy_loop_with_tail:
      CppVecKernel.load
      CppVecKernel.compute
      CppVecKernel.store
    for tail loop of cpp_kernel_proxy_loop_with_tail:
      CppKernel.load
      CppKernel.compute
      CppKernel.store
```

torchinductor/codegen/cpp.py
EikanWang: This logic is `CppVecKernel` specific. It is not applicable for `CppKernel` because the most inner loop is `LoopLevelWithTail`

torchinductor/codegen/cpp.py
EikanWang: Actually, it is possible to reduction here because I have filtered the reduction out. So I removed the comment and am supporting reduction now.

torchinductor/codegen/cpp.py
```diff
@@ -504,6 +596,239 @@ def write_to_suffix(self):
         (self.loads, self.compute, self.stores, self.cse) = prior
 
 
+class CppSimdVecKernel(CppKernel):
```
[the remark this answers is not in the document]
EikanWang: Done

torchinductor/metrics.py
```diff
@@ -1,8 +1,12 @@
 # counter for tracking how many kernels have been generated
 generated_kernel_count = 0
+generated_simd_vec_kernel_count = 0
 
 
 # reset all counters
 def reset():
     global generated_kernel_count
+    global generated_simd_vec_kernel_count
```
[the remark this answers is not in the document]
EikanWang: Done

test/inductor/test_torchinductor.py
```diff
@@ -3725,6 +3726,75 @@ def set_num_threads(num_threads):
...
+                x1 = torch.randn((10, 7))
+                x2 = torch.randn((10, 7))
+                traced = make_fx(fn, tracing_mode="symbolic")(x1, x2)
+                compiled = compile_fx_inner(traced, [x1, x2])
+                assert same(fn(x1, x2)[0], compiled(x1, x2)[0], equal_nan=True)
+                assert metrics.generated_simd_vec_kernel_count == 1
+
+                metrics.reset()
+                x1 = x1[:, 2:3]
+                x2 = x2[:, 2:3]
```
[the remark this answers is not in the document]
EikanWang: If the last time is non-contiguous, it cannot pass check that `f(x) - f(x+1) == 1`.

torchinductor/codegen/cpp.py
```diff
@@ -420,10 +511,11 @@ def codegen_loops(self, code, worksharing):
 
         if config.cpp.simdlen:
             # TODO(jansel): detect stride-1 dimension and vectorize that
-            if reductions:
+            if reductions and len(reductions.loops) > 0:
```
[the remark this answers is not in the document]
EikanWang: The `__bool__` only checks the `loops` is `None` or not but not make sure the `loops` is not empty.

torchinductor/codegen/cpp.py
```diff
@@ -420,10 +511,11 @@ def codegen_loops(self, code, worksharing):
...
                 reductions.loops[-1].simd = True
             else:
-                loops.loops[-1].simd = True
+                if len(loops.loops) > 0:
```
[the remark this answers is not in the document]
EikanWang: ditto

torchinductor/codegen/cpp.py
EikanWang: Done

torchinductor/codegen/cpp.py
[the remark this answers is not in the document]
jgong5: No, bool(list) checks both None and empty, both of which would return `False`.

PR #1573, opened 2022-10-10
Unskip tests

ngimel:


SherlockNoMad (approved):
Nice... which PR fixed them? 
log_softmax.dtype was failing last time I checked ...

ngimel:
I don't know, but I checked them today and they are fine. `test_tmp_not_defined_issue2_cpu` failure doesn't seem related (cc @deserfire just in case), so I'm landing.

PR #1574, opened 2022-10-10
Handle the case where all the inputs are scalar

desertfire:


test/inductor/test_torchinductor.py
```diff
@@ -1019,6 +1019,18 @@ def fn(a, b):
             ),
         )
 
+    def test_div8(self):
+        def fn(a, b):
+            return (
+                aten.div(a, b, rounding_mode=None),
```
ngimel, line 1025: I've just checked out this PR, and it produces wrong results - `div(1000, 320, rounding_mode=None)` produces `3.125` in eager, and `3.0` in inductor

desertfire: If I plug in `1000, 320` in the test here, it passes. I also wrote a simple local test and it does produce `3.125`. Can you share your test code so that I can take a look?

ngimel:
```
@torchdynamo.optimize("inductor")
def div(x, y):
    return torch.ops.aten.div(x, y, rounding_mode=None)


x = 1000
y = 320
print(div(x,y))
```
The generated cpp code is 
```
extern "C" void kernel(float* __restrict__ out_ptr0)
{
    {
        {
            auto tmp0 = static_cast<long>(1000);
            auto tmp1 = static_cast<long>(320);
            auto tmp2 = tmp0 / tmp1;
            out_ptr0[0] = tmp2;
        }
    }
}
```
which indeed would produce 3.0

ngimel: But actually
``` 
@torchdynamo.optimize("inductor")
def mul(x, y):
    return torch.ops.aten.mul(x, y)
```
fails with `4, 3.5` and passes with `3.5, 4` because `make_pointwise` takes dtype from the first arg.

desertfire: Good catch! The `div` operator is really a rabbit hole, lol, so many corner cases to handle...

ngimel: Cool, can you also add tests for e.g. `mul` that doesn't have `override_return_dtype`?

ngimel (approved):

PR #1579, opened 2022-10-10
fix negative indexing value for select_scatter

### LINKED ISSUE #1567: [inductor] tts_angular - bad accuracy for `select_scatter`
Repro

~~~

import torch
import torch.fx as fx
from math import inf

# torch version: 1.14.0a0+git65b4080
# torch cuda version: 11.6
# torch git version: 65b408074f4ecc99faf5720ea5b3570a483ec9f4


# CUDA Info: 
# nvcc: NVIDIA (R) Cuda compiler driver 
# Copyright (c) 2005-2022 NVIDIA Corporation 
# Built on Thu_Feb_10_18:23:41_PST_2022 
# Cuda compilation tools, release 11.6, V11.6.112 
# Build cuda_11.6.r11.6/compiler.30978841_0 

# GPU Hardware Info: 
# NVIDIA A100-SXM4-40GB : 8 


from torch.nn import *
class Repro(torch.nn.Module):
    def __init__(self):
        super().__init__()

    
    
    def forward(self, add, new_zeros):
        select_scatter = torch.ops.aten.select_scatter.default(new_zeros, add, 1, -1);  new_zeros = add = None
        return (select_scatter,)
        
args = [((2, 256), (256, 1), torch.float32, 'cuda'), ((2, 50, 256), (12800, 256, 1), torch.float32, 'cuda')]
args = [rand_strided(sh, st, dt, dev) for (sh, st, dt, dev) in args]
mod = make_fx(Repro().to(device="cuda"))(*args)

from torchdynamo.debug_utils import same_two_models

compiled = compile_fx_inner(mod, args)
assert same_two_models(mod, compiled, args, only_fwd=True), "Accuracy failed"
~~~

cc @desertfire @SherlockNoMad

ngimel:
Fixes #1567
However, this creates a guard on the indexed dimension - don't see a way of fixing it otherwise, symbolic ops fail. My attempts to leave only symbolic manipulation led to 
```
  File "/scratch/ngimel/work/pytorch/torch/fx/_symbolic_trace.py", line 344, in create_arg
    return super().create_arg(a)
  File "/scratch/ngimel/work/pytorch/torch/fx/proxy.py", line 140, in create_arg
    return type(a)(self.create_arg(elem) for elem in a)
  File "/scratch/ngimel/work/pytorch/torch/fx/proxy.py", line 140, in <genexpr>
    return type(a)(self.create_arg(elem) for elem in a)
  File "/scratch/ngimel/work/pytorch/torch/fx/_symbolic_trace.py", line 344, in create_arg
    return super().create_arg(a)
  File "/scratch/ngimel/work/pytorch/torch/fx/proxy.py", line 165, in create_arg
    raise NotImplementedError(f"argument of type: {type(a)}")
NotImplementedError: argument of type: <class 'sympy.core.add.Add'>
```
in inductor compilation.
cc @chillee, this will be problematic for dynamic shapes.

### DIFF
--- a/torchinductor/lowering.py
+++ b/torchinductor/lowering.py
@@ -1207,12 +1207,10 @@ def select_scatter(x, src, dim: int, index: int):
     assert x.get_dtype() == src.get_dtype()
     x_loader = x.make_loader()
     dim = _validate_dim(x, dim, 0)
-    # to check that index is valid we need to freeze the size
-    size_hint = V.graph.sizevars.size_hint(x.get_size()[dim])
-    V.graph.sizevars.guard_equals(x.get_size()[dim], size_hint)
     if index < 0:
-        index = index + size_hint
-    assert 0 <= index and index < size_hint
+        index = index + x.get_size()[dim]
+    V.graph.sizevars.guard_leq(0, index)
+    V.graph.sizevars.guard_lt(index, x.get_size()[dim])
     src = expand(unsqueeze(src, dim), x.get_size())
     src_loader = src.make_loader()
 
@@ -1220,7 +1218,7 @@ def select_scatter(x, src, dim: int, index: int):
         return ops.where(
             ops.eq(
                 ops.index_expr(idx[dim], torch.int32),
-                ops.constant(index, torch.int32),
+                ops.index_expr(index, torch.int32),
             ),
             src_loader(idx),
             x_loader(idx),
--- a/torchinductor/lowering.py
+++ b/torchinductor/lowering.py
@@ -1207,6 +1207,12 @@ def select_scatter(x, src, dim: int, index: int):
     assert x.get_dtype() == src.get_dtype()
     x_loader = x.make_loader()
     dim = _validate_dim(x, dim, 0)
+    # to check that index is valid we need to freeze the size
+    size_hint = V.graph.sizevars.size_hint(x.get_size()[dim])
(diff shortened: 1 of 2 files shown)


Chillee:
Let's not land this lol 🤔 

Perhaps the issue is that we're asserting that `index` is a constant? https://github.com/pytorch/torchdynamo/pull/1579/files#diff-71816da3ef3537afea5fde19d6ede5f9d932f57a38e78b61f465ed481730a561R1223

torchinductor/lowering.py
```diff
@@ -1207,6 +1207,12 @@ def select_scatter(x, src, dim: int, index: int):
     assert x.get_dtype() == src.get_dtype()
     x_loader = x.make_loader()
     dim = _validate_dim(x, dim, 0)
+    # to check that index is valid we need to freeze the size
+    size_hint = V.graph.sizevars.size_hint(x.get_size()[dim])
+    V.graph.sizevars.guard_equals(x.get_size()[dim], size_hint)
+    if index < 0:
```
jansel, line 1213: Perhaps we could only guard_static_shape inside this condition to limit impact?

ngimel:
Ok yeah, making `index` into `indexing_exp` works.

jansel (approved):

Chillee (approved):

PR #1581, opened 2022-10-11
Fix DDPOptimizer for custom layers

wconstab:
- attribute storage used by params that are accessed via get_attr nodes and require grad. These would be the case for params of custom layers that are traced through instead of captured as top level modules
- cosmetic fix for recording an empty graph of size 0 params before the real graphs in the splitter
- add a test for the custom layer and confirm it failed before the change

### DIFF
--- a/torchdynamo/optimizations/distributed.py
+++ b/torchdynamo/optimizations/distributed.py
@@ -44,13 +44,13 @@ class DDPOptimizer:
         bucket_actual_sizes = []
         node_splits = [[]]
         for node in reversed(gm.graph.nodes):
+            if node.op == "output" or node.op == "placeholder":
+                continue
+
             if bucket_bytes >= self.bucket_bytes_cap:
                 bucket_actual_sizes.insert(0, bucket_bytes)
                 bucket_bytes = 0
                 node_splits.insert(0, [])
-
-            if node.op == "output" or node.op == "placeholder":
-                continue
 
             elif node.op == "call_module":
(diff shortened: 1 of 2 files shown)


torchdynamo/optimizations/distributed.py
```diff
@@ -63,6 +63,10 @@ def compile_fn(self, gm: fx.GraphModule, example_inputs: List[torch.Tensor]):
                 )
                 bucket_bytes += params_size_b
                 # print(f"accumulated {params_size_b} b from {node}")
+            elif node.op == "get_attr":
+                maybe_param = getattr(gm, node.target)
```
aazzolini, line 67: cool, i think theoretically this would also catch buffers (which we generally also want for DDP -- in case we didn't pass synchronize_buffers=False) , but it will also catch other tensors that happen to be attributes of an nn.Module even if they are neither parameters or buffers. I think that's ok .

aazzolini (approved):
Thanks for the quick follow up on the issue reported!

torchdynamo/optimizations/distributed.py
wconstab: hmm, what are 'buffers' used for and is there any other way to identify them? i'm guessing we don't have that information if we've captured them out of a custom layer that's been traced through, so this may be the best we can do.

davidberard98 (approved):
looks like it fixes graph breaks for hf_T5_large and hf_GPT2_large!

PR #1582, opened 2022-10-11
Update CI skips/fails

wconstab:
(This is partly a smoke test so i can observe the CI hang issue.  Will it happen on a PR or it only happened after merged?)

- clean up skipifs that should not be relevant anymore since said APIs have been landed for a while


mlazos:
You can create a PR that points the pin in pytorch core to any arbitrary commit and it repro'd in core CI

davidberard98:
running https://github.com/pytorch/pytorch/pull/86217 pointed at this commit.

wconstab:
as i look at this, quite a few of the dynamo CI jobs have been running 'test' for >25 min without producing much output.  is that the hang? or the hang only happens on the pytorch CI side and i just need to wait longer on dynamo side

mlazos:
> as i look at this, quite a few of the dynamo CI jobs have been running 'test' for >25 min without producing much output. is that the hang? or the hang only happens on the pytorch CI side and i just need to wait longer on dynamo side

It didn't repro on the dynamo side at all which was strange. I don't remember the exact jobs, but it was one of the two jobs with torchdynamo in the name in Pytorch CI that would fail after like 4 hours. Here's an image of what I had 
image

I knew it was one of those three test suites and guessed it was this one.

wconstab:
i wonder if it is a bad interaction with init/destroy process group happening more than once inside a process.

i noticed in the runs above `test_misc.py:2664` failed `test_empty_graph` which afaict is just initializing a dist pg for the purpose of testing that dynamo can call 'get_world_size' from inside a graph.

I'm moving that test into test_distributed.py so it can share the class-wide init/destroy processgroup and see if that helps.

wconstab:
@davidberard98  @mlazos does this look good to you or is there another test i can run to confirm?

mlazos:
> @davidberard98 @mlazos does this look good to you or is there another test i can run to confirm?

This looks good to me, as long as you have run it in pytorch core CI pointing at this commit and it passes. That's the only test I saw failing.

davidberard98 (approved):
I just updated the pin in https://github.com/pytorch/pytorch/pull/86217 (rebased on viable/strict), as long as that passes then it seems good to me

mlazos (approved):

PR #1583, opened 2022-10-11
Add minifier test for failures during AOTAutograd

williamwen42:
Title


test/dynamo/test_minifier.py
```diff
@@ -58,7 +58,7 @@ def inner():
         inner()
         self.assertTrue(os.path.exists(repro_file))
 
-    def test_after_aot(self):
+    def _test_around_aot(self, before_aot):
```
anijain2305, line 61: `after_aot` is used because of the flags `REPRO_AFTER` aot and dynamo. So, the test name might be confusing.

anijain2305 (approved):

PR #1586, opened 2022-10-11
Add optimizer to OSS models

mlazos:


anijain2305 (approved):

benchmarks/common.py
```diff
@@ -918,6 +918,13 @@ def setup_amp(self):
             self.grad_scaler = torch.cuda.amp.GradScaler()
             self.autocast = torch.cuda.amp.autocast
 
+    def init_optimizer(self, device, params):
+        if device == "cuda":
+            # capturable is only supported on cuda at the moment
+            self.optimizer = torch.optim.Adam(params, capturable=True)
+        else:
+            self.optimizer = None
```
voznesenskym, line 926:
```suggestion
    def init_optimizer(self, device, params):
        self.optimizer = torch.optim.Adam(params, capturable=True) if device == "cuda" else None
```

benchmarks/torchbench.py
```diff
@@ -344,6 +346,8 @@ def forward_and_backward_pass(self, mod, inputs, collect_outputs=True):
             pred = mod(*cloned_inputs)
             loss = self.compute_loss(pred)
         self.grad_scaler.scale(loss).backward()
+        if self.optimizer is not None:
+            self.optimizer.step()
```
voznesenskym, line 350: you do this twice, just make a `self.optimizer_step()`?

voznesenskym (approved):

benchmarks/common.py
mlazos: I actually have to change this condition due to some failures, I probably won't make it one line, since it will get longer now.

PR #1587, opened 2022-10-11
Unified symbolic shape variables between AOTAutograd and Inductor

Chillee:
Still somewhat WIP.

Goes together with this PR: https://github.com/pytorch/pytorch/pull/86659

Some notes:

1. I'm switching over to using `sympy.Symbol(..., integer=True, positive=True)`. I think there's benefits in simplifications here.

For ex:
image

2. I'm passing `ShapeEnv` to `SizeVarAllocator` and reusing `var_to_val`. Right now that's the only thing that's shared, but I'll probably rewrite some more of inductor's `SizeVarAllocator` to reuse `ShapeEnv`.

3. i moved the `symbolic_sizes_strides` logic into core: https://github.com/pytorch/pytorch/pull/86659/files#diff-03557db7303b8540f095b4f0d9cd2280e1f42f534f67d8695f756ec6c02d3ec7R332

### DIFF
--- a/torchinductor/config.py
+++ b/torchinductor/config.py
@@ -87,7 +87,7 @@ class cpp:
 class triton:
 
     # Use cudagraphs on output code
-    cudagraphs = True
+    cudagraphs = False
 
     # choose conv backend, "aten" or "triton" or "autotune"
     convolution = "aten"
--- a/torchinductor/lowering.py
+++ b/torchinductor/lowering.py
@@ -21,7 +21,9 @@ from . import ir
 from . import overrides
 from .decomposition import decompositions
 from .decomposition import get_decompositions
-from .ir import ExpandView, IndexingConstant, IndexingDiv
+from .ir import ExpandView
+from .ir import IndexingConstant
+from .ir import IndexingDiv
 from .ir import PermuteView
 from .ir import Pointwise
 from .ir import Reduction
@@ -257,16 +259,17 @@ def promote_constants(inputs):
     out = []
     for x in inputs:
         if isinstance(x, (int, float)):
-            out.append(ExpandView.create(
-                ir.Constant(x, ex.get_dtype(), ex.get_device()), list(ex.get_size())
-            ))
+            out.append(
+                ExpandView.create(
+                    ir.Constant(x, ex.get_dtype(), ex.get_device()), list(ex.get_size())
+                )
+            )
         elif isinstance(x, sympy.Expr):
             out.append(IndexingConstant(x, ex.get_dtype(), ex.get_device()))
         else:
             out.append(x)
 
     return out
-
 
 
 def make_pointwise(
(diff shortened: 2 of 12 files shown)


torchinductor/lowering.py
```diff
@@ -430,12 +436,13 @@ def where(cond, a, b):
...
-        b = constant_like(b)(a)
-
-    dtype = torch.promote_types(a.get_dtype(), b.get_dtype())
+    a, b = promote_constants([a, b])
+    # if isinstance(a, (float, int)):
+    #     a = constant_like(a)(b)
+    # if isinstance(b, (float, int)):
+    #     b = constant_like(b)(a)
+    # dtype = torch.promote_types(a.get_dtype(), b.get_dtype())
+    dtype = a.get_dtype()
```
Chillee, line 445: This line causes a confusing error for some reason, need to debug.

voznesenskym: What's the error?

torchinductor/config.py
```diff
@@ -87,7 +87,7 @@ class cpp:
 class triton:
 
     # Use cudagraphs on output code
-    cudagraphs = True
+    cudagraphs = False
```
ngimel, line 90: is this intended?

torchinductor/lowering.py
```diff
@@ -251,19 +253,23 @@ def broadcast_symbolic_shapes(a, b):
 
 
 def promote_constants(inputs):
-    if not any(isinstance(x, (int, float)) for x in inputs):
+    if not any(isinstance(x, (sympy.Expr, int, float)) for x in inputs):
```
ngimel, line 256: this would need special handling when there are no tensor inputs (@desertfire recently landed a PR with a change here)

torchinductor/sizevars.py
```diff
@@ -43,11 +44,11 @@ class PositiveGuard:
 
 
 class SizeVarAllocator(object):
-    def __init__(self, prefix="s", zero_one_const=True):
+    def __init__(self, shape_env, zero_one_const=True):
```
ngimel, line 47: don't we need prefix?

Chillee: No, `SizeVarAllocator` is only used to allocate symbolic shapes at the input, nowhere else.

torchinductor/config.py
Chillee: Just a temporary thing.

torchinductor/lowering.py
Chillee: image

Chillee: Found the issue - the python dispatcher throws this error if you hit a aten op with no tensor inputs.

Chillee:
I tested this on torchbench, and no performance decreases (I actually see a slight increase). Maybe since I turned `torchinductor.config.dynamic_shapes` to False by default.

jansel (commented):
Approach LGTM, though test failures seem real:
```
AttributeError: 'FakeTensorMode' object has no attribute 'shape_env'
```

Chillee:
@jansel Yeah it depends on changes made in core (https://github.com/pytorch/pytorch/pull/86659)

jansel:

PR #1590, opened 2022-10-11
Check gradient with OpInfo tests

SherlockNoMad:
Compare inductor's computed gradient with eager computed gradient with OpInfo samples.


test/inductor/test_torchinductor.py
```diff
@@ -243,6 +244,43 @@ def run(*ex, **kwargs):
...
+                if arg.requires_grad:
+                    leaf_tensors.append(arg)
+            return leaf_tensors
+
+        def compute_grads(args, kwrags, results):
+            flat_results, _ = tree_flatten(results)
+            flat_diff_results = [r for r in flat_results if r.requires_grad]
+            assert len(flat_diff_results) > 0
+
+            grads = [torch.ones(r.shape, device=r.device, dtype=r.dtype) for r in flat_diff_results]
```
ngimel, line 265: better to have random grads, `torch.ones` can hide the bugs

test/inductor/test_torchinductor.py
```diff
@@ -243,6 +244,43 @@ def run(*ex, **kwargs):
...
+
+            leaf_tensors = gather_leaf_tensors(ref_inputs, ref_kwargs)
+            assert len(leaf_tensors) > 0
+            return torch.autograd.grad(flat_diff_results, leaf_tensors,
+                                    grads, allow_unused=True, retain_graph=True)
+
+        correct_grad = compute_grads(ref_inputs, ref_kwargs, correct)
```
ngimel, line 272: should we be using gradcheck as @albanD suggested?

SherlockNoMad: IIUC, autograd.gradcheck computes gradients via small finite differences. It's useful for checking gradient's correctness when a reference implementation is not available. 

For inductor, we can use eager's result as the reference.

test/inductor/test_torchinductor.py
```diff
@@ -243,6 +244,52 @@ def run(*ex, **kwargs):
...
+            leaf_tensors = []
+            args, args_spec = tree_flatten(args)
+            kwargs, kwargs_spec = tree_flatten(kwargs)
+            args = args + kwargs
+            for arg in args:
+                if not isinstance(arg, torch.Tensor):
+                    continue
+                if arg.requires_grad:
+                    leaf_tensors.append(arg)
+            return leaf_tensors
```
voznesenskym, line 259: spec is unused, and args and kwargs are both flattened - can we do something like:

`leaf_tensors = [arg for arg in tree_flatten(args, kwargs)[0] if isinstance(arg, torch.Tensor)]`

(Note: did not run the above, but list comprehension should be able to one line it)

test/inductor/test_torchinductor.py
```diff
@@ -243,6 +244,52 @@ def run(*ex, **kwargs):
...
+            assert len(flat_diff_results) > 0
+
+            grads = [
+                torch.ones(r.shape, device=r.device, dtype=r.dtype)
+                for r in flat_diff_results
+            ]
+
+            leaf_tensors = gather_leaf_tensors(ref_inputs, ref_kwargs)
+            assert len(leaf_tensors) > 0
+            return torch.autograd.grad(
```
voznesenskym, line 273: should len of leaf_tensors match len of grads?

test/inductor/test_torchinductor.py
```diff
@@ -243,6 +244,52 @@ def run(*ex, **kwargs):
...
+            kwargs, kwargs_spec = tree_flatten(kwargs)
+            args = args + kwargs
+            for arg in args:
+                if not isinstance(arg, torch.Tensor):
+                    continue
+                if arg.requires_grad:
+                    leaf_tensors.append(arg)
+            return leaf_tensors
+
+        def compute_grads(args, kwrags, results):
```
voznesenskym, line 261: Nit: maybe pull to top level? It took me a second to realize that it wasn't calling itself in the compute_grads below.

test/inductor/test_torchinductor_opinfo.py
```diff
@@ -394,6 +403,11 @@ def wrapper_set_seed(op, *args, **kwargs):
     "new_empty_strided": {"assert_equal": False},
     "randn": {"assert_equal": False},
     ("nn.functional.tanhshrink", "cuda", f16): {"atol": 3e-4, "rtol": 0.001},
+    "gradient": {"check_gradient": False},  # segfault on check_gradient
+    "linalg.solve_triangular": {"check_gradient": False},  # segfault on check_gradient
+    "linalg.lu_factor": {"check_gradient": False},
+    "linalg.lu_factor_ex": {"check_gradient": False},
+    "linalg.lu_solve": {"check_gradient": False},
```
voznesenskym, line 410: Can you comment why these are skipped as well?

voznesenskym (approved):
LGTM - comments for code cleanliness and brevity.

test/inductor/test_torchinductor.py
SherlockNoMad: update.

test/inductor/test_torchinductor.py
SherlockNoMad: no... 
leaf_tensors are the inputs that requires grad
`grads` are the outputs that requires grad

test/inductor/test_torchinductor_opinfo.py
```diff
@@ -443,18 +458,22 @@ def test_comprehensive(self, device, dtype, op):
             self.skipTest(f"{op_name} in {dtype} not supported")
         elif dtype in inductor_expected_failures_single_sample[device_type].get(
             op_name, set()
+        ) or dtype in inductor_gradient_expected_failures_single_sample[
```
voznesenskym, line 461: If there's a large list of grad failures, we should consider splitting the test into forward or exclude, and then backward or exclude (instead of forward+backward or exclude from either). So as not to regress forward tests.

test/inductor/test_torchinductor.py
SherlockNoMad: ok.

test/inductor/test_torchinductor.py
albanD: The benefit of using gradcheck is that you don't have to write all the code to get proper grad outputs and do the comparison by hand.

test/inductor/test_torchinductor.py
albanD: This might be numerically unstable for big Tensor, for gradcheck, we use a unit-norm Tensor as it proved to be the most stable.

test/inductor/test_torchinductor_opinfo.py
```diff
@@ -394,6 +430,11 @@ def wrapper_set_seed(op, *args, **kwargs):
     "new_empty_strided": {"assert_equal": False},
     "randn": {"assert_equal": False},
     ("nn.functional.tanhshrink", "cuda", f16): {"atol": 3e-4, "rtol": 0.001},
+    "gradient": {"check_gradient": False},  # segfault on check_gradient
+    "linalg.solve_triangular": {"check_gradient": False},
+    "linalg.lu_factor": {"check_gradient": False},
+    # "linalg.lu_factor_ex": {"check_gradient": False},
```
albanD, line 436: remove these?

test/inductor/test_torchinductor_opinfo.py
```diff
@@ -489,24 +534,35 @@ def fn(*args, **kwargs):
...
-                        copy_to_cuda=False,
-                        reference_in_float=False,
-                        **additional_kwargs,
+                        **adjusted_kwargs,
                     )
                 elif device_type == "cpu":
+                    adjusted_kwargs = {
+                        "check_lowp": False,
+                        "nopython": True,
+                        # skip checking gradient on CPU for now
```
albanD, line 556: Why?

test/inductor/test_torchinductor_opinfo.py
SherlockNoMad: commented.

test/inductor/test_torchinductor_opinfo.py
SherlockNoMad: We have 31 new ops failing due to gradient check... which is manageable...

test/inductor/test_torchinductor.py
SherlockNoMad: As discussed offline with Alban, I will stick with current approach, coz gradcheck doesn't work well with low precision floats.

ngimel: Good point about fp32 (not really low precision :-) )

test/inductor/test_torchinductor.py
SherlockNoMad: Updated to use unit norm now.

test/inductor/test_torchinductor_opinfo.py
SherlockNoMad: Inductor with CPU is not the focus now... We can enable this when CPU becomes a priority.

PR #1600, opened 2022-10-11
Free inputs to inductor when they are no longer needed

eellison:
Free inputs when they are no longer needed in torchinductor. 

Follow up to https://github.com/pytorch/pytorch/pull/83137/, for more context read [this](https://github.com/pytorch/pytorch/pull/83137/#issuecomment-1211320670).

cc @anijain2305

### DIFF
--- a/test/inductor/test_torchinductor.py
+++ b/test/inductor/test_torchinductor.py
@@ -6,12 +6,14 @@ import importlib
 import random
 import sys
 import unittest
+import weakref
 from unittest.mock import patch
 
 import torch
 from torch.fx.experimental.proxy_tensor import make_fx
 from torch.nn import functional as F
 from torch.testing._internal.common_utils import TestCase as TorchTestCase
+from torch.utils._python_dispatch import TorchDispatchMode
 from torch.utils._pytree import tree_flatten
 from torch.utils._pytree import tree_unflatten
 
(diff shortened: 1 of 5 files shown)


test/inductor/test_torchinductor.py
```diff
@@ -3655,6 +3656,63 @@ def fn(x, y):
...
+                    a = x + y
+                    return (a @ a,)
+
+                fn_fx = make_fx(fn)(inps[0], inps[1])
+                fn_compiled = compile_fx_inner(fn_fx, inps)
+
+                test_self = self
+                matmul_seen = False
+
+                class TestRefMode(TorchDispatchMode):
```
Chillee, line 3688: Lol, clever test.

Chillee (commented):
Looks good to me overall, probably want to verify memory improvements before landing.

Chillee:
@eellison I think I verified that all the references in AOTAutograd were already cleared (at least at the time of the PR), since I was able to see memory reductions with the `aot_eager` backend.

anijain2305:
@Chillee 
image

Is this optimization already enabled for `aot_eager`? The numbers seem to suggest, that it's not. Should we try to enable it with `aot_eager` first just as a confirmation that nothing new has creeped into the memory footprint?

eellison:
@Chillee I verified memory reduction on a small example, although there may be other issues with the larger models

Chillee:
@eellison I've verified that this seems to clear the memory as expected using Sanket's memory profiler. On the second run, this is how the memory looks like.

image

This is what it looks like before this PR

image

There appears to still be a memory spike at the beginning of the backwards pass (before we run any code). Might be caused by some extra copying, or perhaps by Triton compilation.

image

Chillee (approved):

PR #1585, opened 2022-10-11
Move the triton pin

desertfire:
WARNING: you need to upgrade triton after this PR.

Summary: 
* The pinned version has a proper fix for https://github.com/pytorch/torchdynamo/issues/1450
* Also fixes #1564

PR #1608, opened 2022-10-12
Introduce a triton version checking

desertfire:
Summary: to avoid confusion after a recent triton upgrade in https://github.com/pytorch/torchdynamo/pull/1585

### DIFF
--- a/torchinductor/utils.py
+++ b/torchinductor/utils.py
@@ -1,5 +1,6 @@
 import collections
 import functools
+import logging
 import operator
 import time
 from importlib import import_module
@@ -15,6 +16,7 @@ from torch.fx.immutable_collections import immutable_list
 
 from . import config
 
+log = logging.getLogger(__name__)
 VarRanges = Dict[sympy.Expr, sympy.Expr]
 
 # We import torchdynamo modules indirectly to allow a future rename to torch.dynamo
@@ -32,6 +34,15 @@ def has_triton():
         return False
     try:
         import triton
+
+        if triton is not None:
+            # This is temp fix to hint people upgrad triton after https://github.com/pytorch/torchdynamo/pull/1585
+            triton_compiler_version = triton.runtime.version_key().split("-")[6]
+            if triton_compiler_version != "7929002797455b30efce6e41eddc6b57":
+                log.error(
+                    "Please update your triton version to af76c989eb4799b015f8b288ccd8421558772e56"
+                )
+                exit()
 
         return triton is not None
     except (ImportError, ModuleNotFoundError):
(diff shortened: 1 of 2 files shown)


torchinductor/utils.py
```diff
@@ -33,6 +35,15 @@ def has_triton():
     try:
         import triton
 
+        if triton is not None:
+            # This is temp fix to hint people upgrad triton after https://github.com/pytorch/torchdynamo/pull/1585
+            triton_compiler_version = triton.runtime.version_key().split("-")[6]
+            if triton_compiler_version != "7929002797455b30efce6e41eddc6b57":
+                log.error(
+                    "Please update your triton version to af76c989eb4799b015f8b288ccd8421558772e56"
```
jansel, line 43: Let's include the pip command needed to install it in the error message.

torchinductor/utils.py
```diff
@@ -33,6 +35,15 @@ def has_triton():
     try:
         import triton
 
+        if triton is not None:
+            # This is temp fix to hint people upgrad triton after https://github.com/pytorch/torchdynamo/pull/1585
+            triton_compiler_version = triton.runtime.version_key().split("-")[6]
+            if triton_compiler_version != "7929002797455b30efce6e41eddc6b57":
```
albanD, line 41: Requirering a specific version sounds a bit harsh? If you only want to prevent that error, you can just detect that the function takes more arguments now.
Also you can have an env variable to override this check so that devs that know what they're doing can set it to skip this.

torchinductor/utils.py
```diff
@@ -33,6 +35,15 @@ def has_triton():
...
+                log.error(
+                    "Please update your triton version to af76c989eb4799b015f8b288ccd8421558772e56"
+                )
+                exit()
```
albanD, line 45: Why not raise RuntimeError()

torchinductor/utils.py
```diff
@@ -33,6 +35,15 @@ def has_triton():
     try:
         import triton
 
+        if triton is not None:
+            # This is temp fix to hint people upgrad triton after https://github.com/pytorch/torchdynamo/pull/1585
+            triton_compiler_version = triton.runtime.version_key().split("-")[6]
```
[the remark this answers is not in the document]
desertfire: When we sync to fbcode this time, we will have to update trition. So I guess it should be fine.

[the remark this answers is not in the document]
jansel: Yeah, but then when we want to update the next time this will prevent that upgrade.

torchinductor/utils.py
desertfire: Switched to checking launch-time exception instead. Should be general enough unless the next time triton upgrade triggers exactly the same error. This can also address Jason's FBCODE concern.

albanD (approved):
SGTM

jansel (approved):

Issue #1588, opened 2022-10-11
Migration to PyTorch core


jansel:
We have merged `github.com/pytorch/torchdynamo` into `github.com/pytorch/pytorch`.  Being in a separate repo has allowed us to move fast, but it creates some installation roadblocks for users and version pinning has been a growing headache as the project scales.

- `import torchdynamo` is now `import torch._dynamo`
- `import torchinductor` is now `import torch._inductor`
- The standalone C extensions are added to `torch._C`
- `torchdynamo/benchmarks` is now `pytorch/benchmarks/dynamo`
- `torchdynamo/test/dynamo` is now `pytorch/test/dynamo`
- `torchdynamo/test/inductor` is now `pytorch/test/inductor` 

# Instructions for migrating your PR from `pytorch/torchdynamo` to `pytorch/pytorch`

1) Merge/rebase your PR to the correct hash of TorchDynamo: https://github.com/pytorch/torchdynamo/commit/6380959be21851bfda99424392cc08fda29d073d
```
cd ../torchdynamo
git pull origin 6380959be21851bfda99424392cc08fda29d073d
```

2) Checkout a new branch with the correct version of PyTorch: https://github.com/pytorch/pytorch/commit/054a2fd6c2fc361796663eed4772368c287d6c83
(the two versions must match so only your changes are copied over):
```
cd ../pytorch
git fetch origin master 
git checkout -b <your_new_branch_name> 054a2fd6c2fc361796663eed4772368c287d6c83
```

3) Run the rsync+auto-format script
```
bash ../torchdynamo/copy_to_core.sh
```

4) Commit your changes, rebase to the latest version of PyTorch, and create a pull request as normal for the PyTorch repo.

If you are copying files by hand, make sure to run `make setup_lint && lintrunner -a` to switch to the new auto-formatting.

# The Plan

## Phase 1 (complete)

In the first phase we will move the code over to core, which consists of these two PRs:
https://github.com/pytorch/torchdynamo/pull/1538
https://github.com/pytorch/pytorch/pull/86461

In this phase (now complete), the dynamo source of truth remained `pytorch/torchdynamo`.  Any changes will be copied over by the `copy_to_core.sh` script which we will rerun at the end of phase one.

After these initial PRs land, dynamo will be working in core -- but most of the testing will be disabled.  The initial PR will have testing disable, but then we will enable tests in smaller followup PRs.

## Phase 2 (going on now)

[...]

PR #1609, opened 2022-10-12
add a skipping feature to operatorbench.py

Krovatkin:
This PR adds a skip file feature to easily skip segfaulting or other not easily recoverable tests if any. It works in the conjuction with the verbose option.

The verbose option additionally prints the exact test inputs e.g.

```
Running with the following inputs: ((T([256, 8, 31, 33], f16), -1, False), {})
```

this inputs can be copied as is into a skip file.
The skip file format is 
```
op,dtype,serialized_arguments
```

```
aten._softmax.default;float16;((T([16, 12, 128, 128], f16), -1, False), {})
```

### DIFF
--- a/benchmarks/microbenchmarks/operator_inp_utils.py
+++ b/benchmarks/microbenchmarks/operator_inp_utils.py
@@ -2,6 +2,7 @@ import functools
 import logging
 import math
 import os
+import collections
 from collections import Counter
 from collections import defaultdict
 from functools import partial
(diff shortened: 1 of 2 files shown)


Krovatkin:
We could do that as well. I wasn't sure we would always get the same ordering over multiple runs.

eellison:
It is deterministic

desertfire:
Is the goal here to enable run `all` by skipping segfault failures? We could simply add a skip list like in test_torchinductor_opinfo.py, and eventually make the list empty.

jansel:

PR #1605, opened 2022-10-12
Added softplus and softplus_backwards to decomps

fdrocha:
Turns out the decomps had been in pytorch for a while, and benchmarks look good.

<details><summary>Benchmarks compared to eager</summary>

```
[--------------------------- softplus ---------------------------]
                                               |  eager  |  decomp
32 threads: ------------------------------------------------------
      (torch.Size([1024]),), (2.0, 20.0)       |      6  |      8
      (torch.Size([1048576]),), (2.0, 20.0)    |     31  |     31
      (torch.Size([33554432]),), (2.0, 20.0)   |    925  |    948
      (torch.Size([67108864]),), (2.0, 20.0)   |   1848  |   1900
      (torch.Size([134217728]),), (2.0, 20.0)  |   3691  |   3783
      (torch.Size([268435456]),), (2.0, 20.0)  |   7380  |   7540

[---------------------------------- softplus_backward -----------------------------------]
                                                                       |  eager  |  decomp
32 threads: ------------------------------------------------------------------------------
      (torch.Size([1024]), torch.Size([1024])), (2.0, 20.0)            |      6  |      9
      (torch.Size([1048576]), torch.Size([1048576])), (2.0, 20.0)      |     45  |     45
      (torch.Size([33554432]), torch.Size([33554432])), (2.0, 20.0)    |   1351  |   1388
      (torch.Size([67108864]), torch.Size([67108864])), (2.0, 20.0)    |   2699  |   2771
      (torch.Size([134217728]), torch.Size([134217728])), (2.0, 20.0)  |   5397  |   5533
      (torch.Size([268435456]), torch.Size([268435456])), (2.0, 20.0)  |  10790  |  11080
```

</details>

For future reference, filed an issue with triton about this: https://github.com/openai/triton/issues/773

PR #1633, opened 2022-10-13
New way to compile to libdevice calls for f64

fdrocha:
This is partly to generate cleaner triton code,
and partly to workaround a triton bug.

See https://github.com/pytorch/torchdynamo/pull/1605

### DIFF
--- a/torchinductor/lowering.py
+++ b/torchinductor/lowering.py
@@ -312,7 +312,7 @@ def make_pointwise(
             assert len(index) == len(ranges), f"wrong ndim {index} {ranges}"
             if dtype == torch.bool and override_fn_when_input_bool is not None:
                 return override_fn_when_input_bool(*[load(index) for load in loaders])
-            elif override_fn_when_cuda_float64 and is_cuda and dtype  == torch.float64:
+            elif override_fn_when_cuda_float64 and is_cuda and dtype == torch.float64:
                 return override_fn_when_cuda_float64(*[load(index) for load in loaders])
             else:
                 return fn(*[load(index) for load in loaders])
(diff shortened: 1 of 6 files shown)


torchinductor/lowering.py
```diff
@@ -3204,25 +3212,33 @@ def sum_(x, axis=None, keepdims=False, *, dtype=None):
...
+    use_libdevice_for_f64=True,
 )
 register_pointwise(
-    aten.sin, type_promotion_kind=ELEMENTWISE_TYPE_PROMOTION_KIND.INT_TO_FLOAT
+    aten.sin,
+    type_promotion_kind=ELEMENTWISE_TYPE_PROMOTION_KIND.INT_TO_FLOAT,
+    use_libdevice_for_f64=True,
 )
-register_pointwise(aten.abs)
+register_pointwise(aten.abs, use_libdevice_for_f64=True)
```
ngimel, line 3241: Can you please check if `abs` requires libdevice at all? I tried it at some point and seems like `tl.abs` accepts fp64 args.

ngimel (approved):
This looks good, I left minor comment about `abs`, can you please check if some of the functions on the skip list in `test_torchinductor_opinfo` can now pass or can be moved to xfail list because they no longer segfault?

ngimel:
silu error looks real.

torchinductor/lowering.py
fdrocha: You are right! Will remove the override for `abs`.

fdrocha:
> silu error looks real.

silu error happens because both `silu` and `sigmoid` are defined in `codegen/common.py` and they end up bypassing the overrides and dispatching to `tl.exp` even for `float64`.

I feel like the cleaner way to fix this is to make both `silu` and `sigmoid` come from primtorch decompositions instead, unless there is some downside in terms of compilation efficiency or something like that.

fdrocha:
FYI, the triton issue is already fixed in master, see https://github.com/openai/triton/issues/773

We no longer need this change but still feels worth making because it leads to cleaner generated triton.

fdrocha:
@ngimel could you please have a look at last commit?
I want to benchmark some of the changes I made, but I wanted to check if it passes CI first.

jansel:

PR #1639, opened 2022-10-13
Use Fake Tensors For Shape/Stride Propagation

eellison:
@Chillee observed peak memory increase from this invocation which I have been unable to repro upon rebase.. Nevertheless, we should do this anyway because it will be necessary in the dynamic shapes case, and I think we have some other issues today with running kernels like `lu_solve` which will error on inputs being zero.


jansel:

PR #1645, opened 2022-10-13
Convolution(_backward) bias decomposition for nvprims

jjsjann123:
Added a decomposition in nvprims code path, so we'll have bias/sum detached from convolution(_backward) op where more fusion opportunity is exposed.


IvanYashchuk:
Since dynamo was merged into https://github.com/pytorch/pytorch, let's open a PR there?

jansel:

PR #1647, opened 2022-10-13
Update the naming convention for module_key for easier parsing

SherlockNoMad:
Quantization team is paring the module structure. 

This change produces  `'self_convs_0_conv': 'Conv2d'` instead of `'self_convs_0__conv': 'Conv2d'`


jerryzh168 (approved):
should we add a test as well?

voznesenskym (approved):

PR #1651, opened 2022-10-13
(not for land) Add dist_bench.py for simple experiments

wconstab:
Currently hacks fake tensor, so isn't for land

- supports printing ddp buckets per model
- runs timing experiments on N nodes with dynamo/ddp/fsdp


jansel:

PR #1655, opened 2022-10-13
Fix FX graph stack traceback issues

### LINKED ISSUE #1167: Very first node in FX graph has misattributed stack
Steps to reproduce:
1. Build PyTorch at https://github.com/pytorch/pytorch/pull/84246 (at time of writing, I was on ce220330e5d69899ee89ea8dd533ae81e08e0667 ); this step may not be necessary but I did not try
2. Make log.info print stack traces for the traced graph with this patch

```
diff --git a/torchdynamo/output_graph.py b/torchdynamo/output_graph.py
index db84c3e..54d26a3 100644
--- a/torchdynamo/output_graph.py
+++ b/torchdynamo/output_graph.py
@@ -373,7 +373,7 @@ class OutputGraph(fx.Tracer):
             # the call to tabulate can cause a lot of memory to be allocated
             if config.log_level <= logging.INFO:
                 log.info(
-                    f"TRACED GRAPH\n {name} {gm.forward.__code__.co_filename} {format_graph_tabular(gm.graph)}\n"
+                    f"TRACED GRAPH\n {name} {gm.forward.__code__.co_filename} {gm.print_readable()}\n"
                 )
         except ImportError:
             log.warning(
```

(btw, we should consider making this the default)

3. Inspect the printed graph from `python benchmarks/torchbench.py --only BERT_pytorch --accuracy-aot-nop`:

```
    def forward(self, inputs_0_ : torch.Tensor, inputs_1_ : torch.Tensor):                                      
                                                                                                                
        # Module stack: {'mod': 'BERT'}, File: benchmarks/torchbench.py:336, code: return mod(*inputs)          
        gt = inputs_0_ > 0                                                                                      
                                                                                                                
        # Module stack: {'mod': 'BERT'}, File: /raid/ezyang/benchmark/torchbenchmark/models/BERT_pytorch/bert_pytorch/model/bert.py:40, code: mask = (x > 0).unsqueeze(1).repeat(1, x.size(1), 1).unsqueeze(1)                  
        unsqueeze = gt.unsqueeze(1);  gt = None                                                                 
        repeat = unsqueeze.repeat(1, 128, 1);  unsqueeze = None                                                 
        unsqueeze_1 = repeat.unsqueeze(1);  repeat = None
```

You can see that the `inputs_0_ > 0` operation has a useless stack trace, and the next set of lines (unsqueeze, repeat, unsqueeze) have the correct stack trace (luckily for me, this is the right stack trace for gt too).

This suggests that dynamo doesn't have the correct stack trace setup at the very beginning of a graph capture.

cc @SherlockNoMad

williamwen42:
Fixes #1167 and related bugs where the current translator object is not properly given to FX graph nodes.

Example: https://gist.github.com/williamwen42/1efa7ce86b7ac797e90a13356668d90d
(Note: running the above script before this change requires this modification to the code:
```
--- a/torchdynamo/output_graph.py
+++ b/torchdynamo/output_graph.py
@@ -373,7 +373,7 @@ class OutputGraph(fx.Tracer):
             # the call to tabulate can cause a lot of memory to be allocated
             if config.log_level <= logging.INFO:
                 log.info(
-                    f"TRACED GRAPH\n {name} {gm.forward.__code__.co_filename} {format_graph_tabular(gm.graph)}\n"
+                    f"TRACED GRAPH\n {name} {gm.forward.__code__.co_filename} {gm.print_readable()}\n"
                 )
```
Running after this change requires setting `torchdynamo.config.output_graph_code = True`).

Running `python benchmarks/torchbench.py --only BERT_pytorch --performance --verbose` now gives:
```
class GraphModule(torch.nn.Module):
    def forward(self, inputs_0_ : torch.Tensor, inputs_1_ : torch.Tensor):
        
        # Module stack: {'mod': 'BERT'}, File: /scratch/williamwen/work/torchbenchmark/torchbenchmark/models/BERT_pytorch/bert_pytorch/model/bert.py:40, code: mask = (x > 0).unsqueeze(1).repeat(1, x.size(1), 1).unsqueeze(1)
        gt = inputs_0_ > 0
        unsqueeze = gt.unsqueeze(1);  gt = None
        repeat = unsqueeze.repeat(1, 128, 1);  unsqueeze = None
        unsqueeze_1 = repeat.unsqueeze(1);  repeat = None
        
        # Module stack: {'mod': 'BERT', 'mod_embedding': 'BERTEmbedding', 'mod_embedding_token': 'TokenEmbedding'}, File: /scratch/williamwen/work/torchdynamo-env/lib/python3.8/site-packages/torch/nn/modules/sparse.py:161, code: input, self.weight, self.padding_idx, self.max_norm,
        mod_embedding_token_weight = self.mod_embedding_token_weight
        
        # Module stack: {'mod': 'BERT', 'mod_embedding': 'BERTEmbedding', 'mod_embedding_token': 'TokenEmbedding'}, File: /scratch/williamwen/work/torchdynamo-env/lib/python3.8/site-packages/torch/nn/modules/sparse.py:160, code: return F.embedding(
        embedding = torch.nn.functional.embedding(inputs_0_, mod_embedding_token_weight, 0, None, 2.0, False, False);  inputs_0_ = mod_embedding_token_weight = None
```

### DIFF
--- a/torchdynamo/output_graph.py
+++ b/torchdynamo/output_graph.py
@@ -104,6 +104,7 @@ class OutputGraph(fx.Tracer):
         self.compiler_fn = compiler_fn
         self.root_globals = f_globals
         self.root_tx = root_tx
+        self._current_tx = []
         self.cleanups = []
         self.should_exit = False
         self.random_values_var = None
@@ -117,6 +118,16 @@ class OutputGraph(fx.Tracer):
     @property
     def fake_mode(self):
         return self.root_tx.fake_mode
+
+    def push_tx(self, tx):
+        self._current_tx.append(tx)
+
+    def pop_tx(self):
+        self._current_tx.pop()
+
+    @property
+    def current_tx(self):
+        return self.root_tx if not self._current_tx else self._current_tx[-1]
 
     def copy_graphstate(self):
         """Create a checkpoint of the current state by copying everything"""
@@ -526,7 +537,7 @@ class OutputGraph(fx.Tracer):
         )
 
         # append stack trace to fx node
-        tx = current_tx if current_tx else self.root_tx
+        tx = self.current_tx
 
         nn_module_stack = getattr(tx, "nn_module_stack")
         if nn_module_stack:
--- a/torchdynamo/output_graph.py
+++ b/torchdynamo/output_graph.py
@@ -206,14 +206,7 @@ class OutputGraph(fx.Tracer):
                 name,
             )
 
-    def register_attr_or_module(
-        self, mod: torch.nn.Module, *names, current_tx=None, **options
-    ):
-        from .symbolic_convert import InstructionTranslatorBase
-
-        if not isinstance(current_tx, InstructionTranslatorBase):
-            current_tx = None
-
+    def register_attr_or_module(self, mod: torch.nn.Module, *names, **options):
         if is_dynamic_nn_module(mod):
             return variables.UnspecializedNNModuleVariable(mod, **options)
 
@@ -228,7 +221,10 @@ class OutputGraph(fx.Tracer):
                 return TensorVariable.create(
                     self,
                     self.create_proxy(
-                        "get_attr", module_key, tuple(), {}, current_tx=current_tx
+                        "get_attr",
+                        module_key,
+                        tuple(),
+                        {},
                     ),
                     example_value=mod,
                     **options,
@@ -246,8 +242,6 @@ class OutputGraph(fx.Tracer):
             def wrap_name(module_key):
                 self.output.update_co_names(module_key)
                 self.root_globals[module_key] = mod
-                # FIXME we lose track of current_tx here, so stack traces produced by
-                # OutputGraph.create_proxy through this variable won't be correct.
                 return VariableBuilder(self, ConstantSource(source_name=module_key))(
                     mod
                 )
@@ -530,7 +524,6 @@ class OutputGraph(fx.Tracer):
         name=None,
         type_expr=None,
         proxy_factory_fn=None,
-        current_tx=None,
     ):
         rv = super().create_proxy(
             kind, target, args, kwargs, name, type_expr, proxy_factory_fn
--- a/torchdynamo/output_graph.py
+++ b/torchdynamo/output_graph.py
@@ -195,7 +195,9 @@ class OutputGraph(fx.Tracer):
                 name,
             )
 
-    def register_attr_or_module(self, mod: torch.nn.Module, *names, current_tx=None, **options):
+    def register_attr_or_module(
+        self, mod: torch.nn.Module, *names, current_tx=None, **options
+    ):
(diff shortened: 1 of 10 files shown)


torchdynamo/symbolic_convert.py
```diff
@@ -979,28 +985,28 @@ def MAKE_FUNCTION(self, inst):
...
-            for i in reversed(seq.items):
+            for i in reversed(seq.unpack_var_sequence(self)):
                 self.push(i)
         elif seq.is_python_constant() and isinstance(seq, ConstantVariable):
-            val = seq.as_python_constant()
+            val = seq.unpack_var_sequence(self)
             assert len(val) == inst.argval
             for i in reversed(val):
-                self.push(ConstantVariable(i, **options))
+                self.push(i)
```
jansel, line 998: Can some of these cases be combined now?   This seems to repeat a lot.

jansel (changes requested):

jansel (approved):

mlazos:
what's the difference between this new stack of tx's that you're adding to `OutputGraph` and the existing stack (see `tx.parent`) Could that stack be used instead? Otherwise, LGTM.

torchdynamo/output_graph.py
```diff
@@ -118,6 +119,16 @@ def output(self):
     def fake_mode(self):
         return self.root_tx.fake_mode
 
+    def push_tx(self, tx):
```
voznesenskym, line 122: Ah, you went with the stack after all :) glad it worked

torchdynamo/output_graph.py
```diff
@@ -118,6 +119,16 @@ def output(self):
...
+        self._current_tx.append(tx)
+
+    def pop_tx(self):
```
voznesenskym, line 125: Nit: idiomatic for pop to return what it popped

torchdynamo/variables/tensor.py
```diff
@@ -392,15 +392,18 @@ def var_getattr(self, tx, name):
 
         return result
 
-    def unpack_var_sequence(self, tx):
+    def unpack_var_sequence_range(self, tx, range):
```
voznesenskym, line 395: Hmm I wonder if we eventually need to do this for all unpacks? As in, make them all take range or provide a default?

voznesenskym (approved):
Lgtm

jansel (changes requested):
I'm about to switchover to core, lets hold off on landing

jansel:

williamwen42:
> what's the difference between this new stack of tx's that you're adding to `OutputGraph` and the existing stack (see `tx.parent`) Could that stack be used instead? Otherwise, LGTM.

The output graph only keeps a reference to a root (first) tx, not the most current one. The tx stack replaces the current_tx kwarg, since passing around a translator object as a kwarg is quite prone to errors.

PR #1660, opened 2022-10-14
add benchmarking utility + remove condition for reduction dim to be an integer for us to unroll

Chillee:
We're already guarding on the reduction ranges, so we might as well lift the requirement that it's an integer.

### DIFF
--- a/torchinductor/ir.py
+++ b/torchinductor/ir.py
@@ -711,8 +711,7 @@ class Reduction(Loops):
             return Pointwise.create(device, dst_dtype, fn, ranges)
 
         if (
-            isinstance(reduction_numel, sympy.Integer)
-            and V.graph.sizevars.size_hint(reduction_numel)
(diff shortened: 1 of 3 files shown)


torchinductor/ir.py
```diff
@@ -711,8 +711,7 @@ def fn(index):
             return Pointwise.create(device, dst_dtype, fn, ranges)
 
         if (
-            isinstance(reduction_numel, sympy.Integer)
-            and V.graph.sizevars.size_hint(reduction_numel)
+            V.graph.sizevars.size_hint(reduction_numel)
```
ngimel, line 714: should you guard on reduction_numel in this case?

Chillee: We guard on `reduction_numel` within `_unroll_reduction_fn`

jansel (approved):

jansel:

Issue #1257, opened 2022-09-16
pytorch core test failure: RuntimeError: Can't call numpy() on Tensor that requires grad. Use tensor.detach().numpy() instead.


desertfire:
Repro:
cherry-pick https://github.com/pytorch/pytorch/pull/85183
```
PYTORCH_TEST_WITH_INDUCTOR=1 pytest test/distributions/test_distributions.py  -k  test_pareto
```
Error:
```
Traceback (most recent call last):
  File "/fsx/users/binbao/pytorch/test/distributions/test_distributions.py", line 2556, in test_pareto
    self._check_log_prob(Pareto(scale, alpha), ref_log_prob)
  File "/fsx/users/binbao/pytorch/test/distributions/test_distributions.py", line 837, in _check_log_prob
    asset_fn(i, val.squeeze(), log_prob)
  File "/fsx/users/binbao/pytorch/test/distributions/test_distributions.py", line 2550, in ref_log_prob
    def ref_log_prob(idx, x, log_prob):
  File "/fsx/users/binbao/conda/envs/release/lib/python3.9/site-packages/scipy/stats/_distn_infrastructure.py", line 1884, in logpdf
    def logpdf(self, x, *args, **kwds):
  File "/fsx/users/binbao/conda/envs/release/lib/python3.9/site-packages/scipy/stats/_distn_infrastructure.py", line 1907, in <graph break in logpdf>
    args, loc, scale = self._parse_args(*args, **kwds)
  File "/fsx/users/binbao/conda/envs/release/lib/python3.9/site-packages/scipy/stats/_distn_infrastructure.py", line 1908, in <graph break in logpdf>
    x, loc, scale = map(asarray, (x, loc, scale))
  File "/fsx/users/binbao/pytorch/torch/_tensor.py", line 945, in __array__
    return self.numpy()
RuntimeError: Can't call numpy() on Tensor that requires grad. Use tensor.detach().numpy() instead.
----------------------------------------------------------------------------------------------------- Captured log call -----------------------------------------------------------------------------------------------------
ERROR    torchdynamo.convert_frame:convert_frame.py:357 WON'T CONVERT test_pareto /fsx/users/binbao/pytorch/test/distributions/test_distributions.py line 2535 
due to: 
Traceback (most recent call last):
  File "/fsx/users/binbao/torchdynamo-tip/torchdynamo/variables/tensor.py", line 266, in create
    assert (
[...]

PR #1661, opened 2022-10-14
[WIP] Fix .view().detach() not handled correctly by AOT autograd

sangongs:
This is to fix issue https://github.com/pytorch/torchdynamo/issues/1257

Depending on https://github.com/pytorch/pytorch/pull/86838

### DIFF
--- a/torchinductor/decomposition.py
+++ b/torchinductor/decomposition.py
@@ -28,7 +28,6 @@ decompositions = get_decompositions(
         aten.col2im,
         aten.cudnn_batch_norm,
         aten.cudnn_batch_norm_backward,
-        aten.detach,
         aten.dot,
         aten.elu,
         aten.elu_backward,
--- a/torchinductor/ir.py
+++ b/torchinductor/ir.py
@@ -1323,6 +1323,7 @@ class ReinterpretView(BaseView):
     """Pretend our storage has a different layout"""
 
     layout: "Layout"
+    detached = False
 
     def __str__(self):
         return self.str_helper(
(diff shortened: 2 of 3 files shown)


Chillee:
@sangongs I don't think this is the right solution. Inductor should not be responsible for handling things like autograd.

sangongs:
@Chillee Thanks for having an eye on this.

> Inductor should not be responsible for handling things like autograd.

Agreed.

This is just a tentative workaround for [an AOT autograd issue](https://github.com/pytorch/functorch/issues/1052 ). AOT autograd currently does not handle `detach()` correctly. According to @bdhirsh:
> it's because we're compiling the whole thing (including the `detach()` call) into an `autograd.Function`, and `autograd.Function` will unconditionally mark all of its forward outputs as requiring gradients.

 @bdhirsh made [a fix](https://github.com/pytorch/pytorch/pull/86838): 
> We can use autograd.function's mark-nondifferentiable [API](https://pytorch.org/docs/stable/generated/torch.autograd.function.FunctionCtx.mark_non_differentiable.html#torch.autograd.function.FunctionCtx.mark_non_differentiable), to (statically) mark those outputs as not requiring gradients

But the fix does not work as expected if the output is a view,  because `mark_non_differentiable` takes no effect in such a case:
https://github.com/pytorch/pytorch/blob/ae45dab57e22e3d04516e7dd81ef8dbefd51bfe3/torch/csrc/autograd/custom_function.cpp#L290-L299

See the AOT autograd issue for more details: https://github.com/pytorch/functorch/issues/1052

In my view, there are four options to fix this:
1. Leave `autograd.Function` as it is. Work around the `is_view()` problem in Inductor like this PR.
2. Work around the problem in AOT autograd. We are discussing this in the AOT autograd issue. But apparently @bdhirsh has concerns.
3. Work around the problem in Dynamo. This may take a lot of efforts.
4. Update `autograd.Function` such that `mark_non_differentiable()` takes effect when an output is a view. But I am not sure if this is viable.

@Chillee Do you have any suggestions?

bdhirsh:
Oh tbc, my vote is probably still for (2) - manually add in some extra `.detach()` calls in AOTAutograd, so inductor isn't forced to worry about requires-grad-ness ([thread](https://github.com/pytorch/functorch/issues/1052#issuecomment-1278222869))

jansel:

PR #1665, opened 2022-10-14
fix masked_fill and where type promotion

### LINKED ISSUE #1619: [AMP] Bad decomp for softmax
Repro

~~~
from math import inf

import torch
import torch.fx as fx
from torch import device
from torch import tensor

import torchdynamo
from torchdynamo.debug_utils import run_fwd_maybe_bwd
from torchdynamo.debug_utils import same_two_models

args = [
    ((1, 12, 128, 128), (196608, 16384, 128, 1), torch.float16, "cuda", True),
    ((1, 12, 128, 128), (128, 0, 0, 1), torch.bool, "cuda", False),
    ((), (), torch.float32, "cpu", False),
]
args = [
    rand_strided(sh, st, dt, dev).requires_grad_(rg) for (sh, st, dt, dev, rg) in args
]


from torch.nn import *


class Repro(torch.nn.Module):
    def __init__(self):
        super().__init__()

    def forward(self, matmul, expand_as, tensor):
        masked_fill = matmul.masked_fill(expand_as, tensor)
        softmax = torch.nn.functional.softmax(masked_fill, dim=-1)
        return (softmax,)


mod = Repro().cuda()
opt_mod = torchdynamo.optimize("aot_inductor_debug")(mod)


with torch.cuda.amp.autocast(enabled=True):
    ref = run_fwd_maybe_bwd(mod, args)
    res = run_fwd_maybe_bwd(opt_mod, args)


~~~

ngimel:
Fixes #1619, turned out nothing to do with softmax, masked_fill and where type promotion was wrong.

### DIFF
--- a/test/inductor/test_torchinductor.py
+++ b/test/inductor/test_torchinductor.py
@@ -1940,6 +1940,21 @@ class CommonTemplate:
             ),
         )
 
+    def test_masked_fill_promotion(self):
+        def fn(mask, value):
+            return aten.masked_fill(value, mask, torch.tensor(float("-inf")))
+
+        opt_fn = torchdynamo.optimize("inductor")(fn)
+        inputs = (
+            torch.randint(0, 1, [1, 16], dtype=torch.bool, device=self.device),
+            torch.randn(
+                [16, 16],
+                dtype=torch.float16 if self.device == "cuda" else torch.float32,
+                device=self.device,
+            ),
+        )
+        self.assertEqual(fn(*inputs), opt_fn(*inputs))
+
     def test_fill1(self):
         def fn(x):
             tmp = torch.ones_like(x)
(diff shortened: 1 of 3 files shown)


anijain2305 (approved):
Thanks! I was looking at it simultaneously. I was trying to fix `masked_fill` by adding one more `to` if the `other` and `value` dtypes did not match. But, I did not see that its actually `where` which is causing the problem.

ngimel:
Just fixing masked_fill is enough to make the repro not error out, but the output dtype is wrong in this case (because of where), and where is super annoying because it needs manual broadcasting and type promotion. 
Looks like we'll need to have a way to more precisely specify type promoting arguments to `register_lowering` if we want to avoid these manual type promotion situations.

jansel (approved):

ngimel:
@anijain2305 you are right, we should convert `other` to `value`'s type in `masked_fill`, otherwise filling e.g. integer tensor with float value would cause wrong type promotion in `where`, I'll push a fix. 
`where` type promotion was still wrong before this PR (and it's still a little bit wrong now), so I'll leave `where` fixes even though they are not necessary for `masked_fill`.

malfet (approved):

PR #1669, opened 2022-10-14
fused bias for cpu convolution path

XiaobingSuper:
* __->__ #1669

For aten.convolution CPU path, the bias always can be fused, so this PR adds a device check: if inputs's device is CPU, we will fuse it.

### DIFF
--- a/torchinductor/lowering.py
+++ b/torchinductor/lowering.py
@@ -1065,11 +1065,15 @@ def convolution(
     output_padding: List[int],
     groups: int,
 ):
+    # For cpu path, bias is alway can be fused.
+    inputs_is_cpu = x.get_device().type == "cpu" and weight.get_device().type == "cpu"
+    if bias is not None:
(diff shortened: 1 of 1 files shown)


torchinductor/lowering.py
```diff
@@ -1065,11 +1065,15 @@ def convolution(
...
+    # For cpu path, bias is alway can be fused.
+    inputs_is_cpu = x.get_device().type == "cpu" and weight.get_device().type == "cpu"
+    if bias is not None:
+        inputs_is_cpu = inputs_is_cpu and bias.get_device().type == "cpu"
     result = TensorBox.create(
         ir.Convolution.create(
             x,
             weight,
-            None,  # bias handled below
+            bias if inputs_is_cpu else None,  # bias handled below for gpu path
```
jgong5, line 1076: You may explain the reason "For cpu path, bias can always be fused" here.

jgong5 (changes requested):

jansel:

PR #1674, opened 2022-10-14
Copy inputs in minifier as inductor now clears the list

anijain2305:
@eellison

### DIFF
--- a/torchdynamo/debug_utils.py
+++ b/torchdynamo/debug_utils.py
@@ -404,17 +404,18 @@ def wrap_compiler_debug(compiler_fn, compiler_name: str):
                     raise NotImplementedError(
                         "Accuracy minification is supported for inductor only"
                     )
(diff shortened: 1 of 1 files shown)


torchdynamo/debug_utils.py
```diff
@@ -404,24 +404,27 @@ def deferred_for_real_inputs(real_inputs):
                     raise NotImplementedError(
                         "Accuracy minification is supported for inductor only"
                     )
+                cloned_real_inputs = [x.clone() for x in real_inputs]
```
eellison, line 407: you can just clone the list - don't need to clone the inputs. `x.clone()` won't preserve strides (or use clone_preserve_strides or whatever it's called)

`list(real_inputs)`

eellison (approved):
lgtm but needs one small fix

torchdynamo/debug_utils.py
anijain2305: Good catch!

PR #1682, opened 2022-10-15
Don't only apply DDP optimizer on forward frames

davidberard98:
Previously a check would only apply DDP optimizer on frames named "forward".

But on hf_T5_large, a graph break causes some frames like:
```
<graph break in _shift_right>
<graph break in forward>
```

So instead, apply DDP optimizer on all frames.


jansel:

PR #1696, opened 2022-10-18
include numels in divisible_by_16 descriptor tuple

stumpOS:
I am not sure how Triton code generation is tested so I added a class that extracts the kernels from inductor so I can check the descriptor given a function and inputs. @Chillee what is the preferred way to test?

Also I'm not sure what guards need to be added along with this because I'm not sure how this annotation is used yet (TODO: autotune.py contains the decorators that accept these arguments)

### DIFF
--- a/test/inductor/test_torchinductor.py
+++ b/test/inductor/test_torchinductor.py
@@ -4088,7 +4088,7 @@ if HAS_CUDA:
             from torchinductor.graph import GraphLowering
             from torchinductor.virtualized import V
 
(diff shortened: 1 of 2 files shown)


Chillee:
TorchDynamo has been merged into core, so please re-open this PR on pytorch/pytorch.

jansel (changes requested):
Instructions for porting this PR to pytorch/pytorch are in https://github.com/pytorch/torchdynamo/issues/1588

PR #1700, opened 2022-10-18
Bug Issue Template

msaroufim:
Click `files changed`, `view file` to see what the template will look like

Curious how people feel about a template in general, I'm thinking our Troubleshooting.md will be more of a dev facing doc to debug nuanced issues whereas this template should cover 90%+ of the bugs we've been fixing

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -9,6 +9,7 @@ body:
       value: >
         #### Before submitting a bug, please make sure the issue hasn't been already addressed by searching through [the
         existing and past issues](https://github.com/pytorch/torchdynamo/issues)
+        The below template should close the vast majority issues you're seeing but if it doesn't please check out our advanced troubleshoothing guide https://github.com/pytorch/torchdynamo/blob/main/documentation/TROUBLESHOOTING.md
   - type: textarea
     attributes:
       label: 🐛 Describe the bug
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -0,0 +1,47 @@
...
+      required: true
+  
+  - type: textarea
+    attributes:
+      label: Minified repro
+      description: |
+        Please run the minifier on your example and paste the minified code below 
+        Learn more here https://github.com/pytorch/torchdynamo/blob/main/documentation/TROUBLESHOOTING.md
+      placeholder: |
+        >> torch._dynamo.config.repro_after="aot" python your_model.py
```
jansel, line 37: I think users need to do:

```suggestion
        env TORCHDYNAMO_REPRO_AFTER="aot" python your_model.py
```

.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -0,0 +1,47 @@
...
+        >> torch._dynamo.config.repro_after="dynamo" python your_model.py
```
jansel, line 38:
```suggestion
        or
        env TORCHDYNAMO_REPRO_AFTER="dynamo" python your_model.py
```

yanboliang:
I'd suggest to add instruction on trying different backends and report if user's code passes on these backends: eager, aot_eager, inductor. This is the most straight forward way to locate which is the problematic component.

yanboliang (approved):

PR #1702, opened 2022-10-18
LRU for cache management

zyq8709:
This is just an initial proposal of LRU implementation for trace cache management. If I was right, the caches are organized into a single linked list without any explicit ordering. It is presumed that if LRU is applied to trace cache management, the better lookup performance may be achieved. There is no doubt that other better algorithms can be chosen for the ordering through elaborate research.

### DIFF
--- a/torchdynamo/_eval_frame.c
+++ b/torchdynamo/_eval_frame.c
@@ -180,6 +180,17 @@ inline static const char *name(PyFrameObject *frame) {
 }
 #endif
 
+inline static CacheEntry *get_extra(PyCodeObject *code) {
+  CacheEntry *extra = NULL;
+  _PyCode_GetExtra((PyObject *)code, extra_index, (void *)&extra);
+  return extra;
+}
+
+inline static void set_extra(PyCodeObject *code, CacheEntry *extra) {
+  // TODO(jansel): would it be faster to bypass this?
+  _PyCode_SetExtra((PyObject *)code, extra_index, extra);
+}
+
 static void call_guard_fail_hook(PyObject *hook, CacheEntry *e,
                                  PyObject *f_locals) {
   // call debugging logic when a guard fails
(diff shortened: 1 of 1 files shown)


torchdynamo/_eval_frame.c
```diff
@@ -216,12 +228,36 @@ static PyCodeObject *lookup(CacheEntry *e, PyObject *f_locals) {
   }
   Py_DECREF(valid);
   if (valid == Py_True) {
-    return e->code;
+    CacheEntry *extra = get_extra(frame->f_code), *next = e->next, *head = extra;
+    if (extra != e) {
```
jansel, line 232: Add some comments explaining what this is doing.

torchdynamo/_eval_frame.c
```diff
@@ -216,12 +228,36 @@ static PyCodeObject *lookup(CacheEntry *e, PyObject *f_locals) {
...
+        PyObject *check_fn_tmp = e->check_fn;
+        PyCodeObject *code_tmp = e->code;
+        memcpy(e, next, sizeof(CacheEntry));
+        next->check_fn = check_fn_tmp;
+        next->code = code_tmp;
+        next->next = extra;
+        head = next;
+      } else {
+        CacheEntry *cur = extra;
+        while (cur) {
```
jansel, line 243: We should be able to do this in O(1) time.  We would need a pred pointer for that though.

It would also be a lot cleaner if we could handle it with a single case rather than two.

jansel (changes requested):
We moved this file to `pytorch/pytorch` here:
https://github.com/pytorch/pytorch/blob/master/torch/csrc/dynamo/eval_frame.c

So you will need to resubmit to the new repo.

torchdynamo/_eval_frame.c
zyq8709: Hi, thanks for your advice! As you know, the caches are managed by a single linked list, so it is difficult to adjust the data structure directly. If possible, we can change it into a double linked list for O(1) and less case handling.

torchdynamo/_eval_frame.c
zyq8709: Fine. I can add some comments. Simply this is just a process to retrieve the original head, adjust the single linked list and set the new head.

torchdynamo/_eval_frame.c
jansel: One alternative to making it double-linked would be to pass the pred pointer of the current node in as an arg to the lookup function.  Since we are walking the list, we should already know the prior node (it was the one we looked at right before the current one).

jansel:
Closing this, please resubmit to pytorch/pytorch:

File moved to:
https://github.com/pytorch/pytorch/blob/master/torch/csrc/dynamo/eval_frame.c

PR #1704, opened 2022-10-18
Capture and parse torchbench logs

wconstab:
put process exit error into logs


jansel (changes requested):
This PR needs to go to core, see https://github.com/pytorch/torchdynamo/issues/1588

PR #1720, opened 2022-10-19
Update readme for alpha users

williamwen42:


README.md
```diff
@@ -52,7 +52,27 @@ performance of different training backends.  You can read more in the
...
+that you have installed locally matches the PyTorch version you are running. For
+the command below, you will need CUDA 11.7.
+
+`pip3 install --pre torchtriton --extra-index-url https://download.pytorch.org/whl/nightly/cu117` 
+
+### Install from local source
+
+Build PyTorch from source: https://github.com/pytorch/pytorch#from-source, which has TorchDynamo included.
+
+## Requirements and Setup (old)
```
msaroufim, line 75: We should just deprecate the old instructions, it's gonna confuse folks

README.md
```diff
@@ -52,7 +52,27 @@ performance of different training backends.  You can read more in the
...
+
+TorchDynamo is included in the nightly binaries of PyTorch, for reference, https://pytorch.org/get-started/locally/
+
+### Install GPU/CUDA version requirements
+
+To use GPU back ends (and in particular Triton), please make sure that the cuda
+that you have installed locally matches the PyTorch version you are running. For
+the command below, you will need CUDA 11.7.
+
+`pip3 install --pre torchtriton --extra-index-url https://download.pytorch.org/whl/nightly/cu117`
```
msaroufim, line 69: Mention CPU version which is just installing pytorch nightly

msaroufim (changes requested):

README.md
williamwen42: I made it clear that the old instructions are deprecated, but I think it's good to leave it for reference for now.

msaroufim (approved):

msaroufim:
There's a few more sections we should remove
* https://github.com/pytorch/torchdynamo#install-from-local-source
* https://github.com/pytorch/torchdynamo#verify-installation may have issues since it checks for dynamo version which is no longer a thing
* https://github.com/pytorch/torchdynamo#developer-setup is outdated

msaroufim (approved):

mlazos (approved):
I think this looks good. @msaroufim one thought I have is to move the guards piece to the FAQ under a question "What is JIT?" or something like that.

msaroufim:
Yeah I'll do that in a follow up PR

PR #1852, opened 2022-11-10
Docker installation instructions

msaroufim:
Working container is here https://github.com/orgs/pytorch/packages/container/package/pytorch-nightly so if you see this PR just go to that link and download the latest docker container

There are some issues with Docker nightly images
* Nightly latest have not been published in a while: https://github.com/pytorch/pytorch/issues/88833 so will need to point people to the page with teh docker images instead of just linking to it
* There might be an issue with larger compile times: https://github.com/pytorch/pytorch/pull/88741

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,8 +105,7 @@ We also provide all the required dependencies in the PyTorch nightly binaries wh
 
 And for ad hoc experiments just make sure that your container has access to all your GPUs
 
-`docker run --gpus all -it pytorch-nightly /bin/bash`
-
+`docker run --gpus all -it ghcr.io/pytorch/pytorch-nightly:latest /bin/bash`
 
 ## Getting started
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -97,6 +97,17 @@ python verify_dynamo.py
...
 
+## Docker installation
+
+We also provide all the required dependencies in the PyTorch nightly binaries which you can download with
+
+`docker pull ghcr.io/pytorch/pytorch-nightly`
+
+And for ad hoc experiments just make sure that your container has access to all your GPUs
+
+`docker run --rm --gpus all -it pytorch-nightly /bin/bash`
```
malfet, line 108: I wonder if it's better to omit `--rm` option here, as it allows folks to continue work from an accidentally stopped container

malfet (approved):

msaroufim:
Here's an example of dynamo working on a docker container that I'm attaching to in VS code

Screen Shot 2022-11-14 at 4 34 01 PM

